PR #6, opened 2016-07-13
Building Hours

elijahverdoorn:
Add the building hours work to master.

This formats the building hours data (including some of the images, which should be hosted on Oleville), adds a component or two, and builds out the basics of the hours display.


drewvolz:
You might want to consider hosting the images back on github and then accessing them through GitHub'a raw URL. That way it is much less likely that it would disappear at the hands of a new Oleville web developer as folders are moved and purged, and it would be easier to track changes to the files to have an audit of the images rather than the URL names. Makes changing them much easier overtime, too, as they can be committed right to the GitHub project.

Another thing -- I know this is version one of the translation, but you might want to consider the indicator that something is opening or closing soon (the yellow in the current version). I have had countless requests to include text on the table row for each building indicating "opens in 20 minutes" or "closes in 5 minutes" which could easily replace the yellow color. 

One more reason we are hoping to host things on GitHub is that other developers could make PRs more easily and fork the project to develop functionality. We are hoping to host the building hours as JSON here to allow an audit, a reliable spot to host both in uptime and reliability of it not being deleted, and for others to make changes to files, such as building hours.

elijahverdoorn:
@drewvolz Thanks for the comment.

I'll be looking to make another pass at all the major views before I publish, so the "closing soon" idea can be hit then. I am looking to get the first version out rather quickly, then iterate after it's published, so if not for the first release then for a version soon after that.

As for hosting, I'm still making decisions on where the content should come from. I've created a "tech" site on Oleville specifically for the app (and publishing updates from the team related to the app to the student body) and have set the permissions of that site to be just me and my direct team for now, so I am not super worried about anyone deleting or moving content at this time. I see the benefit of the content coming from GitHub, but also like the idea of reducing dependencies. The app is going to depend on Oleville no matter what, as I am going to pull content from the branch pages, so consolidating to that one place makes sense to me.

hawkrives:
@elijahverdoorn Out of curiosity: are you publishing this as an update to AAO or as a new app?

elijahverdoorn:
@hawkrives I'm looking at publishing it as a new app. I think that the two versions are sufficiently different, and with the publication of the Android version I'd rather just make a distinction between the two.

PR #7, opened 2016-07-13
Contacts v2

elijahverdoorn:
Add safe ride card
Re-order cards according to feedback from initial screenshots.


drewvolz:
See PR #6 for ideas and thoughts related to hosting of JSON/image files.

PR #8, opened 2016-07-29
Directory

elijahverdoorn:
Add a webview for the directory. Stolaf.edu already has a mobile view (although it's a bit wonky), so let's just use that until we can come up with something better.

### DIFF
--- a/views/directory.js
+++ b/views/directory.js
@@ -22,7 +22,7 @@ const styles = StyleSheet.create({
   }
 })
 
-const URL = 'https://www.stolaf.edu/personal/'
+const URL = 'https://www.stolaf.edu/personal/directory/index.cfm'
 
 export default class DirectoryView extends React.Component {
   constructor() {
--- a/views/directory.js
+++ b/views/directory.js
@@ -1,18 +1,54 @@
 /**
  * All About Olaf
- * iOS Directory page
+ * Directory page
  */
 
 import React from 'react'
 import {
   StyleSheet,
   View,
-  Text,
+  NetInfo,
+  WebView,
 } from 'react-native'
+
 import NavigatorScreen from './components/navigator-screen'
-import queryStalkernet from '../lib/stalkernet'
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+  },
+  webView: {
+    flex: 1,
+  }
+})
+
+const URL = 'https://www.stolaf.edu/personal/'
 
 export default class DirectoryView extends React.Component {
+  constructor() {
+    super()
+    this.scalesPageToFit = true
+    this.url = URL
+    this.state = {
+      isConnected: true
+    }
+  }
+
+  componentWillMount() {
+    NetInfo.isConnected.fetch().then(isConnected => {
+      this._onConnectivityChange(isConnected)
+    })
+    NetInfo.isConnected.addEventListener('change', this._onConnectivityChange.bind(this))
+  }
+
+  componentWillUnmount() {
+    NetInfo.isConnected.removeEventListener('change', this._onConnectivityChange.bind(this))
+  }
+
+  _onConnectivityChange(isConnected) {
+    this.setState({isConnected})
+  }
+
   render() {
     return <NavigatorScreen
       {...this.props}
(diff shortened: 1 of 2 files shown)


hawkrives:
@drewvolz and I found an xml endpoint for stalkernet and convinced Dan Beach to leave it expressly for use on-campus.

hawkrives:
… which is part of what `/lib/stalkernet.js` does.

drewvolz:
Making use of that endpoint that @hawkrives mentioned is ideal. We want to aim to be as native as possible in react-native! Shouldn't be more than a simple HTTP request to that endpoint

elijahverdoorn:
I understand the purpose of react native and would agree that there it's better to do implement this as you described.
**Two problems:**
- Stalkernet added the need to login to perform a search. This is in accordance with FERPA law. The existing helper file does not deal with this, and even if the endpoint allowed it I would not publish something that was potentially invading student privacy.
- I am running out of time. Version 1 of the app needs to be ready by the end of August. I am currently thinking of version 1 as a Minimum Viable Product, so since the current implementation is functional I am going to focus on other areas.

Addressing this section will be high on the list of improvements for v2.

hawkrives:
Wait - even on campus you have to log in?

Historically, on-campus has bypassed the login requirement.

**Edit:** Testing with the VPN shows that on-campus still bypasses the login requirement. This API that we found is only available on-campus. I'd been envisioning that the entire Stalkernet section would block you out when it didn't detect a 130.71 IP on the device, but even if it didn't the API wouldn't return anything if you were off-campus.

hawkrives:
Also, what would you most like help with? I'd be happy to pitch in.

elijahverdoorn:
@hawkrives From my VPN, yes. I support that change too, it was rather uncomfortable to me to allow anyone to access that data who was on the network, as I know that there are external people given that access.

I'll make a few more changes to the readme with things that you could do. Thanks for the offer!

hawkrives:
Huh. I just logged in with my Olaf VPN and searched for you and didn't have to log in.

Hmm. I agree that it's a bit uncomfortable… perhaps we should communicate with Dan to check on expectations.

Maybe it's time to move the SIS login up to the app-level. We (read: @drewvolz) never really had a reason to before, since nothing else needed login.

Thanks for writing out a list of stuff!

elijahverdoorn:
Accidental close :(

Yeah, App-level login would be good. I want it usable for anyone that doesn't have an Olaf login, so we would have to keep that in mind during implementation.

hawkrives:
Yeah. How about - and this is spitballing off the top of my head - login is managed in "settings" (which we'd need to add), and we prompt for it whenever you try to open a section that needs login?

So, you open the app, and it works just fine. 

The first time you open SiS or Stalkernet, it pops up a interstitial modal for login, then stores that – somewhere, in persistent storage. You can log out by going to settings.

(If we wanted to be really fancy, we could add an lock icon or something to the auth-required buttons so you know before you tap.)

elijahverdoorn:
Right, that sounds good. I don't know how the XML endpoint for stalkernet handles the login... If it returns the data without login we cannot use it until it gets fixed, that would open sensitive data to the public. I'll speak with Dan.

hawkrives:
IIRC, it just doesn't exist unless you've got a 130.71 IP (aka on-campus/vpn). If it exists, it doesn't authenticate. LMK if that's incorrect.

elijahverdoorn:
@hawkrives That's what I'm getting. I think that it would be best if it worked off-campus (for those with a login, of course). If the endpoint doesn't exist. maybe the webview is the best option? It's not pretty, but it seems to provide the most functionality. Do you have other ideas?

elijahverdoorn:
Also updated the readme, feel free to take on any of the tasks that are listed in the TODO marked _PRs Welcome_.

hawkrives:
… To be honest, I really just don't like the web interface for Stalkernet, so that's coloring my opinion on this quite heavily.

So. With that qualifier, I think it's quite reasonable to restrict Stalkernet use to the campus network only. If it's important that it be available to anyone with a login, anywhere, well… I suppose we could either tell them to go use the web browser, or just (since webviews are easy in RN) add a button to fall back to the web view?

elijahverdoorn:
@hawkrives Fair enough, that sounds reasonable. I'll see if I can reconfigure my VPN to get passwordless access, then build this out a bit.

hawkrives:
SInce I'm the one pushing for it, if you want to say "you want it, you build it" that's fair too 😜

elijahverdoorn:
@hawkrives I'm already on it, I'll do it... You do streaming media.

hawkrives:
Lol k

drewvolz:
Would it be fair to ask that the directory URL end up in some sort of resources JSON file? Things break all too often when we can't control who makes changes and how often. If they suddenly change the directory address, it will break, so quick editing on the app's part might be handy. This is the design principle I am hoping to aim for in the SIS portion as well (which I have not started on) -- where the URLs are extracted into a resource JSON file. I have been burned by this before!

I would be willing to create the separate resources JSON file and add that URL to it. Just let me know.

elijahverdoorn:
@drewvolz I think that a URLs.json would be a good idea. Go for it!

PR #25, opened 2016-08-05
Implement the Maps screen

### LINKED ISSUE #11: Campus Map
- The iOS version of this is easy, there is a built-in component
- Android version could be harder, find an existing component?

---

IIRC, when I last looked at this, I came to the conclusion that we'd need to get a Google Maps API key for Android, at least.

hawkrives:
using a web view against the St. Olaf maps page.

Closes #11.

screen shot 2016-08-05 at 10 09 36 pm

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -126,7 +126,6 @@ dependencies {
     compile fileTree(dir: "libs", include: ["*.jar"])
     compile "com.android.support:appcompat-v7:23.0.1"
     compile "com.facebook.react:react-native:+"  // From node_modules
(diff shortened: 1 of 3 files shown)


hawkrives:
@drewvolz: I'm going to have to request an expedited review for this.

I'm having so many problems with my environment and `react-native-maps` that I can't make any headway without removing it first.

(Or @elijahverdoorn, but I know Drew's awake.)

drewvolz:
Thanks for the proof. Go ahead and patch this as it sounds as if we all agree the webview is the way to move forrward from our conversation in #11  :shipit:

Issue #23, opened 2016-08-03
Settings


drewvolz:
Working on a basic settings view. It will have a spot for system-wide credentials (username/password for SIS and off-campus directory). An about page and licensing info can be stored here as well.
- [x] Credentials
- [x] Contact
- [x] Legal
- [x] Credits
- [x] Privacy

drewvolz:
Could make use of this if we wanted [react-native-settings-list](https://github.com/evetstech/react-native-settings-list/blob/master/README.md)

elijahverdoorn:
I like the settings idea, and have a sketch of what I want on the about page. I'd be curious about how the package that you linked would function on Android. If it works and looks ok I say use it, if not I'd find another way.

hawkrives:
We're not going to find identical packages for Andoid/iOS – at least, not according to my experience earlier this summer. That might at least let us only build one custom screen, and might inform us building the android screen.

hawkrives:
also – post mockups! Maybe we need a #mockup label.

elijahverdoorn:
My mockup at this point is really just the text... Not worth posting, I'll just make changes to the code. As for the package, I'm ok with using it assuming that we can replicate the patterns on Android. I like the idea of keeping the app feeling native, but I prioritize features over look and feel, so spending too much time re-implementing feels counter-productive to me.

drewvolz:
The settings view should hopefully not be too complicated to design, so perhaps it does not need "react-native-settings-list" as I suggested.

hawkrives:
If it looks that much like a UITableView, I want to use it everywhere!

I've had so much trouble reproducing normal list views.

hawkrives:
Or maybe there's something else that just does nice lists.

drewvolz:
These, maybe? I know there are a plethora of options out there already. How do we pick the best one is the problem. I like what @elijahverdoorn  implemented for the Contacts view.
- [listview](https://facebook.github.io/react-native/docs/listview.html)
- [tableview](https://github.com/aksonov/react-native-tableview)

hawkrives:
https://github.com/APSL/react-native-item-cell maybe

hawkrives:
Tableview maybe. List view just gives you the list behavior. I've been struggling with the iOS styles (see: dictionary)

drewvolz:
Nothing I have found can capture the flexibility that Obj-c/Swift offered with tableviews. This needs more research.

hawkrives:
[...]

PR #30, opened 2016-08-07
First version of the settings menu

drewvolz:
- Added the settings page when the info icon is pressed (maybe should change to gear for settings)
- Added `react-native-tableview-simple` so that we can simply create stylized tables and cells
- Added the credentials fields (username/password) for app-wide login. Currently saves with `react-native-keychain`. Able to add, edit, and delete logins/passwords. At the moment, they save when both username and password are present, and only when 'Done' is hit on the password.
- Added the rows to launch views for support, version, credits, legal, and privacy policy
- Added email function using `react-native-communications` for support to `odt@stolaf.edu`
  settings


hawkrives:
I don't think we necessarily need to change the icon – it's more info than settings right now, no? :P

But I also see the want to change it. I care not.

hawkrives:
(this is part of #23)

drewvolz:
Right. It is a negligible change. Thinking about how to route the user to this view, I remember @elijahverdoorn mentioned he wanted certain portions of the app unlocked to everyone, so that cuts out making the user log-in from the start. So if they enter the SIS portion of the app, do we immediately route them to the Settings view?

We could present a button saying 'Let's get started' in the SIS view, and have that route them to the Settings so that it is hopefully less confusing. Thoughts?

elijahverdoorn:
Everyone should be able to use the app, regardless of if they have a St. Olaf account or not. I am planning on implementing an "intro" screen using this https://github.com/FuYaoDe/react-native-app-intro , what if we had the option to input username and password there (skippable) and also in the settings? If they start the SIS view without having a valid username and password then we can just display a message and link to the settings view for them to enter it.

As for the icon, I'd prefer a gear.

hawkrives:
aao sis

This is something along the lines of what I'm thinking. I'm all for letting anyone use the app and putting as little as possible behind the auth barriers.

I'm thinking something along the lines of,

```
<username>
<password>

<short paragraph about why we need olaf credentials for this part of the app>
```

Once you've logged into one section, we share your credentials to the other sections.

drewvolz:
Note: need to add user data to support email using [react-native-device-info](https://github.com/rebeccahughes/react-native-device-info)
- Manufacturer
- Model
- ID
- System Name
- Device Version
- App Version
- Build Number

PR #33, opened 2016-08-08
Cage menu

elijahverdoorn:
Add the cage menu using the data that I got when I went to campus last week. Note that only the items on the menu are included, the bakery items and the ones in the cooler are not. We might add those later, but I don't have the data right now.

### DIFF
--- a/data/cage-menu.json
+++ b/data/cage-menu.json
@@ -82,7 +82,7 @@
       {
         "name": "Soup & 1/2 Deli Sandwich",
         "price": "5.25"
-      },
+      }
     ]
   },
   {
@@ -104,7 +104,7 @@
       {
         "name": "Caesar Salad & Chicken",
         "price": "5.60"
-      },
+      }
     ]
   },
   {
@@ -322,9 +322,9 @@
       {
         "name": "Vanilla",
         "price": ""
-      },
+      }
     ]
-  }
+  },
   {
     "name": "Breakfast",
     "subtext": "",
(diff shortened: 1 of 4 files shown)


hawkrives:
Should the prices be floats, instead of strings?

elijahverdoorn:
@hawkrives I'd rather keep them as strings, mostly because I'm lazy. The only category that has more than one size right now is the drinks, and denoting the two sizes difference in price was super easy with strings, it would take more code as floats. Since the most logical grouping of items on the menu has nothing to do with the price, I don't see a time that we will ever want to use them for numerical comparison or any kind of math, so for the sake of getting it done I kept them as strings. Could be changed later if we decide to add sorting by price or something like that, but I don't see that being very high priority.

drewvolz:
Should the menu accordion load with the sections expanded automatically? I am trying to see what the most intuitive option is. I feel as if tapping on things in that view initially is not apparent.

elijahverdoorn:
@drewvolz I'm ok either way. I think that the use case is slightly stronger for them closed - less scrolling for the bottom categories - but I'm happy to change it if you feel strongly... (or maybe @hawkrives has an opinion?)

hawkrives:
I'll have to take a look in a bit

elijahverdoorn:
Seeing no further comments, I'm merging this one. If someone feels strongly that it should change, open an issue and we can discuss it.

hawkrives:
I haven't actually looked yet. Now that I've got Travis working, I'll be able to check out master. I'll open an issue if I don't like something 😉

PR #38, opened 2016-08-12
change how config vars are loaded

### LINKED ISSUE #29: Convert key loading to read from a .env file
I'm looking at using https://www.npmjs.com/package/react-native-config.

I'll make a `.env.example`, which has the key names and some instructions, but it'll read from a `.env` file in the root directory. `.env` itself will also be ignored by git.

hawkrives:
Really, this PR is more about a filename change than anything else.

I centralized config loading in `/lib/config.js` - it's up to that file to figure out how to import config stuff. That way, we can experiment with alternative approaches without changing the callsites.

I took `keys.js` and changed it to `.env.js`. I also created a sample `.env.sample.js`, for reference purposes. I left `keys.js` in the .gitignore, so that we wouldn't accidentally commit an existing one.

``` js
// Make a copy of this file as '.env.js'
// Fill in the variables as appropriate
export const GOOGLE_CALENDAR_API_KEY = 'key goes here'
```

The main change is in the names of the keys, so you can't quite just rename `keys.js` to `.env.js`.

---

I experimented with `react-native-config`, but it's a bit wonky. Also it's completely broken for Android currently. Yeah. Fun.

---

Closes #29

### DIFF
--- a/.env.sample.js
+++ b/.env.sample.js
@@ -1,0 +1,3 @@
+// Make a copy of this file as '.env.js'
+// Fill in the variables as appropriate
+export const GOOGLE_CALENDAR_API_KEY = 'key goes here'
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,4 @@ android/keystores/debug.keystore
 
 # KEYS
 keys.js
+.env.js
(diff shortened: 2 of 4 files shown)


elijahverdoorn:
Cool. Approved.

Issue #36, opened 2016-08-11
Contacts view funkiness


drewvolz:
![contacts](https://cloud.githubusercontent.com/assets/5240843/17577194/dcb10028-5f48-11e6-8ee1-61b802433837.png)

drewvolz:
And `react-native-card-view` did not push an update recently.

elijahverdoorn:
I'm seeing the same on android. I can look into it later tonight.

elijahverdoorn:
It has to do with the amount of text in the card. If I shorten the text in the card, it renders fine. I'm not really impressed with that, so I'm just going to make something similar myself. :(

elijahverdoorn:
Finished.

PR #50, opened 2016-08-15
Remove card view, implement it manually

elijahverdoorn:
Fixing bugs the hard way, but it looks about the same.

Closing #36

### DIFF
--- a/data/contact-info.json
+++ b/data/contact-info.json
@@ -1,1 +1,30 @@
-{}
+[
+  {
+    "title":"St. Olaf Public Safety",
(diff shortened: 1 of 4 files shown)


hawkrives:
LGTM. My only comment is, if you merge my local building branch, we don't need the full paths to the images in here.

With that branch, you can just refer to a relative path `images/<folder>/<image>.(jpg|png)` instead of `../../data/images/<folder>/<image>`

elijahverdoorn:
I'm not using the images now anyway, I left them in case I want to add them later.

PR #57, opened 2016-08-15
Update license to give ownership to StoDevX

drewvolz:
Credit is due to the community that works on the app.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2015 Drew Volz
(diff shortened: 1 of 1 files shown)


drewvolz:
I want to make sure that was okay with @elijahverdoorn and @hawkrives to remove my name from it and put StoDevX.

hawkrives:
LGTM.

elijahverdoorn:
I agree with the notion. Since this is also an SGA thing, I may make mention of that in the license, but I agree with the explicit ownership being to this group.

PR #62, opened 2016-08-16
Fixed padding on calendar view

drewvolz:
Removed the margin bottom as we are handling this with padding on the Tabbed view component for all tabbed views.


hawkrives:
LGTM :shipit:

PR #73, opened 2016-08-17
Changed animation config style to Push from Float

drewvolz:
Transitioning from Float to Push.


hawkrives:
This should change the default style to be in line with iOS, as well as returning it to React Native's default value.

LGTM.

elijahverdoorn:
I have no real opinion on this, go for it.

PR #76, opened 2016-08-17
Speed up android builds on Travis

hawkrives:
I enabled a more … selective build matrix on Travis.
- JS builds only need JS tooling, so we use Travis' docker-based containers for fast boot. this trims the runtime for js-tests from around 3-4 minutes to about 45-50 seconds.
- Android builds have a dedicated `android` environment that we can take advantage of. use it. this drops the time from ~20 minutes to ~9 minutes.
- I added the tooling to support running an android emulator on the new builds. however, it takes about 20 minutes to run the android tests on this system because of the emulator. when we decide to build more tests for the emulator, this will be useful.
- OS X / iOS builds still use the same image.

None of this affected Appveyor, whose builds are currently averaging 7 minutes (windows, android-only).

Oh, and the linux image-based builds also start up faster 😉. They usually run about a minute ahead of the OS X-based images.

---
### Before ([build 138](https://travis-ci.org/StoDevX/AAO-React-Native/builds/152887135))
- **JS:** 1 min, 59 sec
- **iOS:** 8 min, 56 sec
- **Android:** 20 min, 25 sec
- **Total:** 31 min, 20 sec
### After ([build 163](https://travis-ci.org/StoDevX/AAO-React-Native/builds/152953913))
- **JS:** 0 min, 48 sec
- **iOS:** 10 min, 44 sec
- **Android:** 9 min, 5 sec
- **Total:** 20 min, 37 sec

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,11 @@ cache:
     - /usr/local/Cellar/.git
 
 install:
+  - echo "Now testing $TEST_TYPE on $TRAVIS_OS_NAME"
+  - echo "Using android emulator $EMULATOR"
+  - echo "Travis branch is $TRAVIS_BRANCH"
+  - echo "Travis branch is in pull request $TRAVIS_PULL+REQUEST"
+
   # work around android build failing - https://github.com/travis-ci/travis-ci/issues/6307
   - if [[ "$TEST_TYPE" = android ]]; then rvm get head; fi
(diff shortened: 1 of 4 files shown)


hawkrives:
Actually merged with ab7b07f205dc04e7cb6abdcfa98ce73ae8e59210

PR #80, opened 2016-08-18
Allow login from Settings view

hawkrives:
Should allow login from the settings screen.

Haven't actually tested yet.


hawkrives:
image

It's ready to merge now, I think.

drewvolz:
Fantastic work integrating a more dynamic set of elements. :shipit:

PR #85, opened 2016-08-19
Center images in story view

drewvolz:
This centers the images on the individual news view.

### DIFF
--- a/views/news/news-item.js
+++ b/views/news/news-item.js
@@ -29,6 +29,8 @@ function NewsItemContents({content}: StoryType) {
         height: auto;
         display: block;
         margin: auto;
+        margin-top: 12px;
+        margin-bottom: 12px;
       }
       p {
         line-height: 1.2em;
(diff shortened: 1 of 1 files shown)


hawkrives:
Could you also add some margin-top and margin-bottom, to put space around above/below the images?

``` css
img {
  margin-top: 12;
  margin-bottom: 12;
}
```

hawkrives:
Just, while we're styling the images ;)

drewvolz:
Done.

hawkrives:
~~LGTM. :shipit: if it passes.~~

Address the one comment I just left, and I'll LGTM it ;)

views/news/news-item.js
```diff
@@ -27,6 +27,10 @@ function NewsItemContents({content}: StoryType) {
       img {
         max-width: 100%;
         height: auto;
+        display: block;
+        margin: auto;
+        margin-top: 12px;
+        margin-bottom: 12px;
```
hawkrives: Could you use a shorthand `margin` prop here? I'm thinking `margin: 12px auto;` instead of the three declarations. 

That way we don't override our initial declaration.

hawkrives:
LGTM. :shipit: if it passes!

PR #87, opened 2016-08-19
Get some financial data

hawkrives:
This is the first step towards loading financial data into the SIS view.

### DIFF
--- a/lib/login.js
+++ b/lib/login.js
@@ -1,4 +1,5 @@
 // @flow
+import { AsyncStorage } from 'react-native'
 import buildFormData from './formdata'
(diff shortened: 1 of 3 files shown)


hawkrives:
> then you can use the new stuff I store in `AsyncStorage` with `getFinancialData` to show fancy stuff

Recommendations for loading the financial data into the SIS view: In `componentWillMount`, load the data from `AsyncStorage`, but also kick off a request to `getFinancialData`, and update `this.setState` after both have finished. 

`AsyncStorage` will be done first, so the numbers will magically update when the loading is done.

drewvolz:
:shipit:

PR #94, opened 2016-08-20
Home tweaks

hawkrives:
- Get "Important Contacts" onto one line on 5s screens
- Found and removed an extraneous layer of view nesting around the buttons


elijahverdoorn:
Cool.

PR #101, opened 2016-08-20
Add ability for buildings to close the next day

hawkrives:
This PR adds the ability to mark builds as closing "nextDay", that is, as closing in the morning.

I switched to using [MomentJS](http://momentjs.com/) for time parsing in BuildingHours, because times are nasty things.

This PR also evaluates times in Central only, rather than wherever your device happens to be. This is a small touch.

In short, if an hour period is marked `nextDay`, a day is added to the closing time.


hawkrives:
This PR also requires that the times to be updated to reflect this new reality. I added a minute to them, so they close at midnight, but I don't recall when the buildings actually close.

drewvolz:
:shipit:

PR #102, opened 2016-08-20
Use local building images

### LINKED ISSUE #52: Using local BuildingImages on Android doesn't work
Current error:

```
Could not get BatchedBridge, make sure your bundle is packaged properly
```

hawkrives:
(Closes #52)

This uses the local copies of the building images.

I have tested and seen it working on my emulators for both iOS and Android.


hawkrives:
Drew says there's a problem with this when scrolling.

drewvolz:
I've noticed that it _sometimes_ shows the list view after you've scrolled, otherwise it is a blank screen. But I do not know whether it will show when I try again. So, I can get it to appear, but it is consistent. Maybe it's me.

drewvolz:
I am noticing the view not displaying until I scroll on Important Contacts screen as well now -- repeatability seems random.

hawkrives:
Ok. This still has the problem about sometimes being blank until you start scrolling, but we're not sure what causes that, and it's not limited to just this view. So. Ignoring that:

iOS screenshot

Android screenshot

This PR reformats the building info screen. It also uses local building images, instead of remote ones. It also fixes a few more bugs in time parsing (they only affected people in different timezones.)

I believe that this is now ready for review.

drewvolz:
:shipit:

elijahverdoorn:
Thanks guys.

PR #106, opened 2016-08-20
Shrink iOS build times

hawkrives:
Drops about a minute by not installing xctool – it's already installed on the Travis images. iOS now builds in about ~8 minutes.


drewvolz:
:shipit:

PR #113, opened 2016-08-20
Added html decoding to Oleville title property

drewvolz:
This takes care of html being rendered in the Latest view as well as the title of the story view.

### DIFF
--- a/views/oleville/index.js
+++ b/views/oleville/index.js
@@ -20,9 +20,8 @@ import NavigatorScreen from '../components/navigator-screen'
 import LoadingView from '../components/loading'
 import LatestView from './latestView'
 import * as c from '../components/colors'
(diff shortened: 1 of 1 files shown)


views/oleville/index.js
```diff
@@ -125,7 +127,7 @@ export default class OlevilleView extends React.Component {
   }
 
   renderRow(data: Object) {
-    let title = data.title.rendered
+    let title = entities.decode(data.title.rendered)
```
hawkrives: Didn't we need to remove the HTML tags?

drewvolz:
It seemed to render fine. I can adjust to remove the tags instead if you'd like?

hawkrives:
LGTM!

PR #114, opened 2016-08-20
Contact view button stylings

drewvolz:
Added an overflow of hidden to the styles. And a bit more margin for good faith.


hawkrives:
LGTM, but I want to look locally before merging

PR #115, opened 2016-08-20
Calendar stylings and html tags

drewvolz:
This makes the view have less html and more consistency.

### DIFF
--- a/views/calendar/calendar.js
+++ b/views/calendar/calendar.js
@@ -144,6 +144,7 @@ let styles = StyleSheet.create({
     backgroundColor: '#ffffff',
   },
   row: {
+    minHeight: 88,
     marginLeft: 10,
     marginRight: 10,
     borderBottomWidth: 1,
--- a/views/calendar/event.js
+++ b/views/calendar/event.js
@@ -7,6 +7,7 @@ import {
 
 import * as c from '../components/colors'
 import padEnd from 'lodash/padEnd'
+import {getText, parseHtml} from '../../lib/html'
 
 let styles = StyleSheet.create({
   itemTitle: {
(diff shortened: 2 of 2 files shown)


views/calendar/event.js
```diff
@@ -53,8 +53,15 @@ function getString(date) {
   return `${month}/${day} ${hours}:${min}`
 }
 
+function stripAndClean(text: string) {
```
hawkrives: What happened to HTMLparser?

drewvolz:
If you are referring to this

```
let Entities = require('html-entities').AllHtmlEntities
const entities = new Entities()
```

...then it won't work. It does not strip out the `<i></i>` tags.

hawkrives:
I'm referring to the parseHtml/getText combo from /lib that I mentioned earlier.

drewvolz:
I missed that completely. I'll make the changes happen with that, then. Thanks.

hawkrives:
I'm confused by one thing. 

We've got two PRs that need to do the same thing, yet the HTML entities decoding only works to remove the tags in one case?

drewvolz:
They both can probably be solved by the HTML parser you have in lib. That was my misunderstanding.

views/calendar/calendar.js
```diff
@@ -144,6 +144,7 @@ let styles = StyleSheet.create({
     backgroundColor: '#ffffff',
   },
   row: {
+    minHeight: 88,
```
hawkrives, line 147: What does `minHeight` do to the title in the one-line case? Does the title stay at the top of the row?

drewvolz:
It keeps it at the top, correct.
screen shot 2016-08-20 at 10 56 40 pm

hawkrives:
LGTM

PR #116, opened 2016-08-20
Webcams styles

drewvolz:
Try to display webcams full size in a scrollview. We still have to deal with them not playing after you play them once, and no thumbnails. We should try to use react-native-video in the hopes of being able to use native HLS M3U8 attributes such as timed tags and thumbnails that are streamed.

### DIFF
--- a/views/streaming/webcams.js
+++ b/views/streaming/webcams.js
@@ -13,7 +13,6 @@ import {
   WebView,
 } from 'react-native'
 import webcamInfo from '../../data/webcams'
-import * as c from '../components/colors'
 
 // const inlineVideo = url => `
 //   <style>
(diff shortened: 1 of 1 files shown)


views/streaming/webcams.js
```diff
@@ -75,14 +76,26 @@ let styles = StyleSheet.create({
...
   video: {
-    // position: 'absolute',
-    // top: 0,
-    // left: 0,
-    // right: 0,
-    // height: 210,
+    top: 0,
+    left: 0,
+    right: 0,
+    height: 210,
```
hawkrives, line 96: Do these styles actually change anything for you?

drewvolz:
The height. Otherwise, the other props can go.

hawkrives:
👍 LGTM.

PR #118, opened 2016-08-21
Add spaces to the app names

hawkrives:
This PR adds spaces to the app names on both iOS and Android.

screen shot 2016-08-21 at 7 06 04 pm


drewvolz:
:shipit:

PR #119, opened 2016-08-21
Android tweaks

hawkrives:
- Make a couple of android tweaks
- Remove the extra padding from tab bars and tab views
- Add back button support
- Make the listview in Courses scroll faster for android


drewvolz:
:shipit:

PR #120, opened 2016-08-21
Show version in Settings

hawkrives:
This uses the `package.json` file as the authoritative version. We will need to ensure that this is continued by whatever packaging system we use to build the final apps.


drewvolz:
:shipit:

PR #121, opened 2016-08-21
Bug fixes

hawkrives:
My time stuff from yesterday was using local times for the buildings, then converting that time to Central time. This evaluates the time zone in Central time, avoiding the issue.

This PR also properly handles multi-line information from courses.


drewvolz:
:shipit:

PR #125, opened 2016-08-21
Confirmation prompt to Important Contacts

drewvolz:
Add a confirmation prompt to check whether the user actually wants to make the phone call.
- We need to make sure the user actually wants to dial the number. This confirms their decision. 
- It also shows them the number.

### DIFF
--- a/views/components/contactCard.js
+++ b/views/components/contactCard.js
@@ -3,6 +3,7 @@ import {
   StyleSheet,
   View,
   Text,
+  Alert,
 } from 'react-native'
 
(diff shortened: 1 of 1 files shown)


elijahverdoorn:
On android it used to just dump the number in the calling app, then you'd have to push the call button. Was it different on iOS? Either way, good change.

drewvolz:
It would immediately open the phone app and make the call.

elijahverdoorn:
I'm glad that we have people testing each platform. Approved.

PR #128, opened 2016-08-21
Change the background color of the text on oleville list view

elijahverdoorn:
The gold text was hard to read.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "react-native-animatable": "0.6.1",
     "react-native-button": "1.6.0",
     "react-native-calendar": "0.6.4",
+    "react-native-cli": "1.0.0",
     "react-native-collapsible": "0.7.0",
     "react-native-communications": "2.1.0",
     "react-native-keychain": "0.3.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -40,6 +40,7 @@
     "react-native-animatable": "0.6.1",
     "react-native-button": "1.6.0",
     "react-native-calendar": "0.6.4",
+    "react-native-cli": "1.0.0",
```
hawkrives: What is this and why did you add it to the package?

elijahverdoorn: This is in response to the issues that I was having building when I pulled from master. This is the windows CLI implementation that you need to build with the command `react-native run-android`. (I think)

hawkrives:
`react-native` itself comes with the cli implementation. That was merged into the main `react-native` package a while ago.

Please remove it?

hawkrives:
Regardless of whether or not we need it, it doesn't really belong in this PR 😃

elijahverdoorn:
Fair. Removing.

hawkrives:
LGTM! :shipit:

Issue #27, opened 2016-08-06
Map not loading on Android


elijahverdoorn:
On my android phone, the map view is not loading. It shows this:
<img src="https://cloud.githubusercontent.com/assets/12042702/17458859/59c1c80a-5be7-11e6-80f1-19cd023450dc.jpg" width="320">

Any ideas? @hawkrives I think that this was your code.

hawkrives:
It is. I'll fire it up and take a look at it in a bit.

hawkrives:
![screen shot 2016-08-12 at 2 58 33 pm](https://cloud.githubusercontent.com/assets/464441/17613954/418816c6-609d-11e6-9481-40da3ec47243.png)

There appears to be a JS error in their code somewhere, although the map works in the stock browser.

hawkrives:
Or not. I haven't been able to reproduce that error yet.

Next up, I'm going to compare the inspectors from the native browser and the webview. But I need to put this away for now.

hawkrives:
Current analysis says that they've got some code that resizes the canvas they draw to, and that code is setting said canvas to `0` height in our webview.

elijahverdoorn:
Would injecting our own resize code fix the issue?

hawkrives:
Maybe, but I can't figure out where their stuff is getting called.

Actually. You gave me an idea.

If I trigger a layout with `document.body.style.height = "500px"`, I get this:

<img width="543" alt="screen shot 2016-08-19 at 11 41 13 pm" src="https://cloud.githubusercontent.com/assets/464441/17813464/6d7951a4-6666-11e6-816e-34dba917af0e.png">

Which gets us part of the way there.

hawkrives:
<img width="543" alt="screen shot 2016-08-19 at 11 43 42 pm" src="https://cloud.githubusercontent.com/assets/464441/17813559/c86b5580-6666-11e6-868a-a5529912b93d.png">

Well hmm.

`document.querySelector('#map-size').style.height = '500px'`

hawkrives:
oh god. they're drawing _multiple `<canvas>`s_

![screen shot 2016-08-19 at 11 45 52 pm](https://cloud.githubusercontent.com/assets/464441/17813632/1691797e-6667-11e6-850c-48cebffeb484.png)

hawkrives:
slash multiple image tiles

[...]

PR #130, opened 2016-08-22
Hack around the android map issue

elijahverdoorn:
Someday we'll fix this, but until then we need something that works so
that we can ship a first version.


elijahverdoorn:
Not meant to close it, but referencing #27

hawkrives:
I like it. If the builds pass, :shipit:!

hawkrives:
I'll move #27 to the `next` milestone then.

elijahverdoorn:
It's the same linter error that @hawkrives agreed to fix, so I'm merging.

hawkrives:
ok but you could fix it sooner than I can :P

just add `schedule: React.PropTypes.any,` to propTypes in busLine.js

hawkrives:
i'll update the proptype to actual typing info later

hawkrives:
oh fine I'll add it

hawkrives:
see #132

PR #137, opened 2016-08-22
Add a webview for KSTO

elijahverdoorn:
Better than nothing! Do more with this once I communicate with KSTO management.


elijahverdoorn:
Thanks for the update @drewvolz. I like that better. Merging assuming that you're ok with the change including your commit.

drewvolz:
Thanks!

PR #144, opened 2016-08-23
Fixes and tweaks to Transportation

hawkrives:
I did these last night. I was trying to solve an extra padding bug that still escapes me… but I think these are still worthwhile.

Mostly self-contained per-commit.


drewvolz:
:shipit:

PR #151, opened 2016-08-23
Android header

elijahverdoorn:
Made a few changes to better fit Android styles.
Removed edit link on home screen for all platforms.

### DIFF
--- a/app.js
+++ b/app.js
@@ -111,7 +111,7 @@ const styles = StyleSheet.create({
   },
   titleText: {
     color: 'white',
-    fontSize: 16,
+    fontSize: Platform.OS === 'ios' ? 16 : 20,
     fontWeight: 'bold',
     marginTop: 14,
   },
--- a/app.js
+++ b/app.js
@@ -77,7 +77,7 @@ const styles = StyleSheet.create({
     backgroundColor: c.iosLightBackground,
   },
   navigationBar: {
-    backgroundColor: c.olevilleGold,
+    backgroundColor: c.gold,
     paddingTop: 22,
     flexDirection: 'row',
     justifyContent: 'center',
--- a/app.js
+++ b/app.js
@@ -4,7 +4,15 @@
  */
 
 import React from 'react'
-import {Navigator, BackAndroid, StyleSheet, TouchableOpacity, Text, Platform} from 'react-native'
+import {
+  Navigator,
+  BackAndroid,
+  StyleSheet,
+  TouchableOpacity,
+  Text,
+  Platform,
+  View,
+} from 'react-native'
 
 import AboutView from './views/about'
 import CalendarView from './views/calendar'
@@ -143,7 +151,7 @@ function LeftButton(route, navigator, index, navState) {
           style={[styles.backButton, {marginLeft: 10, marginTop: 14}]}
           onPress={() => navigator.pop()}
         >
-          <Text style={styles.backButtonText}>Close</Text>
+          <Text style={styles.backButtonText}>Close</Text>}
         </TouchableOpacity>
       )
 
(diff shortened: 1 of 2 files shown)


app.js
```diff
@@ -155,31 +163,33 @@ function LeftButton(route, navigator, index, navState) {
...
-          <Text style={styles.backButtonText}>{backTitle}</Text>
-        </TouchableOpacity>
-      )
+      if (Platform.OS === 'android') {
+        return (
+          <TouchableOpacity
+            style={styles.backButton}
+            onPress={() => navigator.pop()}
+          >
+              <Icon style={styles.backButtonIcon} name='md-arrow-back' />
```
elijahverdoorn, line 172: I think that this is the right icon pack, but it's giving me this: 
image
Not what I expect. Insights?

app.js
```diff
@@ -143,7 +151,7 @@ function LeftButton(route, navigator, index, navState) {
           style={[styles.backButton, {marginLeft: 10, marginTop: 14}]}
           onPress={() => navigator.pop()}
         >
-          <Text style={styles.backButtonText}>Close</Text>
+          <Text style={styles.backButtonText}>Close</Text>}
```
hawkrives: Why is there an unmatched closing brace after this line?

app.js
hawkrives: That looks right. I'll take a look in a bit.

app.js
```diff
@@ -155,31 +163,33 @@ function LeftButton(route, navigator, index, navState) {
...
-      return (
-        <TouchableOpacity
-          style={styles.backButton}
-          onPress={() => navigator.pop()}
-        >
-          <Icon style={styles.backButtonIcon} name='ios-arrow-back' />
-          <Text style={styles.backButtonText}>{backTitle}</Text>
-        </TouchableOpacity>
-      )
+      if (Platform.OS === 'android') {
```
hawkrives: Could you change the `if` to only operate on the icon, instead of swapping out the entire touchableOpacity?

Like, `name={Platform.OS === 'android' ? 'md-arrow-back' : 'ios-arrow-back'}`. 

Thanks.

hawkrives: Or at least the inner block of the TouchableOpacity.

app.js
```diff
@@ -155,31 +163,33 @@ function LeftButton(route, navigator, index, navState) {
...
+        return (
+          <TouchableOpacity
+            style={styles.backButton}
+            onPress={() => navigator.pop()}
+          >
+              <Icon style={styles.backButtonIcon} name='md-arrow-back' />
+          </TouchableOpacity>
```
hawkrives, line 173: Don't Android apps (thinking Gmail) usually have some text next to the button? Phone is currently dead so can't check.

hawkrives: My bad. They don't.

hawkrives:
You could also experiment with adding the `backgroundColor` prop to the StatusBar element in HomeView – I don't know if it automatically darkens the background color, but it would extend it into the status bar like good android apps do.

app.js
hawkrives: If you use a diff. icon, say, `md-body`, does it still give you the … ?

app.js
elijahverdoorn: Typo. I'll remove it.

app.js
elijahverdoorn: Yes.

hawkrives: Weird. Really weird.

views/home.js
```diff
@@ -51,6 +51,7 @@ export default function HomePageScene({navigator, route}: ScenePropsType) {
     >
       <StatusBar
         barStyle='light-content'
+        backgroundColor={c.gold}
```
hawkrives, line 54: Does it automatically do the darkening thing?

elijahverdoorn: Yeah. 
image

hawkrives: Nice. Last ~~question~~ request: I used `olevilleGold` for the navbar; I think you used `gold` for the status bar? Could you make that consistent, please?

elijahverdoorn: If I do that, It doesn't do the darkening.

hawkrives: … ok that's weird. Oh well. It looks fine.

app.js
elijahverdoorn: Weirder: It renders the icon on my device, but it's unaligned.
screenshot_2016-08-23-23-01-22

hawkrives:
Do you want to move the Settings gear down a bit, too, or shall I do that later?

elijahverdoorn:
Just made a comment about that. I'll do it quick.

hawkrives:
For v+1, I'm thinking about moving to a drawer (android-only) that goes between Home and Settings.

views/home.js
```diff
@@ -51,7 +51,7 @@ export default function HomePageScene({navigator, route}: ScenePropsType) {
     >
       <StatusBar
         barStyle='light-content'
-        backgroundColor={c.gold}
+        backgroundColor={c.olevilleGold}
```
hawkrives: I thought this didn't do the shading?

elijahverdoorn:
I've had the same thought, but would like to have more than that in such a menu. I don't think that having just two options in those menus looks good. Maybe we could move the entire settings menu into the drawer? There's not that much stuff in there now. (Or maybe we'll have more content when we get that far)

hawkrives:
Or we could move About into there or something, yeah. Two is a bit few.

You can see how it would look with http://pokedex.org/ – it's got a drawer with two items (and a title).

views/home.js
elijahverdoorn: Gah. That's what I get for testing on two devices and not refreshing both. Fixing.

hawkrives:
Also it would be really weird to have input fields in a drawer?

app.js
hawkrives: Can you move it down, then? (did you already fix that?)

It's really odd that it doesn't show up in the sim. I'll dig into that a bit tonight.

hawkrives:
Alright. Do we have everything wrapped up in here?

elijahverdoorn:
:shipit:

hawkrives:
LGTM.

PR #159, opened 2016-08-24
Add exception for KSTO on iOS

### LINKED ISSUE #156: Add exception for KSTO website to Info.plist
Otherwise ATS blocks it.

elijahverdoorn:
I've never done this before, thought that I should take a crack at it. Let me know if there is something that I've messed up.

Closes #156.


hawkrives:
Only that we don't let Xcode edit that file… look at that diff.

hawkrives:
I always edit it in Sublime because of that.

hawkrives:
Actually.

I didn't realize that KSTO was on pages.stolaf.

Everything there also runs over HTTPS, so all we need to do it change the link to `https://pages.stolaf.edu/ksto/listen/` and it'll work.

elijahverdoorn:
Funny, it didn't look like that when I ran `git diff`. It only showed a few lines different. Can we keep it or should I revert and add the diff?

hawkrives:
Did it look like https://github.com/StoDevX/AAO-React-Native/pull/159/files?w=0 ?

hawkrives:
I'd prefer to switch to loading over HTTPS, since that's doesn't require the exception.

If we do continue with the exception, we should switch to `NSTemporaryExceptionAllowsInsecureHTTPLoads` instead of `NSThirdPartyExceptionAllowsInsecureHTTPLoads`, since "we" control that domain.

elijahverdoorn:
@hawkrives It did look like that, yeah.

hawkrives:
Xcode has a habit of rearranging that file, in addition to editing the whitespace. I think there's a PR in here somewhere where Drew's copy, like, flipped everything.

Why? Because Xcode.

elijahverdoorn:
I'd rather switch to HTTPS. Closing this PR and opening another for that soon.

hawkrives:
Kk

PR #161, opened 2016-08-24
Add iOS notifications

elijahverdoorn:
This PR adds notification support on iOS. There's a lot of files in the PR that are added by the library that I added from OneSignal to make this happen, but all the changes that we care about are in AppDelegate.m, AppDelegate.h, and index.ios.js on the javascript side. All the rest of it is just config changes to get the new library in there.

PR #164, opened 2016-08-24
Make Travis work with Cocoapods

hawkrives:
Part of #161.

Travis needs to be updated to work with the cocoapods from iOS notifications.
- [x] Use the `xcworkspace` file.
- [x] ???
- [x] Build successfully
- Profit

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -144,7 +144,7 @@ script:
   - if [[ "$TEST_TYPE" == "js" ]]; then npm run test:js; fi
 
   # iOS-only tests - building and specs
-  - if [[ "$TEST_TYPE" == "ios" ]]; then npm run test:ios; fi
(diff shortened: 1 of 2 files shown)


drewvolz:
Looking into this, I see two helpful things. I will try this much later today.
- [One](https://github.com/travis-ci/travis-ci/issues/3001)
- [Two](https://github.com/facebook/xctool/issues/415#issuecomment-62013753)

hawkrives:
I'll try switching to xcodebuild, but I switched away from that because of testing problems, IIRC.

hawkrives:
@elijahverdoorn: I'm currently seeing if just having travis `build`, instead of `test`, the iOS app will allow it to pass.

However, when we merge the push notifications PR, I won't be able to run it locally anymore – it requires that I add the Apple Developer account that's linked with the build identifier. So we'll need to figure something out about that.

hawkrives:
Well, it passed.

Issue #147, opened 2016-08-23
Strange padding with ListView and TabbarIOS


drewvolz:
- [ ] Transportation/Transit
- [ ] Webcams

We've noticed that our solution of bumping up the margin/padding on the ListView seems to backfire in two known-instances. The scrollbar and content do not agree. 

We thought we could fix it using [this explanation](url) which suggests using this to fix the ListView

`<ListView
  contentInset={{bottom:49}}
  automaticallyAdjustContentInsets={false}
`

But that did not work. It's not above the bottom, but it's at the bottom. And adding padding brings us right back where we started.

hawkrives:
Closed by #168.

PR #168, opened 2016-08-25
Tabbar fix

drewvolz:
Fixed the tabbar content overlap, and adds consistent styles to other transport options.

This fixes two stylistic issues involving views on iOS.


hawkrives:
Travis is struggling right now, but I tested locally and it works. It works so well 😭!!!

drewvolz:
closes #147

PR #169, opened 2016-08-25
Begin support for Xcode 8 (and fix iOS launch images)

hawkrives:
This PR fixes the backwards-compatible changes that Xcode 8 makes to the project file.

It also gives in to the inevitable and re-indents Info.plist according to Xcode's style. There's a method to the madness, though: it's sorted alphabetically by key, and indented with tabs. so it should be easy enough to keep around.

It also enables the launchscreens that we've had forever and have never used. I don't know how to remove the Android equivalent, but it's probably not hard.

If this passes Travis, I'm going to merge it, as I'm categorizing it as "cleanups".


hawkrives:
I'd like to merge this – I can rebase the changes for the push notifications branch if you'd like.

elijahverdoorn:
I'm going to get new launch images for the next version (I want them to match the icon) but for this version it's too late - our graphic artist is overworked right now. The config changes look fine, and I'd rather have the old launch screen than the default. Approved.

PR #172, opened 2016-08-25
Tweak listviews

hawkrives:
This adds refreshing support to all of our listviews.

It also fixes the blank-screen bug in buildingHours, I believe.

Squash and merge this one, please, rather than just merging.

### DIFF
--- a/views/building-hours/index.js
+++ b/views/building-hours/index.js
@@ -5,9 +5,10 @@
  */
 
 import React from 'react'
-import {ListView} from 'react-native'
+import {View, Text, ListView, RefreshControl} from 'react-native'
 import BuildingView from './building'
 
+import delay from 'delay'
 import type {BuildingInfoType} from './types'
 import hoursData from '../../data/building-hours.json'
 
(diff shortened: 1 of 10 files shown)


views/building-hours/index.js
```diff
@@ -54,6 +75,13 @@ export default class BuildingHoursView extends React.Component {
         renderRow={this._renderRow.bind(this)}
         pageSize={4}
         initialListSize={6}
+        removeClippedSubviews={false}  // remove after https://github.com/facebook/react-native/issues/8607#issuecomment-241715202
```
drewvolz, line 78: _Not an issue but a suggestion for later_

We will want to add this to every list view temporarily, perhaps? It has happened on more than one list view (Important Contacts for example). 

Come to think of it, we should implement our own wrapper around ListView to handle our customization. We would need a `prop` to tell if this were to exist inside of a Tabbar or not in the case of the extra bottom inset, but it should be doable.

That can come in a later PR. I am okay with this for now. Nice find!

drewvolz:
Found an issue with the `componentWillMount` not calling `refresh` but instead the child method still. I am fixing this one.

views/building-hours/index.js
hawkrives: Can you make a list of all the listviews that you've seen this on? Are there any you haven't?

I should add this prop to all the affected ones now, as this PR.

hawkrives:
I'm going to hold off on merging this until tonight – I trust drew that he had a reason for making those changes, but I don't understand it fully and I want to take another peek tonight.

hawkrives:
Before Drew's changes:


After Drew's changes:


I like them enough to keep them (for now, at least.)

hawkrives:
Good work, @drewvolz!

drewvolz:
I made 3 one-liners, so the real credit goes to you! Thank you for adding the refreshables to the lists, @hawkrives

PR #174, opened 2016-08-26
Add SIS credential error handling

elijahverdoorn:
For the case where a user hasn't entered any credentials for the SIS in settings.

Add a button that redirects them to settings, with an error message.

### DIFF
--- a/views/sis/index.js
+++ b/views/sis/index.js
@@ -15,7 +15,8 @@ import {
 import TabbedView from '../components/tabbed-view'
 import Button from 'react-native-button'
 import tabs from './tabs'
-import {loadLoginCredentials} from '../../lib/login'
+import LoadingView from '../components/loading'
+import {isLoggedIn} from '../../lib/login'
 import * as c from '../components/colors'
 const styles = StyleSheet.create({
   container: {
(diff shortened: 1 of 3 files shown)


views/sis/index.js
```diff
@@ -5,20 +5,66 @@
...
-  return <TabbedView style={styles.container} tabs={tabs} />
+export default function SISView({navigator, route}: {navigator: typeof Navigator, route: Object}) {
+  // Check if the user has a valid username and password
+  let {username, password} = loadLoginCredentials()
+  if (username && password) {
+    return <TabbedView style={styles.container} tabs={tabs} />
+  } else {
+    return (
+      <View style={styles.error}>
+        <Text>Sorry, we couldn't find your login credentials. Did you set them up in the settings?</Text>
```
elijahverdoorn, line 78: This is really weird - I'm not supposed to escape the `'`, am I? It works fine for me this way.

hawkrives: No, that's just Github's highlighting. It's fine.

views/sis/index.js
```diff
@@ -5,20 +5,66 @@
...
+  Text,
+  View,
+  Navigator,
+} from 'react-native'
 
 import TabbedView from '../components/tabbed-view'
+import Button from 'react-native-button'
 import tabs from './tabs'
-
+import {loadLoginCredentials} from '../../lib/login'
```
hawkrives: loadLoginCredentials is an asynchronous function, right?

So it'll return a promise.

views/sis/index.js
```diff
@@ -5,20 +5,66 @@
...
+    borderRadius: 6,
+    overflow: 'hidden',
+  },
 })
 
-export default function SISView() {
-  return <TabbedView style={styles.container} tabs={tabs} />
+export default function SISView({navigator, route}: {navigator: typeof Navigator, route: Object}) {
+  // Check if the user has a valid username and password
+  let {username, password} = loadLoginCredentials()
```
hawkrives: Which means that username and password will always be `undefined` here.

hawkrives: We'll have to rebuild this as a class component, with username and password in state, then fetch in `componentWillMount` and render the views differently based on those state variables.

hawkrives: We might also want to add a `loaded` state var, so we can show the "loading" screen until it's loaded the credentials – shouldn't take long, but better to be prepared than to have someone really confused by the flashing of the "please log in" screen.

elijahverdoorn: You're right, that sounds better. I'll fix it before I cut the release.

hawkrives:
I'll just need to test this tonight, but it's what I was thinking about.

hawkrives:
Alright. Downside to the current approach is that when you come back from Settings, the view doesn't update automatically.

I'm not sure how to solve that in this context; normally, this sort of data would be stored in a higher-level data store such as Redux, but I've never gotten around to setting that up.

I'm voting that we take this for _now_, and try to figure something out\* between rc1 and release.

*such as using redux, which will be pretty simple if we limit it to only storing credentials and login status.

hawkrives:
I'm going to merge this as soon as the builds pass.

PR #181, opened 2016-08-29
Allow for iPads

elijahverdoorn:
It's better to have the app scaled up than to use an iPhone app on the iPad.


hawkrives:
Drew and I will need to test this before merging. We will need to ensure that there are no remaining screen size assumptions – we've found at least four so far, only noticeable when switching screen sizes.

say, on a 5s when the screen was designed on a 6+.

drewvolz:
The justified text might be something we want to deal with.

hawkrives:
Ooh, good catch. Max-width on the text?

elijahverdoorn:
This is something that I'm not expecting for v2.0.0, just something that I've been trying. Thanks for finding the issues, I'd be happy to fix them if you post them here.

hawkrives:
Oh, ok. Yeah, we'll do an audit after v2 is released.

elijahverdoorn:
Closing this since it's old and we're not really ready. Another effort will be made in the future.

drewvolz (approved):
Let’s explore making it look good on iPad!

hawkrives (approved):
@elijahverdoorn I've decided that I agree with your original sentiment: it's nice to let it scale up. 

I've also convinced Drew that this is the first step to doing anything with iPad. 

As such, let's merge this baby!

PR #184, opened 2016-08-30
Oauth stuff

hawkrives:
I can't do any more tonight. This is what I've got so far, but _both_ simulators are fighting me. iOS says it can't find CFBundleIdentifier, and Android is throwing an invariant of "please link the damned library with `rnpm link`" (paraphrased).

But this might work? It's a copy of https://github.com/ryanmcdermott/react-native-login/tree/master/ReactNativeLogin, and if I understand correctly, the cookies are cached by the underlying network layer, not the webview, so once we get a login _here_ it should just work for the other `fetch` requests. (see https://github.com/facebook/react-native/issues/1274#issuecomment-102221311)

One of you is going to have to take this on, I'm afraid.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -135,7 +135,6 @@ android {
 }
 
 dependencies {
-    compile project(':react-native-cookies')
     compile project(':react-native-vector-icons')
     compile project(':react-native-keychain')
     compile project(':react-native-android-tablayout')
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -135,6 +135,7 @@ android {
 }
 
 dependencies {
+    compile project(':react-native-cookies')
     compile project(':react-native-vector-icons')
     compile project(':react-native-keychain')
     compile project(':react-native-android-tablayout')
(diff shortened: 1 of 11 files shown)


drewvolz:
Speaking from the iOS side of things, I see the same error on iOS if I try building and running with `npm run ios` (without pulling in these changes). However, if I run with Xcode, all is well. That is a band-aid and it does not fix the problem, but it is a way that we can still build and run on iOS at the very least.

elijahverdoorn:
This is the cause of the `rnpm link` issues: https://github.com/joeferraro/react-native-cookies/issues/37

hawkrives:
I used react-native link first… let's try removing the commit that added them, then installing the package, then react-native link. 

And make sure not to commit the Info.plist changes. React Native and Xcode have different ideas about how it should be formatted.

hawkrives:
Can I merge this to `master`?

drewvolz:
I found one more condition.
1. Login via settings
2. Logout via settings
3. Try to load the SIS tab. It loads, and has a cached token. Oops.
4. Pull down to load the financials view, or load the courses
5. Crash/red screen, and errors/warnings

Financials
screen shot 2016-08-31 at 8 28 39 am

Courses
screen shot 2016-08-31 at 8 31 27 am

hawkrives:
Fixed in the latest commit.

drewvolz:
:shipit:

PR #210, opened 2016-09-05
Fix the iOS navbar shadow

hawkrives:
It was close before, but not quite there. (Next up, Android's drop shadow 😜.)
###### Before

screen shot 2016-09-05 at 9 44 04 pm
###### After

screen shot 2016-09-05 at 9 43 46 pm
###### Before

screen shot 2016-09-05 at 9 44 29 pm
###### After

screen shot 2016-09-05 at 9 44 20 pm

PR #211, opened 2016-09-05
Make Android nicer

hawkrives:
(An ongoing project.)

This PR includes the changes from #210 … and implements them in a different way.

It also makes a bunch of tweaks the the base Android experience, to make it feel more native.

This PR focus mostly on the home screen and navbar. I'm finally happy with the navbar on both iOS and Android 😄.

This PR also makes a few tweaks to the Courses view.
## Before / After

home
settings
courses


elijahverdoorn:
Nice.

Issue #143, opened 2016-08-23
Add dropdown of building hours


elijahverdoorn:
It might be good to show the dropdown of the hours when clicking on the items, so that we can help with the "I missed it, when is it open next?" problem.

hawkrives:
Closed by #225

PR #225, opened 2016-09-10
Revamp building hours

mattk410:
This adds a collapsible view of building hours.

Issue #143

simulator screen shot sep 10 2016 11 51 34 am


views/building-hours/building.js
```diff
@@ -16,75 +15,106 @@ import * as c from '../components/colors'
 import type {BuildingInfoType} from './types'
 import {isBuildingOpen} from './is-building-open'
 import {formatBuildingHours} from './format-building-hours'
+import CollapsibleBlock from '../components/collapsibleBlock'
+import {allBuildingHours} from './all-building-hours.js'
+import Dimensions from 'Dimensions'
```
hawkrives: I think this also exists as an import from `"react-native"` itself; could you import from there, please?

views/building-hours/all-building-hours.js
```diff
@@ -0,0 +1,29 @@
...
+} from 'react-native'
+
+import moment from 'moment'
+
+export function allBuildingHours(info, style) {
+  let dayTimes = []
+  let hoursString = ''
+  for (let i = 0; i < 7; i++) {
+    let day = moment().add(i, 'days').format('dddd')
+    let d = moment().add(i, 'days').format('ddd')
```
hawkrives: Two notes here: one, could you create one moment and query it twice, instead of creating two moments? I have an irrational fear of the two moments getting created across a day divide so they don't match… 

The other note is, could you use moment-timezone set to … I think we use America/Winnipeg? Otherwise is used the device's timezone, which is fine but the rest of the building hours stuff uses the explicit America/Winnipeg zone. I think it'll be really confused when I try it in Japan if half of it's in JST and half is in CST…

hawkrives:
Thanks Matt!

I'll do another look at it in the morning. 

Congrats on your first feature!

elijahverdoorn:
Thanks for the PR, Matt. This is a good feature! I'll review it soon.

views/building-hours/all-building-hours.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+import {
+  Text,
+} from 'react-native'
+
+import moment from 'moment'
+
+export function allBuildingHours(info, style) {
+  let dayTimes = []
+  let hoursString = ''
```
hawkrives: Why are you defining `hoursString` all the way out here? Could you move it a bit closer to where it's used, or at least to inside the loop? Thx.

views/components/collapsibleBlock.js
```diff
@@ -0,0 +1,59 @@
...
+  static propTypes = {
+    borderColor: React.PropTypes.string.isRequired,
+    children: React.PropTypes.array.isRequired,
+  }
+
+  state = {
+    isCollapsed: true,
+  }
+
+  style() {
```
elijahverdoorn: This should be in a stylesheet, as it is in the rest of the project.

views/building-hours/building.js
```diff
@@ -16,75 +15,106 @@ import * as c from '../components/colors'
...
 
 type PropsType = {
   info: BuildingInfoType,
   image: number,
   name: string,
 };
+
+
+
+let style = function(){
```
hawkrives: Why did you make style a function? If it's so that `getContainerWidth` is updated when the screen moves, I'd recommend moving just _that_ part of the style into an inline style in the `render` function.

views/components/collapsibleBlock.js
```diff
@@ -0,0 +1,59 @@
...
+        alignSelf: 'center',
+        backgroundColor: this.statusColor(),
+        
+      },
+      collapsedContainer: {
+      },
+    }
+  }
+
+  statusColor(){
```
elijahverdoorn: This function is redundant, it's clearer to just use `this.props.borderColor` wherever you need it.

views/building-hours/building.js
```diff
@@ -16,75 +15,106 @@ import * as c from '../components/colors'
...
+    contents: {
+      paddingBottom: 10,
+      paddingTop: 5,
+      width: getContainerWidth(),
+    },
+  }
+}
+
+let getBorderColor = function(){
+  borderColor = borderColors[openStatus]
```
hawkrives: I don't like global variables. Could you move this discovery process back into the BuildingView function, please?

views/components/collapsibleBlock.js
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  style() {
+    return {
+      container: {
+        marginBottom: 0,
+      },
+      collapsedContent: {
+        alignSelf: 'center',
+        backgroundColor: this.statusColor(),
```
hawkrives: Why don't you just use `this.props.borderColor` here, instead of calling a function?

views/building-hours/building.js
```diff
@@ -16,75 +15,106 @@ import * as c from '../components/colors'
...
-          <Text style={styles.status}>{hours}</Text>
+        resizeMode='cover'>
+        <View style={style().inner}>
+          <Text style={style().name}>{name}</Text>
+          <Text style={style().status}>{hours}</Text>
         </View>
       </Image>
     </View>
+
+    <View style={style().contents}>
```
elijahverdoorn: I think that for consistency we should keep the styling in the `StyleSheet.create()` syntax. (unless there's a good reason to change the entire codebase to use this style that I don't know about!)

views/building-hours/building.js
```diff
@@ -16,75 +15,106 @@ import * as c from '../components/colors'
...
-    color: c.white,
-    fontSize: 14,
-    paddingTop: 2,
-    paddingBottom: 2,
-    textAlign: 'center',
-    backgroundColor: 'rgba(0, 0, 0, 0.45)',
-  },
-})
+
+let openStatus
```
elijahverdoorn: Why is this defined out here? Can it be defined within the function that it is used in?

views/components/collapsibleBlock.js
hawkrives: The advantage to `StyleSheet.create` is that RN can process the stylesheet once, then reference a cache to load the styles in the future, rather than needing to build the entire style tree every time. Could you extract `this.style` into an external `styles` variable created with `StyleSheet.create`, then reference it in the `render` method, please?

That will require that you use the array form of `style={[styles.collapsedContent, {backgroundColor: this.props.borderColor}]}` for the `collapsedContent` block.

views/building-hours/building.js
```diff
@@ -16,75 +15,106 @@ import * as c from '../components/colors'
...
+let borderColor
 
 let borderColors = {
-  'Open': styles.buildingOpen,
-  'Almost Closed': styles.buildingAlmostClosed,
-  'Closed': styles.buildingClosed,
+  'Open': c.pastelGreen,
```
elijahverdoorn: Just a question: should the color be changed now that it is being used for the background of the text? It looked ok in small amounts on the borders, but I wonder if we want a different shade for the backgrounds. Opinions?

views/components/collapsibleBlock.js
```diff
@@ -0,0 +1,59 @@
...
+} from 'react-native'
+
+import Collapsible from 'react-native-collapsible'
+
+
+export default class CollapsibleBlock extends React.Component {
+
+  static propTypes = {
+    borderColor: React.PropTypes.string.isRequired,
+    children: React.PropTypes.array.isRequired,
```
hawkrives, line 32: `children` is better typed as `React.PropTypes.node`, as `this.props.children` is not in an array [if there's only one child](https://facebook.github.io/react/tips/children-props-type.html). `node` covers anything React can render, so elements, text, and arrays of the above.

views/building-hours/building.js
mattk410: I agree. I changed the color scheme to match the GitHub red/green. I think it looks better with the lighter colors.

hawkrives:
Better! I've got a few comments to make… thanks for sticking with us!

views/building-hours/all-building-hours.js
```diff
@@ -0,0 +1,35 @@
...
+    if (timesArray) {
+      let open = moment(timesArray[0], 'H:mm').format('h:mm a')
+      let close = moment(timesArray[1], 'H:mm').format('h:mm a')
+      hoursString = open + ' - ' + close
+    } else {
+      hoursString = 'Closed'
+    }
+
+    let dayString = day + ': ' + hoursString
+    dayTimes.push(<Text key={d} style={style}>{dayString}</Text>)
```
hawkrives, line 32: Where does `style` come from here?

views/building-hours/all-building-hours.js
```diff
@@ -0,0 +1,35 @@
...
+export function allBuildingHours(info, style) {
+  let dayTimes = []
+  let current = moment().tz(CENTRAL_TZ)
+
+  for (let i = 0; i < 7; i++) {
+    let hoursString = ''
+    let day = current.format('dddd')
+    let d = current.format('ddd')
+
+    current = current.add(1, 'days')
```
hawkrives: Why are we adding a day to `current` here when it doesn't appear to be used again?

views/building-hours/all-building-hours.js
```diff
@@ -0,0 +1,35 @@
...
+
+    let timesArray = info.times.hours[d]
+
+    if (timesArray) {
+      let open = moment(timesArray[0], 'H:mm').format('h:mm a')
+      let close = moment(timesArray[1], 'H:mm').format('h:mm a')
```
hawkrives: Don't forget to instantiate these with the final `tz` parameter, too. See https://github.com/StoDevX/AAO-React-Native/blob/master/views/building-hours/get-building-hours.js#L20 for a reference?

views/building-hours/all-building-hours.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react'
+import {
+  Text,
+} from 'react-native'
+
+import moment from 'moment-timezone'
+const CENTRAL_TZ = 'America/Winnipeg'
```
hawkrives, line 7: :+1:

views/building-hours/building.js
```diff
@@ -6,86 +6,101 @@
...
+
+  type PropsType = {
+    info: BuildingInfoType,
+    image: number,
+    name: string,
+  };
+
+  let openStatus = isBuildingOpen(info)
+
+  let styles = StyleSheet.create({
```
hawkrives: I know I said I didn't really like global variables… but stylesheets are kinda an exception? 

Let me try to explain: StyleSheet.create is interesting. Once it's been created, a reference to something like `styles.container` doesn't return an object; it returns a number. That number is a reference into the global stylesheet cache-thing. It's good practice to extract as many styles into a stylesheet as possible, so that they can be created once and just referenced after that. 

Now, that obviously doesn't work if something is dependent on an outside variable – for instance, `borderColor`. That's fine– you just switch to an inline style, but only for the properties that need it.

views/building-hours/building.js
```diff
@@ -6,86 +6,101 @@
...
+    container: {
+      flex: 1,
+      justifyContent: 'center',
+      borderLeftWidth: 5,
+      borderRightWidth: 5,
+      marginTop: 10,
+      height: 100,
+      borderColor: borderColors[openStatus],
```
hawkrives: So everything _but_ this line would be extracted into a global `styles` object in this file.

views/building-hours/building.js
```diff
@@ -6,86 +6,101 @@
...
+      backgroundColor: 'rgba(0, 0, 0, 0.45)',
+    },
+    hoursText: {
+      textAlign: 'center',
+      fontSize: 13,
+    },
+    contents: {
+      paddingBottom: 10,
+      paddingTop: 5,
+      width: Dimensions.get('window').width,
```
hawkrives: Oh, and excluding this, too, since it needs to be able to respond to size changes.

views/building-hours/building.js
```diff
@@ -6,86 +6,101 @@
...
-      >
-        <View style={styles.inner}>
-          <Text style={styles.name}>{name}</Text>
-          <Text style={styles.status}>{hours}</Text>
-        </View>
-      </Image>
-    </View>
-  )
+    <CollapsibleBlock style={styles} backgroundColor={borderColors[openStatus]}>
+      <View style={[styles.container]}>
```
hawkrives: This would become `style={[styles.container, {borderColor: borderColors[openStatus]}]}` – so that the borderColor is the only thing set by the inline style.

views/building-hours/building.js
```diff
@@ -6,86 +6,101 @@
...
+          style={{width: undefined, height: 100}}
+          resizeMode='cover'>
+          <View style={styles.inner}>
+            <Text style={styles.name}>{name}</Text>
+            <Text style={styles.status}>{hours}</Text>
+          </View>
+        </Image>
+      </View>
+
+      <View style={styles.contents}>
```
hawkrives: This would become `style={[styles.container, {width: Dimensions.get('window').width}]}`

views/building-hours/building.js
```diff
@@ -6,86 +6,101 @@
...
-  let open = isBuildingOpen(info)
-  let borderColor = borderColors[open]
+
+  let borderColors = {
+    'Open': '#CEFFCE',
+    'Almost Closed': '#FFFC96',
+    'Closed': '#F7C8C8',
+  }
+
+  type PropsType = {
```
hawkrives: This `type` declaration should live outside of the function.

views/building-hours/index.js
```diff
@@ -5,7 +5,7 @@
  */
 
 import React from 'react'
-import {View, Text, ListView, RefreshControl} from 'react-native'
+import {ListView, RefreshControl} from 'react-native'
```
hawkrives, line 8: Thanks

views/components/collapsibleBlock.js
```diff
@@ -0,0 +1,53 @@
...
+  StyleSheet,
+} from 'react-native'
+
+import Collapsible from 'react-native-collapsible'
+
+
+export default class CollapsibleBlock extends React.Component {
+
+  static propTypes = {
+    backgroundColor: React.PropTypes.string.isRequired,
```
hawkrives: Here, a more general approach might be nice – what do you think about accepting a `style` property, instead of just a single `backgroundColor` prop? If you don't want to, that's fine – it's unlikely that we'll need to pass more styles down, but it might be nice ð

views/components/collapsibleBlock.js
```diff
@@ -0,0 +1,53 @@
...
+    collapsedContainer: {
+    },
+  })
+
+  render() {
+    return (
+        <TouchableOpacity
+          onPress={() => this.setState({isCollapsed: !this.state.isCollapsed})}
+        >
+          {this.props.children[0]}
```
hawkrives, line 45: Um… Let me think about this and get back to you. It's generally a bad idea to do anything to `props.children`, as it won't be an array if there's only one child… which there shouldn't be, but … hmm. Referencing https://facebook.github.io/react/tips/children-props-type.html…

It looks like we should use the React.children.toArray helper function to ensure that this is an array: https://facebook.github.io/react/docs/top-level-api.html#react.children.toarray

views/components/collapsibleBlock.js
```diff
@@ -0,0 +1,53 @@
...
+          <Collapsible collapsed={this.state.isCollapsed} style={[this.styles.collapsedContent, {backgroundColor: this.props.backgroundColor}]}>
```
hawkrives: Anyway, if you wanted to make the styles more general, you could just replace this with `style={[styles.collapsedContent, this.props.style]}`.

views/components/collapsibleBlock.js
```diff
@@ -0,0 +1,53 @@
...
+  static propTypes = {
+    backgroundColor: React.PropTypes.string.isRequired,
+    children: React.PropTypes.array.isRequired,
+  }
+
+  state = {
+    isCollapsed: true,
+  }
+
+  styles = StyleSheet.create({
```
hawkrives: Please remove this from the class?

hawkrives:
Ok that might have been more than a few comments. Sorry about that. It's shaping up nicely!

views/building-hours/building.js
mattk410: How do I give StyleSheet.create access to borderColors and openStatus if they are in BuildingView and StyleSheet.create is not?

mattk410: Never mind

elijahverdoorn:
I second everything Hawken said. Thanks for making the changes.

views/building-hours/all-building-hours.js
mattk410: It's just an argument passed into the function for the purpose of styling the Text. From building.js, it is styles.hoursText.

views/building-hours/all-building-hours.js
mattk410: It is for the next iteration of the loop.

views/building-hours/all-building-hours.js
hawkrives: Ah, ok. I see it now.

views/building-hours/building.js
hawkrives: :+1:

views/building-hours/all-building-hours.js
hawkrives: … that makes a lot of sense. This is what I get for doing code reviews at 3am.

views/components/collapsibleBlock.js
mattk410: There should be exactly 2 children in the array. One is the TouchableOpacity "header" and the other is the dropdown contents

hawkrives:
looks good to me! great work, @mattk410!
- [x] Does it pass `npm run lint` with no warnings in these files? (answer: [yes](https://travis-ci.org/StoDevX/AAO-React-Native/jobs/159482641#L214))
- [x] is yoda green? 📗 🍏 💚 

I say :shipit:!

(Um, @elijahverdoorn, what's our review policy again? have two people sign off or just one?)

{edit: not sure why I did the yoda thing. i'll blame it on the tired brain. and the rain.}

elijahverdoorn:
@hawkrives Review policy is one, but for someone's first change, having a few eyes on it throughout the life of the PR is good. I also approve it.

hawkrives:
🎉 🎉 🎉

hawkrives:
Closes #143

PR #261, opened 2016-10-07
Removing unused react-tween-state dependency

drewvolz:
@hawkrives suggested that this be removed, we do not use it yet depend upon it. I added this one way back when to try and get SIS numbers to animate. Never did anything with it.


hawkrives:
I was going to blame you for this one! Finally, something that I didn't add-and-not-use.

hawkrives (approved):

PR #263, opened 2016-10-07
Added image check for posts not containing images

### LINKED ISSUE #237: Latest cover images
If an Oleville Latest post does not have an image, can we supplement a graphic as a default in the app? I understand latest stories are supposed to have images, but mistakes happen.

screenshot of Oleville latest

drewvolz:
Closes #237. This checks to see if an image id exists from the `article.featured_media` prop. If it does not, we return the default image.

### DIFF
--- a/views/oleville/index.js
+++ b/views/oleville/index.js
@@ -80,11 +80,9 @@ export default class OlevilleView extends React.Component {
       if (responseJson.media_details.sizes.medium.source_url) {
         // if there is a featured image, use that url
         return responseJson.media_details.sizes.medium.source_url
-      } else {
-        // otherwise default to the oleville logo
-        return 'http://oleville.com/wp-content/uploads/2015/12/Oleville-Logo.png'
       }
     }
+    return 'http://oleville.com/wp-content/uploads/2015/12/Oleville-Logo.png'
   }
 
   fetchData = async () => {


drewvolz:
This can be done in a cleaner way.

views/oleville/index.js
```diff
@@ -85,6 +85,9 @@ export default class OlevilleView extends React.Component {
         return 'http://oleville.com/wp-content/uploads/2015/12/Oleville-Logo.png'
       }
     }
+    else {
+        return 'http://oleville.com/wp-content/uploads/2015/12/Oleville-Logo.png'
+    }
```
hawkrives: What about _only_ having this case, and removing the inner else-statement?

Actually: Given https://github.com/StoDevX/AAO-React-Native/blob/6f7b172ddde96b1960077e0f8d8ffc22e47372e3/views/oleville/index.js#L74-L91, 

``` diff
  async getImageUrl(id: number) {
    if (id) {
      let response = await fetch(`http://oleville.com/wp-json/wp/v2/media/${id}`)
      let responseJson = await response.json()
      this.setState({loaded: true})

      if (responseJson.media_details.sizes.medium.source_url) {
        // if there is a featured image, use that url
        return responseJson.media_details.sizes.medium.source_url
+      }
-      } else {
-        // otherwise default to the oleville logo
-        return 'http://oleville.com/wp-content/uploads/2015/12/Oleville-Logo.png'
-      }
     }
-    else {
-        return 'http://oleville.com/wp-content/uploads/2015/12/Oleville-Logo.png'
-    }
+    return 'http://oleville.com/wp-content/uploads/2015/12/Oleville-Logo.png'
  }
```

That is, removing _both_ else-blocks entirely and just letting the function fall back to a single value?

hawkrives (changes requested):
Just the once comment

views/oleville/index.js
elijahverdoorn: Either way is fine by me.

hawkrives:
Build failed because ([link](https://travis-ci.org/StoDevX/AAO-React-Native/jobs/165724439#L233))

``` console
./views/oleville/index.js
  89:9   error    Expected indentation of 6 spaces but found 8                                  indent
```

aka you indented with four spaces in your else block.

drewvolz:
Thanks for the reviews and feedback. I had forgoten to `lint`, but now all is well. I've now added your recommended change of short circuiting with the default return if we haven't returned already. Here is the final product.

screenshot of Oleville latest

hawkrives (approved):
Looks good! :shipit:

PR #294, opened 2016-11-02
react-native@0.36

### LINKED ISSUE #179: Android Tab Layouts Crashing
Every time I switch to the second or third tab of a view on Android, the app crashes. 
_This is a urgent fix needed for 2.0.0._

hawkrives:
> ~~Note: **This is broken. Do not merge!!!**~~
> Update: This is fixed.

Closes #179.

This PR upgrades us from `react-native@0.31` to `react-native@0.36`.

## Broken Things
- The collapsible rows won't open; see https://github.com/oblador/react-native-collapsible/issues/51 and https://github.com/facebook/react-native/issues/10603
  - If we use my custom collapsible component from #295, we'll be ready to go.


hawkrives:
If this builds, it should be ready for merging into `master`.

hawkrives:
I don't know about the push stuff, but … since we don't use that / barely have it hooked up … maybe it doesn't matter?

hawkrives:
I have tested all of the Android screens, and none of them crash when I tap on things.

hawkrives:
This is green. I want to merge it. I have tested it on both iOS and Android, as far as I am able – this includes the SIS portion. 

The only part I have not been able to test is the push notifications things… however, I would humbly submit that we don't care about that right now. OneSignal will upgrade their stuff if they need to, to target the latest react-native version, and we don't currently **use** the push notifications for anything.

We can revert this if we really need to, if you feel _really strongly_ about testing and not breaking the push notifications thing at this time, @elijahverdoorn.

hawkrives:
Also I'm currently working on making the iOS builds faster and as part of that I'm working on Xcode 8, which means that I need to upgrade react native (or manually apply patches to some library files).

drewvolz (approved):
:+1: :shipit:

hawkrives:
If we need to revert this, I am available this weekend + now have data on my phone so I can actually be reached.

PR #296, opened 2016-11-04
Revamp menu

### LINKED ISSUE #65: Display native menu screen for Caf?
BonApp publishes "menus" online for both the Caf and the Cage. The cage's is … never right, but the caf's is in sync with the web menu.

I would like to, for a future version, build a native Caf menu screen that took advantage of the menu data for the Caf to render locally, instead of loading the giant webpage in a webview.

Caf JSON Menu: http://legacy.cafebonappetit.com/api/2/menus?cafe=261
Cage JSON Menu: http://legacy.cafebonappetit.com/api/2/menus?cafe=262

Unfortunately, the Caf menu only seems to include _some_ of the menu items.

mattk410:
Native menu view for Stav and the Cage

Fixes #65 

stav
cage


views/menus/stav.js
```diff
@@ -4,67 +4,199 @@ import React from 'react'
...
+        stationItems.push(temp)
+      }
+    }
+    return stationItems
   }
 
-  onNavigationStateChange(navState: {url: string}) {
-    this.setState({url: navState.url})
+
+  fetchData () {
```
hawkrives: You'll want to make this a bound instance function with `fetchData = () => {` – it's an ES2017(?) change that binds the function to the class instance. We can use it because the react native processor turns it on for us. 

Without that, the call to `this.setState` – well, I'd think that it would never work, but it might? I think it's a bug if it does.

views/menus/dietaryFilters.js
```diff
@@ -0,0 +1,37 @@
+module.exports = function(){
+  return {
```
hawkrives: Why is this a function instead of just an exported object?

views/menus/stav.js
```diff
@@ -4,67 +4,199 @@ import React from 'react'
...
+          }
+        }
+      }
+
+      this.setState({
+        dataSource: this.state.dataSource.cloneWithRowsAndSections(dataBlob, sectionIDs, rowIDs),
+        loaded: true,
+      })
+
+    }).done()
```
hawkrives: Does the `.done()` method exist on our promises? I thought it had been removed from the standard…

views/menus/stav.js
```diff
@@ -4,67 +4,199 @@ import React from 'react'
...
 
-const URL = 'http://stolaf.cafebonappetit.com/cafe/stav-hall#Lunch'
-// const DEFAULT_URL = 'http://legacy.cafebonappetit.com/print-menu/cafe/261/menu/112732/days/today/pgbrks/0/'
+import DietaryFilters from './dietaryFilters'
+const dietaryFilters = DietaryFilters()
+
+import buildingHours from '../../data/building-hours.json'
 
 export default class StavMenuView extends React.Component {
-  state = {
```
hawkrives, line 24: Please make an initial state object with initial values for each thing you've got in state, so there's one place for us to look and see what we've got available.

hawkrives (commented):
AMAZING WORK OVERALL

(Could you post a screenshot of how it looks, please?)

views/menus/stav.js
mattk410: Isn't that what the constructor does?

hawkrives: Ah yes, apologies. I looked for that but didn't see it. Now I do. Criticism withdrawn!

hawkrives:
Appveyor error is eslint errors; `npm run lint` on your machine should show them to you as well. `npm run lint -- --fix` might even fix some of them.

hawkrives (approved):
APROOVED :shipit:

drewvolz:
👍 :shipit:

hawkrives:
Woot woot😻🤧😤🤠😇👾🐴🚍👹

elijahverdoorn:
Nice!

PR #297, opened 2016-11-04
Add some eslint rules  for managing whitespace

hawkrives:
This enables two eslint rules: 

- `no-trailing-spaces` requires that there be no whitespace characters at the end of a line
- `eol-last` requires that the file end in a newline

`eol-last`, in particular, makes git diffs cleaner because without a newline at the end of the file, the diff will always show it as changed.


hawkrives:
I also updated the current repository to follow the rules. There was only one violation.

drewvolz (approved):
:+1: :shipit:

PR #305, opened 2016-11-08
Add "Ongoing" section to the calendars

elijahverdoorn:
I changed something on the calendar. The old calendar had multi-day events listed at their start day, rather than today. This makes the St. Olaf Master Calendar view look silly, since it seems like it's stuck in the past. These changes set the start time of anything in the past to the current time. Since the API will never send us anything that has already ended, we're not going to show past events. As a side effect, any event that is happening now will display the current time and the end time, which I considered changing but kinda liked, since it emphasizes the currency of the event. If there are objections, I'd be willing to compromise.

### DIFF
--- a/views/calendar/calendar.js
+++ b/views/calendar/calendar.js
@@ -86,8 +86,9 @@ export default class CalendarView extends React.Component {
         event.endTime = moment(event.end.date || event.end.dateTime)
       })
       data.forEach(event => {
-        if (event.startTime < Date.now())
+        if (event.startTime < Date.now()) {
           event.startTime = moment()
+        }
       })
       let grouped = groupBy(data, event => event.startTime.format('ddd  MMM Do'))
       this.setState({events: this.state.events.cloneWithRowsAndSections(grouped)})
--- a/views/calendar/calendar.js
+++ b/views/calendar/calendar.js
@@ -85,6 +85,10 @@ export default class CalendarView extends React.Component {
         event.startTime = moment(event.start.date || event.start.dateTime)
         event.endTime = moment(event.end.date || event.end.dateTime)
       })
+      data.forEach(event => {
+        if (event.startTime < Date.now())
+          event.startTime = moment()
+      })
       let grouped = groupBy(data, event => event.startTime.format('ddd  MMM Do'))
       this.setState({events: this.state.events.cloneWithRowsAndSections(grouped)})
     } else if (data && !data.length) {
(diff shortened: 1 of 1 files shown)


hawkrives:
You missed a curly brace https://travis-ci.org/StoDevX/AAO-React-Native/jobs/174105961#L233

views/calendar/calendar.js
```diff
@@ -93,6 +85,10 @@ export default class CalendarView extends React.Component {
         event.startTime = moment(event.start.date || event.start.dateTime)
         event.endTime = moment(event.end.date || event.end.dateTime)
       })
+      data.forEach(event => {
+        if (event.startTime < Date.now())
```
hawkrives: All blocks require curly braces in this project.

``` js
if (x) {
  z
}
```

instead of

``` js
if (x)
  z
```

drewvolz: I have a second, I'll fix this up for @elijahverdoorn

hawkrives:
I have a piece of feedback.

Rather than editing the `startTime` variable, which will change both the grouped time and the displayed start time, would you be opposed to just grouping on the current time (and leaving the displayed time the same)?

<hr>

**Replacing** 

``` js
data.forEach(event => {
  if (event.startTime < Date.now())
    event.startTime = moment()
})
let grouped = groupBy(data, event => event.startTime.format('ddd  MMM Do'))
```

<hr>

**with** 

``` js
let grouped = groupBy(data, event => {
    let start = (event.startTime < Date.now())
        ? moment()
        : event.startTime
    return start.format('ddd  MMM Do')
})
```

<hr>

([`views/calendar/calendar.js:88-92`](https://github.com/StoDevX/AAO-React-Native/blob/ea5c6778ef7c7b1fa1fb6a986da48b5b21468a50/views/calendar/calendar.js#L88-L92))

should do it.

hawkrives:
central time (bad safari)

hawkrives:
My other comment for this PR is that there's no use of moment-timezone here.

I think the original stuff didn't have an issue because Google provides the times in a standardized timezone (UTC?), but this change would use a different time zone when it's normalizing the times of the early-start events (device timezone).

I think we should set the timezone in the call to `moment()` to whatever Google returns, either UTC or US/Central.

hawkrives:
But! I appreciate this change. The current behavior was bugging me. :+1: @elijahverdoorn.

views/calendar/calendar.js
elijahverdoorn: Thanks @drewvolz.

drewvolz (approved):

PR #313, opened 2016-11-11
Enable Bitrise for faster OS X builds

hawkrives:
> Bitrise is a bit different from Appveyor and TravisCI, as it doesn't use the committed config file. Instead, you can import and export a config using their web interface. 
>
> I would like to keep this as the canonical version, and copy it into their system whenever it changes.

Bitrise finishes the build in roughly 5 minutes, and has almost no startup time, unlike Travis. I intend to disable Travis' OS X builds for now, leaving it with just JS and Android/unix build duties.

Eventually, I want to start writing tests for Xcode to run against different versions of iOS, to make sure things work – things like "tapping on this button opens this screen". When that happens, I will probably re-enable Travis, since it makes it so much easier to start multiple builds at once.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "flow": "flow",
     "bundle:ios": "react-native bundle --entry-file index.ios.js --dev true --platform ios --bundle-output ./ios/AllAboutOlaf/main.jsbundle --assets-dest ./ios",
     "bundle:android": "react-native bundle --entry-file index.android.js --dev true --platform android --bundle-output android/app/src/main/assets/index.android.bundle --assets-dest android/app/src/main/res/",
-    "build:ios": "xctool -configuration Release -workspace ios/AllAboutOlaf.xcworkspace -scheme AllAboutOlaf -sdk iphonesimulator -destination platform='iOS Simulator,name=iPhone 5s,OS=9.2'",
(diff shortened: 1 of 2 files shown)


hawkrives:
I will also add more people to the Bitrise team. I currently have Drew, as I was just testing it out. @elijahverdoorn should I only add you, or should I add the rest of your team as well?

hawkrives:
Bitrise (iOS): `5m 1s`
Travis (Android, Linux): `5m 33s`
Appveyor (Android, Windows): `6m 59s`
Travis (iOS): `30m 10s` (waiting: `26m 40s`, running: `9m 30s`)

package.json
```diff
@@ -12,7 +12,7 @@
     "flow": "flow",
     "bundle:ios": "react-native bundle --entry-file index.ios.js --dev true --platform ios --bundle-output ./ios/AllAboutOlaf/main.jsbundle --assets-dest ./ios",
     "bundle:android": "react-native bundle --entry-file index.android.js --dev true --platform android --bundle-output android/app/src/main/assets/index.android.bundle --assets-dest android/app/src/main/res/",
-    "build:ios": "xctool -configuration Release -workspace ios/AllAboutOlaf.xcworkspace -scheme AllAboutOlaf -sdk iphonesimulator -destination platform='iOS Simulator,name=iPhone 5s,OS=9.2'",
+    "build:ios": "xcodebuild -configuration Release -workspace ios/AllAboutOlaf.xcworkspace -scheme AllAboutOlaf -sdk iphonesimulator -destination platform='iOS Simulator,name=iPhone 5s,OS=9.2' | xcpretty",
```
hawkrives, line 15: This will need to make its way onto `master` anyway, soon, so that Travis can start working with xcode8.

PR #323, opened 2016-11-14
Add initialListSize to ContactsView

### LINKED ISSUE #321: ListViews are sometimes blank
If you open the Transportation view, scroll to the bottom of the list, then go back, and open Important Contacts, it's blank. Part of the time.

Screenshot proof included below.

hawkrives:
> Closes #321

I think that setting this to 3 should tell RN to start by rendering 3 rows, rather than 0.

I believe that, without this prop, RN starts at 0 rows, and continues to render them until the `scrollRenderAheadDistance` prop is reached. For some reason, sometimes it never goes beyond 0 rows. Therefore, I think that starting it at 3 should kick-start that process.

### DIFF
--- a/views/contacts.js
+++ b/views/contacts.js
@@ -41,6 +41,7 @@ export default class ContactView extends React.Component {
     return (
       <View style={styles.container}>
         <ListView
+          initialListSize={3}
           renderRow={this._renderRow.bind(this)}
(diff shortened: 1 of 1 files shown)


drewvolz (approved):
:shipit:  :+1:

PR #326, opened 2016-11-15
Fixed multiple settings menus

### LINKED ISSUE #155: Settings button can be pushed multiple times
> It is weirdly possible to hit the settings button quickly enough to summon multiple instances of the settings menu. They appear to be pushed onto the view stack. I can mash the mutton 4 or 5 times, and  become stuck in that state and cannot pop-back. If you push another view (credits/privacy/legal) you can get it to let you pop the settings view.

I'm thinking … disable the settings button after a tap?

But why does it take so long to open?

Hmm.

(oh lol drew suggested disabling the button too. it's a stopgap fix, at least.)

drewvolz:
This prevents the settings button from being mashed quickly which results in pushing two or more views to the stack. This closes #155.

### DIFF
--- a/app.js
+++ b/app.js
@@ -135,6 +135,16 @@ const styles = StyleSheet.create({
   },
 })
 
+let settingsButtonActive = false
+
+function openSettings(route, navigator) {
+  navigator.push({
+    id: 'SettingsView',
+    title: 'Settings',
+    index: route.index + 1,
+    sceneConfig: Navigator.SceneConfigs.FloatFromBottom,
+  })
+}
 
 function LeftButton(route, navigator, index, navState) {
   switch (route.id) {
@@ -142,12 +152,10 @@ function LeftButton(route, navigator, index, navState) {
       return (
         <TouchableOpacity
           style={[styles.backButton]}
-          onPress={() => navigator.push({
-            id: 'SettingsView',
-            title: 'Settings',
-            index: route.index + 1,
-            sceneConfig: Navigator.SceneConfigs.FloatFromBottom,
-          })}
+          onPress={settingsButtonActive ? () => {} : () => {
+            settingsButtonActive = true
+            openSettings(route, navigator)
+          }}
         >
           <Icon style={styles.settingsIcon} name='ios-settings' />
         </TouchableOpacity>
@@ -157,7 +165,10 @@ function LeftButton(route, navigator, index, navState) {
       return (
         <TouchableOpacity
           style={[styles.backButton, {marginLeft: 10, marginVertical: Platform.OS === 'android' ? 21 : 16}]}
-          onPress={() => navigator.pop()}
+          onPress={() => {
+            navigator.pop()
+            settingsButtonActive = false
+          }}
         >
           <Text style={styles.backButtonText}>Close</Text>
         </TouchableOpacity>
--- a/app.js
+++ b/app.js
@@ -149,8 +149,8 @@ function openSettings(route, navigator) {
 }
 
 function closeSettings(navigator) {
-  navigator.pop()
   settingsButtonActive = false
+  navigator.pop()
 }
 
 function LeftButton(route, navigator, index, navState) {
@@ -169,9 +169,7 @@ function LeftButton(route, navigator, index, navState) {
       return (
         <TouchableOpacity
           style={[styles.backButton, {marginLeft: 10, marginVertical: Platform.OS === 'android' ? 21 : 16}]}
-          onPress={() => {
-            closeSettings(navigator)
-          }}
+          onPress={() => closeSettings(navigator)}
         >
           <Text style={styles.backButtonText}>Close</Text>
         </TouchableOpacity>
--- a/app.js
+++ b/app.js
@@ -36,6 +36,7 @@ import PrivacyView from './views/settings/privacy'
 import LegalView from './views/settings/legal'
 
 import NoRoute from './views/components/no-route'
+import noop from 'lodash/noop'
 
 // Render a given scene
 function renderScene(route, navigator) {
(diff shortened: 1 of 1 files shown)


app.js
```diff
@@ -135,19 +135,27 @@ const styles = StyleSheet.create({
   },
 })
 
+let settingsButtonActive = false
```
hawkrives: Make this a state variable in App, and pass it into the LeftButtom function as a prop.

hawkrives: Along with a callback to a member function in App to toggle the variable.

app.js
```diff
@@ -135,19 +135,27 @@ const styles = StyleSheet.create({
...
         <TouchableOpacity
           style={[styles.backButton]}
-          onPress={() => navigator.push({
-            id: 'SettingsView',
-            title: 'Settings',
-            index: route.index + 1,
-            sceneConfig: Navigator.SceneConfigs.FloatFromBottom,
-          })}
+          onPress={settingsButtonActive ? () => {} : () => {
+            settingsButtonActive = true
```
hawkrives: Then call that callback here

app.js
```diff
@@ -135,19 +135,27 @@ const styles = StyleSheet.create({
...
       return (
         <TouchableOpacity
           style={[styles.backButton]}
-          onPress={() => navigator.push({
-            id: 'SettingsView',
-            title: 'Settings',
-            index: route.index + 1,
-            sceneConfig: Navigator.SceneConfigs.FloatFromBottom,
-          })}
+          onPress={settingsButtonActive ? () => {} : () => {
```
hawkrives: You might also name the noop function, since we're naming the other.

app.js
```diff
@@ -157,7 +165,10 @@ function LeftButton(route, navigator, index, navState) {
       return (
         <TouchableOpacity
           style={[styles.backButton, {marginLeft: 10, marginVertical: Platform.OS === 'android' ? 21 : 16}]}
-          onPress={() => navigator.pop()}
+          onPress={() => {
+            navigator.pop()
+            settingsButtonActive = false
```
hawkrives: This online function should probably be extracted to a member function,`closeSettingsScreen` or something.

hawkrives (dismissed):

drewvolz:
From chatting with you, this sounds as if it will be a more complicated fix than we first imagined.

app.js
hawkrives: Import `noop` from lodash and use it here, and move the `settingsButtonActive = true` line into the `openSettings` function.

`onPress={settingsButtonActive ? noop : () => openSettings(route, navigator)}`

app.js
hawkrives: This is still a comment I'd like to have addressed.

hawkrives (changes requested):
Moar feedbacks

app.js
hawkrives: Ignore these

app.js
hawkrives: Ignore this comment

app.js
hawkrives: Ignore this comment

app.js
```diff
@@ -146,16 +148,18 @@ function openSettings(route, navigator) {
   })
 }
 
+function closeSettings() {
+  navigator.pop()
+  settingsButtonActive = false
```
hawkrives, line 152: Flip these?

app.js
```diff
@@ -146,16 +148,18 @@ function openSettings(route, navigator) {
...
   switch (route.id) {
     case 'HomeView':
       return (
         <TouchableOpacity
           style={[styles.backButton]}
-          onPress={settingsButtonActive ? () => {} : () => {
-            settingsButtonActive = true
-            openSettings(route, navigator)
-          }}
+          onPress={settingsButtonActive ? noop : () => openSettings(route, navigator)}
```
hawkrives, line 162: ✅

app.js
```diff
@@ -166,8 +170,7 @@ function LeftButton(route, navigator, index, navState) {
         <TouchableOpacity
           style={[styles.backButton, {marginLeft: 10, marginVertical: Platform.OS === 'android' ? 21 : 16}]}
           onPress={() => {
-            navigator.pop()
-            settingsButtonActive = false
+            closeSettings()
           }}
```
hawkrives: Almost

We need to give `navigator` to the function

`onPress={() => closeSettings(navigator)}`

app.js
hawkrives: It also needs to accept `navigator` as a prop

hawkrives (changes requested):
Almost there

app.js
```diff
@@ -157,7 +169,9 @@ function LeftButton(route, navigator, index, navState) {
       return (
         <TouchableOpacity
           style={[styles.backButton, {marginLeft: 10, marginVertical: Platform.OS === 'android' ? 21 : 16}]}
-          onPress={() => navigator.pop()}
+          onPress={() => {
```
hawkrives: put this on one line and remove the outer set of braces?

app.js
```diff
@@ -135,19 +136,30 @@ const styles = StyleSheet.create({
...
+    id: 'SettingsView',
+    title: 'Settings',
+    index: route.index + 1,
+    sceneConfig: Navigator.SceneConfigs.FloatFromBottom,
+  })
+}
+
+function closeSettings(navigator) {
+  navigator.pop()
+  settingsButtonActive = false
```
hawkrives, line 152: and should we flip these two?

I don't think there'll be any functional difference

app.js
drewvolz: Sure

app.js
drewvolz: Done.

app.js
drewvolz: We could.

hawkrives (approved):

PR #329, opened 2016-11-15
remove automaticallyAdjustContentInsets

hawkrives:
The `automaticallyAdjustContentInsets` prop is supposed to be all you need to allow the ScrollViews to sit above a tab bar, but on RN`@0.29` it was broken, so we removed it.

It is now fixed, so we can allow it to work its magic once again.

### DIFF
--- a/views/directory.js
+++ b/views/directory.js
@@ -57,7 +57,6 @@ export default class DirectoryView extends React.Component {
     return (
       <View style={styles.container}>
         <WebView
-          automaticallyAdjustContentInsets={false}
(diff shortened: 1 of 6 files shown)


drewvolz (changes requested):
I'd like to see them disappear from `ListViews` and `WebViews` too, please. https://github.com/StoDevX/AAO-React-Native/search?utf8=%E2%9C%93&q=automaticallyAdjustContentInsets

drewvolz (approved):

PR #335, opened 2016-11-16
Meals remaining

### LINKED ISSUE #303: Display Meal Plan Information
IIRC, the meal plan balance info used to live at https://www.stolaf.edu/apps/olecard/checkbalance/index.cfm, but at the end of summer when @drewvolz and I were wrapping stuff up the meal plan info was either missing or disabled temporarily. 

We'd need someone else to go re-implement the data fetching.

<hr>

_(edit from @drewvolz -- it is worthy to note that with the info being hidden, it is of great importance that we make sure to not assume the data is available... so when scraping the table it is displayed in, we should take care to ensure that we don't necessarily expect it to be there before proceeding.)_

elijahverdoorn:
Ok. I've got this working. The code's not beautiful, but it does work. I made a few decisions in there that we might want to discuss:

* Layout of settings
* Redirection to settings from SIS
* Re-naming variables in the SIS

I'd appreciate suggestions and tweaks.
(I'm especially not proud of the error checking code, there's got to be a better way to do that)

Closes #303

### DIFF
--- a/lib/financials.js
+++ b/lib/financials.js
@@ -104,17 +104,25 @@ async function getFinancialDataFromServer(): Promise<FinancialDataShapeType> {
     AsyncStorage.setItem(FINANCIALS__CACHE__BALANCES, JSON.stringify(new Date())),
   ])
 
-  return {flex, ole, print}
+  return {flex, ole, print};
 }
 
 
 // TODO: come up with a better story around auth for olecard
-export default async function getWeeklyMealsRemaining() {
+export async function getWeeklyMealsRemaining() {
   let {username, password} = await loadLoginCredentials()
   let form = buildFormData({
     username: username,
     password: password,
   })
-  let loginResult = await api.post('/apps/olecard/checkbalance/authenticate.cfm', {body: form})
-  return loginResult
+  let fetchResult = await api.post('/apps/olecard/checkbalance/authenticate.cfm', {body: form})
+
+  let dom = parseHtml(fetchResult.body)
+  let data = cssSelect('.accountrow', dom)
+  let values = []
+  data.map((item) => {
+    values.push(item.next.next.children[0].data)
+  })
+
+  return({"weeklyMeals": values[4], "dailyMeals": values[3]})
 }
--- a/lib/financials.js
+++ b/lib/financials.js
@@ -104,7 +104,7 @@ async function getFinancialDataFromServer(): Promise<FinancialDataShapeType> {
     AsyncStorage.setItem(FINANCIALS__CACHE__BALANCES, JSON.stringify(new Date())),
   ])
 
-  return {flex, ole, print};
+  return {flex, ole, print}
 }
 
 
@@ -120,14 +120,13 @@ export async function getWeeklyMealsRemaining() {
   try {
     let dom = parseHtml(fetchResult.body)
     let data = cssSelect('.accountrow', dom)
-    let values = []
-    data.map((item) => {
-      values.push(item.next.next.children[0].data)
+    let values = data.map((item) => {
+      return item.next.next.children[0].data
     })
-    return({"weeklyMeals": values[4], "dailyMeals": values[3]})
+    return({'weeklyMeals': values[4], 'dailyMeals': values[3]})
   } catch (error) {
-    console.log("error in fetching meal data")
-    return({"weeklyMeals": null, "dailyMeals": null})
+    console.log('error in fetching meal data')
+    return({'weeklyMeals': null, 'dailyMeals': null})
   }
 
 }
--- a/lib/financials.js
+++ b/lib/financials.js
@@ -117,12 +117,17 @@ export async function getWeeklyMealsRemaining() {
   })
   let fetchResult = await api.post('/apps/olecard/checkbalance/authenticate.cfm', {body: form})
 
-  let dom = parseHtml(fetchResult.body)
-  let data = cssSelect('.accountrow', dom)
-  let values = []
-  data.map((item) => {
-    values.push(item.next.next.children[0].data)
-  })
+  try {
+    let dom = parseHtml(fetchResult.body)
+    let data = cssSelect('.accountrow', dom)
+    let values = []
+    data.map((item) => {
+      values.push(item.next.next.children[0].data)
+    })
+    return({"weeklyMeals": values[4], "dailyMeals": values[3]})
+  } catch (error) {
+    console.log("error in fetching meal data")
+    return({"weeklyMeals": null, "dailyMeals": null})
+  }
 
-  return({"weeklyMeals": values[4], "dailyMeals": values[3]})
 }
--- a/lib/financials.js
+++ b/lib/financials.js
@@ -120,13 +120,12 @@ export async function getWeeklyMealsRemaining() {
   try {
     let dom = parseHtml(fetchResult.body)
     let data = cssSelect('.accountrow', dom)
-    let values = data.map((item) => {
+    let values = data.map(item => {
       return item.next.next.children[0].data
     })
-    return({'weeklyMeals': values[4], 'dailyMeals': values[3]})
+    return {'weeklyMeals': values[4], 'dailyMeals': values[3]}
   } catch (error) {
     console.log('error in fetching meal data')
-    return({'weeklyMeals': null, 'dailyMeals': null})
+    return {'weeklyMeals': null, 'dailyMeals': null}
   }
-
 }
--- a/lib/login.js
+++ b/lib/login.js
@@ -29,32 +29,43 @@ export function clearLoginCredentials(): Promise<any> {
 
 export async function isLoggedIn(): Promise<boolean> {
   let result = await AsyncStorage.getItem('credentials:valid')
+  result = result && await AsyncStorage.getItem('olafCredentials:valid')
+
   return JSON.parse(result)
 }
 
 export async function performLogin(username: string, password: string): Promise<{result: boolean, message?: string}> {
-  let {result, message} = await stOlafLogin(username, password)
-  if (!result) {
-    await AsyncStorage.setItem('credentials:valid', JSON.stringify(false))
-    return {result, message}
+  if (username && password) {
+    let {result, message} = await stOlafLogin(username, password)
+
+    if (!result) {
+      await AsyncStorage.setItem('olafCredentials:valid', JSON.stringify(false))
+      return {result, message}
+    }
+
+    await Promise.all([
+      saveLoginCredentials(username, password),
+      AsyncStorage.setItem('olafCredentials:valid', JSON.stringify(true)),
+    ])
+
+    return {result}
+  } else {
+    return {false}
   }
-
-  await Promise.all([
-    saveLoginCredentials(username, password),
-    AsyncStorage.setItem('credentials:valid', JSON.stringify(true)),
-  ])
-
-  return {result}
 }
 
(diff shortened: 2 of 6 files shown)


elijahverdoorn:
screen shot 2016-11-16 at 10 45 28 am

screen shot 2016-11-16 at 10 45 28 am

lib/financials.js
```diff
@@ -104,17 +104,30 @@ async function getFinancialDataFromServer(): Promise<FinancialDataShapeType> {
     AsyncStorage.setItem(FINANCIALS__CACHE__BALANCES, JSON.stringify(new Date())),
   ])
 
-  return {flex, ole, print}
+  return {flex, ole, print};
```
zimmer18: ; ?

views/sis/balances.js
```diff
@@ -71,6 +75,7 @@ export default class BalancesView extends React.Component {
   }
 
   loadIfLoggedIn = async () => {
+    console.log('TESTING')
```
drewvolz: :zap:

lib/financials.js
```diff
@@ -104,17 +104,30 @@ async function getFinancialDataFromServer(): Promise<FinancialDataShapeType> {
...
+  try {
+    let dom = parseHtml(fetchResult.body)
+    let data = cssSelect('.accountrow', dom)
+    let values = []
+    data.map((item) => {
+      values.push(item.next.next.children[0].data)
+    })
+    return({"weeklyMeals": values[4], "dailyMeals": values[3]})
+  } catch (error) {
+    console.log("error in fetching meal data")
```
drewvolz: Maybe we should present an alert, or some information for the user to know that it could not be fetched?

views/settings/index.js
```diff
@@ -36,50 +43,81 @@ export default class SettingsView extends React.Component {
...
     })
   }
 
-  logOut = async () => {
-    this.setState({loading: true})
+  logOutGoogle = async () => {
+    this.setState({loadingGoogle: true})
     CookieManager.clearAll((err, res) => {
       if (err) {
         console.log(err)
```
drewvolz, line 96: Maybe we could let the user know we could not log them out?

lib/financials.js
hawkrives: Bad semicolon :)
And good eye, @zimmer18.

lib/financials.js
```diff
@@ -104,17 +104,30 @@ async function getFinancialDataFromServer(): Promise<FinancialDataShapeType> {
...
-  return loginResult
+  let fetchResult = await api.post('/apps/olecard/checkbalance/authenticate.cfm', {body: form})
+
+  try {
+    let dom = parseHtml(fetchResult.body)
+    let data = cssSelect('.accountrow', dom)
+    let values = []
+    data.map((item) => {
+      values.push(item.next.next.children[0].data)
+    })
```
hawkrives, line 125: Please don't use `map` for side-effects; `map` should be a pure function, not mutating anything outside of its scope.

Instead, you could do 

``` js
let values = data.map((item) => {
  return item.next.next.children[0].data
})
```

lib/login.js
```diff
@@ -28,5 +29,43 @@ export function clearLoginCredentials(): Promise<any> {
...
+    }
+
+    await Promise.all([
+      saveLoginCredentials(username, password),
+      AsyncStorage.setItem('olafCredentials:valid', JSON.stringify(true)),
+    ])
+
+    return {result}
+  } else {
+    return {false}
```
hawkrives: According to your type defs, this should be `return {result: false}`

views/settings/index.js
```diff
@@ -36,50 +43,81 @@ export default class SettingsView extends React.Component {
...
       }
       console.log(res)
```
hawkrives: Remove this `log`

drewvolz:
Try running `npm run lint` to diagnose linting issues which is what CI failures in Travis and Appveyor are choking on.

views/settings/index.js
```diff
@@ -36,50 +43,81 @@ export default class SettingsView extends React.Component {
...
       })
     })
     await AsyncStorage.setItem('credentials:valid', JSON.stringify(false))
   }
 
+  logInStOlaf = async () => {
+    this.setState({loadingStOlaf: true})
+    let {username, password} = this.state
+    if (username) {
+      username.replace('@stolaf.edu', '') // just in case someone uses their full email, throw the part we don't need away
```
hawkrives: Might as well just make this generic; remove anything after `@`? Something like `.replace(/@.*$/, '')`?

views/settings/index.js
```diff
@@ -102,36 +140,106 @@ export default class SettingsView extends React.Component {
...
     )
 
     let accountSection = (
       <View>
-        <Section>
+        <Section header='LOGIN' footer='Note: This application requires logging in to Google and inputting your St. Olaf username and password in order use all features.'>
+          {usernameCell}
+          {passwordCell}
+          {loginButton}
           {actionCell}
```
hawkrives, line 279: I don't like having both of the login buttons right next to each other. Can we pull the "Sign In with Google" button into its own `<Section/>`?

views/sis/balances.js
hawkrives: Remove this `log`, please

views/sis/balances.js
```diff
@@ -59,6 +61,8 @@ export default class BalancesView extends React.Component {
     flex: null|number,
     ole: null|number,
     print: null|number,
+    weeklyMeals: null|number,
+    dailyMeals: null|number,
```
hawkrives, line 65: Yay flow types!

views/sis/balances.js
```diff
@@ -92,6 +97,13 @@ export default class BalancesView extends React.Component {
         let {flex, ole, print} = data
         this.setState({flex, ole, print})
       }
+      data = await getWeeklyMealsRemaining()
```
hawkrives: Maybe don't reuse the variable? No harm, I suppose, but I dislike reusing variables in the same scope…

views/sis/balances.js
```diff
@@ -174,12 +186,12 @@ export default class BalancesView extends React.Component {
           <Section header='MEAL PLAN'>
             <Cell cellStyle='RightDetail'
               title='Daily Meals Left'
-              detail='Not Available'
+              detail = {dailyMeals}
```
hawkrives: No spaces around the `=`

views/sis/balances.js
```diff
@@ -174,12 +186,12 @@ export default class BalancesView extends React.Component {
...
             />
 
             <Cell cellStyle='RightDetail'
               title='Weekly Meals Left'
-              detail='Not Available'
+              detail = {weeklyMeals}
```
hawkrives: No spaces around the `=`

views/settings/index.js
```diff
@@ -102,36 +140,106 @@ export default class SettingsView extends React.Component {
...
+              ? 'Sign Out of St. Olaf'
+              : 'Sign In to St. Olaf'}
         </Text>
       </CustomCell>
     )
 
     let accountSection = (
       <View>
-        <Section>
+        <Section header='LOGIN' footer='Note: This application requires logging in to Google and inputting your St. Olaf username and password in order use all features.'>
```
hawkrives: I feel like this should say `[…] requires both logging into Google and […]`, with a `both` in there somewhere?

hawkrives (changes requested):

hawkrives:
My biggest tweak suggestion would probably be that we could extract the St. Olaf login section into its own component – it displays the two fields and the button, and accepts an onLogin prop that it calls when the user pushes the button.

lib/financials.js
hawkrives: Nah. It'll just say N/A in the view.

drewvolz: I really think being clear is important here. Otherwise, you get 1-star reviews on the store saying it does not work.

hawkrives: I'm just not sure what someone would expect. We have no idea what the failure cases would be, really, so we can't provide tailored error messages. The most common case will be that your password has changed, but that should be dealt with similarly to the SIS data, I think.

views/settings/index.js
drewvolz: "in order **to** use all features"

drewvolz:
Should we change the text on the login buttons? I will say that I have become fairly desensitized to most login forms, and multiple options either means register/login (not the case here), or login with EITHER of these methods. I feel that some might glide past our "note" below the buttons. This would lead to problems when they login with only one method... because we are asking for both.

I am proposing that we make this more clear. Can we list the buttons as: 

[**First**, Sign In to St. Olaf]
[**Next**, Sign In with Google]

elijahverdoorn:
I think I've addressed all the major issues with this one.

hawkrives:
I'll check the merge conflict locally

elijahverdoorn:
@hawkrives I was just doing that.

hawkrives:
@elijahverdoorn Try something for me: tap on the text "username" (the label for the username field)?

I'm afraid it might crash.

hawkrives:
More to the point, it's still calling the `focusUsername` and `focusPassword` functions that I wrote for the original login fields, but they don't exist anymore.

views/settings/index.js
```diff
@@ -102,36 +140,106 @@ export default class SettingsView extends React.Component {
...
+          autoCorrect={false}
+          autoCapitalize='none'
+          style={styles.customTextInput}
+          placeholderTextColor='#C7C7CC'
+          disabled={disabledStOlaf}
+          placeholder='username'
+          value={username}
+          returnKeyType='next'
+          onChangeText={text => this.setState({username: text})}
+          onSubmitEditing={this.focusPassword}
```
hawkrives, line 229: `this.focusPassword` needs to be restored from the history

views/settings/index.js
```diff
@@ -102,36 +140,106 @@ export default class SettingsView extends React.Component {
...
+            : loggedInGoogle
+              ? 'Sign Out of Google'
+              : 'Sign In with Google'}
+        </Text>
+      </CustomCell>
+    )
+
+    let usernameCell = (
+      <CustomCell contentContainerStyle={styles.loginCell}>
+        <Text onPress={this.focusUsername} style={styles.label}>Username</Text>
```
hawkrives, line 216: `this.focusUsername` needs to be restored from the history

views/settings/index.js
```diff
@@ -102,36 +140,106 @@ export default class SettingsView extends React.Component {
...
+          returnKeyType='next'
+          onChangeText={text => this.setState({username: text})}
+          onSubmitEditing={this.focusPassword}
+        />
+      </CustomCell>
+    )
+
+    let passwordCell = (
+      <CustomCell contentContainerStyle={styles.loginCell}>
+        <Text onPress={this.focusPassword} style={styles.label}>Password</Text>
```
hawkrives, line 236: `this.focusPassword` needs to be restored from the history

hawkrives (changes requested):
These functions

views/settings/index.js
elijahverdoorn: Gotcha. I've run out of time for the moment, but I'll be back on it later. Good catch.

hawkrives (approved):

views/settings/index.js
hawkrives: done

views/settings/index.js
hawkrives: done

views/settings/index.js
hawkrives: done

PR #338, opened 2016-11-16
Edit building hours stuff

hawkrives:
> Closes #337 

This is a pretty big refactoring of the building hours time tooling stuff. It cleans up more than just #337, but it fixes that as part of the cleanups.

- I removed the `<Text>` generation from `allBuildingHours()`
- I added `now` as a prop and argument to everything that needs the "current" time, so there's one spot to define the current moment. This eliminates the edge-case of one moment being in one hour or day, and the next function using a moment from the next hour or day.
- I made a new function `parseBuildingHours`, which takes over the actual logic of checking if a building is open or closed. I reworked the other functions to use it, as appropriate.

I think that's everything. I'll edit this comment once I see the diff again.

PR #341, opened 2016-11-17
Fix post-midnight times

hawkrives:
> Sorry about this bring blank at first.

This PR fixes a bug with building hours.

The current implementation always uses the current day as the base for finding the open status, but when a building is open onto the next day, that starts to all apart.

Namely, when someone opens the Building Hours view after Midnight on Friday night (which is technically Saturday Morning), it will switch to evaluating the hours on Saturday. Which results in (for example) the Pause being closed, because it's checking (saturday: open 10:30am, close Sunday 2am) instead of (friday: open 10:30am, close Saturday 2am).

This PR introduces the assumption that no buildings will be open past 3am.

With that assumption, though, if it is currently before 3am, we subtract a day from the evaluation, allowing the logic to work correctly.

I will rebase this PR to be smaller after #338 is dealt with.

### DIFF
--- a/views/building-hours/get-building-hours.js
+++ b/views/building-hours/get-building-hours.js
@@ -10,7 +10,7 @@ type HourPairType = {open: momentT, close: momentT};
 
 export function parseBuildingHours(hours: BuildingHoursType, now: momentT): HourPairType {
   let dayOfYear = now.dayOfYear()
-  let [startTimeString, closeTimeString, options={nextDay: false}] = hours
+  let [startTimeString, closeTimeString] = hours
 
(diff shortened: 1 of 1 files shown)


hawkrives:
If you want to see a smaller diff, https://github.com/StoDevX/AAO-React-Native/compare/a1fd1b8...c23719da428a1c8d49e34e97b7c8fac33c589e54 is just the two commits from this PR.

hawkrives:
I'm leaving this un-merged for now because I'd like some other eyes to look at my new algorithm and see if they see any issues with it.

drewvolz (approved):
👍 :shipit:

mattk410 (approved):
Looks good

PR #342, opened 2016-11-17
Faster transit updates

hawkrives:
This PR does two things:

- It slows down the update frequency of the Bus views, since they were lagging (open the Transit view, then slowly perform the swipe back gesture on iOS; you will see the animation judder about every second, as the view updates)
- It splits the bus lines into three tabs, one per line
  - ~~I think I'm still using my silly icons for the different tabs. I can change that, if anyone cares.~~ Not any longer

---

There are two important architectural changes in this PR:

One is that I've started applying Flow types to the imported data.

```js
import defaultBusLines from '../../../data/bus-times.json'
(defaultBusLines: BusLineType[])
```

Because we're just importing from a JSON file, Flow has no idea what we're bringing in. So, I apply a [Flow typecast](https://flowtype.org/docs/builtins.html#note-about-typecast-syntax) to the imported variable. This allows Flow to properly propagate that type information to the rest of the program. (Side note: we're getting really close to being able to turn Flow on in the CIs. One more RN version and we should be good!!!)

The second change is that in these root components, which deal with the imported data, I've begun passing the data in as a `prop`. This is to allow us to more easily test the components with mock data later, because we can override the default with custom data easily.

This is accomplished by taking advantage of the "defaultProps" special value, which sets a default value for a prop if no value is given (think of a default argument value in c++).

```js
import defaultBusLines from '../../../data/bus-times.json'
(defaultBusLines: BusLineType[])

export class BusView extends React.Component {
  // First, we sketch out the type for the runtime propTypes check
  static propTypes = {
    busLines: React.PropTypes.arrayOf(React.PropTypes.object).isRequired,
  }

  // Then, we set the `static` defaultProps key.
  // The `static` keyword sets the value on the class, instead of a particular instance
  static defaultProps = {
    busLines: defaultBusLines,
  }

  // Then, we specify the flow types to allow us earlier and stronger typechecks
  // Make sure that `propTypes` and `props` check for the same keys;
  // `eslint` will help with a lot of making sure you do this right. (It will complain if
  // you try to access a prop or state value that you haven't declared before.)
  props: {
    busLines: BusLineType[],
  }

  // Finally, you just use it like any other prop, because that's all it is!
  render() {
     return this.props.busLines.map(busLine => busLine.line)
  }
}
```

### DIFF
--- a/data/bus-times.json
+++ b/data/bus-times.json
@@ -1,6 +1,6 @@
 [
   {
-    "line": "Express",
+    "line": "Express Bus",
     "schedules": [
       {
         "days": ["mo", "tu", "we", "th", "fr", "sa"],
@@ -31,7 +31,7 @@
     ]
   },
   {
-    "line": "Red",
+    "line": "Red Line",
     "schedules": [
       {
         "days": ["su", "mo", "tu", "we", "th", "fr", "sa"],
@@ -58,7 +58,7 @@
     ]
   },
   {
-    "line": "Blue",
+    "line": "Blue Line",
     "schedules": [
       {
         "days": ["su", "mo", "tu", "we", "th", "fr", "sa"],
--- a/views/transportation/bus/bus-line-title.js
+++ b/views/transportation/bus/bus-line-title.js
@@ -1,0 +1,26 @@
+// @flow
+import React from 'react'
+import {View, StyleSheet, Text} from 'react-native'
+
+let styles = StyleSheet.create({
+  busLineTitle: {
+    paddingTop: 5,
+    paddingBottom: 5,
+    paddingLeft: 15,
+    borderBottomWidth: StyleSheet.hairlineWidth,
+    borderColor: '#c8c7cc',
+  },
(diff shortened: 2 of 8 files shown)


drewvolz (approved):
👍 :shipit:

hawkrives:
Oh! Right. I also made this look more android-y on Android.

Screenshots incoming.

hawkrives:
The Android settings app:
screen shot 2016-11-18 at 6 03 34 pm

Since it has space between the lists, I mimicked that on the Android view (bottom):
screen shot 2016-11-18 at 6 07 07 pm

There's no space above the top of the list:
screen shot 2016-11-18 at 6 07 10 pm

hawkrives:
My only qualm is that the Android 7 Settings app has no space between lists, nor at the bottom of the view, but seeing as nothing _else_ on Android can agree, either, and this looks a _lot_ more like Android than the current version, I think it's an improvement.

screen shot 2016-11-18 at 6 10 30 pm

PR #346, opened 2016-11-21
Removing unused About file and references

drewvolz:
We do not use this file as our AboutView. Removing to clean up the base.


hawkrives (approved):
Nothing uses it, and it kinda got subsumed into SettingsView anyway. 

:shipit:

PR #347, opened 2016-11-21
Added device info in support email

drewvolz:
* Adds brand, model, device id, system name, version, and app version to emails sent to support
* Adds `react-native-device-info` to the project

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -135,6 +135,7 @@ android {
 }
 
 dependencies {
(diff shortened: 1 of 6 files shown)


hawkrives (approved):
I see this as a valuable change. @elijahverdoorn? Since you're on the group that will actually get these emails?

elijahverdoorn:
Yeah @hawkrives, I totally think this is good, as I see you assumed.  Good change, @drewvolz.

PR #348, opened 2016-11-22
Change bus name string comparison

drewvolz:
Change Express Line to Express Bus


drewvolz:
The color was red when it should have been green.

**Red on green**
red
**Green on green**
green

hawkrives:
Can you post an After screenshot, too, Drew?

drewvolz:
Updated both before and after to make the change clear.

hawkrives (approved):
:shipit:

PR #349, opened 2016-11-22
Remove usable prop on Views

drewvolz:
I believe it is time to toss this prop out as we removed the styling for it in September as of [#211.](https://github.com/StoDevX/AAO-React-Native/commit/8f8e7afbdda096f3c51246125df975b8dd1853f3#diff-80e9d8542d380a15519f77778bd9c1c1) 

We are no longer rendering styles based upon the `usable` prop on Views. This prop is a relic from development when these buttons were mostly non-functional, providing us a way to know what worked and what did not.

### DIFF
--- a/views/home.js
+++ b/views/home.js
@@ -21,20 +21,20 @@ import * as c from './components/colors'
 const Dimensions = require('Dimensions')
 let Viewport = Dimensions.get('window')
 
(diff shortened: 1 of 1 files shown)


hawkrives (approved):
:shipit:

PR #353, opened 2016-11-22
Change building hours closing text to say Midnight

drewvolz:
We should replace 12:00am with Midnight


hawkrives:
After

screen shot 2016-11-22 at 4 11 54 pm

hawkrives (approved):

PR #352, opened 2016-11-22
Buildings Closing In Countdown

mattk410:
Added 'Closing In' countdown to buildings. Got the ListView to refresh automatically and on pull-down.

PR #360, opened 2016-11-23
Report "minute" sometimes

hawkrives:
Say "minute" when there is only one minute

Addresses @drewvolz's feedback from #352.

### DIFF
--- a/views/building-hours/format-building-hours.js
+++ b/views/building-hours/format-building-hours.js
@@ -18,7 +18,7 @@ export function formatBuildingHours(hoursInfo: BuildingInfoType, now: momentT): 
   if (openStatus === 'Almost Closed') {
     let closingIn = close.diff(now, 'minutes') + 1
 
-    return `Closing in ${closingIn} minutes`
+    return `Closing in ${closingIn} ${closingIn === 1 ? 'minute' : 'minutes'}`
   } else {
     let openString = open.format(RESULT_FORMAT)
     let closeString = close.format(RESULT_FORMAT)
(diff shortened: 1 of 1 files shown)


hawkrives:
Actually.

I did some research in the Moment docs.

https://runkit.com/hawkrives/5835bcd2472b6500147cc597
http://momentjs.com/docs/#/displaying/to/

```js
const moment = require('moment')

let now = moment()
let in5 = now.clone().add(5, 'minutes')
let in1 = now.clone().add(1, 'minutes')
let past10 = now.clone().subtract(10, 'minutes')

console.log("Closes " + now.to(in5))
console.log("Closes " + now.to(in1))
console.log("Closes " + now.to(past10))
```

```
"Closes in 5 minutes"
"Closes in a minute"
"Closes 10 minutes ago"
```

In short, we can replace

```js
if (openStatus === 'Almost Closed') {
  let closingIn = close.diff(now, 'minutes') + 1
  
  return `Closing in ${closingIn} minutes`
}
```

with

```js
if (openStatus === 'Almost Closed') {
  return `Closing ${now.to(close)}`
}
```

> (because `now` and `close` already exist in the scope)

That will return `Closing in X minutes` or `Closing in a minute`.

The only question is, how much do we dislike `a minute` versus `1 minute`?

drewvolz:
Using "a minute" sounds fine to me. I don't mind it, and I like that `moment` covers that case.

drewvolz (approved):
👍 :shipit:

hawkrives:
Alright. I've updated this PR to use moment for formatting that case.

OmarShehata:
"in a minute" sounds more natural anyway I think.

Issue #220, opened 2016-09-07
ErrorView state/view not refreshed unless view is popped in special case


drewvolz:
If you load the Financials and Courses views while the view ErrorView is present, then your view of either the Courses or Financials view is blocked after you log-in with the ErrorView on the stack. The state/view of the screen you are _not_ on does not get refreshed after logging in -- only the view that we log-in on is loaded. The user has to pop the navigator and return back to the SIS view if they would like to see and update that view. 

**In short:**
Home -> SIS -> tap both tabs -> hit 'open settings' -> login -> tap the other tab -> can't access -> go back home -> load SIS -> can access

This is unaffected by actually heading to Settings and logging in.

drewvolz:
Not a shipping bug -- does not cause a crash, or loss of data. Simply an annoyance in an edge case. I say ship our current bug fixes without this one for now.

hawkrives:
Oh drat. Yes, that's an unintentional side effect of how it's designed. 

I'll think about it on the way in.

hawkrives:
Has this been avoided by pushing a view over the entire screen from the bottom? I think this is no longer an issue.

hawkrives:
Nope, still an issue.

hawkrives:
I'm taking this – I think it will be resolved by one of my redux branches, when I move login state into a global store, so both screens can access a canonical state.

PR #378, opened 2016-11-27
Allow home view to reorder

drewvolz:
* Adds a view to reorder list of main views on home view
* Saves order to async storage to use back on the home view
* Reorders home view with any changes made
* Adds @hawkrives' WIP async storage higher-order component

- [x] Works on Android
- [x] Works on iOS

opening-in

### DIFF
--- a/views/editHome.js
+++ b/views/editHome.js
@@ -45,6 +45,8 @@ class Row extends React.Component {
     style: {
       shadowRadius: new Animated.Value(2),
       transform: [{scale: new Animated.Value(1)}],
+      opacity: new Animated.Value(1.0),
+      elevation: new Animated.Value(2),
     },
   }
 
@@ -76,6 +78,16 @@ class Row extends React.Component {
         easing: Easing.out(Easing.quad),
         toValue: 10,
       }),
+      Animated.timing(style.opacity, {
+        duration: 100,
+        easing: Easing.out(Easing.quad),
+        toValue: 0.65,
+      }),
+      Animated.timing(style.elevation, {
+        duration: 100,
+        easing: Easing.out(Easing.quad),
+        toValue: 4,
+      }),
     ]).start()
   }
 
@@ -88,6 +100,16 @@ class Row extends React.Component {
         toValue: 1,
       }),
       Animated.timing(style.shadowRadius, {
+        duration: 100,
+        easing: Easing.out(Easing.quad),
+        toValue: 2,
+      }),
+      Animated.timing(style.opacity, {
+        duration: 100,
+        easing: Easing.out(Easing.quad),
+        toValue: 1.0,
+      }),
+      Animated.timing(style.elevation, {
         duration: 100,
         easing: Easing.out(Easing.quad),
         toValue: 2,
@@ -157,6 +179,7 @@ let styles = StyleSheet.create({
     shadowOpacity: 1,
     shadowOffset: {height: 2, width: 2},
     shadowRadius: 2,
+    opacity: 1.0,
     elevation: 2,
   },
   rectangleButtonIcon: {
--- a/views/editHome.js
+++ b/views/editHome.js
@@ -139,7 +139,7 @@ function EditHomeView(props: {
       contentContainerStyle={styles.contentContainer}
       data={objViews}
       order={props.order}
-      onChangeOrder={(order: {[key: string]: ViewType}) => props.onSaveOrder(order)}
+      onChangeOrder={(order: ViewType[]) => props.onSaveOrder(order)}
       renderRow={({data, active}: {data: ViewType, active: boolean}) =>
         <Row data={data} active={active} />}
     />
--- a/views/editHome.js
+++ b/views/editHome.js
@@ -13,7 +13,6 @@ import {
   Navigator,
   StyleSheet,
   Text,
-  View,
 } from 'react-native'
 
 import * as c from './components/colors'
--- a/views/editHome.js
+++ b/views/editHome.js
@@ -131,12 +131,8 @@ export default class EditHomeView extends React.Component {
 
   loadData = async () => {
     this.setState({loaded: false})
-
     let savedOrder = JSON.parse(await AsyncStorage.getItem('homescreen:view-order'))
-
-    // check to see if we have a modified view order or not
     savedOrder = savedOrder || Object.keys(objViews)
-
     this.setState({loaded: true, order: savedOrder})
   }
 
@@ -177,7 +173,6 @@ let styles = StyleSheet.create({
     marginHorizontal: 15,
     paddingVertical: 12,
     borderRadius: 4,
-
     shadowColor: 'rgba(0,0,0,0.2)',
     shadowOpacity: 1,
     shadowOffset: {height: 2, width: 2},
@@ -186,13 +181,11 @@ let styles = StyleSheet.create({
   },
   rectangleButtonIcon: {
     marginRight: 20,
-    // borderRadius: 20,
     color: c.white,
     paddingLeft: 10,
     paddingRight: 10,
   },
   listButtonIcon: {
-    // borderRadius: 20,
     color: c.black,
     paddingLeft: 10,
     paddingRight: 10,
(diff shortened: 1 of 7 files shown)


drewvolz:
@hawkrives pointed out that making use of `componentWillReceiveProps` as used within this PR might allow us to solve #220

views/editHome.js
```diff
@@ -0,0 +1,207 @@
...
+    if (!this.state.loaded) {
+      return <LoadingView />
+    }
+
+    return (
+      <SortableList
+        contentContainerStyle={styles.contentContainer}
+        data={objViews}
+        order={this.state.order}
+        onChangeOrder={this.onOrderChange}
```
elijahverdoorn: I wonder if it would be better to store the order of the items in memory, then call the `AsyncStorage` stuff when the user hits the done button? Seems unnecessary to be writing to storage every time they change one when they only see the result when they hit done. (I might be missing a very good reason for it to work this way!)

views/home.js
```diff
@@ -37,51 +44,94 @@ const views: ViewType[] = [
   {view: 'OlevilleView', title: 'Oleville', icon: 'mouse-pointer', tint: c.grapefruit},
 ]
 
+const Touchable = ({children, onPress}: {onPress: () => any, children?: Element<any>}) => {
+  return Platform.OS === 'ios'
+    ? <TouchableOpacity onPress={onPress} activeOpacity={0.65}>{children}</TouchableOpacity>
+    : <TouchableNativeFeedback onPress={onPress}>{children}</TouchableNativeFeedback>
```
elijahverdoorn, line 36: This is nice. Maybe we should extract the logic of `TouchableOpacity` vs `TouchableNativeFeedback` out to a component?

views/editHome.js
elijahverdoorn: I wonder if it would be better to store the order of the items in memory, then call the AsyncStorage stuff when the user hits the done button? Seems unnecessary to be writing to storage every time they change one when they only see the result when they hit done. (I might be missing a very good reason for it to work this way!)

views/home.js
elijahverdoorn: By this I mean in the `/components` folder so that we can use it elsewhere where we now just use `TouchableOpacity`.

elijahverdoorn (commented):
Looks really good overall, just a few suggestions/questions in the review.

views/editHome.js
drewvolz: That is a fair suggestion, and should be easily doable. You are making a decent suggestion -- not missing any point. Thanks.

views/home.js
drewvolz: @hawkrives thoughts? Any drawbacks from extracting into our own component to represent the two?

drewvolz:
@hawkrives tested this one out on Android, and it crashes when reordering at this point in time.

views/home.js
hawkrives: That's a good idea. I've got a similar component built in the BuildingHours PR. 

I'll take a look at extracting it out.

views/editHome.js
drewvolz: I have tried this now -- saving it to memory, and then calling AsyncStorage stuff when the user hits the done button. In short, it is not quick enough. 

There's no downside to saving on reorder. It's fast and asynchronous, it's just saving a list of strings. There are upsides -- it saves the order on crash of that view... which is nice if you spent a bit of time thinking and reordering, or if the app suddenly dies (as is the case of Android).

You have pointed out something that needs to change as a whole, as using `componentWillReceiveProps()` may not be the correct way to do this... it may not fire every time. 

**So** the correct answer here is "use redux or some other method of lifting application state above react", as confirmed by @hawkrives

drewvolz:
@hawkrives and I did research and a lot of testing as to why Android fails with this component. The problem occurs when we try to modify `zIndex` in `SortableList` within `react-native-sortable-list`.

[SortableList.js#L193](https://github.com/gitim/react-native-sortable-list/blob/master/src/SortableList.js#L193)
```
if (active || released) {
  Object.assign(style, {zIndex: 100});
}
```

[SortableList.js#L176](https://github.com/gitim/react-native-sortable-list/blob/master/src/SortableList.js#L176)

`Object.assign(style, {width: rowWidth, zIndex: 0});
`

A [few open issues](https://github.com/facebook/react-native/issues?utf8=✓&q=is%3Aissue%20is%3Aopen%20zindex) on the React Native project reveal others have seen odd behavior with `z-index`.

There[ is a PR](https://github.com/facebook/react-native/pull/10954) that attempts to fix this, maybe? I tried applying their changes to my copy of React Native Android but I still saw the same issue.

> Commit 3d3b067 implemented zIndex support for Android. However after zIndex resorting of the views, the JS doesn't know about the new order. When dynamically removing the view from an array of views, JS sends the command to remove the view at certain index, which is not the same in Java since the order of items changed.

drewvolz:
I am now patching the `react-native-sortable-list` project by hosting a copy of the package [on my own GitHub](http://github.com/drewvolz/react-native-sortable-list). 

The patch simply comments out the `zindex` modifications that occurred when rows in the list moved. All that provided was a sense of depth, putting one element on top of the other.

That is not a sufficient reason to delay this functionality as a whole, so I will watch it for changes in the near future. I would consider this review-worthy and approval worthy at this point.

drewvolz:
I've added `easing` animations for the `opacity` and `elevation` styles, as well as re-enabled the `zindex` changes for iOS devices. @hawkrives and I have concluded that the rows' opacity combined with the drop shadow is enough of a 3D replacement for now.

hawkrives:
… I hadn't thought of platform-checking for the zIndex thing. Good idea!

I'll pull and take another look locally.

hawkrives (approved):
We're positive that the stock Android crashes are from a React Native bug.

This workaround's not _perfect_, but it's workable, IMHO.

hawkrives:
To expound on how redux works a bit, and why it's a good thing:

Our current app setup requires each component to manage its own state: download its own data, organize it, persist it if needed, and reload it from the persisted datastore.

That's a lot of code to include in every component, and it makes it tricky to communicate between two components — for example, the two screens under the SIS view; when one logs in, it can't tell the other tab to update.

---

Redux is an improvement because it follows what it calls the "three principles": 

1. a [single source of truth][sst], 
2. the [state is read-only][sro], except by emitting actions that describe what happened
3. all [changes must be made by pure functions][pure].

With Redux, instead of each component being responsible for both visually rendering some data, _and_ managing the lifecycle of that data, the React components are only responsible for rendering the data they are given. Management of the data is handed over to the "store", which is a representation of the state of the app at any given time. 

The components themsevles never actually update the store, though. They call "actions", which describe how to update the store. Those "actions" are then dispatched to the "store" by Redux, and the store handles updating itself from the contents of the actions.

When the store updates, it returns the new state to Redux. Redux knows which components are subscribed to which parts of the store, and it provides the new values from the state as props to those components.

```
Component -> Action -> Store -> Component
```

This all harks back to the core of React, IMHO: that a component is the function (in the mathematical sense) of its props and its state. Props are easier to reason about than state, though, since the component itself can never modify them. When Redux extracts some of the state of the component and provides it as props, then, it simplifies that data flow.

[sst]: http://redux.js.org/docs/introduction/ThreePrinciples.html#single-source-of-truth
[sro]: http://redux.js.org/docs/introduction/ThreePrinciples.html#state-is-read-only
[pure]: http://redux.js.org/docs/introduction/ThreePrinciples.html#changes-are-made-with-pure-functions

---

With Redux, the SIS view would not load `credentials:valid` from AsyncStorage in componentWillMount; instead, it would subscribe to the (hypothetical) `store.settings.credentials.valid` value, probably provided as a `credentialsValid` prop (via the `mapStateToProps` function). Then, when someone is logged in, the value is `true`; if they're logged out, it's `false`. When one of the SIS data-fetching actions fails due to token expiry, the "store" updates `settings.credentials.valid` because it knows it needs to when it handles a … flex dollar fetching failure, for example.

---

The "single source of truth" that a Redux store implies is wonderful, IMHO.

Only having one way for data to change in the store, and one way to get the data out, removed _so many_ bugs in Gobbldygook. 

I'm going to stop writing this before it becomes a novel, but I can expound on any confusing bits if requested.

elijahverdoorn:
I was wondering how long it would take for us to add Redux. Good work @hawkrives and @drewvolz, this is awesome.

PR #380, opened 2016-11-28
Tweaks

hawkrives:
This PR does a bunch of things, but I really don't want to make separate PRs for each one.

# Fix Bus View
There was a bug in the BusView that would result in the code using the "last bus of the day" logic during the period between routes.

For instance, between 8:12pm and 8:15pm, it would say "Last Route", instead of the correct status of having three laps left.

This fixes that.

## Fix Flow
With this PR, there are 0 flow errors.

```console
$ flow

> AllAboutOlaf@2.0.1 flow /Users/hawken/AAO-React-Native
> flow

No errors!
```

This means that we can actually enable Flow on Travis and enfore type correctness. :soon:

## Fix Eslint Warnings
With this PR, there are no more lint warnings.

```console
$ npm run lint

> AllAboutOlaf@2.0.1 lint /Users/hawken/AAO-React-Native
> eslint app.js index.*.js lib/ views/

```

## Others
- I made a `<Separator />` component, for rendering separators between items in ListViews.
- I reworked the Settings button in app.js to remove the need for a global variable
	- This also allows other views to create from-bottom views with custom logic to dismiss themselves, by simply adding an `onDismiss` function to the object they push to the navigator.


drewvolz (approved):
I've gone through and looked at the changes as well as verified `flow` and `lint` being clear, and the bus schedule bug being fixed. Nice work. 👍 :shipit:

PR #386, opened 2016-11-29
Update fastlane

hawkrives:
This adds two fastlane features:

1. Local builds will now automatically pick up the next version number from HockeyApp, instead of being set to 1.
2. There is now (theoretically) a `bundle exec fastlane register` command, which _should_ get new, unauthorized devices from HockeyApp and go generate a new profile for them from Apple.

Tl;dr: Hawken want; plz merge.


drewvolz (approved):
:+1: :shipit: great changes

PR #393, opened 2016-11-30
Fancy Menus

### LINKED ISSUE #320: Cage menu overridden by new cage menu
I really like the new menus and think that we should extend it to the cage, but the changes have overridden the standing items on the cage menu. We should re-enable those somehow, I know that people use that section.

hawkrives:
> This **should not** be a blocker for this release. I ~~will~~ can split the visual tweaks out from the rest of this mess, and I _would_ like to get those into this release.

## Remaining issues
- [x] [question] Are there items on our Cage menu that aren't on the online cage menu?
  - Resolution: it is what it is. We should bug BonApp if their data is wrong.
- [x] [data] Make the Pause menu in Bonapp format
- [x] ~~[code] Build the Carleton Cafs list screen~~
- [x] ~~[code] Make the filter view faster~~
- [x] [bug] Fix prices separated by HTML entities (`&nbsp;`, namely)
- [x] [bug] Fix stav being blank at 4:39pm
- [x] [bug] Fix iOS listviews hiding behind the tab bar
- [x] [bug] Fix dietary filters so they work (thanks @mattk410)
- [x] [bug] Filter by station doesn't work if "only specials" is toggled on (thanks @mattk410)
- [x] [bug] Show correct menu name in filter toolbar
- [x] [bug] Hide filter sections when they are blank
- [x] [feature] Show station description notes in the list somewhere

## This PR
… reworks Matt's bonapp-menu parsing stuff.

- It picks a menu based on the times provided by Bonapp
- It allows filtering of the menu items by category
- Adds support for the Carleton caf menus (all 4 of them)
- Migrates the Pause menu to the Bonapp format, and thus the same `<Menu/>` component

This is what happens when I have too much time on my hands… which I don't have, so I'm not sure how this happened 😆


drewvolz:
* I would prioritize fixing the filters (dietary / stations), and providing the pause menu in the same format.
* If we had to pick something to slash in a time crunch, perhaps slash Carleton menus for now. 
* I would be happy to take the pause menu one and implement it as I know we need it ASAP.

hawkrives:
@drewvolz If you could get the Pause, that'd be great.

I'll push up another commit that removes Carleton from the tabs again.

I can point at the code that needs changing for the filters, but I don't have time today/tomorrow to fix it myself.

I will also push a commit to remove prices from the menu, per an out-of-band discussion with @drewvolz.

hawkrives:
Well, I _said_ I didn't have time, but I guess what I meant was that I didn't have time to test it. The last commit there (d8d7bf0) fixed up the filters so they should be able to apply multiple filters at once (eg, only gluten-free specials).

hawkrives:
The next thing is, "Fix dietary filters so they work".

If I had to hazard a guess, there's probably some alphabetic case difference between the dietary filters we provide from the filter, and the ones in `item.cor_icon`.

I'd advise tossing some `log`s in around line 144 of fancy-menu.js ([`45c2adbb#fancy-menu.js:144`](https://github.com/StoDevX/AAO-React-Native/pull/393/files#diff-45c2adbb46d30dbd33d54edbaaee17a2R144)) to check.

hawkrives:
The next thing is, "Fix dietary filters so they work".

If I had to hazard a guess, there's probably some alphabetic case difference between the dietary filters we provide from the filter, and the ones in `item.cor_icon`.

I'd advise tossing some `log`s in around line 144 of fancy-menu.js to check.

hawkrives:
The dietary filters appear to have been fixed with d8d7bf0. 

I just pushed a commit to show the correct meal name in the toolbar, instead of always showing "Lunch".

hawkrives:
I am going to make a new beta build.

If you'd like to test the beta on your devices, I just need an email. (cc @StoDevX/all-about-olaf).

hawkrives:
I just pushed a change so that it only shows "special" items when first loading.

hawkrives:
This last change tweaks how the filter view filters dietary restrictions.

Before, if you wanted only "gluten-free" and had unchecked "vegan", it would show you items that were both gluten-free _and_ vegan.

But, if you're filtering, you expect that unchecking a line will hide _all_ items that match that filter.

Now, if you want only gluten-free items, you only see gluten-free items. Nothing else.

:-)

hawkrives:
As of a375227, `<Filter/>` now hides empty sections correctly.

hawkrives:
Drew has now incorporated the Pause's online menu into a file for us.

The only thing left is to render the station notes in the view somewhere. Then this PR will be ready for final review.

hawkrives:
I'm going to work on splitting this PR into smaller chunks, for easier review. I will comment here when that happens.

PR #397, opened 2016-12-02
Enable flow on travis

hawkrives:
Flow, in addition to catching type problems, also catches missing function calls.

It is how I noticed the missing definitions in #335.

This makes it part of the Travis flow.


drewvolz (approved):
:+1: :shipit:

PR #408, opened 2016-12-05
Moves button check to global scope

drewvolz:
This got shuffled around in a reorganization of files a bit ago, and belongs in the global scope in order to retain its value.

### DIFF
--- a/app.js
+++ b/app.js
@@ -146,8 +146,9 @@ const styles = StyleSheet.create({
   },
 })
 
+let settingsButtonActive = false
+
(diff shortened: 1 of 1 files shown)


hawkrives:
It's not so much that it *belongs* in the global scope, as … the function that was wrapping it gets re-called each time the Left/RightButton is pushed. 

So it gets a new copy of the variable every time it's pushed, as the button rerenders. 

Until we can figure out a way around __that__, this will have to live in global scope.

hawkrives (approved):

PR #416, opened 2016-12-06
Small tweaks

hawkrives:
This PR is made up of two changes:
- reduce font weight of ios navbar
- darken app tint color

We have been using a heavier font weight (`700`, or `bold`) in our navbar since … well, since always. That's because my brain said that 600 == bold, not 700, but iOS uses 600 for navbars, not 700.

And secondly, the app's tint color is too light IMHO. This changes it to the same color used for tinting the app in other places, most noticeably the iOS tab bars. This makes it much easier to read the titlebar, especially when night shift is enabled, IMO.

diff-1

| Before | After |
|:---:|:---:|
before | after

---

Comment period: until someone from Olaf comments.

### DIFF
--- a/app.js
+++ b/app.js
@@ -89,7 +89,7 @@ const styles = StyleSheet.create({
     backgroundColor: Platform.OS === 'ios' ? c.iosLightBackground : c.androidLightBackground,
   },
   navigationBar: {
-    backgroundColor: c.olevilleGold,
+    backgroundColor: c.tint,
     ...navbarShadows,
   },
   backButton: {
--- a/app.js
+++ b/app.js
@@ -131,7 +131,7 @@ const styles = StyleSheet.create({
     color: 'white',
     fontFamily: Platform.OS === 'ios' ? 'System' : 'sans-serif-light',
     fontSize: Platform.OS === 'ios' ? 17 : 20,
-    fontWeight: Platform.OS === 'ios' ? 'bold' : '500',
+    fontWeight: Platform.OS === 'ios' ? '600' : '500',
     marginVertical: Platform.OS === 'ios' ? 12 : 14,
   },
   rightButton: {
--- a/app.js
+++ b/app.js
@@ -89,7 +89,7 @@ const styles = StyleSheet.create({
     backgroundColor: Platform.OS === 'ios' ? c.iosLightBackground : c.androidLightBackground,
   },
   navigationBar: {
-    backgroundColor: c.tint,
+    backgroundColor: c.olevilleGold,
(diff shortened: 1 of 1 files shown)


drewvolz:
~~@hawkrives I know it might be a challenge, but would you be able to photograph what you consider difficult to read with your nightshift enabled? I know screenshotting will not work for this task.~~

~~I believe it may be hard to read on your particular device with a certain tint of nightshift, but I am not completely convinced that it is warranted to adjust the tint to be darker if it is readable on other settings of nightshift.~~

~~Lastly, I believe your tint change is only relevant on one platform... someone would have to apply a blue-light filter to an android device?~~

~~How does everyone else in @StoDevX/all-about-olaf else feel about these changes? I personally am a fan of the previous `olevilleGold` color, as it was extracted from St. Olaf's website, but if it really is beneficial, I could be persuaded.~~

drewvolz (approved):
I have had a change of mind. It is just a color, and if @hawkrives has a preference and would like to change it, that is fine by me.

hawkrives:
Hawken would like to change it.

If anyone wants to _not_ change it, then I'll remove that change.

app.js
```diff
@@ -89,7 +89,7 @@ const styles = StyleSheet.create({
     backgroundColor: Platform.OS === 'ios' ? c.iosLightBackground : c.androidLightBackground,
   },
   navigationBar: {
-    backgroundColor: c.olevilleGold,
+    backgroundColor: c.tint,
```
elijahverdoorn: What if we change it to `c.olevilleGold`? It's a bit darker, but not as dark as the `tint`

elijahverdoorn (changes requested):

app.js
drewvolz: The proposal is to change it to a darker color named tint away from OlevilleGold. It is more orange than gold. For context, [here is the PR ](https://github.com/StoDevX/AAO-React-Native/pull/416).

elijahverdoorn: Wow I'm blind. Thanks @drewvolz. I think I'd like to keep in the same color family as `olevilleGold`, but if we want to go with a darker variation that's ok. Maybe one of these:
* `#C28F00`
* `#987000`

@hawkrives what do you think of those?

hawkrives: olevilleGold | tint (mandarin) | `#C28F00` | `#987000`
---|---|---|---
tint mandarin | olevillegold |  | c28f00

hawkrives:
Hawken is dropping the color change part of this PR.

PR #423, opened 2016-12-09
updated flow typedefs

hawkrives:
The flow typedefs needed updating.

I committed this as a different user so I don't get these giant files added to my stats.


hawkrives:
Looks like I need to do some work with flow types on master, first. https://travis-ci.org/StoDevX/AAO-React-Native/jobs/182642673#L246-L266

PR #426, opened 2016-12-11
Fix blank ordering on init of the Edit Home view

drewvolz:
* The order passed in would remain forever blank if we did not initialize it first.
* This was changed by accident during the redux/flux additions.
* Adding back a default order in place of a blank one solves this.

### DIFF
--- a/flux/index.js
+++ b/flux/index.js
@@ -6,7 +6,7 @@ import {AsyncStorage} from 'react-native'
 import {allViews} from '../views/views'
 
 export const SAVE_HOMESCREEN_ORDER = 'SAVE_HOMESCREEN_ORDER'
-let defaultViewOrder = allViews.map(v > v.view)
+let defaultViewOrder = allViews.map(v => v.view)
 
 export const loadHomescreenOrder = async () => {
   let savedOrder = JSON.parse(await AsyncStorage.getItem('homescreen:view-order')) || defaultViewOrder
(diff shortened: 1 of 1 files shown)


hawkrives (commented):
All we need is the list of view IDs, with? `view.view`?

What about doing 

```js
let defaultViewOrder = allViews.map(v > v.view)
```

and then using that in the `||`, instead of Object.keys?

drewvolz:
@hawkrives I will try using that instead. Would you want it defined only within `loadHomescreenOrder` or is the current placement okay with you?

hawkrives:
It can be outside the function, sure.

flux/index.js
```diff
@@ -3,11 +3,13 @@ import createLogger from 'redux-logger'
 import reduxPromise from 'redux-promise'
 import reduxThunk from 'redux-thunk'
 import {AsyncStorage} from 'react-native'
+import {allViews} from '../views/views'
 
 export const SAVE_HOMESCREEN_ORDER = 'SAVE_HOMESCREEN_ORDER'
+let defaultViewOrder = allViews.map(v > v.view)
```
hawkrives: `=>`?

hawkrives (changes requested):

flux/index.js
drewvolz: What are you asking here? I haven't had a chance to run anything, so I went with your changes listed above.

hawkrives: Ah. My typo, then. Apologies.

You'll need `=>`, instead of `>`.

This is currently doing `allViews.map(false)`, essentially, because we've got a boolean comparison instead of a lambda function.

drewvolz: Thanks for the explanation. I'll fix that up.

hawkrives (approved):

PR #437, opened 2016-12-20
Fix new home additions

drewvolz:
The home screen became editable as of a few PRs ago! You can reorder the views to your liking, or to your friend's disliking!

So what happens when we add a new view to the home screen? It doesn't show in the list of views you can reorder. Oh no... how do we fix this? Well, we could check if we have a differing amount of views between what we _know_ is your preferred order, and the list of views we know are displaying by default.

That seems to work pretty well! We can now get away with safely adding or removing views from the home screen without any issues. This is good news, because we want the user to be able to reorder all of the views, not just some of them.

This was a combo effort by @hawkrives and myself. We even have tests to prove it works (thank you hawken).

tl;dr: adding/removing views did not work with the edit stuff, and this fixes that.

### DIFF
--- a/flux/__tests__/update-view-order.test.js
+++ b/flux/__tests__/update-view-order.test.js
@@ -1,0 +1,66 @@
+/* eslint-env jest */
+import {updateViewOrder} from '../index'
+
+test('it should return the current order if no screens have changed', () => {
+  let savedOrder = ['a', 'b', 'c']
+  let defaultOrder = savedOrder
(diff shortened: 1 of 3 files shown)


hawkrives:
This will also do its best to maintain the sort order that you had before we changed things around.

It will remove views from their spots in the sorted list, and add any new ones to the end.

`saved: [a b c]` `new: [a c d]` -> `result: [a c d]`

hawkrives (approved):
LGTM

PR #452, opened 2016-12-23
𝕱𝖆𝖓𝖈𝖞𝕸𝖊𝖓𝖚𝖘 ① — 𝑻𝒉𝒆 𝑭𝒂𝒏𝒄𝒚𝒊𝒏𝒈

### LINKED ISSUE #320: Cage menu overridden by new cage menu

hawkrives:
Hello there! It's nearly done!

This PR breaks Matt's nice menu screens into three parts:
- `<Menu/>`
- `<FancyMenu/>`
- `<BonAppHostedMenu/>`

It also applies new styles to the menus.

## `<Menu/>`
Menu is the core of it all. It's what the styles are applied to, and it configures the root `<ListView/>` just the way we need it.

## `<FancyMenu/>`
FancyMenu takes an array of FoodItems, sorts them by station and then by ID, then groups them into an object for the Menu ListView to render.

In a future PR, FancyMenu will also take on responsibility for filtering the food items based on the user's selections.

## `<BonAppHostedMenu/>`
BonAppHostedMenu's only job is to get the menu data from BonApp. It takes a café ID (Stav is 261 and the Cage is 262), then fetches the data from the server and passes it to FancyMenu.

It is in charge of rendering the loading screens, while the data loads. It also shows a "closed" message, if BonApp provides one.

Closes #320
---

Screenshots! Note that this view has not been optimized for Android. There are also item descriptions and highlighting of "special" items.

a | b
---|---
caf | cage
descriptions | specials


hawkrives:
I added rendering of item descriptions, and highlighting of "special" items.

specials | descriptions
---|---
specials | descriptions

drewvolz (dismissed):
Keep these as comments, or remove them completely?

drewvolz (commented):
Keep these as comments, or remove them completely?

elijahverdoorn (approved):
Dang, this is good. Let's do it.

PR #455, opened 2016-12-24
𝕱𝖆𝖓𝖈𝖞𝕸𝖊𝖓𝖚𝖘 ② — 🅃🄷🄴 🅻🅸🅾🅽’🆂 🄿🄰🅄🅂🄴 🅼🅴🅽🆄

hawkrives:
> Depends on #452.
> Do not merge this PR. Once it's approved, and #452 is merged, this PR will be switched to target `master`, instead. Then it can be merged.

This PR edits the Pause menu view. Thanks to @drewvolz for the data.

New components:
- `<GitHubHostedMenu/>`

## `<GitHubHostedMenu/>`
GitHubHostedMenu has the same responsibilities as the BonAppHostedMenu, but instead of fetching from the BonApp API it fetches a `{stationMenus, foodItems, corIcons}` object from GitHub. Or really anywhere on the internet; the file could be hosted on the Pause's servers just as easily.

If the data fetch fails, the app falls back to the local version of the github-hosted pause menu.

This also includes a data-upgrader, because the BonApp data format is very verbose. The upgrader takes a minimal description of the food items and adds the remaining keys, allowing a much simpler data format.

This PR removes the collapsible Pause menu.

---

Screenshots:

a | b
---|---
pause sweets | pause pizza


mattk410 (approved):
This looks great. I like the way you are doing the GitHub menu.

drewvolz (approved):
Very nice job.

elijahverdoorn (approved):
This is good! I think that the menus will finally stabilize with this new system, it seems maintainable.

views/menus/tabs.js
```diff
@@ -36,14 +36,14 @@ const stolaf = [
...
-    //   ],
-    // },
+    component: GitHubHostedMenu,
+    props: {
+      name: 'pause',
+      loadingMessage: [
+        'Mixing up a shake…',
+        'Spinning up pizzas…',
+      ],
+    },
```
hawkrives, line 46: @drewvolz This is why I left those commented in #452 – thinking it might make the diff smaller.

PR #456, opened 2016-12-24
Revert to a good copy of the sortable-list dep

hawkrives:
This fixes stuff.

Drew has promised to not touch this package again unless it actually tries to fix things.


drewvolz (approved):

drewvolz:
Who would have thought that an update would break things?

PR #458, opened 2016-12-28
change a bunch of descriptions in the Pause menu

hawkrives:
Also we should make a `menus` Github label. Or maybe `data`.

### DIFF
--- a/docs/menus/pause.json
+++ b/docs/menus/pause.json
@@ -1,6 +1,6 @@
 {
   "stationMenus": [
-    {"label": "Pizza", "note": "Includes a cup of Marinara, Ranch, BBQ, or Honey Mustard sauce. For orders of 4 or more pizzas, please visit Oleville 24 hours in advance."},
+    {"label": "Pizza", "note": "For orders of 4 or more pizzas, please visit Oleville 24 hours in advance."},
     {"label": "Specialty Pizza"},
     {"label": "Toppings"},
     {"label": "Quesadilla", "note": "Salsa & Sour Cream are extra"},
--- a/docs/menus/pause.json
+++ b/docs/menus/pause.json
@@ -1,17 +1,38 @@
 {
+  "stationMenus": [
+    {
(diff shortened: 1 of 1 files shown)


docs/menus/pause.json
```diff
@@ -1,17 +1,38 @@
...
+      "label": "Shakes",
+      "note": "Mixins: M&Ms, Oreos, Reese’s Peanut Butter Cups"
+    },
+    {
+      "label": "Soft-Serve",
+      "note": "Hand-swirled in a cup or a cone",
+    },
+    {
+      "label": "Pizza",
+      "note": "Includes a cup of Marinara, Ranch, BBQ, or Honey Mustard sauce"
```
drewvolz: This is the note for Chicken Strips. Perhaps we do not need a note here.

docs/menus/pause.json
```diff
@@ -100,39 +121,30 @@
...
       "station": "Quesadilla"
     },
     {
+      "description": "Pizza built on pita bread; includes two toppings",
       "label": "Pita Pizza",
       "special": true,
       "station": "Pita"
     },
     {
+      "description": "Pizza built on pita bread; includes two toppings",
```
drewvolz: I do not think "specialty" items include an "extra two toppings". We need to check with a kitchen manager or worker to be sure.

drewvolz (changes requested):

docs/menus/pause.json
drewvolz: On another note, large orders are a pain-point for the pause. We could use the note here to communicate "For orders of 4 or more pizzas, visit Oleville"

hawkrives: Can do

hawkrives:
@drewvolz I addressed your comments.

(I need to merge this soon, so my stationMenus stuff can be tested before I submit the filters PR)

drewvolz (approved):

PR #463, opened 2016-12-31
Thumbnails on news stories

drewvolz:
This adds 50x50 thumbnails to stories with images embedded into their content. It seeks out the first image within the story content and uses it as the image's source. Stories without images will display as they did previously.

Screenshots:

Before | After
---|---
no-thumbs |  thumbs

### DIFF
--- a/views/news/news-container.js
+++ b/views/news/news-container.js
@@ -23,8 +23,9 @@ const entities = new AllHtmlEntities()
 
 type NewsRowPropsType = {
   story: StoryType,
-  onPress: (t: string, s: StoryType) => any,
+  onPress: (title: string, story: StoryType) => any,
 };
+
 class NewsRow extends React.Component {
   state = {
     thumbnailUrl: null,
@@ -46,7 +47,7 @@ class NewsRow extends React.Component {
       const url = this.findImage(props.story['content:encoded'][0])
       // if we didn't find a valid URL, return null
       if (!url) {
-        resolve(null)
+        return null
       }
 
       // Image.getSize is a callback-based API
--- a/views/news/news-container.js
+++ b/views/news/news-container.js
@@ -45,12 +45,12 @@ export default class NewsContainer extends React.Component {
     let re = '<img.*?src="([^"]*)"[^>]*>(?:</img>)?'
     let imageMatch = content.match(re)
 
-    if (imageMatch !== null && imageMatch !== undefined) {
+    if (imageMatch) {
       let src = imageMatch[1]
       return <Image source={{uri: src}} style={styles.image} />
     }
 
-    return false
+    return null
   }
 
   fetchData = async () => {
--- a/views/news/news-container.js
+++ b/views/news/news-container.js
@@ -42,12 +42,14 @@ export default class NewsContainer extends React.Component {
   };
 
   findImage = (content: string) => {
-    let re = '<img.*?src="([^"]*)"[^>]*>(?:</img>)?'
-    let imageMatch = content.match(re)
+    let reUrl = /^https?:\/\/[^\s/$.?#].[^\s]*$/
+    let reImg = /<img.*src=["'](.*?)["'].*?>/g
+    let images = content.match(reImg)
 
-    if (imageMatch) {
-      let src = imageMatch[1]
-      return <Image source={{uri: src}} style={styles.image} />
+    if (images) {
+      let imageUrls = images.map(tag => reImg.exec(tag))
+      let validUrls = imageUrls[0].filter(url => reUrl.test(url))[0]
+      return <Image source={{uri: validUrls}} style={styles.image} />
     }
 
     return null
--- a/views/news/news-container.js
+++ b/views/news/news-container.js
@@ -2,118 +2,19 @@
 import React from 'react'
 import {
   StyleSheet,
-  Image,
   ListView,
   Platform,
   RefreshControl,
 } from 'react-native'
-import {fastGetTrimmedText} from '../../lib/html'
 import delay from 'delay'
 import {parseXml} from './parse-feed'
 import type {StoryType} from './types'
 import LoadingView from '../components/loading'
-import {Row, Column} from '../components/layout'
-import {ListRow, ListSeparator, Detail, Title} from '../components/list'
+import {ListSeparator} from '../components/list'
 import {NoticeView} from '../components/notice'
+import {NewsRow} from './news-row'
 import type {TopLevelViewPropsType} from '../types'
 import {tracker} from '../../analytics'
-import {AllHtmlEntities} from 'html-entities'
-
-const entities = new AllHtmlEntities()
-
-type NewsRowPropsType = {
-  story: StoryType,
-  onPress: (title: string, story: StoryType) => any,
-};
-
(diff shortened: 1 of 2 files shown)


views/news/news-container.js
```diff
@@ -43,6 +44,18 @@ export default class NewsContainer extends React.Component {
...
+  findImage = (content: string) => {
+    let re = '<img.*?src="([^"]*)"[^>]*>(?:</img>)?'
+    let imageMatch = content.match(re)
+
+    if (imageMatch !== null && imageMatch !== undefined) {
+      let src = imageMatch[1]
+      return <Image source={{uri: src}} style={styles.itemImage} />
+    }
+
+    return false
```
hawkrives: Generally, an optional react component is denoted with `null` for the no-component case, not `false`. Please switch this to use `null`.

views/news/news-container.js
```diff
@@ -43,6 +44,18 @@ export default class NewsContainer extends React.Component {
     this.refresh()
   }
 
+  findImage = (content: string) => {
+    let re = '<img.*?src="([^"]*)"[^>]*>(?:</img>)?'
+    let imageMatch = content.match(re)
+
+    if (imageMatch !== null && imageMatch !== undefined) {
```
hawkrives: Just do `if (imageMatch)`?

hawkrives (changes requested):

views/news/news-container.js
drewvolz: Good point, thanks.

views/news/news-container.js
drewvolz: Done.

hawkrives (commented):
- [x] I would like to request a screenshot of a list with a mix of both thumbnailed and non-thumbnailed stories. 

- [x] Furthermore, since most of these run on Wordpress, what happens when the image to be thumbnailed is smaller than the thumbnail?

- [x] What about when it's the "N comments" image that Wordpress sticks at the end of posts in an RSS feed sometimes?

I just want to make sure we've got the edge cases here.

drewvolz:
> I would like to request a screenshot of a list with a mix of both thumbnailed and non-thumbnailed stories.

randomized

> Furthermore, since most of these run on Wordpress, what happens when the image to be thumbnailed is smaller than the thumbnail?

We won't show an image for it. I have adjusted to check the image height and width based upon reading your question.

> What about when it's the "N comments" image that Wordpress sticks at the end of posts in an RSS feed sometimes?

We are now making sure that icons are at least 50x50, so I doubt they would show up. If it is that large, we will display it. But I doubt it will be there.

views/news/news-container.js
```diff
@@ -58,7 +58,13 @@ export default class NewsContainer extends React.Component {
...
     }
 
-    return <Image source={{uri: imageUrl}} style={styles.itemImage} />
+    let icon = <Image source={{uri: imageUrl}} style={styles.itemImage} />
+
+    Image.getSize(imageUrl, (width, height) => {
+      icon = (width >= 50 && height >= 50) ? icon : null
+    })
+
+    return icon
```
hawkrives: This won't work correctly. `Image.getSize` is asynchronous – the callback may not (should not, I think) have executed by the time we return from the function.

We'll have to make `findImage` return a promise, instead, I think, and store some state in the component.

hawkrives (changes requested):

hawkrives:
Thanks for the investigation and comments. It's coming along nicely!

elijahverdoorn:
I like where this is going!

drewvolz:
@hawkrives added functionality to check the image sizes asynchronously. Looks like we're good to go.

drewvolz:
We may need to wrap image fetching with the global fetch, and handle what is returning vs resolving. I am noticing errors occasionally.

screen shot 2017-01-06 at 1 45 35 am

views/news/news-container.js
```diff
@@ -21,6 +22,98 @@ import * as c from '../components/colors'
...
+
+  props: NewsRowPropsType
+
+  makeThumbnail = async (props: NewsRowPropsType) => {
+    const thumbnailUrl = await new Promise(resolve => {
+      // grab the URL
+      const url = this.findImage(props.story.content)
+      // if we didn't find a valid URL, return null
+      if (!url) {
+        return null
```
hawkrives: You should resolve() here too. You never want to `return` from the body of a Promise constructor.

views/news/news-container.js
```diff
@@ -21,6 +22,98 @@ import * as c from '../components/colors'
...
+        // if getSize fails, we also resolve with `null`
+        () => resolve(null))
+    })
+
+    if (!thumbnailUrl) {
+      return
+    }
+
+    this.setState({thumbnailUrl})
+
```
hawkrives: Nit: extra blank line.

views/news/news-container.js
```diff
@@ -21,6 +22,98 @@ import * as c from '../components/colors'
 let Entities = require('html-entities').AllHtmlEntities
 const entities = new Entities()
 
+type NewsRowPropsType = {
+  story: StoryType,
+  onPress: (title: string, story: StoryType) => any,
+};
+
+class NewsRow extends React.Component {
```
hawkrives: I would like to pull this component into its own file. News-item-Row.js

hawkrives (changes requested):

views/news/news-container.js
drewvolz: Resolving here caused a problem where null was being returned somewhere down below

views/news/news-container.js
drewvolz: Great idea

views/news/news-container.js
hawkrives: I'm struggling to understand that, since we do the `if (!thumbnailUrl)` which should catch any falsy values including `null`.

drewvolz: Manitou Messenger error if I have `resolve(null)` on line 51

screen shot 2017-01-08 at 4 05 27 pm

drewvolz: What about a `return` after the `resolve(null)`? It seems to behave well.

      if (!url) {
        resolve(null)
        return
      }

hawkrives: That tells us that this is a timing issue

hawkrives: That tells us that this is a timing issue

hawkrives: That tells us that this is a timing issue

hawkrives: Because, if nothing else, our check in render should prevent us from rendering the image with "null" as the source.

hawkrives:
My current feelings on this:

http://stackoverflow.com/questions/31651610/return-value-of-promises-resolve-reject-functions

The return value of a new Promise is ignored. That brings me back to, there's an underlying timing bug somewhere, which is being exposed by resolving.

And until we resolve that bug I'm uncomfortable merging it. 

Because, if nothing else, our check in `render` ought to not render the Image of the URL is null, and that's obviously not working for some reason.

I want to know why that's not working.

On the bright side, at least it's reproducible and not intermittent. 😄

hawkrives:
Closed in favor of #693

PR #464, opened 2016-12-31
Shrink the size of the home screen buttons

drewvolz:
Removes two pixels of padding from the buttons so 5s users can see icons below, now that we have included Student Orgs.

Screenshots:

Before | After
---|---
 original |  shrunken


hawkrives (approved):
Woot! Thanks!

PR #465, opened 2016-12-31
Activity indicator on loading views

drewvolz:
This handless the case where the loading view has the text "Loading…". It adds a spinner below the text.

It tries to handle the case where the loading message is anything else e.g. the menus custom messages. Although those can be "loading" messages, it would be confusing to display a "closed" message and the spinner at the same time. That could be handled in the future by importing those messages? I dunno. Suggestions?

Screenshots:

Before | After
---|---
 before |  after

### DIFF
--- a/views/components/loading.js
+++ b/views/components/loading.js
@@ -4,6 +4,6 @@ import {NoticeView} from './notice'
 
 export default function LoadingView({text='Loading…'}: {text?: string}) {
   return (
-    <NoticeView text={text} />
+    <NoticeView text={text} spinner={true} />
   )
 }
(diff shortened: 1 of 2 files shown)


hawkrives:
This feels like it should be part of LoadingView, rather than NoticeView. 

What about adding a showSpinner prop to NoticeView, that defaults to false, which LoadingView would set to true?

hawkrives (changes requested):
Please add a new prop to `NoticeView`, called, say, `activityIndicator`, that accepts a boolean.

Then edit LoadingView to call `<NoticeView activityIndicator />`.

hawkrives:
Should the spinner be below the text? I keep expecting it to be above the text…

hawkrives (approved):

PR #475, opened 2017-01-03
Clean up the data/ folder

hawkrives:
Whoof. Um. I moved things around?

> @hawkrives: How should I describe this PR?
> @drewvolz: A brighter future / Preparing for the cloud / API here we come

You'll want to read this introduction to YAML at some point: http://camel.readthedocs.io/en/latest/yamlref.html

---

The core of these changes is in a reorganization of the single-file structure of the data/ folder.

Before:

	images/
	about.md
	building-hours.json
	bus-times.json
	cage-menu.json
	calendar.json
	contact-info.json
	credits.json
	dictionary.json
	dietary-filters.js
	legal.json
	map-coordinates.json
	map.json
	new-building-hours.js
	news.json
	privacy.json
	transportation.json
	webcams.json

After: 

	building-hours/
	bus-times/
	contact-info/
	dictionary/
	map-coordinates/
	transportation/
	webcams/
	breaks.yaml
	chapel.yaml
	credits.yaml
	faqs.md
	legal.md
	map.yaml
	pause-menu.yaml
	privacy.md

I took the data files that were merely an array of objects, such as dictionary, building-hours, or bus-times, and converted them into folders with one-file-per-item, with each file converted to YAML.

The other data files I determined to be either plain text, or semi-structured stuff. I really really dislike editing JSON by hand, so the semi-structured stuff I converted to YAML files; the plain text ones I changed to Markdown.

(I added `js-yaml` and `marked` as devDependencies, since they're run at development time and not at runtime.)

When you run `npm run bundle-data`, which you should do after editing a data file, it automatically runs through the entire data/ folder and compiles them to docs/.

Individual files within a folder are compiled to a JSON array named after the folder – filenames are irrelevant, except to order the items in the generated array, as `{data: []}`.

Standalone files in data/ are compiled to one of two styles of output: either `{text: blah}` for a markdown input (e.g., legal.json), or `{data: blah}` for a yaml input (e.g., credits.json).

---

The main benefit, as I see it, to storing the files separately is that git is far better at tracking file changes in individual files than in one big blob.

I implemented a check in Travis to make sure that the data files have been regenerated in a PR, to make sure we don't accidentally ship old generated files.

---

The other change in this PR is that I moved the images from /data/images to just /images.

---

I feel that this PR sets us up nicely for putting more of our content on GitHub and fetching it dynamically in the app, like the Pause menu does.

---

TL;DR: go read the stuff and give me some feedback. I've got an FAQ view ready to go, too.

And finally some stats:

[...]

PR #481, opened 2017-01-04
Add schemas to allow automatic validation of data files

hawkrives:
> This PR builds on top of #475.

Using [json-schema.org](http://json-schema.org) and [epoberezkin/ajv](https://github.com/epoberezkin/ajv), I created schema files for each data file in `data/`.

I also wrote a script to automatically validate the contents of the files against the schema, so we'll know quickly if we break something.

This setup actually caught an issue in the Tom Porter file that slipped by me during my data-splitting PR.

What to say about the schemas… They're actually written in YAML, not JSON, but I find YAML so much clearer than JSON to write and edit that I converted them over. (See https://gist.github.com/hawkrives/659d9e0af8f781533cba3e24cafb8370 for a comparison.) It's no more work to use YAML for the schema, since `ajv` runs in JavaScript-land anyway so it operates on JS objects in the first place.

I highly recommend https://spacetelescope.github.io/understanding-json-schema/basics.html for getting started with JSON-Schema, and http://camel.readthedocs.io/en/latest/yamlref.html for an overview of YAML.

I am willing to answer any questions about this PR, up to any including debates about if we even need schemas for the files (personally, I think we do.)

TL;DR: automatic validation of data in the data files. good? bad? silly?


drewvolz (approved):

elijahverdoorn (approved):
I like schemas in general, I don't see a reason not to use them for the data files.

hawkrives:
Thanks all! I'll merge this after #475 is merged.

hawkrives:
now, github, really?

PR #482, opened 2017-01-04
Northfield calendar

drewvolz:
Northfield events from the http://visitingnorthfield.com calendar.

thisisnorthfield

### DIFF
--- a/views/calendar/tabs.js
+++ b/views/calendar/tabs.js
@@ -10,17 +10,17 @@ export default [
     props: {calendarId: 'le6tdd9i38vgb7fcmha0hu66u9gjus2e%40import.calendar.google.com'},
   },
   {
-    id: 'northfield',
-    title: 'Northfield',
-    rnVectorIcon: {iconName: 'pin'},
-    component: CalendarView,
-    props: {calendarId: 'thisisnorthfield%40gmail.com'},
-  },
-  {
     id: 'oleville',
     title: 'Oleville',
     rnVectorIcon: {iconName: 'happy'},
     component: CalendarView,
     props: {calendarId: 'stolaf.edu_fvulqo4larnslel75740vglvko@group.calendar.google.com'},
+  },
+  {
+    id: 'northfield',
+    title: 'Northfield',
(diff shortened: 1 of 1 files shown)


drewvolz:
We may have to make an adjustment to the calendar. A few of these times look wrong with the start/end time being the same.

hawkrives:
Sure. How about a thing that says "hide the end time if start === end"?

Great work!

hawkrives:
(Such a change should probably be a separate PR.)

drewvolz:
Could there be a way to increase the end time by one if they are listed as the same? To remain consistent with the other calendars?

hawkrives:
We don't know that they're an hour long, though? They might be two or something. At first blush, I'd prefer to just hide the end time if we don't know how long they'll go.

drewvolz:
So you are thinking that events listing start/end as the same may not be one hour in length. Alright, I guess we cannot assume that. I'll aim for your original suggestion.

hawkrives:
Now for the important question: I feel like this should be the third tab, rather than the middle one, in order to provide priority to on-campus events.

[ `St. Olaf` | `Oleville` | `Student Orgs` | `Northfield` ]

drewvolz:
Sure, that will be moved back to its original order.

hawkrives (approved):
Man. Look at that diff. That's a nice diff.

hawkrives:
@elijahverdoorn We've opted to give you final say on adding this, since it kinda falls under the "new feature" umbrella.

elijahverdoorn:
I'm in favor of the idea, but want to make sure that the ordering of tabs is St. Olaf, Oleville, Northfield. That's the order that the code would suggest, but the screenshot that @drewvolz has doesn't match. Assuming that the order is what the code has and not the screenshot, go for it.

hawkrives:
Yes. We updated the order after the screenshots were taken. He couldn't update the screenshots today, but it's definitely Olaf, Oleville, Northfield.

elijahverdoorn:
@hawkrives Thanks for the confirmation.

PR #484, opened 2017-01-04
Hide endTime when startTime === endTime

hawkrives:
Fixes the issue as noted in #482.


hawkrives:
screen shot 2017-01-04 at 9 55 47 am

I don't know. It feels like it's missing information.

Can we put something small in the end-time slot?

Like,

```
6:00pm
   ???
```

hawkrives:
screen shot 2017-01-04 at 10 02 44 am

Maybe something like that?

drewvolz (approved):
Thanks for finding a way to display a better end time. Good work.

PR #491, opened 2017-01-05
Launch image update

drewvolz:
Let's shave off some more data. The old launch image was wonderful, but according to Apple's human interface guidelines, it breaks all of the rules.

Here's what Apple recommends:

> **Design a launch screen that’s nearly identical to the first screen of your app**. If you include elements that look different when the app finishes launching, people can experience an unpleasant flash between the launch screen and the first screen of the app.
> 
> **Avoid including text on your launch screen**. Because launch screens are static, any displayed text won’t be localized.
> 
> **Downplay launch**. People are likely to switch apps frequently, so design a launch screen that doesn’t draw attention to the app launching experience.
> 
> **Don’t advertise**. The launch screen isn’t a branding opportunity. Don’t design an entry experience that looks like a splash screen or an "About" window. Don’t include logos or other branding elements unless they’re a static part of your app’s first screen.

The new launch screen makes use of Xcode's built-in launch screen. This way, we can use the native UI rather than a static image.

Let's try to meet **all** of these suggestions:

- [x] nearly identical to the first screen of the app
- [x] avoids including text
- [x] does not draw attention
- [x] does not advertise the brand

**bonus:** removes the white flash on start

Screenshots:

4s 
---
4s

5 | 5s
---|---
 | 5s

6 | 6s
---|---
| 6s

6s Plus | 7
---|---
 | 6s Plus | 

7 Plus | 7 SE
---|---
 | 7 Plus | 7 SE


hawkrives:
Good work!

The LaunchImages were somewhere between 2-3MB, total, so the iOS app should be approaching 6-7MB now. 

For an app that was 17-18MB just yesterday, this is pretty nice.

hawkrives (approved):
Woot woot!

drewvolz:
There might be a bug in here somewhere. The LaunchScreen view is not working as it did before. Don't merge this yet.

drewvolz:
Switching to a storyboard instead of a xib. Storyboards, in my opinion, are much easier to work with. Also the issues I mentioned earlier are fixed. Good to go.

hawkrives (approved):
I'll test this later today.

hawkrives:
Looks like it works to me!

PR #492, opened 2017-01-05
Fix buttons again again

### LINKED ISSUE #155: Settings button can be pushed multiple times

drewvolz:
Two issues are tackled here here:

(1) Blank views can be pushed on a corrupted stack
  * Happens on **settings** button mashing
  * Happens on **edit** button mashing

(2) Multiple views can be pushed to the navigator stack
  * I am able to push an Edit view and a Settings view to the stack one after the other
  * I can touch either the settings or edit buttons sequentially and very quickly to make this happen.
  * It will render both views, as both were pushed individually in sequential order.... we do not check for that yet.

<hr>

Solution for (1) is to check outside of the called function whether the function should be called instead of within the function.

    let onPressEditHome = (settingsButtonActive || editHomeButtonActive) ? null : openEditHome(route, navigator)
    let onPressSettings = (settingsButtonActive || editHomeButtonActive) ? null : openSettings(route, navigator)

Solution for (2) is to check if the opposite navbar button is active in addition to itself. This is a little iffy, which I would like feedback on. The onDismiss is being called for the `HomeView` and changing their values when the `HomeView` disappears instead of when the `EditHomeView` or `SettingsView` views disappear. The logic works the same, but maybe there is a more elegant solution.

     if (editHomeButtonActive || settingsButtonActive)

Closes #155

### DIFF
--- a/app.js
+++ b/app.js
@@ -223,10 +223,11 @@ function LeftButton(route, navigator, index, navState) {
   }
 
   if (route.id === 'HomeView') {
+    let onPressSettings = (settingsButtonActive || editHomeButtonActive) ? null : openSettings(route, navigator)
     return (
       <TouchableOpacity
         style={[styles.settingsButton]}
-        onPress={openSettings(route, navigator)}
+        onPress={onPressSettings}
       >
         <Icon style={styles.settingsIcon} name='ios-settings' />
       </TouchableOpacity>
@@ -265,6 +266,10 @@ function LeftButton(route, navigator, index, navState) {
 // Leaving the boilerplate here for future expansion
 function RightButton(route, navigator) {
   if (route.onDismiss) {
+    if (route.id === 'SettingsView' || route.id === 'EditHomeView') {
+      settingsButtonActive = false
+      editHomeButtonActive = false
+    }
     return (
       <TouchableOpacity
         style={styles.backButtonClose}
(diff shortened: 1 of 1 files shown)


hawkrives (changes requested):
Nitpick as I can't think atm: instead of providing "null" as a callback function, please use "() => {}", aka an empty lambda

drewvolz:
The empty lambdas are now provided as callbacks. Thank you for your feedback.

drewvolz:
Any other requests?

hawkrives:
I _see_ the changes, but I don't _grok_ the changes. 

I guess I was waiting until I understood why they were needed. 

If you can't reproduce with these, then Ok.

drewvolz:
~~I can verify that this solves the issue.~~

**Update**: You are right. It only gets us part of the way there. We should not accept this until it is fixed in a more elegant way.

drewvolz:
What about utilizing redux to store the state we're manipulating and checking a bunch of times?

hawkrives:
That's gotta be slower than just checking a local variable though, yeah?

PR #496, opened 2017-01-06
Wrapper for trimmed text with spaces

### LINKED ISSUE #462: Filter html from item descriptions
We need to clean up the item descriptions to prevent raw HTML from appearing. I know @hawkrives mentioned possible performance concerns with cleaning up each of those strings.

<hr>

bonappimage

drewvolz:
This solves #462 where we noticed html within item descriptions on the Bon Appetit menus.

Introduces a helper function that wraps `getTextWithSpaces` with `getTrimmedTextWithSpaces`. Allows for us to slim down on code duplication when splitting/joining spaces on the returned strings. The duplication of code would have existed in Orgs, Oleville, Calendar, and Menus.

Thanks @hawkrives for the direction on this.

Closes #462

### DIFF
--- a/lib/html.js
+++ b/lib/html.js
@@ -28,5 +28,5 @@ function getTextWithSpaces(elem: Object|Object[]): string {
 }
 
 export function getTrimmedTextWithSpaces(elem: Object|Object[]): string {
-  return getTextWithSpaces(elem).split(/\s+/).join(' ')
+  return getTextWithSpaces(elem).split(/\s+/).join(' ').trim()
 }
--- a/lib/html.js
+++ b/lib/html.js
@@ -20,9 +20,13 @@ export function getText(elem: Object|Object[]): string {
   return ''
 }
 
-export function getTextWithSpaces(elem: Object|Object[]): string {
+function getTextWithSpaces(elem: Object|Object[]): string {
   if (Array.isArray(elem)) return elem.map(getText).join(' ')
   if (elem.type === 'tag') return getText(elem.children)
   if (elem.type === 'text') return elem.data
   return ''
+}
+
+export function getTrimmedTextWithSpaces(elem: Object|Object[]): string {
+  return getTextWithSpaces(elem).split(/\s+/).join(' ')
 }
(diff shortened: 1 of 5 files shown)


lib/html.js
```diff
@@ -20,9 +20,13 @@ export function getText(elem: Object|Object[]): string {
...
-export function getTextWithSpaces(elem: Object|Object[]): string {
+function getTextWithSpaces(elem: Object|Object[]): string {
   if (Array.isArray(elem)) return elem.map(getText).join(' ')
   if (elem.type === 'tag') return getText(elem.children)
   if (elem.type === 'text') return elem.data
   return ''
 }
+
+export function getTrimmedTextWithSpaces(elem: Object|Object[]): string {
+  return getTextWithSpaces(elem).split(/\s+/).join(' ')
```
hawkrives: Should we `.trim()` the string, too, to remove starting/ending whitespace?

hawkrives:
Do we want to update [`views/calendar/event.js#L59`](https://github.com/StoDevX/AAO-React-Native/blob/master/views/calendar/event.js#L59) and [`views/oleville/index.js#L140`](https://github.com/StoDevX/AAO-React-Native/blob/master/views/oleville/index.js#L140), as well? I think that the Calendar could benefit, at least.

elijahverdoorn (approved):
Good fix.

lib/html.js
drewvolz: Yes, good point.

elijahverdoorn:
I approved this as is, and still stand by that, but want to ask a discussion question: Would we like to filter for `<br>` tags and replace them with something other than a space? It would make sense to me to have them be replaced with a semicolon or comma, since from the data I can see it would seem that they are being used to indicate lists (wish they'd use `<li>`). This is an example of what I'm talking about:

4s

hawkrives:
Alright. I like that.

What would we do about non-`<br/>` tags? Just replace them with spaces?

hawkrives:
I vote for joining with `;` rather than `,`; some of these descriptions are more like what I'm going to call "annotations" than lists of items.

4s

drewvolz:
We are all in agreement that filtering for `<br />` tags and joining with `;` will  be a new issue. I will merge this PR now.

PR #497, opened 2017-01-06
Fix map URL

elijahverdoorn:
A rogue " slipped into the URL for the map, causing it to fail on Android. Removed that issue.


hawkrives:
Yay thanks. I noticed that but hadn't made a PR yet.

Wonder why it didn't break on iOS? Idk.

drewvolz (approved):

hawkrives (approved):

PR #512, opened 2017-01-07
Require that the map URL be valid

hawkrives:
Are there any other schemas that need uri?

- [x] map
- [x] transportation (the "other" options)
- [x] webcams

Or other formats. We can choose from the following:

> "date-time": Date representation, as defined by RFC 3339, section 5.6.
> "email": Internet email address, see RFC 5322, section 3.4.1.
> "hostname": Internet host name, see RFC 1034, section 3.1.
> "ipv4": IPv4 address, according to dotted-quad ABNF syntax as defined in RFC 2673, section 3.2.
> "ipv6": IPv6 address, as defined in RFC 2373, section 2.2.
> "uri": A universal resource identifier (URI), according to RFC3986.

### DIFF
--- a/data/_schemas/map.yaml
+++ b/data/_schemas/map.yaml
@@ -3,5 +3,5 @@ $schema: http://json-schema.org/draft-04/schema#
 additionalProperties: false
 required: [url, description]
 properties:
-  url: {type: string}
+  url: {type: string, format: uri}
   description: {type: string}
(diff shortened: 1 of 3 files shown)


drewvolz:
Phone numbers from Important Contacts could be validated this way.

hawkrives:
Using a regex? Sure. Is the current format just a string of numbers (e.g., 9185789388?) 

If we, we could add a piece to the "definitions" in defs.yaml:

```
phone:
  type: string
  pattern: ^\d{9}$
```

drewvolz:
Yes, just a string as you provided for example.

PR #525, opened 2017-01-09
Center android tabs when there are only 2 tabs

elijahverdoorn:
When there are only 2 tabs in a tabview on android, the tabs are not centered properly. This PR fixes that.

**Before:**
before

**After:**
after


hawkrives (approved):
Nice!

PR #526, opened 2017-01-09
Remove the `collapsible` component

hawkrives:
Nothing uses it anymore. Would anyone like me to keep it?

Comment period: open until 12pm, Tuesday, Jan. 10th.


elijahverdoorn:
I'm partial to starting a StoDevX repo to keep general-use components in, so that we don't lose track of what we've written in the past (and maybe encourage other projects to use them), and removing unused components from this repo. We could add them as dependencies if we needed them in the future.

drewvolz:
You have my vote to nix it.

hawkrives:
@elijahverdoorn: We'd need to figure out how to make npm packages for react-native components first… and I have no idea how to do that atm. 

I'd also want to pull out the fancy-menu stuff, and filters, mmm probably not the tabbed view, and maybe some others.

For this one, though, I'm perfectly fine with just letting it lay in the history of this repo, and if anything ever wants to use it in the future they could pull it out then…

elijahverdoorn:
@hawkrives That's fine with me. I might explore the possibility of publishing with NPM packages at a future time. I'm good with removing it from the repo.

elijahverdoorn (approved):

PR #530, opened 2017-01-09
Enable automatic nightly beta builds from Travis

hawkrives:
This variant checks if we need a new revision before submitting a new beta, by comparing the commit hashes of master/HEAD and the last build on Hockeyapp.

People affected: only the beta testers.

> - [ ] Once this has landed, we need to flip the cron switch back on in Travis's UI.


drewvolz (approved):
👍

Issue #523, opened 2017-01-09
Scrutinize App Transport Security exceptions


hawkrives:
We've got 12 exceptions, and I don't think that we need some of them anymore.

- [unused] `bamco.com`
	- I can't find any uses of this URL in the codebase.
- [unused] `hub.cafebonappetit.com`
	- I can't find any uses of this URL in the codebase.
- [required] `legacy.cafebonappetit.com`
	- Used for the menus
- [required] `localhost`
	- Needed for dev-time react-native reloading
- [required] `manitoumessenger.com`
	- Needed for the MESS's RSS feed
- [required] `oleville.com`
	- Needed because oleville isn't HTTPS (RSS feed)
- [required] `stolaf-flash.streamguys.net`
	- Needed for the streams
- [unused] `stolaf.cafebonappetit.com`
	- Used to be used to load the café menus?
	- I can't find any uses of this URL in the codebase.
- [unused] `template.cafebonappetit.com`
	- I can't find any uses of this URL in the codebase.
- [unused] `van-go.info`
	- Used to … open the link in a browser?
- [required] `wp.stolaf.edu`
	- RSS feed (https) links to images over http
- [unused] `youarriveontime.com`
	- Used to … open the link in a browser?

I think we can eliminate `bamco.com`, `hub.cafebonappetit.com`, `stolaf.cafebonappetit.com`, `template.cafebonappetit.com`, `van-go.info`, and `youarriveontime.com` from our exceptions list.

#### Keep
- `legacy.cafebonappetit.com` – Used for the menus
- `localhost` – Needed for dev-time react-native reloading
- `manitoumessenger.com` – Needed for the MESS's RSS feed
- `oleville.com` – Needed because oleville isn't HTTPS (RSS feed)
- `stolaf-flash.streamguys.net` – Needed for the streams
- `wp.stolaf.edu` – RSS feed (https) links to images over http

#### Remove
- `bamco.com`
- `hub.cafebonappetit.com`
- `stolaf.cafebonappetit.com`
- `template.cafebonappetit.com`

#### Maybe remove?
- `van-go.info`
- `youarriveontime.com`

drewvolz:
Refresh my memory. Do exceptions _only_ apply to webviews?

drewvolz:
Ah. I can answer my own question. No, we'd need them outside of webviews.

drewvolz:
[...]

PR #535, opened 2017-01-10
Remove unneeded NSAppTransportSecurity exceptions

hawkrives:
> part of #523 

- remove bamco.com
- remove hub.cafebonappetit.com
- remove stolaf.cafebonappetit.com
- remove template.cafebonappetit.com


hawkrives:
To clarify: we only need legacy.cafebonappetit.com for the menus, now that we're using the API to make the menus.

drewvolz (approved):

PR #539, opened 2017-01-10
Chapter 1, or: the Creation of ListRow&co

hawkrives:
> Depends on #540. The build will fail until `<Touchable/>` is merged.

Common list styles FTW!

~~Descriptions incoming later.~~

This and #538 are a pair of complementary component sets. This PR introduces the `<ListRow/>`, `<ListSectionHeader/>`, `<ListSeparator/>`, `<Title/>`, and `<Detail/>` components.

These are components that apply basic styling to our ListViews, since we have so many of them. These two PRs together are what produced #499.

All of these components are imported from `view/components/list` – `import {ListRow, Title} from '../../components/list'`, for example.

## `<ListSeparator/>`
It's a `<Separator/>` that hides itself on Android, takes a `fullWidth={true}` and a `spacing={{left: NUMBER}}`` prop to make adjusting the left margin easy, and applies default styling on iOS.

Soon, I'll have to update it for proper 3-line Android ListView styles, but that'll be a subsequent PR.

Example usage:

```js
<ListSeparator />
```

## `<ListSectionHeader/>`
Applies default styling to the SectionHeaders, as appropriate for both iOS and Android.

Takes a `title` prop for supplying the title text, as well as optional `subtitle` (and `separator`) props to supply a secondary title, as seen on Menus. Also takes `bold={true|false}`, to change if it should be bold or not. Finally, you can apply styles to the title and subtitle with `titleStyle` and `subtitleStyle`, and adjust the left margin with `spacing={{left: NUMBER}}`, just like `<ListSeparator/>`.

Example usage:

```js
<ListSectionHeader 
  title='Pizza' 
  subtitle='Orders larger than 4 pizzas should be placed 24 hours in advance, through Oleville' 
/>
```

## `<ListRow/>`
The big kahuna. Applies appropriate row styles to a row for both iOS and Android, as well as displaying a disclosure caret as appropriate on iOS.

Any ListView's renderRow should wind up returning a ListRow, either directly or by supplying a component whose top-level is ListRow.

ListRow takes several props:

- `style`: styles, for the element that contains the children.
- `contentContainerStyle`: styles, for the wrapping View.
- `arrowPosition=center|top|none`: where to show the arrow on iOS. Defaults to `center` if onPress is given; else hides itself.
- `onPress`: called when the row is tapped.
- `fullWidth` / `fullHeight`: remove the default padding around the row; used for the BusTransit rows (both options), and Calendar (just fullWidth), among others.
[...]

PR #540, opened 2017-01-10
An Introduction, or: how <Touchable/> subsumed Touchable{Highlight,NativeFeedback,Opacity}

hawkrives:
`<Touchable/>` is designed to wrap around `TouchableOpacity`, `TouchableHighlight`, and `TouchableNativeFeedback`, as appropriate for the platform.

I believe that we can, in the future, only use this component and never import any of those three again (unless we want opacity on android for some reason i guess).

> This is a prerequisite for #539.

Usage:

```js
<Touchable>
   {stuff}
</Touchable>
```


drewvolz (approved):
:+1: :shipit:

PR #538, opened 2017-01-10
Chapter 2, or: the Introduction of the Structural components (Row and Column)

hawkrives:
~~Descriptions incoming later.~~

This PR introduces two primary new components, `<Row/>` and `<Column/>`. They behave slightly 
differently from what you're used to, as they're intended to be exclusively layout aids.

They can be imported from `components/layout`, e.g. `import {Row, Column} from '../components/layout'`.

Row is a View that defaults to flexDirection:row, while Column is a View that defaults to flexDirection:column. So far, so good.

Now for the tricky part: All props on Row and Column are interpreted as styles. `<Row backgroundColor='red' />` is equivalent to `<View style={{flexDirection: 'row', backgroundColor: 'red'}} />`.

That's all there is to them!

(The special props handling is mainly so it's easy to give custom `justifyContent` / `alignItems` styles, tbh.)

#### FoodItem
```js
<ListRow>
  <Row>
    <Icon style={styles.badgeIcon} name={specialsIcon} />

    <Column flex={1}>
      <Title lines={1}>blah</Title>
      <Detail>blah blah blow</Detail>
    </Column>

    <DietaryFilters />
  </Row>
<ListRow>
```

#### Event
```js
<ListRow>
  <Row>
    <CalendarTimes />
    
    <Bar />

    <Column>
      <Title>blah</Title>
      <Detail>blah blah blow</Detail>
    </Column>

  </Row>
</ListRow>
```

#### BuildingHoursRow
```js
<ListRow arrowPosition='top'>
  <Column>
    <Row>
      <Title lines={1} style={{flex: 1}}>blah</Title>
      <StatusBadge />
    </Row>
    <Detail>blah blah blah hours stuff</Detail>
  </Column>
</ListRow>
```

#### Dictionary item
```js
<ListRow arrowPosition='none'>
  <Column>
    <Title lines={1}>{item.word}</Title>
    <Detail lines={2}>{item.definition}</Detail>
  </Column>
</ListRow>
```

PR #542, opened 2017-01-10
Add ability to specify description to a filter item

hawkrives:
also sets up ability to render an image for a row

This changes the type of data in the list filters from an array of strings to an array of `{title, description?, image?}` objects.

### DIFF
--- a/views/components/filter/__tests__/apply-and-list-filter.test.js
+++ b/views/components/filter/__tests__/apply-and-list-filter.test.js
@@ -1,15 +1,16 @@
 /* eslint-env jest */
 // @flow
 import {applyAndListFilter} from '../apply-filters'
+import {filterValue} from './filter-value.helper'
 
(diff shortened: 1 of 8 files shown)


hawkrives:
Kosher needs a description in another PR.

drewvolz:
1. Can we remove the title from the description? e.g. "~~Vegan:~~ Contains absolutely no.."

2. Can we make the dietary cells custom cells so that they either expand, or have them display the entire description instead of one line of description?

hawkrives:
(1) should be a new PR, probably as part of #543.

hawkrives:
I'd like to tackle (2) as part of the inclusion of images into the rows, since I'll need to make a CustomCell instance for either change.

I'd also like to get #538 (Row and Column) merged before tackling the CustomCell here.

drewvolz (approved):

PR #547, opened 2017-01-10
Disable dietary restrictions correctly

### LINKED ISSUE #521: Dietary Restrictions filter doesn't disable correctly
@drewvolz:
> [Applying] and removing dietary restriction filters will cause an incorrect count of filters, and then mucks up the list of food items displayed. I feel as if there is a bug where the cor icon diets are not unapplying themselves once deselected. It feels reversed after a few toggles?
>
> […]
>
> The number of filters only resets back to zero at one point if the cor icon filters are all checked, when the original state for that is having them all disabled.

Yep. This is a bug.

Each filter has an `enabled` property, which sets whether the filter is enabled or not. I have a bug in the `enabled` setting code for list filters, because I originally didn't distinguish between the AND and OR styles.

In [views/components/filter/section-list.js#L33](https://github.com/StoDevX/AAO-React-Native/blob/2c9a67ae159772ee39f5c319b06bbc1bfd5580ec/views/components/filter/section-list.js#L33)

```js
// `result` is the list of selected options, and `options` is the list of all options
onChange({
  ...filter,
  enabled: result.length !== options.length,
  spec: {...spec, selected: result},
})
```

That `enabled` setting is correct for the OR mode, where a filter is disabled only when all of the options are selected (aka "this filter should show any of these options".)

However, it's incorrect for the AND mode, where a filter should be disabled whenever none of the options are selected (aka "this filter should only show items with all of these selected options".)

I think a good change would be to pull out the `enabled` value into a variable, then do an `if` to check on the `mode` of the filter, and switch between the two algorithms depending.

Something like:

```js
// `result` is the list of selected options, and `options` is the list of all options
let enabled = false
if (mode === 'AND') {
  enabled = result.length !== options.length
} else if (mode === 'OR') {
  enabled = result.length !== 0
}

onChange({
  ...filter,
  enabled: enabled,
  spec: {...spec, selected: result},
})
```

I can do this at some point, or someone else can take it, either way is fine with me.

drewvolz:
It is incorrect for the AND mode, where a filter should be disabled whenever none of the options are selected (aka "this filter should only show items with all of these selected options".)

This pulls out the enabled value into a variable, then performs a check on the mode of the filter, and switches between the two algorithms depending.

Thanks @hawkrives 

Closes #521

### DIFF
--- a/views/components/filter/section-list.js
+++ b/views/components/filter/section-list.js
@@ -28,9 +28,16 @@ export function ListSection({filter, onChange}: PropsType) {
       result = concat(selected, tappedValue)
     }
 
+    let enabled = false
+    if (mode === 'AND') {
+      enabled = result.length !== options.length
+    } else if (mode === 'OR') {
+      enabled = result.length !== 0
+    }
+
     onChange({
       ...filter,
-      enabled: result.length !== options.length,
+      enabled: enabled,
       spec: {...spec, selected: result},
     })
   }


hawkrives (approved):
yaaaaay let's merge it

PR #551, opened 2017-01-11
Update sis and news faqs

drewvolz:
* Explanation on long update delay
* Status update on SIS
* Status update on News


hawkrives:
Make them h1s instead of h2s, please?

The embedded stylesheet has a rule for h1s.

hawkrives (changes requested):
Make them h1s instead of h2s, please?

The embedded stylesheet has a rule for h1s.

hawkrives (approved):
We chatted about this out-of-band.

> @drewvolz: I think the question text is very large. When do you honestly see text that large in mobile apps? When you’re reading something, especially.
> @hawkrives: So make it smaller

And so he did.

drewvolz:
Now that News is fixed, we need a PR to rid of the news FAQ.

PR #499, opened 2017-01-07
Android lists

hawkrives:
> New message
I'm breaking this up into smaller PRs, or "chapters":

- [x] (#541, #564) Preface, or: an AlphabetListView upgrade (to HEAD)
- [x] (#540) Introduction, or: how Touchable subsumed Touchable{Highlight,NativeFeedback,Opacity}
- [x] (#539) Chapter 1, or: the Creation of ListRow&co
- [x] (#538) Chapter 2, or: the Introduction of the Structural components (Row, Column, and List*)
- [x] (#562) Chapter 3.1, or: the Upkeep of Calendar
- [x] ~~(no pr) Chapter 3.2, or: the Maintenance of BusView~~ merged with #573 

---

- [x] ~~(no pr) Chapter 4.1, or: the Androidification of Calendar's list view~~ got merged into #572
- [x] (#571) Chapter 4.2, or: the Androidification of StudentOrgs' Detail view

---

- [x] (#570) Chapter 5.1, or: how to Port BuildingHours to Row+friends
- [x] (#572) Chapter 5.2, or: how to Port Calendar to Row+friends
- [x] (#566) Chapter 5.3, or: how to Port Dictionary to Row+friends
- [x] (#567) Chapter 5.4, or: how to Port Menus to Row+friends
- [x] (#569) Chapter 5.5, or: how to Port Courses to Row+friends
- [x] (#560) Chapter 5.6, or: how to Port News to Row+friends
- [x] (#565) Chapter 5.7, or: how to Port Orgs to Row+friends
- [x] (#573) Chapter 5.8, or: how to Port Transit/Bus to Row+friends

I will close this PR when all of the above are done.

---

> Old message

This PR brings Android up a lot closer to iOS in terms of feeling native.

~~I will include screenshots at a later date.~~

There are likely things that need tweaking. It's time to go difference hunting! If you don't like something in here, please speak up.

This PR centralizes some list styles so they're easier to update.

## iOS

Before | After
-------|------
calendar | calendar
dictionary | dictionary
hours | hours
menus | menus
news | news
org detail | org detail
org list | org list

## Android
Before | After
-------|------
calendar | calendar
dictionary | dictionary
hours | hours
menus | menus
news | news
org detail | org detail
org list | org list

PR #567, opened 2017-01-13
Port MenuView to ListRow&co

hawkrives:
> This is a PR in a series of conversion PRs, split out of #499

I switched from using ✩, a unicode char, to using a star icon in Ionicons. Makes the line height more consistent.

I changed the toolbar from using padding to separate the title and subtitle to just using a space.

That's the substantive changes, I think. Beyond that, I switched to using Row, Column, ListRow, and such.

## Stav
| Platform | Before | After |
| ---:| --- | --- |
| iOS | ios menu stav a | ios menu stav b
| Android | android menu stav a | android menu stav b

## Cage
| Platform | Before | After |
| ---:| --- | --- |
| iOS | ios menu cage a | ios menu cage b
| Android | android menu cage a | android menu cage b

## Pause
| Platform | Before | After |
| ---:| --- | --- |
| iOS | ios menu pause a | ios menu pause b
| Android | android menu pause a | android menu pause b

### DIFF
--- a/views/menus/components/filter-menu-toolbar.js
+++ b/views/menus/components/filter-menu-toolbar.js
@@ -31,9 +31,10 @@ export function FilterMenuToolbar({date, title, filters, onPress}: PropsType) {
   return (
     <Toolbar onPress={onPress}>
       <View style={[styles.toolbarSection, styles.today]}>
-        <Text>{date.format('MMM. Do')}</Text>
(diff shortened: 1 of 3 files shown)


hawkrives:
I think that the Menus should probably be redone on Android to follow the Material guidelines better, but I think that this is much more android-y than HEAD's styles.

drewvolz (approved):

Issue #590, opened 2017-01-15
Top Menu Bar Button Alignment Off on Non-Default Font Sizes


rye:
In the attached screenshots, you can see that the button and text in the top menu bar is/are misaligned.  My phone uses the "Small" Android Display font size, and I believe that is what causes this.  Shouldn't this be dynamically centered vertically, or is that not possible/easy?

I'm running version `2.0.0-beta.1`.

<img width=480 src=https://cloud.githubusercontent.com/assets/1566689/21966414/490a93a4-db38-11e6-9739-b7d801ea1341.png>
<img width=480 src=https://cloud.githubusercontent.com/assets/1566689/21966412/49075f2c-db38-11e6-90f0-9648fdce4a0e.png>
<img width=480 src=https://cloud.githubusercontent.com/assets/1566689/21966413/490a85a8-db38-11e6-9861-9e5245f17d69.png>
<img width=480 src=https://cloud.githubusercontent.com/assets/1566689/21966415/490d2f92-db38-11e6-8a9c-44ed1d6dcf1a.png>

hawkrives:
You would think that it would be easy, wouldn't you? I would have expected that. 

I'll take another stab at it and see if I can make the computer center them vertically instead of the current method of tweaking pixel values. 

Thanks for reporting! We'll see if we can't get this fixed in the next release.

hawkrives:
@drewvolz, @elijahverdoorn We should remember to test v2.1 with alternate font sizes.

hawkrives:
Me, in Slack:

> A quick update on https://github.com/StoDevX/AAO-React-Native/issues/590: The issue is that we respect dynamic font sizes on the navbar. I've checked stock iOS apps, and they don't change the font size on the navbar; I haven't checked on Android yet, but I wouldn't be surprised if they do the same.
> 
> Therefore, I propose that we set `allowFontScaling={false}` on our navbar components. (Where, exactly, is tbd.)
> 
> The biggest problem with that, is that said property is only respected on iOS. Or, it was, until https://github.com/facebook/react-native/commit/4394419b60c7dfe5287e5e3f897e855eb7dc57c2 happened. Unfortunately, said commit is only in v0.41-pre, _but_ it seems pretty self-contained.
> 
[...]

PR #596, opened 2017-01-18
allow applying rn autoSizedText patch

hawkrives:
> Part of #590

Just run `node scripts/apply-test-patch.js`.

That applies https://github.com/facebook/react-native/commit/4394419b60c7dfe5287e5e3f897e855eb7dc57c2 into our local node_modules codebase.

I haven't tested that said patch does what we want it to, but I'll get to that shortly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,6 +76,7 @@
     "eslint-plugin-babel": "4.0.0",
     "eslint-plugin-flowtype": "2.30.0",
     "eslint-plugin-react": "6.8.0",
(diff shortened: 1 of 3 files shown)


drewvolz (approved):

hawkrives:
I'm going to merge this; it's just a script that we don't need to run, but it'll be there if we wind up needing it. Or, if we don't use it before 0.41 comes out… then I'll remove it again.

Issue #543, opened 2017-01-10
Add descriptions to remaining dietary tags


hawkrives:
**In Balance** 
Contains a balanced portion of whole grains, fresh fruits and vegetables, and lean protein with a minimum amount of healthy fat

**Kosher**
Contains only ingredients prepared in accordance with Jewish Dietary Law

hawkrives:
they go in https://github.com/StoDevX/AAO-React-Native/blob/master/images/dietary-filters.js

hawkrives:
Anyone is welcome to do this.

elijahverdoorn:
Interested in a small change? @ekmanzu? @zimmer18?

elijahverdoorn:
I just did this one in #597.

PR #597, opened 2017-01-18
Add kosher and made without gluten descriptions

elijahverdoorn:
Just a data add as described in #543.

We can hold this until post 2.1.


hawkrives:
Great! Thanks.

Do you _want_ to hold it? I'd say this falls squarely into a gray area between "feature" and "bug", since all the other items already had a description…

hawkrives (approved):

PR #599, opened 2017-01-19
Reduxify Settings, SIS, lib, … too much

### LINKED ISSUE #220: ErrorView state/view not refreshed unless view is popped in special case

### LINKED ISSUE #501: Ensure that "Sign In"/"Sign Out" labels work correctly
Sometimes "sign out" appears when the signin process didn't finish successfully. It should still show signin for that case.

hawkrives:
> Closes #501. Closes #220.

I'm not _entirely_ sure how I wound up with ~700 more lines than I started with, but whatevs.

This PR is pretty much a complete refactoring of `lib/`, a restructure of `views/settings` and `views/sis` to use Redux instead of internal state, and a rearchitecture of `views/settings` to be composed of components instead of a single giant render method. It also splits up the reducers in `flux/` so they can live in their own files.

Oh, and it removes both `frisbee` and `react-native-cookies`. And it extracts that neat blue button of yours, @elijahverdoorn, into `components/button`, and uses it in several more places.

## About `lib/`
I reorganized `lib/`. Actually, I guess I rewrote most of it. Primarily, I removed all the logic about what updates when, and cache management, and such. 

Cache management is now structured around `lib/cache`. `lib/cache` and `lib/storage` are the _only_ places that code should access AsyncStorage, as they can provide guarantees about the type of data that's coming out of and going into the store.

> Rule of thumb: If something needs to access AsyncStorage, it should have helper functions added to one of those two files.

Within `lib/courses` and `lib/financials`, which you'll notice are now directories instead of monolithic files, I restructured the logic into `load*`, `fetch*`, and `parse*` functions. The `load*` group contain the logic for switching between fetching data from the server, and loading it from the cache; the `fetch*` functions handle actually loading it from the server, and the `parse*` functions parse data from the responses within the `fetch*` group.

You'll notice that each function returns an `{error, value}` object – I revamped the logic flow to avoid `throw` and `try/catch` as much as possible. 

I moved the actual cache expiry computations into `lib/cache`, so all the calling function gets is `{isCached, isExpired, value}`, which is all I've needed so far. The _timing_ of the cache expiry is handled within `cache.js` itself.

I think cache.js is pretty cool, and you should glance through it at some point.

## The Rearchitecture of `views/settings`
```js
return (
<ScrollView>
  <TableView>
    <CredentialsLoginSection />
    <TokenLoginSection />
    <SupportSection />
    <OddsAndEndsSection />
  </TableView>
</ScrollView>
)
```

I think that pretty much speaks for itself.

There's an argument for moving the redux subscriptions further down the tree, so that there aren't as many props being passed out of Settings, but that's maybe a change for another time.

## `redux-redux-redux`
This is the third time I've tried to port Settings and SIS, among others, to Redux. This particular branch appears to have begun in early December.

I moved the logic of when to update the various components out of said components into the redux reducers.

This also allows us to check if the student's credentials are valid at launch time, along with preloading the various bits of information (courses, meals, balances) before opening their associated screens.

## Screenshots
label | iOS | Android
--- | --- | ---
Balances |  | n/a
Balances (partly logged in) | | n/a
Courses |  | n/a
Logged In Already |  | n/a
Settings |  | n/a

Android screenshots will come when I can get my emulator to behave.


hawkrives:
I am 99% confident that this all works.

Treat this as a complete revamp of the existing areas, rather than a change, maybe? That might make the review a bit easier.

I'm sorry that it's so big. I can see ways that I could have split it up, but I didn't have time, _especially_ not enough time to get this tested before we want to release.

Thanks all!

hawkrives:
Drew also pointed out that we could maybe rate limit ourselves for the SIS, instead of switching to the unofficial transcript; I might revert to that later. That probably won't make this release though.

hawkrives:
The only bug I know about wrt this PR is that, if you log out of Google, go into the SIS/Courses tab, log in there, then go back to the Courses view, it's blank until you start to scroll.

That's a recurrence of #321, and I just solved it with d49f413.

I now know of no issues with this PR.

hawkrives:
well. I say that.

image

It just managed to request _only_ the current term's Unofficial Transcript page.

Guess I get to go poke at that again.

hawkrives:
nooo bad broken build.

edit: darn my strict lints

edit 2: now it should pass

elijahverdoorn:
I'm reviewing this to get it in the build for release... hopefully testing tonight!

hawkrives:
I've got a collection of on-campus friends on the Testflight beta, too, so if we merge it to master for a day or two before release, they could test it out.

Thanks for perusing! I think the new stuff is a lot more straightforward, if you can find where it starts; there are far fewer starting points for the login/sis stuff now.

hawkrives:
so, somehow I added ~300 lines of code, 0 new tests, and yet code coverage _increased_? huh.

elijahverdoorn (approved):
I'm kinda new to Redux code, but overall I think it brings the app forward and will be a good change. I'll have to learn a bit more to be 100% able to read all this code with the same ease I read our regular PRs, but from what I can understand it looks good and I am excited with the direction that this takes us. Good work and thanks to @hawkrives.

hawkrives:
Kk. I'm going to merge it and it'll be in tonight's beta, so I'll ask the campus testers to check it out.

PR #606, opened 2017-01-19
fix custom dimension tracking in analytics

### LINKED ISSUE #601: Analytics is not tracking custom dimensions
Problem is that in https://github.com/StoDevX/AAO-React-Native/blob/master/analytics.js, we're just calling trackEvent, not trackEventWithCustomDimensions. 

I remember that there was a problem somewhere with the custom dimensions, but I don't remember if it was after correctly setting up the custom dimensions on Google's side or not. . 

Without fixing this, we won't know how people have arranged their home screens or what filters they've enabled.

hawkrives:
> closes #601

I like data? Now we collect a bit more?

I've tested it on Android 4.3, where it works and sends stuff to Google. I'm still not sure how to _view_ the custom dimensions, but I think we're at least sending them now.

### DIFF
--- a/analytics.js
+++ b/analytics.js
@@ -19,9 +19,9 @@ disableIfOptedOut()
 // Google requires that custom dimensions be tracked by index, and we only get
 // 20 custom dimensions, so I decided to centralize them here.
 export function trackMenuFilters(menuName: string, filters: mixed) {
-  tracker.trackEvent('menus', 'filter', {label: menuName}, {'1': JSON.stringify(filters)})
+  tracker.trackEventWithCustomDimensionValues('menus', 'filter', {label: menuName}, {'1': JSON.stringify(filters)})
 }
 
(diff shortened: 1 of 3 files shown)


hawkrives:
I pushed a second commit: it shrinks the amount of data we're sending over the network for the menu filters.

Because really, do we need to send 

> Made without Gluten-Containing Ingredients: Due to our open kitchens that handle gluten, we cannot guarantee that items made without gluten-containing ingredients are “gluten-free,” as defined by the FDA.  We make every effort to avoid gluten cross-contact; however there is always the potential for cross-contact with other gluten-containing food items, particularly in our self-serve facilities. We encourage guests to speak to the chef or manager regarding any questions about ingredients.

with every menu filter change?

(Spoiler: no.)

This only sends a `{key: string, type: string, enabled: boolean, spec: 'n/a'|'Title,Title,Title'}` object, which should be a lot smaller and nicer.

Issue #544, opened 2017-01-10
Add Offices to hours view


hawkrives:
Probably should, since we've got literally every room around it listed.

What category should it go in?

```yaml
name: Center for Multicultural Engagement
abbreviation: CMIE
category: Offices

schedule:
  - title: Hours
    hours: 
      - {days: [Mo, Tu, We, Th], from: '8:00am', to: '8:00pm'}
      - {days: [Fr], from: '8:00am', to: '5:00pm'}

breakSchedule:
  fall: []
  thanksgiving: []
  winter: []
  interim: []
  spring: []
  easter: []
  summer: []
```

* [x] CMIE
* [x] OSA
* [ ] Gender & Sexuality Center
* [x] Conferences, Camps and Events
* [ ] Telecom
* [x] Parking
* [x] Admissions
* [x] Piper Center
* [ ] Bon App office
* [x] Business
* [x] Financial Aid
* [x] SARN
* [x] Pastor's Office
* [x] IT Helpdesk
* [x] Registrar

drewvolz:
I think it might be the [Center for Multicultural and International Engagement](http://wp.stolaf.edu/cmie/)

drewvolz:
I could be wrong. Maybe it is the [International Student Life](http://wp.stolaf.edu/buntrock/files/2013/09/MapLevel2_Jul13.pdf) office?

hawkrives:
Second one, yeah.

hawkrives:
That is one outdated floorplan, buntrock

hawkrives:
It might be the CMIE too.

elijahverdoorn:
I propose adding an "Office" category with _some of_ the following:

[edit(hawken): moved to OP]

I might have missed some, anyone have suggestions?

_EDIT: Making this the list of offices to add. Feel free to add to the list as you think of things. Not all of these will go under the "offices" category._

drewvolz:
Telecom, Parking, Admissions, Piper Center, Bon App office... to name a few.

hawkrives:
Helpdesk is already in the hours list

hawkrives:
From Bidit:

> 8am-8pm on Monday-Thursday, 8am-5pm on Friday and closed on Saturday and Sunday.
> 
> It's a pretty popular hangout for international students. So might be worth it.
> 
> It's a student lounge for everyone, some choose to go and some don't, so I wouldn't expect any social effect.

hawkrives:
[...]

PR #608, opened 2017-01-21
Building hours

zimmer18:
Add building hours for
- OSA
- Conferences, Camps and Events
- Parking office
- Admissions
- Piper Center
- Business Office
- Registrar
- Financial Aid
- Pastor's Office
- CMIE

and remove all extra apostrophes from all the building hour yaml files

### DIFF
--- a/data/building-hours/5-2-asc.yaml
+++ b/data/building-hours/5-2-asc.yaml
@@ -1,9 +1,9 @@
-name: 'Academic Support Center'
-abbreviation: 'ASC'
-category: 'Help and Support'
+name: ASC
+subtitle: Academic Support Center
+category: Help and Support
 
 schedule:
-  - title: 'Hours'
+  - title: Hours
     hours:
       - {days: [Mo, Tu, We, Th], from: '8:00am', to: '9:00pm'}
       - {days: [Fr], from: '8:00am', to: '5:00pm'}
--- a/data/building-hours/9-osa.yaml
+++ b/data/building-hours/9-osa.yaml
@@ -1,0 +1,17 @@
+name: 'OSA'
+subtitle: 'Office of Student Activities'
+category: 'Offices'
+
+schedule:
+  - title: 'Hours'
+    hours:
+      - {days: [Mo, Tu, We, Th, Fr], from: '8:00am', to: '5:00pm'}
+
+breakSchedule:
+  fall: []
+  thanksgiving: []
+  winter: []
+  interim: []
+  spring: []
+  easter: []
+  summer: []
--- a/data/building-hours/9-osa.yaml
+++ b/data/building-hours/9-osa.yaml
@@ -1,9 +1,9 @@
-name: 'OSA'
-subtitle: 'Office of Student Activities'
-category: 'Offices'
+name: OSA
+subtitle: Office of Student Activities
+category: Offices
 
 schedule:
-  - title: 'Hours'
+  - title: Hours
     hours:
       - {days: [Mo, Tu, We, Th, Fr], from: '8:00am', to: '5:00pm'}
 
--- a/data/building-hours/9-piper-center.yaml
+++ b/data/building-hours/9-piper-center.yaml
@@ -1,8 +1,8 @@
-name: 'Piper Center'
-category: 'Offices'
+name: Piper Center
+category: Offices
 
 schedule:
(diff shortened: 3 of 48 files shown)


zimmer18:
Closes #544

data/building-hours/5-2-asc.yaml
```diff
@@ -1,9 +1,9 @@
-name: 'Academic Support Center'
-abbreviation: 'ASC'
-category: 'Help and Support'
+name: ASC
+subtitle: Academic Support Center
```
hawkrives, line 2: Why flip this?

data/building-hours/9-osa.yaml
```diff
@@ -0,0 +1,17 @@
+name: OSA
+subtitle: Office of Student Activities
```
hawkrives, line 2: This feels like it should be Name and Abbreviation instead of Name and Subtitle...

hawkrives (commented):
I remarked on two places in the diff. 

It's entirely subjective, but some buildings I think about as the abbreviation, and some I think of by their name. 

Buildings with really long names should probably have the full name as the subtitle, like the CMIE, but the ASC and the OSA should, IMHO, be Name (Abbr). 

In part, that's a judgement call around how common the abbreviation is. I don't think of OSA as OSA, since I was never in an elected position (actually, I think of it as the SGA room, but that's not an option here.)

I'm rambling. Point is, I'd like to hear other people's opinions on those two rooms. 

Thanks for doing this!

data/building-hours/9-piper-center.yaml
```diff
@@ -0,0 +1,17 @@
+name: Piper Center
+category: Offices
+
+schedule:
+  - title: Hours
+    hours:
+      - {days: [Mo, Tu, We, Th, Fr], from: '8:00am', to: '5:00pm'}
+      - {days: [Tu, We], from: '7:00pm', to '9:00pm'}
```
hawkrives: The Travis failure: 
```json
building-hours/9-piper-center.yaml is invalid
{
  "keyword": "additionalProperties",
  "dataPath": ".schedule[0].hours[1]",
  "schemaPath": "#/additionalProperties",
  "params": {
    "additionalProperty": "to '9:00pm'"
  },
  "message": "should NOT have additional properties"
}
```

Looking at the file, I think you missed the colon between "to" and the time.

docs/building-hours.json
```diff
@@ -2063,6 +2063,399 @@
...
+              "from": "8:00am",
+              "to": "5:00pm"
+            },
+            {
+              "days": [
+                "Tu",
+                "We"
+              ],
+              "from": "7:00pm",
+              "to '9:00pm'": null
```
hawkrives: Yeah. This is what it got translated to.

hawkrives (changes requested):

hawkrives:
Finally, do we want to close #544 before we get:

- Gender & Sexuality Center
- Telecom
- Bon App office

added?

data/building-hours/5-2-asc.yaml
zimmer18: This was flipped to match all the other abbreviations. All the other ones use the way I changed it too.

zimmer18:
Let's close this one and mark it as a completed 2.1 release blocker. I have made a new issue #612 with these remaining three and it won't be a 2.1 release blocker.

data/building-hours/5-2-asc.yaml
hawkrives: Oh, the other ASC places? Good point.

data/building-hours/9-osa.yaml
drewvolz: I feel that @hawkrives has a decent point. I might flip the name and subtitle. I bet it is thought of as the "SGA office", so an expansion couldn't hurt.

hawkrives: Actually, I withdraw my complaint. 

Are there any cases left that use the abbreviation field?

hawkrives: The only one left is HOM. 

Could you switch it to Name+subtitle as well, please?

hawkrives (approved):
LGTM! Thanks!

drewvolz:
Thank you @zimmer18

PR #617, opened 2017-01-22
Create hosted building hours

hawkrives:
Building off of the hosted pause menu, which falls back to a local copy if the device can't fetch, this introduces a wrapper around the BuildingHours view that will fetch the hours list from Github if it is able. 

This allows us to update the building hours list remotely.

It is a new feature, that is true. Do we want to pull it into 2.1? I'd like to. I think it brings benefits to this next release cycle.

### DIFF
--- a/views/building-hours/index.js
+++ b/views/building-hours/index.js
@@ -1,14 +1,14 @@
 // @flow
 /**
  * All About Olaf
- * Building Hours list page
+ * Building Hours view. This component loads data from either GitHub or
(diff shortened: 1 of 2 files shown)


drewvolz:
I would like to pull this into 2.1.

elijahverdoorn (approved):
Cool - I think we've been talking about doing this since June!

hawkrives:
Yay!

PR #619, opened 2017-01-22
load local faqs on development

drewvolz:
load the local faqs when developing instead of rendering remote data

### DIFF
--- a/views/faqs/index.js
+++ b/views/faqs/index.js
@@ -35,6 +35,11 @@ export class FaqView extends React.Component {
       tracker.trackException(err.message)
       console.warn(err.message)
     }
+
+    if (__DEV__) {
+      html = faqs
+    }
+
     this.setState({html: html})
   }
 
--- a/views/faqs/index.js
+++ b/views/faqs/index.js
@@ -32,6 +32,7 @@ export class FaqView extends React.Component {
       let blob: {text: string} = await fetchJson(this.url)
       html = blob.text
     } catch (err) {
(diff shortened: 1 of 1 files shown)


views/faqs/index.js
```diff
@@ -35,6 +35,11 @@ export class FaqView extends React.Component {
       tracker.trackException(err.message)
       console.warn(err.message)
```
hawkrives, line 37: Would you add a fallback in here, too, please?

views/faqs/index.js
```diff
@@ -35,6 +35,11 @@ export class FaqView extends React.Component {
...
     }
+
+    if (__DEV__) {
+      html = faqs
```
hawkrives, line 41: This needs to be faqs.text

hawkrives (changes requested):

views/faqs/index.js
drewvolz: It's using the text imported from the json above.

`import {text as faqs} from '../../docs/faqs.json'`

hawkrives: Ah. excdllent point. Carry on.

hawkrives: Thank you for not correcting, autocorrect...

views/faqs/index.js
drewvolz: Done.

hawkrives (approved):
LGTM.

PR #623, opened 2017-01-23
update FAQs

hawkrives:
- add cage-specials note
- add incorrect building hours note


hawkrives:
Adds 

> - Building Hours: It reports the 2am buildings as open too often; for instance, the Pause is open on Saturday at 2am, continuing from Friday, but the app reports it as being open on Friday at 2am, as well. This is incorrect.
> - The Cage menu only shows specials: that's a limitation of the data we get from BonApp. SGA is talking with them about providing the rest of the Cage menu. One day, it'll suddenly work!

drewvolz (approved):

elijahverdoorn (approved):
I think this clears up all the potential issues we'd have for releasing 2.1.

drewvolz:
@elijahverdoorn the best part is -- if we find any others, we can update it from here.

PR #644, opened 2017-01-27
fix in-menu image loading on android

hawkrives:
… I was providing the `url` prop instead of `uri`.

Now it works.

pasted image at 2017_01_27 04_02 pm


elijahverdoorn (approved):
Please merge this to `2.1.1` also.

PR #645, opened 2017-01-28
Edit safe ride info

erichkauffman:
Change Safe Ride to Safe Guard, add info for Safe Walk, change button text

### DIFF
--- a/data/contact-info/1-safe-ride.yaml
+++ b/data/contact-info/1-safe-ride.yaml
@@ -1,11 +1,12 @@
-title: Safe Ride
+title: Safe Guard
 phoneNumber: '5077863666'
-buttonText: Request Safe Ride
+buttonText: Call Public Safety
 image: safe ride.jpg
 text: >
-  Students concerned about their personal safety may contact Safe Ride to
+  Students concerned about their personal safety may contact Public Safety to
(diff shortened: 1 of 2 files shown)


data/contact-info/1-safe-ride.yaml
```diff
@@ -1,11 +1,12 @@
-title: Safe Ride
+title: Safe Guard
 phoneNumber: '5077863666'
-buttonText: Request Safe Ride
+buttonText: Call Public Safety
```
elijahverdoorn, line 3: This is one that has been debated - while _I_ know that pubsafe = safe ride, does the majority of campus? Is it clear enough?

erichkauffman: I think that if someone uses the app to call a safe ride, they won't be surprised when Pub Safe answers. EDIT: If the button says "Call Public Safety."

hawkrives: I'm good with the new button text.

elijahverdoorn: That's fine by me then, I just wanted another set of eyes.

elijahverdoorn (approved):

hawkrives (approved):

PR #649, opened 2017-01-28
analytics opt out

drewvolz:
* allows for users to opt-in or out of sharing google analytics data
* all credit goes to @hawkrives ... he wrote the entire thing

before


hawkrives (approved):
:shipit:

PR #665, opened 2017-02-06
Reorganize the source code

hawkrives:
This moves the `lib/`, `views/`, and `flux/` folders underneath a new top-level `source/` folder.

This is in preparation for a larger-scale refactoring of code to move related stuff next to each other (such as putting the reducers for a given screen inside that screen's folder).

Thoughts?


drewvolz (approved):
Thank you for taking this one on. In your own words, I have to agree that this reorg is well-suited for new contributors at the expense of existing ones. That is not a bad thing. Grouping *our* contributions within `source/*` will help in the long run.

Thank you again! I'm happy with this reorg.

elijahverdoorn (approved):
I agree with @drewvolz, that this change is a step in the right direction, but will take some getting used to for those of us that have been here since the beginning.

elijahverdoorn (commented):
One quick question - should we use `/src` rather than `/source`? I looked at a few other RN apps and they all seem to use `/src`

hawkrives:
Meh. It doesn't really matter – I came down on the side of adding vowels to the folder name. 

I see src vs. source as, like, dating back to old Unix, with etc instead of etcetera, usr instead of user, var instead of various.

I'd _prefer_ to use source? But I can use src/ instead of we'd rather leave out vowels. 

:smile:

hawkrives:
Also, Travis why u die this time? What did I forget to rename.

elijahverdoorn:
I'll leave it up to you, @hawkrives to decide on `/src` vs. `/source`. Good to merge whatever you decide.

hawkrives:
I'll wait for #682 to be merged, I guess.

Or, if you want to handle the conflicts, @elijahverdoorn, we can merge this PR at any time.

hawkrives:
Closed in favor of #689.

PR #668, opened 2017-02-07
Add CircleCI support

hawkrives:
This adds initial CircleCI support, in addition to Travis.

It's currently only running javascript tests, but we hope to move the iOS tests to it soon.


drewvolz (approved):
Awesome, I think moving towards CircleCI is the right idea. Good work!

PR #670, opened 2017-02-07
Update to Fastlane 2.14

hawkrives:
Because updates.

Nothing big for us, just minor bugfixes around working with iTunes connect (which we don't use).


hawkrives:
We're going to ignore the Circle failure here, since it's a flow issue not a fastlane one

drewvolz (approved):
Fastlane upgrades are always welcome.

PR #671, opened 2017-02-07
Convert CircleCI to macOS builds

hawkrives:
I think we've got them enabled now.

See the build status at https://circleci.com/gh/StoDevX/AAO-React-Native/24

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -6,7 +6,9 @@ general:
 
 machine:
   pre:
-    - brew install rvm
+    - brew install gnupg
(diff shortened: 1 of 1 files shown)


drewvolz (approved):

hawkrives:
I don't think we're properly enabled for their mac builds yet, despite what that email said. We should probably reach out again.

drewvolz:
> Note: this option only works if the project builds under an OS X plan. Configure that option here.

I think you are correct that we need to reach out again.

PR #679, opened 2017-02-09
Update student orgs endpoint

drewvolz:
checkimhere rebranded as presence so we will start using their new endpoint to get data


hawkrives:
Oh good, it actually is a valid URL.

I had a moment of panic where I was afraid that they'd rebranded without updating the URL 😆

elijahverdoorn (approved):

Issue #652, opened 2017-01-29
app transport exceptions in oleville


drewvolz:
Hyperlinks need to be disabled within the Oleville section as they can lead to app transport exceptions for http domains that are not white listed within the app.

Is this the appropriate time to think about aggregating Oleville to the news portion of the app?

<img width="417" alt="Oleville ATS" src="https://cloud.githubusercontent.com/assets/5240843/22408777/c27c8f20-e64c-11e6-8fb2-c5c23215ebcf.PNG">

hawkrives:
Sure. 

I'd also like to investigate either catching navigation events from the WebViews and opening them in a browser, or even using HTMLparser2 to avoid webviews entirely.

drewvolz:
Absolutely, that is a great idea.

PR #682, opened 2017-02-09
Move Oleville to the news section

elijahverdoorn:
Oleville view was really news, and I like having an even number of front-page buttons. Closes part of #652.


views/news/tabs.js
```diff
@@ -14,15 +15,11 @@ export default [
...
-    rnVectorIcon: {iconName: 'megaphone'},
-    component: NewsContainer,
-    props: {
-      url: 'http://oleville.com/politicole/feed/',
-      mode: 'rss',
-      name: 'PoliticOle',
-    },
+    id: 'OlevilleView',
+    title: 'Oleville',
+    rnVectorIcon: {iconName: 'mouse-pointer'},
```
hawkrives: `mouse-pointer` is not a valid icon for the tab bar, apologies. Tab bar uses ionicons, while the homescreen uses Entypo.

elijahverdoorn: That's what I get for developing on android. Fixed.

views/news/tabs.js
```diff
@@ -14,15 +15,11 @@ export default [
...
-    rnVectorIcon: {iconName: 'megaphone'},
-    component: NewsContainer,
-    props: {
-      url: 'http://oleville.com/politicole/feed/',
-      mode: 'rss',
-      name: 'PoliticOle',
-    },
+    id: 'OlevilleView',
+    title: 'Oleville',
+    rnVectorIcon: {iconName: 'ion-mouse'},
```
hawkrives: Sadly, it has to be an icon that's available in an `ios-` variant.

See https://github.com/StoDevX/AAO-React-Native/blob/master/views/components/tabbed-view/index.ios.js#L35 – the tab bar loads `ios-${name}-outline` and `ios-${name}` variants.

You also won't need the `ion-` prefix, as rn-vector-icons removes that.

drewvolz: Changing the icon to `happy` which matches the icon used in `calendar`

drewvolz (approved):

hawkrives (approved):

hawkrives:
If you want a different icon, @elijahverdoorn, glance through http://ionicframework.com/docs/v2/ionicons/ and use the name in the left column.

The `happy` icon was chosen because it's what we were using in the Calendar screen already.

PR #687, opened 2017-02-11
Move carleton views list out of menus/tabs.js

hawkrives:
When they were both in tabs.js, it resulted in a circular import:

`carleton-list.js` imports `carleton` from `tabs.js`, and `tabs.js` imports `CarletonMenuPicker` from `carleton-list.js`.

RN didn't complain about it, but something else I was experimenting with did, and circular imports seem like something that might become an error in the future.


drewvolz (approved):
Good catch.

PR #693, opened 2017-02-12
Update News views

hawkrives:
This PR:

1. Brings the Oleville view in line with the rest of News
2. Adds thumbnail images for featured images for some feeds

---

--- | List | Article
--- | --- | ---
St. Olaf | stolaf-list copy | stolaf-article copy
Oleville | oleville-list copy | oleville-article copy
The Mess | mess-list copy | mess-article copy

---

I thought about using cards, instead, but it turns out that the Material Design guidelines explicitly ask you [not to do that](https://material.io/guidelines/components/cards.html#cards-usage): 

image

---

Um. What else.

I broke NewsContainer into two components, NewsContainer and NewsList.

NewsContainer fetches the data and parses it from RSS/WP-JSON into a single StoryType-shaped object. It passes an array of those to NewsList, which actually renders the list.


hawkrives:
Oh, I also added KSTO's news view, but Pages is currently down, so you won't be able to see it.

hawkrives:
This PR probably also closes https://github.com/StoDevX/AAO-React-Native/pull/463?

hawkrives:
Hi Danger. I'm not sure how I would split it into multiple parts, unfortunately.

drewvolz (approved):

hawkrives:
@drewvolz Could you post that screenshot I sent you of the mixed card/list news views, please?

drewvolz:
Mixed mediums

mixed

hawkrives:
We might want to remove KSTO again? Or at least filter out some of their content - it's kinda weird to have the form-based posts show up. 

We should investigate. If those posts all share a category or a tag, we can filter out all posts with that category / tag from KSTO's view. If they have nothing in common, then... I guess we either turn off the ksto view or ask them to add categories. 

We could probably pressure them into categories, if they want an increased audience size, by showing them our usage stats. (Admittedly, most of our users check the menus, but some open News, I think?)

drewvolz:
KSTO has a few categories. I think you were right in saying we could approach them about their categorization. If we were able to get only reviews within reviews, I would consider querying categories 3 and 38, although we might have to think about what we want to show.

I am thinking album reviews and some form of news but we should decide for sure what we would like to display. Their [twitter](https://twitter.com/ksto_radio) is super active -- that might be a cool thing to include.

|id |category|notes
|---|---|---|
3 | reviews | ideal if application forms weren't tagged as this 
10 | hip-hop/electronica | album reviews
11 | top 100 | album reviews
12 | alternative | album reviews
19 | electronica | album review
24 | news |blank entries, non-working embedded content
25 | weekend preview | non-working embedded content
29 | campus news |blank entries, non-working embedded content
38 | ksto news / dj links | useful announcements to DJs

hawkrives:
@elijahverdoorn I'm going to wait until you've got a moment to look at this.

drewvolz:
I see that we have access to potential metadata about stories. Authors, categories, and date published. Perhaps we could display some of those somewhere.

hawkrives:
@drewvolz I'd be up for that; let's look at other apps for inspiration (after we merge this?)

PR #697, opened 2017-02-13
Allow last char of user/pass fields to be removed

### LINKED ISSUE #696: Cannot delete last character from username login field
**Problem**
I'm playing with the St. Olaf login form in Settings. I can delete all but the first character of my username or password. Further investigation by @hawkrives reports that no redux action is fired when one tries to remove the last character.

We should investigate why no action is fired when the last character is removed. 

**Impact**
Login form in settings

drewvolz:
Removes truthiness check from settings onchange

Closes #696

### DIFF
--- a/source/views/settings/sections/login-credentials.js
+++ b/source/views/settings/sections/login-credentials.js
@@ -50,7 +50,7 @@ export class CredentialsLoginSection extends React.Component {
           label='Username'
           _ref={ref => this._usernameInput = ref}
           disabled={loading}
-          onChangeText={text => text && this.props.onChangeUsername(text)}
+          onChangeText={(text='') => this.props.onChangeUsername(text)}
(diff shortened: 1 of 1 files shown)


source/views/settings/sections/login-credentials.js
```diff
@@ -50,7 +50,7 @@ export class CredentialsLoginSection extends React.Component {
           label='Username'
           _ref={ref => this._usernameInput = ref}
           disabled={loading}
-          onChangeText={text => text && this.props.onChangeUsername(text)}
+          onChangeText={(text='') => this.props.onChangeUsername(text)}
```
hawkrives, line 53: To note: we only had to set the default value here to satisfy Flow. As far as we can tell, it would always be a string anyway, but Flow disagreed, so here we are.

hawkrives (approved):
LGTM.

PR #710, opened 2017-02-19
Add support for recreating the station menus from the food items

### LINKED ISSUE #660: [RFC] Cage menu is blank (again)
> As is the Caf, but we have a solution for the Cage.

Over Christmas break, when we were building and finishing `fancy-menus`, the Cage listed all of their normal menu items for us. As soon as we shipped, though, they quit using them! @drewvolz emailed them, and they directed us to their PDF menus if we wanted to see the normal selection of food.

A bit of background: BonApp's menu api returns both a list of all food items used in the requested day range, which is usually only a single day, as well as a list of menus used for that café. The Cage only has one menu per day, but the Caf has three — Breakfast, Lunch, and Dinner.

The Cage's api response returns their entire set of default menu items, but the response's _menu_ only references the specials.

I propose that, for the Cage, we ignore the listed menus and regenerate them from the big list of all food items. This would be done via a prop in the tabs file, so that it only applied to cafés that need it.

Thoughts?

---

PS: The Caf's menu is blank today because they're on Interim Break hours, but instead of saying "closed for breakfast", they have "breakfast: open, lunch: special hours, dinner: special hours, breakfast: closed". 

The Cafés api returns the next week's schedule for the requested café; each `day` has a status (open | closed) and a message, as well as a list of `dayparts`, which have a `starttime`, `endtime`, `label`, and `message`. 

Today, the response looks something like this:

```json
"days": [
        {
          "status": "open",
          "message": "",
          "date": "2017-02-03",
          "dayparts": [
            {
              "starttime": "07:00",
              "id": "1",
              "label": "Breakfast",
              "endtime": "09:45",
              "message": ""
            },
            {
              "starttime": "12:00",
              "id": "3",
              "label": "Lunch",
              "endtime": "13:00",
              "message": "Interim Break hours"
            },
            {
              "starttime": "17:00",
              "id": "4",
              "label": "Dinner",
              "endtime": "18:00",
              "message": "Interim Break hours"
            },
            {
              "starttime": "07:00",
              "id": "8",
              "label": "Closed",
              "endtime": "09:45",
              "message": "Closed for breakfast Interim Break"
            }
          ]
        },
        {
          "status": "closed",
          "message": "Closed for Interim Break",
          "date": "2017-02-04",
          "dayparts": []
        },
]
```

You can see that there are _two_ entries with the same start/end time for today, one labelled "Breakfast" and one labelled "Closed" (‽). So, the Caf's problem is just bad data. Nothing we can do about it.

hawkrives:
> Closes #660

This PR allows us to opt-in to ignoring the provided list of menu items, in favor of creating the stations from the food items themselves.

This is only necessary because the Cage menu refuses to reference their normal menu items, even though they send them down.

• | Before | After
--- | --- | ---
iOS | screen shot 2017-02-18 at 10 19 00 pm | screen shot 2017-02-18 at 10 18 45 pm

### DIFF
--- a/source/views/menus/menu-bonapp.js
+++ b/source/views/menus/menu-bonapp.js
@@ -12,9 +12,11 @@ import type {
   StationMenuType,
   ProcessedMealType,
   DayPartMenuType,
+  MenuItemContainerType,
 } from './types'
 import sample from 'lodash/sample'
 import mapValues from 'lodash/mapValues'
+import reduce from 'lodash/reduce'
 import type momentT from 'moment'
 import moment from 'moment-timezone'
 import {trimStationName, trimItemLabel} from './lib/trim-names'
@@ -29,6 +31,7 @@ const fetchJsonQuery = (url, query) => fetchJson(`${url}?${qs.stringify(query)}`
 
 type BonAppPropsType = TopLevelViewPropsType & {
   cafeId: string,
+  ignoreProvidedMenus?: boolean,
   loadingMessage: string[],
   name: string,
 };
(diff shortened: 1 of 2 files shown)


drewvolz (approved):
👍 :shipit:

PR #713, opened 2017-02-20
Fix analytics sharing

### LINKED ISSUE #653: Ensure analytics are enabled by default
> This bug is a result of https://github.com/StoDevX/AAO-React-Native/pull/649

there's a chunk in init that makes sure they're set to be enabled by default, but I have a feeling it is not working the way we want it. I want to double check that it is indeed toggled true from a fresh download or a version update.

hawkrives:
> Well hello there, multi-purpose PR. How nice to see you again.
> Closes #653 

Turns out that we had `feedbackEnabled` and `didOptOut` as synonyms – they were the same value, renamed. Which is a problem, because it means that our "on" state – "share analytics? yes" – was actually the "off" state.

This fixes that.

It also fixes a warning from React about something that Codepush does: 

> Warning: Stateless function components cannot be given refs (See ref "rootComponent" in StatelessComponent created by CodePushComponent). Attempts to access this ref will fail.

And it cleans up some sample code I copied from OneSignal's github during the onesignal@3.0 upgrade.

### DIFF
--- a/source/analytics.js
+++ b/source/analytics.js
@@ -5,18 +5,17 @@ import {
 } from 'react-native-google-analytics-bridge'
 import {stringifyFilters} from './views/components/filter'
 
-import * as storage from './lib/storage'
+import {getAnalyticsOptOut} from './lib/storage'
 
 const trackerId = __DEV__ ? 'UA-90234209-1' : 'UA-90234209-2'
 export const tracker = new GoogleAnalyticsTracker(trackerId)
 
 function disableIfOptedOut() {
-  return storage.getAnalyticsOptOut()
-    .then(didOptOut => {
-      if (didOptOut) {
-        GoogleAnalyticsSettings.setOptOut(true)
-      }
-    })
+  return getAnalyticsOptOut().then(didOptOut => {
+    if (didOptOut) {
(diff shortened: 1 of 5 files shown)


source/analytics.js
```diff
@@ -5,18 +5,17 @@ import {
...
 export const tracker = new GoogleAnalyticsTracker(trackerId)
 
 function disableIfOptedOut() {
-  return storage.getAnalyticsOptOut()
-    .then(didOptOut => {
-      if (didOptOut) {
-        GoogleAnalyticsSettings.setOptOut(true)
-      }
-    })
+  return getAnalyticsOptOut().then(didOptOut => {
```
hawkrives, line 14: `didOptOut`?

source/flux/parts/settings.js
```diff
@@ -219,7 +218,7 @@ export function settings(state: Object=initialSettingsState, action: Object) {
       return {...state, theme: payload}
 
     case SET_FEEDBACK:
-      return {...state, feedbackEnabled: payload}
```
hawkrives, line 222: and `feedbackEnabled`. Same value. Opposite meaning.

drewvolz (approved):
Thanks for fixing that.

PR #732, opened 2017-02-22
Rework how tabs work

### LINKED ISSUE #717: No event is sent for the first tab when opening a tabbed view
Analytics sends events when people switch tabs, but it doesn't send one for the first tab. 

We should fix that.

hawkrives:
> Closes #717

I've been getting kinda frustrated with the indirection of our `tabs.js` solution for a while now.

This PR, well… it's a big change.

Before:
```js
const tabs = [
  {
    id: 'BalancesView',
    title: 'Balances',
    rnVectorIcon: {iconName: 'card'},
    component: BalancesView,
    props: {myProp: true},
  },
]

export default function SISView(props: {navigator: any, route: any}) {
  return <TabbedView navigator={props.navigator} route={props.route} style={styles.container} tabs={tabs} />
}
```

After:
```js
export default function SISView({navigator, route}: TopLevelViewPropsType) {
  return (
    <TabbedView style={styles.container}>
      <Tab id='BalancesView' title='Balances' icon='card'>
        {() => <BalancesView route={route} navigator={navigator} myProp={true} />}
      </Tab>
    </TabbedView>
  )
}
```

> You'll want to read ["Function as Child Components"](https://medium.com/merrickchristensen/function-as-child-components-5f3920a9ace9) first.

The new API merges the `<TabbedView>` and `tabs={[]}` pieces of the old API, using the fact that JSX expresses a structure on its own.

A `<Tab/>` needs an `id`, which is used to identify which tab is selected and to identify the tab in analytics. It also needs a `title`, which is the rendered tab title, and an `icon`, which is iOS-only and taken from the Ionicons `ios-` icon set so it can blend in with native iOS apps.

Now, what's going on in the middle there? Are we passing a _function_ as the child of Tab?

Yep! That way, we can create whatever component we want as the child, without it being rendered until we open the tab. If we removed the function, React would go mount the entire component and we'd be doing a lot more work.

We're making JSX elements carry data around for us, instead of writing the structure ourselves.

---

Let's see if CI passes – I've tested this on iOS and Android, and there are no visual changes.

I'd like to do this, but I'd also like to invite some criticism if any, so this PR will be open for a few days as a comment period.


hawkrives:
Gee, thanks for the consistency, flow.

```bash
source/views/calendar/index.js:16
 16:       <Tab id='StOlafCalendarView' title='St. Olaf' icon='school'>
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ object type. This type is incompatible with
 16:       <Tab id='StOlafCalendarView' title='St. Olaf' icon='school'>
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ React$Element
```

hawkrives:
@drewvolz I will update the description in a bit, but do you want to take a look-see at what I've got here?

hawkrives:
I've decided that I'm going to split this into two PRs: one that does the bulk of the work, and one that just moves the tab definitions into the index files.

hawkrives:
Never mind. One of the things I did was to get rid of one stage of the navigator passing dance, so I have to have the tabs in the index files.

source/views/components/tabbed-view/tabbed-view.ios.js
```diff
@@ -0,0 +1,78 @@
...
+import {tracker} from '../../../analytics'
+import {styles} from './styles'
+import type {TabbedViewPropsType, TabPropsType} from './types'
+import * as c from '../../components/colors'
+
+export class TabbedView extends React.Component {
+  state: {selectedTab: string};
+
+  componentWillMount() {
+    this.onChangeTab(this.props.tabs.filter(Boolean)[0].id)
```
drewvolz, line 15: I cannot fault you for this actually working, and I understand that it gives you the option of nullable tabs...

hawkrives:
I am splitting this into two PRs: one for the tabbedview API reworking, and one for the TabbedView internals reworking.

I'm taking the spirit of this PR, although I don't believe that I will be using any of this code directly.

PR #733, opened 2017-02-22
Add a network cache layer

hawkrives:
> "Untold numbers of developers have hacked together an awkward, fragile system for network caching functionality, all because they weren’t aware that NSURLCache could be setup in two lines and do it 100× better. Even more developers have never known the benefits of network caching, and never attempted a solution, causing their apps to make untold numbers of unnecessary requests to the server."

Using https://developer.android.com/reference/android/net/http/HttpResponseCache.html and http://nshipster.com/nsurlcache/ as guides, I implemented a native-level network cache layer.

As I understand it, RN implemented fetch on top of XMLHttpRequest, and they built [XMLHttpRequest on top of the native requests](https://facebook.github.io/react-native/releases/0.26/docs/network.html#xmlhttprequest), so we can just set up the cache at the native code level and it'll "just work". Theoretically.

This still needs some more testing, but… yeah. If it works, this is so simple.

### DIFF
--- a/android/app/src/main/java/com/allaboutolaf/MainApplication.java
+++ b/android/app/src/main/java/com/allaboutolaf/MainApplication.java
@@ -1,6 +1,8 @@
 package com.allaboutolaf;
 
 import android.app.Application;
+import android.net.http.HttpResponseCache;
+import android.os.Bundle;
 import android.util.Log;
 
 // keep these sorted alphabetically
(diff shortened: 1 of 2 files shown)


drewvolz (approved):
In your words, "it doesn't not work" and we don't see a downside to including it. We're going to test on devices for awhile on master. 👍 :shipit:

PR #734, opened 2017-02-22
Read package version from package.json instead of app package

### LINKED ISSUE #712: Pull app version from package.json
We can't update the embedded app versions with CodePush, but I think we can update package.json, so let's move the app version in Settings from rn-device-info to package.json.

hawkrives:
> Closes #712

This PR pulls the app version from package.json instead of `react-native-get-device-info`.

This allows us to update the version via Codepush.


drewvolz (approved):
👍 :shipit:

PR #737, opened 2017-02-23
Restrict moment calculations to the minute

### LINKED ISSUE #695: Conduct a Moment granularity audit
**Problem**
 Moments (the thing that we use to compare datetimes) feel as if they're always a minute fast.

**Solution**
We should probably do an audit to make sure it compares by the minute and not anything more granular.

**Impact**
Everything that makes use of Moment for comparisons. Bus schedules, calendars, building hours, etc.

**Example**
6:00:02 pm is 4 minutes away from 6:05:00 pm, but we would like it to say 5 minutes.

drewvolz:
Moments feel like they're always a minute fast. This changes the moments that we interact with to be more relaxed in terms of our understanding of time.

Closes #695

### DIFF
--- a/source/views/building-hours/__tests__/parse-hours.test.js
+++ b/source/views/building-hours/__tests__/parse-hours.test.js
@@ -31,8 +31,8 @@ it('will add a day to the close time with nextDay:true', () => {
   let input = {days: [], from: '10:00am', to: '2:00am'}
   let {open, close} = parseHours(input, now)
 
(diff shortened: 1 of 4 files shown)


hawkrives:
Do we really only need to change it in three places + tests?

hawkrives:
I guess we do. Cool.

hawkrives (approved):

PR #745, opened 2017-02-23
Remove all <form> articles from the list of news articles

### LINKED ISSUE #744: Crash caused by submitting form in news
Hitting submit on a form that rendered in a KSTO news story view causes the application to crash.

hawkrives:
> closes https://github.com/StoDevX/AAO-React-Native/issues/744

This just does `entries.filter(entry => entry.content.includes('<form>'))`, which should avoid people talking about forms – that'd be encoded as `&lt;form&gt;`.

This does hide articles that include a form in the middle of other content, though… but so far, I think, all of our `form` entries have been entirely forms.

I want to prefer the `form {display: none}` approach, but I don't want the user to see blank stories in the list.

Actually. What about, like, `form {opacity: 0.5; pointer-events: none}`? That would essentially just disable the form, while still leaving it visible.

Thoughts?

### DIFF
--- a/source/views/news/news-list.js
+++ b/source/views/news/news-list.js
@@ -45,8 +45,8 @@ export class NewsList extends React.Component {
 
   init(props: NewsListPropsType) {
     // remove all entries with a <form> from the list
-    const entries = props.entries.filter(entry => entry.content.includes('<form>'))
-    this.setState({dataSource: this.state.dataSource.cloneWithRows(props.entries)})
+    const entries = props.entries.filter(entry => !entry.content.includes('<form>'))
+    this.setState({dataSource: this.state.dataSource.cloneWithRows(entries)})
   }
 
   renderRow = (story: StoryType) => {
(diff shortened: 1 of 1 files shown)


drewvolz:
Disabling the form and leaving it visible would probably confuse. I would think AAO was bork and that it couldn't do something that I thought it should.

It would be cool if we detected that the entire story was a form _before_ hitting the storyview, and maybe rendered it inside of the new link browsers, which could handle them?

hawkrives:
So, our options include:

- hide the forms
- replace the forms with a note
- disable the forms
- skip NewsItem and open a browser directly for stories with a form

drewvolz:
We'll explore the idea of disabling the form, providing a way to open it in another browser, and indicating there is a form on the listview of stories.

drewvolz (approved):
:+1: :shipit:

hawkrives:
In the meantime, we're going to merge this to prevent the app from crashing when people try to submit a form.

PR #746, opened 2017-02-23
Make validate-data quieter

hawkrives:
This adds a `--quiet` flag to the validate-data script, so it will only print lines that are problematic files.


drewvolz (approved):
:+1: :shipit: great!

PR #748, opened 2017-02-23
Send event for the first tab when opening a tab view

### LINKED ISSUE #717: No event is sent for the first tab when opening a tabbed view

hawkrives:
> Closes #717 

This PR sends an analytics event for the first tab opened in a tab view.

The existing behavior only sent events for the tabs you tapped on while in the tabbed view, but not the first tab itself.

### DIFF
--- a/source/views/components/tabbed-view/index.android.js
+++ b/source/views/components/tabbed-view/index.android.js
@@ -1,7 +1,7 @@
 // @flow
 import React from 'react'
-import { StyleSheet, Platform, Dimensions } from 'react-native'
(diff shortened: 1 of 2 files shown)


drewvolz:
~~I cannot get this to work worth anything. Someone else should test this because I can't get it to load. I do not place blame on this PR -- it is not the PR's fault.~~

drewvolz (approved):
Scratch that. It was due to hot reloading. This works fine. 👍 :shipit:

PR #750, opened 2017-02-23
SimpleListView

hawkrives:
`<ListView/>` comes with a lot of boilerplate.

We have to write `componentWillMount` and `componentWillRecieveProps` for each component, we have to manage a `ListView.DataSource` that we don't understand and essentially overwrite every time we get new data, and we've got a bunch of little properties that we have to set on each ListView individually.

I propose that we adopt a `<SimpleListView/>` component that wraps ListView.

It would provide sensible default values for the props we need, as well as eliminating the boilerplate of managing the DataSource.

I've also added a `forceBottomInset={true}` prop that applies the `contentInset={{bottom: Platform.OS === 'ios' ? 49 : 0}}` and `automaticallyAdjustContentInsets={false}` on iOS, so that the ListViews inset themselves above the tab bars.

SimpleListView also doesn't take a `renderRow` parameter; instead, it uses the ["function as a child component"](https://medium.com/merrickchristensen/function-as-child-components-5f3920a9ace9) pattern to render the rows.

```js
<SimpleListView
  style={styles.listContainer}
  forceBottomInset={true}
  data={entries}
  renderSeparator={this.renderSeparator}
>
  {(story: StoryType) =>
    <NewsRow
      onPress={() => this.onPressNews(story.title, story)}
      story={story}
    />}
</SimpleListView>
```

---

As a side note, I'd consider adding an `isRefreshing` and `onRefresh` prop that would enable the RefreshControl, which would further reduce the amount of boilerplate we need to write.


hawkrives:
I just added implicit RefreshControl support to SimpleListView.

If you provide the two props from RefreshControl (`onRefresh` and `refreshing`), SimpleListView will automatically hook them up to a RefreshControl for you.

hawkrives:
So um. I apparently forgot to test this before PRing, and … none of the list views render anything, now.

I'll be working on that soon.

hawkrives:
This is an issue for the Eslint problem: https://github.com/gajus/eslint-plugin-flowtype/issues/201

hawkrives:
Wow. My danger rule for camelCase filenames? It's asking me to turn "listview.js" into "listview-js". 

Guess that rule needs some tweaking.

hawkrives:
I fixed the displaying problems back in 377179b, so this works now. (Just realized I hadn't commented that I'd fixed it.)

drewvolz (approved):

PR #751, opened 2017-02-23
Remove unused Android fonts

hawkrives:
> The Android half of https://github.com/StoDevX/AAO-React-Native/pull/690

We don't use any of these fonts on Android. So let's remove them.


drewvolz (approved):
👍 :shipit:

PR #754, opened 2017-02-24
allow Contacts' ListView to scroll again

hawkrives:
I'm not entirely sure why this fixes it, but I've tested that it does.

> something something flex:1


drewvolz (approved):
👍 :shipit:

PR #757, opened 2017-02-24
Extract openUrl from NewsItem

hawkrives:
It's a pretty generic function, so… yeah.

Pulled it into `components/`.

Also pulled `HtmlView` out of NewsItem.js, so it can be reused, even though nothing else needs it.

> #759 must be merged into this PR before it is merged into master.
> #758 must be merged into this PR before it is merged into master.

Closes #759
Closes #758


drewvolz (approved):
👍

PR #758, opened 2017-02-24
Change Directory and Maps to use SFViewController/Custom Tabs

hawkrives:
instead of using embedded webviews.

As an added advantage, this will make the map work on android (yay!)


drewvolz (approved):
👍

PR #759, opened 2017-02-24
Use native browsers in Transportation's buttons

hawkrives:
Make the buttons in Transportation's "Other Modes" tab open an embedded browser, instead of using a `<WebView/>`.


drewvolz (commented):
👍

Issue #78, opened 2016-08-18
Add webview for Mobile PrintRelease website


hawkrives:
https://papercut.stolaf.edu:9192 (only available on-campus)

hawkrives:
(this could also be done with a native view the just scrapes the webview, because the webview is s-l-o-w.)

[the native view could also be a subsequent addition, after the webview.]

drewvolz:
This idea now lives [here](https://github.com/StoDevX/AAO-React-Native/wiki/Print-release-mobile).

hawkrives:
Hawken Rives, [Aug 3, 2017, 5:09 PM]:
omg I found a papercut rest api

and it's enabled

https://papercut.stolaf.edu:9192/rpc/api/rest/internal/mobilerelease/api/all-printers?username=rives

(on-campus only)

(and it's internal/undocumented, but we really only upgrade papercut once/twice a year, so … I want to keep it at least until it breaks once)

… hmm. my … command line explorations … which have no session … are only working for _my_ username

this is odd

wtf. the _server_ seems to be keeping track of what IP you are logged in with, and authorizing all requests from that IP as you

until you log out

(which it does automatically after a bit)

(but still)

and I think that it's just the mobile print release stuff that behaves that way, … but c'mon, papercut

PR #762, opened 2017-02-24
[RFC] Add link to StoPrint from the homescreen

hawkrives:
I really think this would be useful.

Thoughts?

// cc @elijahverdoorn 

Part of #78


hawkrives:
@drewvolz:
> I would take this approach: check when opening the parts that are relevant to the parts we care about being relevant. If not on the sto network, then render a NoticeView and explain they're not on the right wifi network

We'll push this out later, via codepush, once we have an error view ready.

hawkrives:
Now that I'm on campus, I can actually check to make sure this works… w00t.

drewvolz:
I am willing to approve of this if we can render a `NoticeView` for users that are not on campus.

hawkrives:
Someone stop me before I just implement a native view for it. There's no mobile site; it's just the desktop version.

hawkrives:
OTOH, it's really simple – a plain html table, with a link to "release" and "cancel" for each row, which are implemented as just simple HTTP requests. No JS magic here.

drewvolz:
Initial work on a native implementation is happening on the [stoprint-take-2](https://github.com/StoDevX/AAO-React-Native/tree/stoprint-take-2) branch.

PR #764, opened 2017-02-24
Make redux connections more precise

### LINKED ISSUE #715: Only persist login credentials when submitting the login form
Instead, it should only send the credentials to redux when the form is submitted. 

When the form is being edited, the credentials should be local state variables. 

This will als fix the bug of you changing the username, but not submitting, leaving the app, and coming back and having the app fail because it's now using the new wrong username.

hawkrives:
> closes #715

There was no reason for me to do all the connections in SettingsView, when SettingsView itself didn't need any of the data.

So, this PR moves the connect() calls down to the views that actually use the data.


hawkrives:
And look at how simple `SettingsView` is now!

```js
    <ScrollView
      contentContainerStyle={styles.container}
      keyboardShouldPersistTaps='always'
      keyboardDismissMode='on-drag'
    >
      <TableView>
        <CredentialsLoginSection />

        <TokenLoginSection
          navigator={props.navigator}
          route={props.route}
        />

        <SupportSection />

        <OddsAndEndsSection
          navigator={props.navigator}
          route={props.route}
        />
      </TableView>
    </ScrollView>
```

drewvolz (approved):

PR #766, opened 2017-02-24
Get all Financials data from the olecard page

hawkrives:
`/apps/olecard` now offers the data that was previously only available through the SIS' Balances page, so this changes the SIS section to read all balances from the OleCard page.

The general approach here was to realize that the financials data is presented in a table like this:

Account | Available Balance
--- | ---
STO Flex Dollars | $0.00
STO Ole Dollars | $0.00
STO Student Printing | $35.00
something about meals | unknown

Each row has an id of `accountrow#`, where # is the row number. The first column of each row has the class `.accountrow`, and the second column has the id `value` (yes, every cell in the right column has the same id. grr.)

So, we grab each cell with `.accountrow`, then get its parent (the entire row). We then extract the text content of that row. 

For each one of those items, which now look something like `"STO Flex Dollars $0.00"`, we know that the "name" of the row is everything before the last space, and the "value" is everything after the last space.

However, the exact names have a tendency to change without warning, so what I've done instead is to create a little /keyword/ regex hash.

```ts
const lookupHash: Map<RegExp, string> = new Map([
  [/ flex /i, 'flex'],
  [/ ole /i, 'ole'],
  [/ print/i, 'print'],
  [/daily/i, 'daily'],
  [/weekly/i, 'weekly'],
])
```

Each of the regexes are a little case-insensitive thing that we use to figure out what the current row is named. I'm pretty sure that each of those are both short enough to not be duplicated across rows, and special enough to be present for everything up to a complete rebranding.

That's about it.

I've tested that this seems to work, but I can't check flex dollars or the meal plan.

@elijahverdoorn / @ekmanzu / @zimmer18, could one of you please check out this branch (`origin/financials-from-olecard`) and see if it works for you? Thanks!


hawkrives:
I'm going to leave this open for two more days, and then I'll merge and have my on-campus beta contacts test it.

PR #770, opened 2017-02-24
Fix Travis command exit statuses

### LINKED ISSUE #761: Propogate exit statuses so travis builds can fail again
Currently, travis JS builds will have a really hard time being marked as "failed". 

All the commands whose output we collect with tee? So that Danger can report it? Tee changes the exit code (or rather, that's how pipes work.)

Anyway, a quick google search says that calling `set -o pipefail` (http://stackoverflow.com/a/6872163/2347774) will probably propogate the exit status so travis can see which commands fail.

hawkrives:
> closes #761

Adds `set -o pipefail` to the beginning of Travis so that failing commands in pipes will produce error codes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -139,6 +139,9 @@ script:
   # These sections will grow over time.
   # Each line is collapsed nicely in the travis output.
 
+  # Make sure that a failing command in a pipe fails the build
+  - set -o pipefail
+
   # ensure the env file exists and fill it out
   - touch .env.js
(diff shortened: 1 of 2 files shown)


hawkrives:
We have _both_ a failing build (https://travis-ci.org/StoDevX/AAO-React-Native/builds/205152088) and a Danger log!

image

drewvolz (approved):

PR #771, opened 2017-02-25
Fix button titles

hawkrives:
I forgot to pass them as `title=` during the last refactoring pass, so they were set to "Push Me!"…


drewvolz (approved):
👍 :shipit:

PR #772, opened 2017-02-25
Set the build number in package.json too

### LINKED ISSUE #742: Version is reported as "2.1.2" in beta
It should be 2.1.2.$build

Is the beta process not incrementing the number in package.json?

hawkrives:
closes #742


drewvolz (approved):
👍 :shipit:

PR #779, opened 2017-02-25
Trim carleton menu icons

drewvolz:
Carleton's bonapp adds `(H) (F2F) (VG) (V) (↓G)` and various other dietary markers to food item labels. We already provide the icons in the view which means there is duplication of information.

This PR proposes we check for any open and closing paren `(` and `)` and remove any including the contents that have 3 or less characters. `F2F` appears to be the largest of the bunch. 

The regex is `/\([^)]?.?.\)/g`. 

There might be a better way, but this seems efficient to say the least. @hawkrives ?

| Before | After
--- | ---
 |

### DIFF
--- a/source/views/menus/lib/trim-names.js
+++ b/source/views/menus/lib/trim-names.js
@@ -7,7 +7,7 @@ export function trimStationName(stationName: string) {
 
 function removeParenTags(str: string) {
   const parensRegex = / \([^)]*?\)$/
-  while (str !== str.replace(parensRegex, '')) {
+  while (str.match(parensRegex)) {
     str = str.replace(parensRegex, '')
   }
   return str
--- a/source/views/menus/lib/trim-names.js
+++ b/source/views/menus/lib/trim-names.js
@@ -7,5 +7,5 @@ export function trimStationName(stationName: string) {
 
 export function trimItemLabel(label: string) {
   // remove extraneous whitespace and title-case the bonapp titles
-  return toLaxTitleCase(label.replace(/\s+/g, ' '))
+  return toLaxTitleCase((label.replace(/\s+/g, ' ')).replace(/\([^)]?.?.\)/g, ''))
 }
--- a/source/views/menus/lib/trim-names.js
+++ b/source/views/menus/lib/trim-names.js
@@ -5,7 +5,17 @@ export function trimStationName(stationName: string) {
   return stationName.replace(/<strong>@(.*)<\/strong>/, '$1')
 }
 
+function removeParenTags(str: string) {
+  const parensRegex = / \([^)]*?\)$/
+  while (str !== str.replace(parensRegex, '')) {
+    str = str.replace(parensRegex, '')
+  }
+  return str
+}
+
 export function trimItemLabel(label: string) {
   // remove extraneous whitespace and title-case the bonapp titles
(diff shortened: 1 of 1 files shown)


drewvolz:
In an out of band conversation, @hawkrives suggested that we could run it in a loop until the output doesn't change to remove the ending parens. This approach is pretty cool and this PR will change to reflect that.

```js
while (str !== str.replace(regex, '')) {
  str = str.replace(regex, '')
}
```

source/views/menus/lib/trim-names.js
```diff
@@ -7,5 +7,5 @@ export function trimStationName(stationName: string) {
 
 export function trimItemLabel(label: string) {
   // remove extraneous whitespace and title-case the bonapp titles
-  return toLaxTitleCase(label.replace(/\s+/g, ' '))
+  return toLaxTitleCase((label.replace(/\s+/g, ' ')).replace(/\([^)]?.?.\)/g, ''))
```
hawkrives: Could we split this into two variables, instead?

```js
 export function trimItemLabel(label: string) {
    // remove extraneous whitespace and title-case the bonapp titles
    const evenedWhitespace = label.replace(/\s+/g, ' ')
    const noParens = evenedWhitespace.replace(/\([^)]?.?.\)/g, '')
    return toLaxTitleCase(noParens)
}
```

hawkrives:
So to take the new approach into account, how about this:

```js
function removeParenTags(str: string) {
  const parensRegex = / \([^)]*?\)$/
  while (str !== str.replace(parensRegex, '')) {
    str = str.replace(parensRegex, '')
  }
  return str
}

export function trimItemLabel(label: string) {
  // remove extraneous whitespace and title-case the bonapp titles
  const evenedWhitespace = label.replace(/\s+/g, ' ')
  const noParens = removeParenTags(evenedWhitespace)
  return toLaxTitleCase(noParens)
}
```

drewvolz:
Your approach works well. Thanks!

hawkrives (approved):

source/views/menus/lib/trim-names.js
```diff
@@ -5,7 +5,17 @@ export function trimStationName(stationName: string) {
   return stationName.replace(/<strong>@(.*)<\/strong>/, '$1')
 }
 
+function removeParenTags(str: string) {
+  const parensRegex = / \([^)]*?\)$/
+  while (str !== str.replace(parensRegex, '')) {
```
hawkrives: I just had a thought:

Instead of running the regex replacement and doing a string comparison, why not just use the `RegExp#test` function?

I tested this in [Runkit](https://runkit.com/hawkrives/58b1ca46e302c300142e1caa) and it works just like the comparison version.

```diff
-while (str !== str.replace(parensRegex, '')) {
+while (parensRegex.test(str)) {
```

So, we're checking if the regex would match the string, and replacing again if it would, instead of having to actually do the replacement twice.

drewvolz: I like that better.

hawkrives: Apparently it's better to use String#match instead of RegExp#test, when it might be run over the same string multiple times: https://github.com/facebook/jest/pull/3001

So let's do this instead:

```diff
-while (str !== str.replace(parensRegex, '')) {
+while (str.match(parensRegex)) {
```

hawkrives:
Great work!

Issue #333, opened 2016-11-16
TextView contents cannot be highlighted/selected on iOS


drewvolz:
This is a limitation of ReactNative at the moment. It has been implemented [for Android](https://github.com/facebook/react-native/pull/8028) but not iOS.

Keeping track of this issue:
* [Product Pains](https://productpains.com/post/react-native/support-selecting-text-within-text-elements-for-copypaste-etc)
* [GitHub](https://github.com/facebook/react-native/pull/8028)
* [GitHub again](https://github.com/facebook/react-native/issues/10897)

iOS is a bit more complicated. The guy who implemented it for Android has this to say

> I have implemented this feature for Android, waiting for it to be merged. iOS has proved to be *much* more difficult since their Text widget works by drawing raw characters to the screen, instead of using the native iOS UITextView, so I'm hoping that someone else more experienced with RN Shadow Views will be able to take that on.

## Status

- [x] Building hours detail (title, subtitle, badge, hours day/schedule)
- [x] Notice snapshot test
- [x] Notice (message)
- [x] Card (header/footer) 
- [x] Contacts card (title, description)
- [x] Dictionary (word, definition)
- [x] SIS balances (flex, ole, print balances)
- [x] KSTO (title, description, notice)
- [x] Transportation (title, description)
- [ ] Orgs
  - [ ] Android 
    - [x] contact, description, error message
    - [x] name
    - [ ] category
    - [ ] meeting time, location are not possible atm b/c of `<Cell>` (see https://github.com/Purii/react-native-tableview-simple/issues/30)
  - [ ] iOS
    - [x] name
    - [ ] description
    - [ ] category
    - [ ] meeting time, location are not possible atm b/c of `<Cell>` (see https://github.com/Purii/react-native-tableview-simple/issues/30)

drewvolz:
As of 0.39, we are able to mark `<Text>` as selectable. This means we are now able to do this!

http://facebook.github.io/react-native/releases/0.39/docs/text.html#selectable

hawkrives:
What views would we like to make selectable?

elijahverdoorn:
Maybe the event info?

It would be really cool if you could highlight an event, have all the info copied to your clipboard, then got to your texting app (for example) and past the copied data and have a nicely-formatted, clean description of the event to text to friends.

drewvolz:
At that rate, we probably would need to add a share button to each row. But I like where you were headed with that idea.

[...]

PR #782, opened 2017-02-27
Allow text to be selected in Detail views

drewvolz:
Lots of text that is now selectable on iOS (and currently is selectable on Android).

Part of #333.


hawkrives:
Jest failed (it looks like I need to update the logic for when to include its log). 

You'll need to update the snapshots; just run `jest -u` locally and commit the updated snapshot files.

hawkrives:
Rebase this onto master once #785 lands.

hawkrives (approved):

PR #784, opened 2017-02-27
Cleanups

hawkrives:
I removed an extra `[style]` prop from Course > Row, and returned a list of objects from `getDetailedBuildingStatus` (instead of a list of `[bool, string, string]` tuples).


hawkrives:
Rebase this onto master once #785 lands.

drewvolz (approved):

PR #796, opened 2017-03-01
Check for fastlane release

drewvolz:
Ensure that we only deploy when we mean to with a truthiness check.


hawkrives (approved):
Yeah, whoops. Thanks!

PR #808, opened 2017-03-02
Clear cache on memory warning

drewvolz:
One of the most frequent cause of crashes seen in apps is being ejected for not freeing up enough memory when a memory warning comes in. When the app receives a memory warning, we should purge the shared cache to free up memory.

### DIFF
--- a/ios/AllAboutOlaf/AppDelegate.m
+++ b/ios/AllAboutOlaf/AppDelegate.m
@@ -65,7 +65,7 @@
 }
 
 - (void)applicationDidReceiveMemoryWarning:(UIApplication *)application {
-    [[NSURLCache sharedURLCache] removeAllCachedResponses];
+  [[NSURLCache sharedURLCache] removeAllCachedResponses];
 }
 
 @end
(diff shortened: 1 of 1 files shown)


hawkrives (commented):
Shouldn't we only clear the RAM cache, not the disk cache? Is that possible?

drewvolz:
It is possible to clear the RAM cache. Instead of the call we're making, we would set the RAM cache to `0` much in the same way we allocated it.

```objective-c
- (void)applicationDidReceiveMemoryWarning:(UIApplication *)application {
  NSURLCache *URLCache = [NSURLCache sharedURLCache];
  const NSUInteger memoryCapacity = URLCache.memoryCapacity;
  URLCache.memoryCapacity = 0;
  URLCache.memoryCapacity = memoryCapacity;
}
```

As far as responding to low-memory warnings, we should base our decisions off of the [Responding to Low-Memory Warnings in iOS docs](https://developer.apple.com/library/content/documentation/Performance/Conceptual/ManagingMemory/Articles/MemoryAlloc.html#//apple_ref/doc/uid/20001881-SW1)

hawkrives:
So… should we only reset the memory cache, or should we just clear the entire cache?

drewvolz:
I don't know the right answer either.

hawkrives:
I'm willing to wait for research into that question, as (according to analytics) we have had 0 crashes due to low-memory conditions.

drewvolz:
Closing this. We do not have any evidence that we have crashed from a memory warning yet. We can revisit this in the future if we need to.

PR #811, opened 2017-03-03
Rework Fastlane files

### LINKED ISSUE #806: Changelog creation fails when prior build wasn't on the same branch
We should edit it to find the previous build on the same branch, if any, with a fallback to the most recent "master" build.

hawkrives:
> … and maybe this is a bit much, but …

This PR reworks the structure behind our Fastlane setup. 

Primarily, it splits up the Fastfile into individual files: the actions stay in `actions/`, the utility lanes turned into functions and moved into `lib/util.rb`, the platform-agnostic lanes moved to `platforms/agnostic.rb`, and the platform-specific lanes moved to `platforms/{android,ios}.rb`.

This also uses Fastlane's `Action.lane_context` shared global hash and global environment variables to reduce the need for passing around parameters to every. single. function. (like `:platform`. I don't have to pass `:platform` four levels down anymore.)

For anyone who reviews, if you could approach this as a new implementation of Fastlane, instead of as a diff, I think you'll be less confused.

---

A sample:

```ruby
platform :ios do
  desc 'Submit a new Beta Build to HockeyApp'
  lane :beta do
    activate_rogue_team

    set_version(
      version: lane_context[:VERSION_NUMBER],
      build: lane_context[:BUILD_NUMBER],
    )

    match(readonly: true)

    build

    hockey(notes: release_notes)
  end
end
```

---

Closes #806.


drewvolz:
I think this failed because:

> Code signing is required for product type 'Application' in SDK 'iOS 10.1'

hawkrives:
Huh. I wonder what I changed for that to happen. 

I will dig into that later.

drewvolz:
Closing/reopening due to [a bug Travis was having](https://www.traviscistatus.com/incidents/ccp8bj1x2n6y)

hawkrives:
I'm going to take another tack on this. I think I know what change is causing the issues, but I'm going to try and split this into several PRs, ideally small enough that Danger won't yell at me about them :smile:

PR #812, opened 2017-03-03
Adjust style of news share button

### LINKED ISSUE #804: iOS News share button needs visual tweaks
Notes | AAO
----- | ---
 |

drewvolz:
This changes the news touchable to be larger and eliminates the background color in favor of changing opacity on press.

Closes #804

Before | After
--- | ---
before | after

### DIFF
--- a/source/views/components/navigation/button-right.js
+++ b/source/views/components/navigation/button-right.js
@@ -43,7 +43,7 @@ const styles = StyleSheet.create({
     color: 'white',
     ...Platform.select({
       ios: {
-        fontSize: 26,
+        fontSize: 32,
       },
       android: {
         fontSize: 24,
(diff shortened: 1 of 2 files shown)


hawkrives:
Please move the share style into the RightButton stylesheet in the navigation components stuff, and remove any unneeded things from there, instead of introducing new items at the use locations of the RightButton. 

Because by doing them in one place, we can recieve the styles in every location and be consistent.

drewvolz:
Much simpler. Thanks for the suggestion.

hawkrives:
Grumble the icon still looks too far to the bottom. 

If you overlay the toolbar with the Notes toolbar, are they the same vertical position?

drewvolz:
Here is a comparison of Apple's photos vs AAO's news. It looks pretty spot-on to me.

photos v news

hawkrives (approved):

PR #816, opened 2017-03-06
Feature new student orgs

drewvolz:
This creates a grouping for new student organizations to be featured at the top of the orgs list.

orgs


source/views/student-orgs/list.js
```diff
@@ -77,9 +77,20 @@ export class StudentOrgsView extends React.Component {
           $groupableName: head(startCase(sortableName)),
         }
       })
+      // All orgs
       let sorted = sortBy(withSortableNames, '$sortableName')
       let grouped = groupBy(sorted, '$groupableName')
-      this.setState({orgs: grouped})
+      // New orgs
+      let newOrgs = sorted.filter(org => org.newOrg)
+      let newOrgsSortableNames = map(newOrgs, item => {
```
hawkrives: Join this into one statement, if you would?

sorted
  .filter()
  .map()

hawkrives (changes requested):
One requested change, and one comment: would you look into AlphabetListView and see if you can specify the letters on the right side separately?  IIRC, you can specify an "extractor" function which will produce the items used for the side; you could do `$groupableName === "New" ? "•" : $groupableName[0]` or something to change "New" into a single char.

drewvolz:
I will look into how to specify groupnames explicitly.

hawkrives:
We could probably call it "• New" and leave it how it's currently taking the first letter of each item, if we wanted.

**Edit:** Nevermind. That doesn't work, because the section headings and the sidebar use the same labels in the current implementation.

hawkrives:
Given https://github.com/sunnylqm/react-native-alphabetlistview#getsectiontitle, I think that the `getSectionTitle` prop is the the ListSectionHeader, and `getSectionListTitle` is for the things on the right.

drewvolz:
Done! Thanks for your help.

hawkrives (changes requested):
You have cellHeight in here twice, accidentally.

source/views/student-orgs/list.js
```diff
@@ -184,6 +199,7 @@ export class StudentOrgsView extends React.Component {
         data={this.state.orgs}
         cell={this.renderRow}
         // just setting cellHeight sends the wrong values on iOS.
+        getSectionListTitle={this.getSectionListTitle}
```
hawkrives: Please move this line above the comment

source/views/student-orgs/list.js
```diff
@@ -119,6 +129,11 @@ export class StudentOrgsView extends React.Component {
     )
   };
 
+  getSectionListTitle = (...args: any) => {
```
hawkrives: `= (name: string) =>` and get rid of the array access

source/views/student-orgs/list.js
```diff
@@ -83,9 +83,19 @@ export class StudentOrgsView extends React.Component {
...
-      this.setState({orgs: grouped})
+      // New orgs
+      let newOrgs = sorted.filter(org => org.newOrg).map(item => {
+        return {
+          ...item,
+          $groupableName: 'New',
+        }
+      })
+      let newOrgsGrouped = groupBy(newOrgs, '$groupableName')
+      let combined = {...newOrgsGrouped, ...grouped}
```
hawkrives: Please inline this into the setState call (setState({orgs: {...newOrgs, ...allOrgs}})

source/views/student-orgs/list.js
```diff
@@ -83,9 +83,19 @@ export class StudentOrgsView extends React.Component {
           $groupableName: head(startCase(sortableName)),
         }
       })
+      // All orgs
       let sorted = sortBy(withSortableNames, '$sortableName')
       let grouped = groupBy(sorted, '$groupableName')
```
hawkrives: Please rename this to allOrgsGrouped

source/views/student-orgs/list.js
```diff
@@ -83,9 +83,19 @@ export class StudentOrgsView extends React.Component {
...
-      this.setState({orgs: grouped})
+      // New orgs
+      let newOrgs = sorted.filter(org => org.newOrg).map(item => {
+        return {
+          ...item,
+          $groupableName: 'New',
+        }
```
hawkrives: Could you just put this whole thing on one line please? return through the end bracket.

hawkrives (changes requested):

hawkrives:
Can you please merge master into this branch and run prettier on it?

Edit: I ran prettier.

hawkrives:
So the current implementation goes like this:

```js
// All orgs
let sorted = sortBy(withSortableNames, '$sortableName')
let allOrgsGrouped = groupBy(sorted, '$groupableName')
// New orgs
let newOrgs = sorted.filter(org => org.newOrg).map(item => {
  return {...item, $groupableName: 'New'}
})
let newOrgsGrouped = groupBy(newOrgs, '$groupableName')
let combined = {...newOrgsGrouped, ...allOrgsGrouped}
this.setState({orgs: combined})
```

What about changing it to this?

```js
let sorted = sortBy(withSortableNames, '$sortableName')
let allOrgsGrouped = groupBy(sorted, '$groupableName')

let newOrgs = sorted.filter(org => org.newOrg)
let orgs = {New: newOrgs, ...allOrgsGrouped}

this.setState({orgs: orgs})
```

hawkrives (approved):

drewvolz:
That particular redesign resulted in every org being marked as new. Also, the font size for that new icon has decreased.

hawkrives:
I think you commented on the wrong PR?

drewvolz:
I meant to comment here as I thought this PR had something to do with it... although it was introduced in [#830 Add new warnings to linter](https://github.com/StoDevX/AAO-React-Native/pull/830)

PR #824, opened 2017-03-07
Check if menu items are filtered with no items

### LINKED ISSUE #709: Menu shows "try disabling filters" when no filters are enabled
From @elijahverdoorn in #660:

> […] would like to point out that it currently says "No items to show. Try changing the filters" even though I don't have any filters on.

drewvolz:
Fixes the issue where menus showed a "disable filters" message when no filters were enabled... this was due to the case when there were no items to filter.

Closes #709

### DIFF
--- a/source/views/menus/components/fancy-menu.js
+++ b/source/views/menus/components/fancy-menu.js
@@ -136,6 +136,7 @@ class FancyMenuView extends React.Component {
     // group the tuples into an object (because ListView wants {key: value} not [key, value])
     const grouped = fromPairs(filteredByMenu)
 
+    let anyFiltersEnabled = filters.some(f => f.enabled)
     const specialsFilterEnabled = Boolean(
       filters.find(
         f =>
@@ -146,7 +147,6 @@ class FancyMenuView extends React.Component {
     )
 
     let messageView = null
-    let anyFiltersEnabled = filters.some(f => f.enabled)
     if (specialsFilterEnabled && stationMenus.length === 0) {
       messageView = (
         <NoticeView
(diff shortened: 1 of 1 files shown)


source/views/menus/components/fancy-menu.js
```diff
@@ -146,6 +146,7 @@ class FancyMenuView extends React.Component {
     )
 
     let messageView = null
+    let filtersEnabled = filters.some(f => f.enabled)
```
hawkrives: Can we rename this variable to "anyFiltersEnabled" instead? That seems like it would go better with the accompanying "specialsFilterEnabled" variable.

source/views/menus/components/fancy-menu.js
```diff
@@ -146,6 +146,7 @@ class FancyMenuView extends React.Component {
     )
 
     let messageView = null
+    let anyFiltersEnabled = filters.some(f => f.enabled)
```
hawkrives: Could you move this up to right above the specialsFilterEnabled line, please?

hawkrives (changes requested):
One last thing.

source/views/menus/components/fancy-menu.js
drewvolz: Done.

source/views/menus/components/fancy-menu.js
```diff
@@ -136,6 +136,7 @@ class FancyMenuView extends React.Component {
     // group the tuples into an object (because ListView wants {key: value} not [key, value])
     const grouped = fromPairs(filteredByMenu)
 
+    let anyFiltersEnabled = filters.some(f => f.enabled)
```
hawkrives: Make this a `const`, please

hawkrives (changes requested):

hawkrives (approved):

PR #833, opened 2017-03-08
Fix new org badge styles

### LINKED ISSUE #832: New org badge applied to all rows
In a last second effort to **optimize** labeling a student org as new or not, we slipped-in a way to label orgs in a more efficient way. Testing the app now reveals that every org was labeled as new... the correct behavior would be to only label new orgs as new. Additionally, the style for the new dots are not being applied.

drewvolz:
This fixes an issue where the styles for new badges were accidentally applied to all rows. The issue came from applying a styling meant for a `<Text/>` tag to a `<View/>` tag.

Closes #832


drewvolz:
Closed in favor of #834

PR #837, opened 2017-03-09
Deduplicate index entry points

hawkrives:
This also prevents us from registering AllAboutOlaf twice.


drewvolz:
Excellent catch, thanks for fixing and optimizing. ~With the unification, we might want to also tackle changing the other files that reference the `index.x.js` files. I believe those are: `AppDelegate`, `react.gradle`, and `package.json`.~

**Edit**: nevermind. you didn't get rid of them. thanks!

drewvolz (approved):

PR #842, opened 2017-03-09
[RFC] Check script

drewvolz:
command | description
---|---
` check` | prettier, flow, lint, and test


hawkrives:
Yes to check, no to the other two.

drewvolz:
What's your reasoning?

hawkrives:
Those are magical incantations. 

My real reasoning is something along the lines of, I have not needed to run them within the span of my memory. 

I don't see a reason to add them.

drewvolz:
Alright... I'll take out the cleaning scripts.

hawkrives:
If you often use the cleaning script, we can add them. I still think you don't need to run them as often as you think you do.

I will support adding _one_ of the cleaning scripts.

package.json
```diff
@@ -7,6 +7,7 @@
     "bundle-data": "node scripts/bundle-data.js data/ docs/",
     "bundle:android": "react-native bundle --entry-file index.android.js --dev true --platform android --bundle-output ./android/app/src/main/assets/index.android.bundle --assets-dest ./android/app/src/main/res/",
     "bundle:ios": "react-native bundle --entry-file index.ios.js --dev true --platform ios --bundle-output ./ios/AllAboutOlaf/main.jsbundle --assets-dest ./ios",
+    "check:": "npm run prettier && npm run flow && npm run lint && npm run test",
```
hawkrives: We don't need to run both "npm run prettier" and "npm run lint" – eslint is run as part of the prettier command.

drewvolz: I'll change that

hawkrives (approved):

PR #846, opened 2017-03-10
Fix automatic data bundling on Travis

hawkrives:
- [x] Reduce verbosity from `set -x`
- [x] Fix data bundling
- [x] Simplify some string replacements
- [x] Uppercase some variables
- [x] Bundle the data

---

We realized a few days ago, after #839 was merged, that the data bundles weren't being updated.

My original idea was to check out the branch that was being built, commit the bundled data changes on top of the branch, then go back to the commit. Once the build was done, it would push the committed and bundled data back to the branch, with `[skip ci]` in the commit message so as to avoid an endless rebuilding situation.

I had tested that this worked in my original branch, but what I didn't realize was that Travis clones differently if you're building a PR versus a push to a branch.

Travis configures git to only fetch the branch that's being built. If it's building a push, it'll clone the pushed branch; if it's building a PR, it clones `master`. Travis sets the `git config remote.origin.fetch` value to restrict the branches.

So, when I built the test branch, Travis worked perfectly, because it checked out the branch that it needed to. When I built the PRs, though, Travis only knew about `master`, so `git checkout $OTHER_BRANCH` couldn't work.

I solved this by adding OTHER_BRANCH (whatever it may be; it's stored in the `$BRANCH` environment variable) to the `remote.origin.fetch` config value, then running `git fetch` so that Travis knows about both `master` and the other branch.

Then, we do the bundle/checkout-branch/commit/checkout-commit dance, and we set the PUSH_BRANCH env variable so that we know that we need to push to the remote. We don't want to always push, because the PR builds include a merge commit, which would result in a never-ending series of builds (Merge, Merge, Merge, Merge, …).

So! We can see that this works, in the commit history of this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ env:
     # variables that are used later on
     - PUSH_BRANCH=0
     - CAN_DEPLOY=0
+    - USE_EMULATOR=no
     - run_deploy=0
 
 matrix:
@@ -150,7 +151,7 @@ install:
 before_script:
   # Fire up the Android emulator
   - |
-    if [[ $ANDROID && $USE_EMULATOR ]]; then
+    if [[ $ANDROID && $USE_EMULATOR = yes ]]; then
       set -x
       EmuName="react-native"
       mkdir -p "$HOME/.android/avd/$EmuName.avd/"
--- a/.travis.yml
+++ b/.travis.yml
@@ -97,7 +97,8 @@ before_install:
   # ensure that the PR branch exists locally
   - git config --add remote.origin.fetch "+refs/heads/$BRANCH:refs/remotes/origin/$BRANCH"
   - git fetch --depth 10
-  - git branch "$BRANCH"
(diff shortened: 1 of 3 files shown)


drewvolz:
Closing/reopening due to [a bug Travis was having](https://www.traviscistatus.com/incidents/ccp8bj1x2n6y)

drewvolz:
Could you please provide a summary of what these changes do?

hawkrives:
No :p 
See the OP, at the top of the page.

EDIT: I apologize. I thought you were "poking the bear", as it were. I have updated the OP with a detailed accounting of what the problem was, and how it was fixed.

drewvolz:
"What was the issue behind the original bundling not working, and how does this fix that problem?" should have been my way of phrasing my original [question](https://github.com/StoDevX/AAO-React-Native/pull/846#issuecomment-285876356) above.

Thank you for updating this PR with helpful information! I know I appreciate it, and the community can certainly learn from it. Much appreciated, @hawkrives.

PR #854, opened 2017-03-12
Split Travis config into individual scripts

hawkrives:
The current [`.travis.yml`](https://github.com/StoDevX/AAO-React-Native/blob/c44e6a5203b27885087fb5e22475980d4d894436/.travis.yml) config is kinda unwieldy.

Okay, it's really unwieldy. 

245 lines? Too much.

So, this PR splits the travis config into five files, one for each chunk of the travis lifecycle: `before_install`, `install`, `before_script`, `script`, and `after_success`.

Almost all of the logic has been pulled out into the individual scripts, except for a few pieces that need to export values to the children (because subshells apparently can't modify exported variables.)

Another benefit: we get syntax highlighting for the config files :-)


drewvolz (approved):

drewvolz:
@hawkrives: "the PR build passed, it's just the push build that's failing"

PR #867, opened 2017-03-14
Sort presence calendar events

### LINKED ISSUE #866: Fix presence calendar sort order
March 14th
March 31st
March 16th

drewvolz:
This fixes the presence calendar sorting order by sorting before filtering so that the events are in order by date.

Closes #866


drewvolz:
This does not solve it... hold off on merging.

PR #870, opened 2017-03-14
Fix relative links in news articles

### LINKED ISSUE #864: Handle broken links
An Oleville post located here http://oleville.com/blog/2017-spring-elections-voting-day/ contains a URL that crashes AAO. We should handle this situation.

`<a href="www.oleville.com/elections/vote">here</a>` lacks the protocol `http://` or `https://`

The error is:

```
ExceptionsManager.js:71 Exception 'The specified URL has an unsupported scheme.
Only HTTP and HTTPS URLs are supported.' was thrown while invoking show on 
target SafariViewManager with params (
        {
        url = "www.oleville.com/elections/vote";
    },
    97
)
```

hawkrives:
Turns out that `<WebView/>` accepts a `baseUrl` parameter to set the base URL for the webview, allowing us to let the web platform handle relative hrefs.

Um.

This fixes bugs. More detail probably available upon request.

Closes #864.

### DIFF
--- a/source/views/components/html-view.js
+++ b/source/views/components/html-view.js
@@ -18,16 +18,14 @@ export class HtmlView extends React.Component {
       return
     }
 
-    // I had to disable the stopLoading call, because it was cancelling the
-    // image loads.
-    // this._webview.stopLoading()
-    this._webview.goBack()
-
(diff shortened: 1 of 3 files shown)


drewvolz (approved):
:+1: :shipit:

PR #885, opened 2017-03-21
Fix webcam schema

drewvolz:
@hawkrives' fix to the webcams schema

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -101,10 +101,7 @@ before_install:
 
   # run our before_install sequence
   - bash scripts/travis/before_install.sh
-  
-  # test
-  - cat ./data/_schemas/webcams.yaml
-  
+
 install:
   - bash scripts/travis/install.sh
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -101,7 +101,10 @@ before_install:
 
   # run our before_install sequence
   - bash scripts/travis/before_install.sh
-
+  
+  # test
+  - cat ./data/_schemas/webcams.yaml
+  
 install:
   - bash scripts/travis/install.sh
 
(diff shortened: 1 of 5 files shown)


hawkrives:
Sigh.

hawkrives:
I think the next step is to add debugging to `scripts/validate-data` to figure out WTF is going on inside it.

The iteration logic is all before the `// MARK: Helpers` comment.

Let's see what my first attempt at debugging shows us: 916bbf4

drewvolz:
`cat ./data/_schemas/webcams.yaml` produced:

```yaml
$schema: http://json-schema.org/draft-04/schema#

additionalProperties: false
required: [name, pageUrl, streamUrl, thumbnail, tagline, accentColor]
properties:
  name: {type: string}
  pageUrl: {type: string, format: uri}
  streamUrl: {type: string, format: uri}
  thumbnail: {type: string}
  tagline: {type: string}
  accentColor:
    type: array
    items: [{type: number}, {type: number}, {type: number}]
    additionalItems: false

```

and the error it spat out was:

`webcams/0-alumni-hall-west.yaml is invalid`
```json
{
  "keyword": "required",
  "dataPath": "",
  "schemaPath": "#/required",
  "params": {
    "missingProperty": ".url"
  },
  "message": "should have required property '.url'"
}
```

hawkrives:
So it didn't actually run my edited code?

hawkrives:
It should have printed out `file, {schema}` lines for each file, before deciding if it was invalid or not.

hawkrives:
master checks: [name, url]
the branch checks: [name, other stuff]

Travis ran Prettier, which does the git stuff

I think that the git stuff is not putting us back on the commit that we started on

So if I `cat` the schema again, right before the validator runs, it'll be the copy from master

let's go test that theory

hawkrives:
Yeah, the git stuff is borkeden. I fix ish ing.

hawkrives (approved):
woot

PR #888, opened 2017-03-21
chore(package): update danger to version 0.14.1

hawkrives:
I also fixed some danger issues. Let's watch!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,42 +37,42 @@ matrix:
       node_js: '7'
       env: [JS=yes]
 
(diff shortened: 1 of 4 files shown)


hawkrives:
Current status: the flow and ios/android js bundling logs are expected to be wonky; I commented them out, so they're not running at all, and they are logged if they _don't_ contain content, while the others are only (theoretically) logged if they _do_.

hawkrives:
@drewvolz want to review? everything LGTM.

PR #897, opened 2017-03-22
Link react-native-maps

drewvolz:
Adds the officially supported native maps library for future use.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -105,7 +105,7 @@ android {
             manifestApplicationId: "${applicationId}",
             onesignal_app_id: "aa46a500-ab1c-4127-b9ff-e7373da3ce35",
             onesignal_google_project_number: "185558680648",
-            gmaps_key: System.getenv("GOOGLE_MAPS_API_KEY") ?: "",
+            gmaps_key: System.getenv("GMAPS_KEY") ?: "",
         ]
     }
     splits {
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -101,10 +101,12 @@ android {
         ndk {
             abiFilters "armeabi-v7a", "x86"
         }
-        manifestPlaceholders = [manifestApplicationId: "${applicationId}",
-                        onesignal_app_id: "aa46a500-ab1c-4127-b9ff-e7373da3ce35",
-                        onesignal_google_project_number: "185558680648",
-                        gmaps_key: System.getenv("GMAPS_KEY") ?: ""]
+        manifestPlaceholders = [
+            manifestApplicationId: "${applicationId}",
+            onesignal_app_id: "aa46a500-ab1c-4127-b9ff-e7373da3ce35",
+            onesignal_google_project_number: "185558680648",
+            gmaps_key: System.getenv("GMAPS_KEY") ?: "",
+        ]
     }
     splits {
         abi {
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -90,7 +90,7 @@ def enableProguardInReleaseBuilds = false
 
 android {
     compileSdkVersion 23
-    buildToolsVersion "23.0.2"
+    buildToolsVersion "25.0.2"
 
     defaultConfig {
         applicationId "com.allaboutolaf"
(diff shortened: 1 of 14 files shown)


drewvolz:
hawkrives `react-native-maps` depends on `android-25` and we recently downgraded to `android-23` when upgrading `react-native` ... thoughts, @hawkrives?

hawkrives:
I removed the android-25 sdk when we upgraded CodePush because we didn't need it anymore. 

Feel free to add it back, here, if it's needed, but I'd try rebasing this branch onto master first, to make sure it's using the upgraded version of everything.

hawkrives:
Nope. Look in the diff from the codepush update PR.

hawkrives:
Undo the change to the buck config and see if that fixes it.

drewvolz:
thanks @hawkrives for figuring out we needed build-tools@25.0.0 and android 25 sdk

drewvolz:
I have no proof that my environment variable passing works. I will verify this tonight.

drewvolz:
My trick to pass the maps api key into the manifest does not work yet. The library at least builds and runs and renders on both platforms. We're close.

drewvolz:
It works! I had to call the right environment variable.

android/build.gradle
```diff
@@ -15,7 +15,7 @@ buildscript {
 ext {
     travisBuild = System.getenv("TRAVIS") == "true"
     // allows for -Dpre-dex=false to be set
-    preDexEnabled = "true".equals(System.getProperty("pre-dex", "true"))
+    preDexEnabled = System.getProperty("pre-dex", "true") == "true"
```
hawkrives, line 18: (To reviewers: This was an irrelevant change that I made. I made it for consistency's sake with the other variable in this block.)

android/gradle.properties
```diff
@@ -11,6 +11,7 @@
 # The setting is particularly useful for tweaking memory settings.
 # Default value: -Xmx10248m -XX:MaxPermSize=256m
 # org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
+org.gradle.jvmargs=-Xmx1536m
```
hawkrives, line 14: (This removes the "… to build gradle in-process you must increase the heap size" notice from the build process.)

android/app/build.gradle
```diff
@@ -101,9 +101,12 @@ android {
...
             abiFilters "armeabi-v7a", "x86"
         }
-        manifestPlaceholders = [manifestApplicationId: "${applicationId}",
-                        onesignal_app_id: "aa46a500-ab1c-4127-b9ff-e7373da3ce35",
-                        onesignal_google_project_number: "185558680648"]
+        manifestPlaceholders = [
+            manifestApplicationId: "${applicationId}",
+            onesignal_app_id: "aa46a500-ab1c-4127-b9ff-e7373da3ce35",
+            onesignal_google_project_number: "185558680648",
+            gmaps_key: System.getenv("GOOGLE_MAPS_API_KEY") ?: "",
```
hawkrives: @drewvolz So here, you use the env variable `GOOGLE_MAPS_API_KEY`, but in the `scripts/travis/script.sh` file you echo the `GMAPS_KEY` env variable.

According to Travis, the correct name is `GMAPS_KEY`.

I notice that in https://github.com/StoDevX/AAO-React-Native/pull/897/commits/f8b3637c204da811ddc01a57f12de52bf46a5f6d, you changed from `GMAPS…` to `GOOGLE_MAPS…` — are you sure that that's correct on Travis?

hawkrives (changes requested):

android/app/build.gradle
drewvolz: I made the change so that it would work locally (non-travis). It most likely needs to be the same name for it to be correct on both travis and local development.

hawkrives: So you had a local env variable named `GOOGLE_MAPS_API_KEY`? Could you just switch that to GMAPS_KEY? Because that'd be simpler than changing the one on travis, since travis doesn't let you rename…

hawkrives:
Sorry, accidental branch deletion.

android/app/build.gradle
drewvolz: After playing with this for some amount of time, I have convinced myself this is not an issue. Do you think it is an issue? I think the environment variables are setup correctly.

hawkrives (approved):

android/app/build.gradle
drewvolz: Out of band, this is cleared up. We'll change this and consider it done.

PR #930, opened 2017-04-10
Update FAQs

hawkrives:
Add item for the disabled News section.

I also removed the playful "why does the app update so slowly" message.


drewvolz (approved):

hawkrives:
:+1:

PR #945, opened 2017-04-13
Use prettier to remove semicolons, instead of eslint

hawkrives:
Before: `✨  Done in 10.86s.`
After: `✨  Done in 4.16s.`

---

Prettier@1.0 supports a `--no-semi` option, to prevent printing semicolons. This allows us to quit processing every file twice, which is what we currently do: `prettier | eslint`, where we let eslint remove the semicolons for us.

### DIFF
--- a/source/app.js
+++ b/source/app.js
@@ -178,7 +178,7 @@ export default class App extends React.Component {
     console.log('Device info:', device)
   }
 
-  _navigator: Navigator;
+  _navigator: Navigator
 
   registerAndroidBackButton = () => {
     if (this._navigator && this._navigator.getCurrentRoutes().length > 1) {
@@ -186,7 +186,7 @@ export default class App extends React.Component {
       return true
     }
     return false
-  };
+  }
 
   render() {
     return (
--- a/source/globalize-fetch.js
+++ b/source/globalize-fetch.js
@@ -6,9 +6,8 @@ function status(response) {
   if (response.status >= 200 && response.status < 300) {
     return response
   } else {
-    let error = new Error(response.statusText)
-    ;// attach the original response to the thrown error
-    (error: any).response = response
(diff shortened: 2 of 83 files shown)


source/globalize-fetch.js
```diff
@@ -6,8 +6,8 @@ function status(response) {
   if (response.status >= 200 && response.status < 300) {
     return response
   } else {
-    let error = new Error(response.statusText);
-    // attach the original response to the thrown error
+    let error = new Error(response.statusText)
+    ;// attach the original response to the thrown error
```
hawkrives: I filed an issue for this one. https://github.com/prettier/prettier/issues/1245

source/app.js
```diff
@@ -178,15 +178,15 @@ export default class App extends React.Component {
...
+  _navigator: Navigator
 
   registerAndroidBackButton = () => {
     if (this._navigator && this._navigator.getCurrentRoutes().length > 1) {
       this._navigator.pop()
       return true
     }
     return false
-  };
+  }
```
hawkrives, line 189: ESLint didn't know how to remove semicolons from these statements.

source/lib/courses/index.js
```diff
@@ -1,7 +1,3 @@
 // @flow
 export {loadAllCourses} from './load-all-courses'
-export type {
-  CourseType,
-  CourseCollectionType,
-  CoursesByTermType,
-} from './types'
+export type {CourseType, CourseCollectionType, CoursesByTermType} from './types'
```
hawkrives, line 3: The only explanation I have for these is that … the algorithm was counting the semicolon in the line length? which makes sense.

source/globalize-fetch.js
drewvolz: I think you filed an issue against this.

PR #950, opened 2017-04-14
Remove /docs folder

hawkrives:
Since #949 worked properly, it's not needed anymore.

From now on, we can just make PRs … and as soon as they're merged, Travis will update the docs for us.

cc: @elijahverdoorn @drewvolz


hawkrives:
I did not know that we had that much stuff in `/docs`. -4,537 lines??? dang.

hawkrives:
Hi travis. Let's get you fixed up…

hawkrives:
I added a call to `npm run bundle-data` as part of the `fastlane build` script, because our build process still assumes that the files are there (because we bundle a copy of them with the app.)

PR #958, opened 2017-04-15
Revert "update the cage and stav hours for easter break"

hawkrives:
Reverts StoDevX/AAO-React-Native#957

Merge this on Monday afternoon.


hawkrives:
It is now Monday afternoon.

PR #959, opened 2017-04-16
Searchable dictionary for iOS

drewvolz:
This introduces a searchbar for iOS in the AlphabetListView for the Dictionary.

**Todo**
- [x] I would like feedback on my flow types... I know they're not all correct
- [x] Need to find a way to re-highlight the *new* orgs in student orgs.
- [x] Need to reduce duplicates by going back to the way we took the first capitalized letter of all words
- [x] Need a prop that hides the "new" grouping when searching (in orgs)
- [x] Searching for `pda` in the dictionary on iOS reveals a strange layout bug
- [x] Need to hide the "x" icon when the search field is empty
- [x] Search will not filter by word at the moment. Finds any substring.
- [x] Searching for 'rolvaag' in the dictionary does not return 'rølvaag' ?
- [x] Need to create a wrapper that uses rn-search-bar and rn-searchbar so that we can stylize how they look per platform
- [ ] Alphbetlistview internal error when searching (on simulator, at least...) 

**Dictionary Screenshot**

screen shot 2017-04-23 at 1 42 45 pm

**Issues**

X-platform state issue
---
screen shot 2017-04-16 at 4 30 01 pm


source/views/dictionary/list.js
```diff
@@ -77,21 +99,43 @@ export class DictionaryView extends React.Component {
     return <ListSeparator key={`${sectionId}-${rowId}`} />
   }
 
+  handleResults = (results: Object) => {
+    this.setState({results: results})
+  }
+
   render() {
+    if (!terms) {
+      return <LoadingView />
```
hawkrives: Why show the loading spinner? If there are no definitions, they'll never show up.

source/views/dictionary/list.js
```diff
@@ -77,21 +99,43 @@ export class DictionaryView extends React.Component {
...
+        <SearchBar
+          data={terms}
+          handleResults={(results) => this.handleResults(results)}
+          closeButton={<Icon style={styles.closeIcon} name="ios-close" />}
+          showOnLoad={true}
+          hideBack={true}
+          allDataOnEmptySearch={true}
+          autoCorrect={false}
+          focusOnLayout={true}
+          iOSPadding={false}
```
hawkrives: What does this one do?

source/views/student-orgs/list.js
```diff
@@ -97,6 +115,8 @@ export class StudentOrgsView extends React.Component {
       let orgs = {New: newOrgs, ...grouped}
 
       this.setState({orgs: orgs})
+      this.setState({justOrgs: responseData})
+      this.setState({results: responseData})
```
hawkrives: I'd like you to merge these three calls to setState into a single call, please.

source/views/dictionary/list.js
```diff
@@ -41,6 +57,12 @@ export class DictionaryView extends React.Component {
     route: React.PropTypes.object.isRequired,
   }
 
+  state: {
+    results: {[key: string]: Object},
```
hawkrives: So instead of `Object`, you should describe the shape of the object that it holds.

source/views/dictionary/list.js
```diff
@@ -77,21 +99,43 @@ export class DictionaryView extends React.Component {
     return <ListSeparator key={`${sectionId}-${rowId}`} />
   }
 
+  handleResults = (results: Object) => {
```
hawkrives: And then you should reference that same shape from line 61 here.

source/views/student-orgs/list.js
```diff
@@ -186,22 +210,35 @@ export class StudentOrgsView extends React.Component {
...
-        }
-        sectionHeader={this.renderSectionHeader}
-        sectionHeaderHeight={headerHeight}
-        renderSeparator={this.renderSeparator}
-        showsVerticalScrollIndicator={false}
-      />
+      <View style={styles.wrapper}>
+        <SearchBar
+          data={this.state.justOrgs}
+          handleResults={(results) => this.handleResults(results)}
```
hawkrives: I think prettier will remove these parens.

hawkrives (commented):
AFAICT, you didn't introduce any new flow types, and I commented on the two spots you used the generic `Object` type.

I'll need to download and play with this, but yay!

hawkrives:
Regarding dual-linking the rn-vector-icons package, I think you should be able to just … not? Not link the copy included with the rn-searchbar package. Maybe.

drewvolz:
I decided to just use a `close` icon from one of the fonts we already use as to not depend on `MaterialIcons` font. That works just fine. Are you referencing something else in the comment above?

The build is currently failing due to this issue:
```
node_modules/react-native-vector-icons/lib/generate-icon-set-from-css.js:0
generateIconSetFromCss. Duplicate module provider current provider. 
See: node_modules/react-native-searchbar/node_modules/
     react-native-vector-icons/lib/generate-icon-set-from-css.js:0
```

drewvolz:
Another thought -- should the search component replace the title component on Android? I *think* that is how Android uses search.... or at least I've seen it used that way.

hawkrives:
The build is currently failing because we have two copies of the rn-vector-icons in the tree. Ours, and the one that the library depends on. Even though you aren't using the icons they provide, they're still calling require() on the library, so it's in our tree.

My comment above was primarily about the message you sent me earlier about the font files being linked into the tree; as we're not using any of them from the inner library, we don't need to include them in our bundle.

drewvolz:
Okay, so they're still calling `require`. And we do use `rn-vector-icons` outside of this branch... Hmm. How would we avoid this?

hawkrives:
… well, with webpack I'd just make webpack ignore it. I'll think on it.

source/views/dictionary/list.js
drewvolz: This is an attempt to solve a timing issue. The data might not have been available at one point on the initial load. I am not so sure this is the best approach to solve it.

source/views/dictionary/list.js
drewvolz: `iOSPadding` is set to true by default. The size of the searchbar would be equivalent to the android size in the screenshots. I was under the impression it should be smaller. Do you have any thoughts about that?

drewvolz:
https://github.com/oblador/react-native-vector-icons/issues/421

drewvolz:
@hawkrives I am trying to remember what you wanted... we discussed somehow nativifying the iOS view to not appear so Androidy.

Did you suggest splitting into iOS and Android index views and wrapping `react-native-searchbar` and `react-native-search-bar`, or would it also work to display our own search-like view that we then stylize according to platform? Both methods of implementation would still require a wrapper for search of sorts.

hawkrives:
I advised adding `components/search-bar`, which would have `index.ios.js` and `index.android.js`. Both would expose the same, minimal, props.

The iOS one would use `react-native-searchbar`, and the android one would use https://www.npmjs.com/package/react-native-material-design-searchbar.

source/views/student-orgs/list.js
```diff
@@ -28,7 +28,7 @@ import * as c from '../components/colors'
 import startCase from 'lodash/startCase'
 import Icon from 'react-native-vector-icons/Ionicons'
 import SearchBar from 'react-native-searchbar'
-import type {StudentOrgAbridgedType} from './types'
+import type {GroupingArgsType, StudentOrgAbridgedType} from './types'
```
hawkrives: Could you just stick this type into the arguments of groupData instead of putting it in another file, please?

source/views/dictionary/list.js
```diff
@@ -77,21 +101,58 @@ export class DictionaryView extends React.Component {
...
-        sectionHeader={this.renderHeader}
-        sectionHeaderHeight={headerHeight}
-        showsVerticalScrollIndicator={false}
-        renderSeparator={this.renderSeparator}
-      />
+      <View style={styles.wrapper}>
+        <SearchBar
+          ref={ref => this.searchBar = ref}
+          data={terms}
+          handleChangeText={text => this.checkIfSearching(text)}
```
hawkrives: Please remove the arrow functions from these two: it's unnecessary overhead; because the arguments from the callbacks match up with the things you're calling, you can just pass references to the functions instead. `Handleresults={this.handleResults}`, for instance.

source/views/student-orgs/list.js
```diff
@@ -178,30 +185,95 @@ export class StudentOrgsView extends React.Component {
...
   }
 
+  handleResults = (results: {[key: string]: StudentOrgAbridgedType[]}) => {
+    this.setState({data: results})
+  }
+
+  groupData = (
+    data: {[key: string]: StudentOrgAbridgedType[]},
+    groupingArgs: GroupingArgsType,
+  ) => {
```
hawkrives: Oh, you're reusing the argument type. I'd prefer to keep them separate, as how to group data feels like an internal detail of the list view that's not inherently sharable between things.

source/views/dictionary/list.js
drewvolz: Thanks for the idea. I'll clean that up.

source/views/student-orgs/list.js
drewvolz: Could you explain this a little more, please?

drewvolz: Out of band, we figured this out. Thanks!

source/views/student-orgs/list.js
drewvolz: Done.

source/views/dictionary/list.js
```diff
@@ -105,11 +106,17 @@ export class DictionaryView extends React.Component {
   }
 
   performSearch = (text: string) => {
+    // Android clear button returns an object
+    if (!isString(text)) {
+      this.setState({results: terms})
+      return
+    }
```
hawkrives, line 111: 👍

source/views/components/searchbar/index.android.js
```diff
@@ -1,8 +1,16 @@
...
-export const SearchBar = (props: PropsType) => {
-  return <View />
+type PropsType = {
+  placeholder?: string,
+  onChangeText: string => any,
 }
+
+export const SearchBar = (props: PropsType) => (
+  <MaterialSearchBar
+    height={40}
```
hawkrives: Is this the height that the material guidelines specify?

source/views/components/searchbar/index.android.js
```diff
@@ -0,0 +1,19 @@
...
+import React from 'react'
+import {View} from 'react-native'
+import MaterialSearchBar from 'react-native-material-design-searchbar'
+
+type PropsType = {
+  placeholder?: string,
+  onChangeText: string => any,
+}
+
+export const SearchBar = () => <View />
```
drewvolz, line 11: Temporarily displaying a blank view (suggested by @hawkrives) to stall on Android search. It currently works, but we want to focus on nativifying this as much as we can... which will involve rendering the searchbar inside the navigation component.

source/views/components/searchbar/index.ios.js
```diff
@@ -0,0 +1,30 @@
...
+  style?: any,
+  placeholder?: string,
+  onChangeText: string => any,
+}
+
+export const SearchBar = (props: PropsType) => (
+  <NativeSearchBar
+    ref={props.getRef}
+    style={styles.searchbar}
+    hideBackground={true}
```
drewvolz, line 26: `hideBackground` allows us to apply a style to the component. I wanted `iosGray` as the background color and applying our own style was the only way to get around that.

source/views/dictionary/list.js
```diff
@@ -14,16 +14,25 @@ import {
...
+import deburr from 'lodash/deburr'
+import isString from 'lodash/isString'
 import {data as terms} from '../../../docs/dictionary.json'
+import {SearchBar} from '../components/searchbar'
 
 const rowHeight = Platform.OS === 'ios' ? 76 : 89
 const headerHeight = Platform.OS === 'ios' ? 33 : 41
 
 const styles = StyleSheet.create({
+  wrapper: {
```
drewvolz, line 33: Necessary to have both the searchbar and the listview both render. Without the flex, the list disappears.

source/views/dictionary/list.js
drewvolz: Thank you for writing `splitToArray`, `termToArray`, and `performSearch`, @hawkrives !

hawkrives (approved):

PR #997, opened 2017-04-26
Use proper semver notation for build metadata

hawkrives:
> Build metadata MAY be denoted by appending a plus sign and a series of dot separated identifiers immediately following the patch or pre-release version. Identifiers MUST comprise only ASCII alphanumerics and hyphen [0-9A-Za-z-]. Identifiers MUST NOT be empty. Build metadata SHOULD be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence. Examples: 1.0.0-alpha+001, 1.0.0+20130313144700, 1.0.0-beta+exp.sha.5114f85.

TL;DR: changing from `v2.1.2-build.234` to `v2.1.2+234`. This allows semver queries to work correctly against our app versions, which Codepush was having a hard time with (it's tricky to match `version-name` semver identifiers.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -92,10 +92,8 @@ before_install:
     fi
 
   # make sure to use ruby 2.3
-  - if [[ $ANDROID || $IOS ]]; then
-      rvm use 2.3 --install --binary --fuzzy;
-      gem install bundler;
-    fi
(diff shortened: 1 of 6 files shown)


drewvolz (approved):
Approval pending successful build

PR #1006, opened 2017-04-30
Hide keyboard in dictionary search and scroll

drewvolz:
Tweaked the dictionary searching on iOS for a smoother experience.

* Hide the iOS keyboard when hitting the "Search" button
* Dismiss the keyboard when scrolling the search results

### DIFF
--- a/source/views/dictionary/list.js
+++ b/source/views/dictionary/list.js
@@ -130,6 +130,7 @@ export class DictionaryView extends React.Component {
         <SearchBar
           getRef={ref => (this.searchBar = ref)}
           onChangeText={this.performSearch}
+          // if we don't use the arrow function here, searchBar ref is null...
           onSearchButtonPress={() => this.searchBar.unFocus()}
         />
         <StyledAlphabetListView
--- a/source/views/dictionary/list.js
+++ b/source/views/dictionary/list.js
@@ -130,6 +130,7 @@ export class DictionaryView extends React.Component {
         <SearchBar
           getRef={ref => (this.searchBar = ref)}
           onChangeText={this.performSearch}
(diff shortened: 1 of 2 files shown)


hawkrives (approved):
LGTM!

source/views/dictionary/list.js
```diff
@@ -130,6 +130,7 @@ export class DictionaryView extends React.Component {
         <SearchBar
           getRef={ref => (this.searchBar = ref)}
           onChangeText={this.performSearch}
+          onSearchButtonPress={() => this.searchBar.unFocus()}
```
hawkrives: actually one request – could you change this to just `{this.searchBar.unFocus}`, please? Passing in the function reference instead of creating an anonymous function just to call it?

hawkrives (changes requested):
jk, I just saw a thing. still great!

source/views/dictionary/list.js
drewvolz: Passing in the function reference instead of creating an anonymous function resulted in an error. In short, the `searchBar` ref was null.

drewvolz: Is there another option or are you okay with the change as-is?

hawkrives: :is thinking about scopes:

hawkrives: Sure, this is fine. Could you add a comment above it, like

```js
onChangeText=…
// if we don't use the arrow function here, `this.searchBar` is null...
onSearchButtonPress=…
```

Otherwise I'll probably keep trying to change it in the future.

drewvolz: I will make that comment. Thanks.

hawkrives (approved):

PR #1029, opened 2017-05-06
Hawken's stuff for 2.2

### LINKED ISSUE #534: Add button for "reset all data" in Settings
We don't know what bugs may crop up, and it's better to have a fallback than not. 

All this button would need to do is to clear AsyncStorage. Should be simple enough. I think there's an AsyncStorage.clearAll method, even. 

I suppose we should probably show a confirmation dialog,  too.

### LINKED ISSUE #1027: Auto-login for St. Olaf Login isn't working
It should automatically refresh your login every time the app opens.

hawkrives:
Stuff. I don't have time to document this before I'm going to a movie tonight.

I want to get this into 2.2.

Closes #1027
Closes #534

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -95,7 +95,7 @@ android {
     defaultConfig {
         applicationId "com.allaboutolaf"
         minSdkVersion 16
-        targetSdkVersion 22
+        targetSdkVersion 25
         versionCode 1
         versionName "2.1.2"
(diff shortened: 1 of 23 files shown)


hawkrives:
Le sigh. I think I can just revert the change that cause the flow error, and I'll rerun Prettier…

drewvolz (approved):

PR #1040, opened 2017-05-08
Allow travis to run a deploy from a tagged commit

hawkrives:
Currently, travis is set up to only ever deploy from the cron jobs.

This will allow it to deploy from tagged commits as well.


drewvolz (approved):
:+1: :shipit:

PR #1041, opened 2017-05-08
Add badges to ios/android beta builds

hawkrives:
Sample:

image 2017-05-08 12 01 03


hawkrives:
so this isn't ideal but it's not bad I think

we can tweak it later

also drew and I are really going to be the only two who see it

drewvolz (approved):
:+1: :shipit:

PR #1048, opened 2017-05-09
fix codepush quoting

hawkrives:
fix things

it broked


hawkrives:
I somehow managed to have ruby try to run the command

```bash
code-push release-react 'AllAboutOlaf-iOS' ios -d 'nightly' '--targetBinaryVersion ~2.2 || ~2.2.0-rc'
```

which, uh, is kinda not right.

This fixes that to be 

```bash
code-push release-react 'AllAboutOlaf-iOS' ios -d 'nightly' --targetBinaryVersion '~2.2 || ~2.2.0-rc'
```

which is much more correct.

PR #1053, opened 2017-05-09
Clear the Keychain when refreshing the app

### LINKED ISSUE #1052: Reset Everything remembered my username/password
I think this is because those are stored in the keychain and I forgot to wipe our keychain data.

hawkrives:
… turns out that I had a function written to do this already, but wasn't calling it lol.

Closes #1052


drewvolz (approved):
:+1: :shipit:

PR #1054, opened 2017-05-09
Disable KSTO url handler error handling

### LINKED ISSUE #1051: Cancelling the "open KSTO" prompt reports an error to us
We probably don't care to receive a bug report every time someone cancels that button tap.

hawkrives:
I disabled the ga/bugsnag tracking here because dismissing the dialogs causes us to get an error... and I can't really see how these would fail?

Like… yeah. There's no good way to differentiate between cancelled and other failures, which is (sadly) how things should be – this prevents bad actors from continually presenting you with the [Open URL] box if you cancel it.

Closes #1051


drewvolz (approved):
:+1: :shipit:

PR #1064, opened 2017-05-12
Remove the Pause's calendar

drewvolz:
This is their scheduling calendar... it is more for internal use than for our own. It contains times that are relevant to setup/teardown and contains internal gravity forms submissions.

...we also received an app store review saying the times were inaccurate.


hawkrives (approved):
… goodbye, pause. It was nice knowin' ya.

PR #1068, opened 2017-05-14
Calculate and expand webview height

drewvolz:
Calculate height of webview to auto-adjust height of organization descriptions. 

Before | After
--- | ---
screen shot 2017-05-13 at 9 30 39 pm | screen shot 2017-05-13 at 9 29 57 pm

### DIFF
--- a/source/views/components/html-view.js
+++ b/source/views/components/html-view.js
@@ -1,14 +1,14 @@
 // @flow
 
 import React from 'react'
-import {WebView, StyleSheet} from 'react-native'
+import {WebView} from 'react-native'
 import openUrl, {canOpenUrl} from '../components/open-url'
 
 export class HtmlView extends React.Component {
   props: {
     html: string,
     baseUrl?: ?string,
-    style?: ?any,
+    style?: ?Object,
     injectedJavaScript?: ?string,
     javaScriptEnabled?: ?boolean,
     onNavigationStateChange?: ?any,
--- a/source/views/components/html-view.js
+++ b/source/views/components/html-view.js
@@ -8,7 +8,10 @@ export class HtmlView extends React.Component {
   props: {
     html: string,
     baseUrl?: ?string,
-    style?: ?string,
+    style?: ?any,
+    injectedJavaScript?: ?string,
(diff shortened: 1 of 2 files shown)


hawkrives:
I'm assuming that you tested it on a number of the orgs? IIRC, when I tried this, it randomly quit working after like 8 views or something.

source/views/components/html-view.js
```diff
@@ -8,7 +8,10 @@ export class HtmlView extends React.Component {
   props: {
     html: string,
     baseUrl?: ?string,
-    style?: ?string,
+    style?: ?any,
+    injectedJavaScript?: ?string,
+    javaScriptEnabled?: ?boolean,
+    onNavigationStateChange?: ?any,
```
hawkrives: Instead of adding three new props, could you add one called, like, `autoAdjustHeight?: boolean`, please? And then just encapsulate the functionality into the HtmlView itself?

Or, like, make an AutoHeightHtmlView that wraps the HtmlView. Either way works, although I'd probably prefer the single prop.

source/views/components/html-view.js
```diff
@@ -8,7 +8,10 @@ export class HtmlView extends React.Component {
   props: {
     html: string,
     baseUrl?: ?string,
-    style?: ?string,
+    style?: ?any,
```
hawkrives: This shouldn't be any – it should be `number | Object | Array<number | Object>`… which I really ought to make into a, like, StylePropType or something, so we don't have to keep copying this around.

hawkrives (changes requested):
Nice work! Glad you got it to work. A few requests, though.

drewvolz:
Testing a bit more, you're right. I can make the same view have variable heights by refreshing the view... I'll try to fix that.

Measuring "Young Americans For Liberty" the first value is "160" and then the second time around it is "667".

drewvolz:
The calculations simply are not accurate 100% of the time. The `event.jsEvaluationValue` is not going to be accurate enough for me to have confidence that the web views will display at the correct height. Closing this.

PR #1085, opened 2017-05-18
Update SARN hours

drewvolz:
Community time on Thursday listed SARN being open until 12:35am instead of 12:35pm.


hawkrives (approved):
Whoops! Good catch!

PR #1086, opened 2017-05-18
Centralize Moment timezones and fix the default rounding

### LINKED ISSUE #807: Investigate use of Moment's `default` options
We should look into using [`Moment.relativeTimeRounding`][rounding] and [`Moment.relativeTimeThreshold`][threshold] to control the global rounding needs expressed in #that-other-issue. (see also: http://stackoverflow.com/a/42520227/2347774)

[rounding]: http://momentjs.com/docs/#/customization/relative-time-rounding/
[threshold]: http://momentjs.com/docs/#/customization/relative-time-threshold/

As it turns out, Moment Timezone also supports a ["default time zone"][default-timezone]. Which will let us show all of our moments in Central time without having that constant scattered over every file :smile:

[default-timezone]: http://momentjs.com/timezone/docs/#/using-timezones/default-timezone/

hawkrives:
This PR does two things:

1. It centralizes all the `moment.tz(CENTRAL_TZ)` stuff that we had scattered _everywhere_ into a central file that the entry point includes
2. It tells Moment to quit rounding minutes, so there are (for example) always 9 minutes between 5:21pm and 5:30pm
    - By default, it changes minutes at the 45-second mark

Closes #807
Appeases @drewvolz

### DIFF
--- a/source/app.js
+++ b/source/app.js
@@ -6,6 +6,7 @@
 
 // Tweak the global fetch
 import './globalize-fetch'
+import './setup-moment'
 import {tracker} from './analytics'
 import OneSignal from 'react-native-onesignal'
 
--- a/source/setup-moment.js
+++ b/source/setup-moment.js
@@ -1,0 +1,20 @@
+// @flow
+import moment from 'moment-timezone'
+
+// Set our default timezone to "central time"
+moment.tz.setDefault('America/Winnipeg')
+
+// These next values made possible by
+// http://stackoverflow.com/questions/42513276
+
+// Tell Moment to round up, so that Moment will always report the
+// time until something else in a way that matches the clock time.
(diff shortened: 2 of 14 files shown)


hawkrives:
In short, with this PR, the "time until close" on the Building Hours view will always reflect your system clock.

drewvolz (approved):
This is great! Thank you

hawkrives:
I tested this by changing my computer's time to Japan and verifying that the times until things opened didn't change.

Now… with that said.

Didn't we want, like, Calendar to show things in the device's time zone, instead of St. Olaf's time zone?

hawkrives:
I'm going to split this PR into two halves.

hawkrives:
I'm withdrawing this PR in favor of #1087. I would like to discuss which views want a localized time zone in here, however (or in a chat somewhere.)

PR #991, opened 2017-04-24
Port student orgs to st. olaf

hawkrives:
Thanks to @drewvolz for fixing my android problems!

This is now ready to review.

Closes #932.

~ | List | Detail
---|---|---
iOS | screen shot 2017-05-20 at 1 59 51 pm | screen shot 2017-05-20 at 2 00 16 pm
Android | screen shot 2017-05-20 at 1 55 15 pm | screen shot 2017-05-20 at 1 55 30 pm

PR #1093, opened 2017-05-20
List open student-work job postings

### LINKED ISSUE #135: Open job postings
I am not completely sure if this would fit into the spectrum of functionality we would like to provide, but I know finding student job opportunities can be a bit of pain. What if we made that easier, too?

I am imagining we could add a `student work` component to the app with two tabs. The first tab would be the Time Entry System that is being discussed in #74, and the second tab could be an optimized view of St. Olaf's [Student Employment job search tool](https://www.stolaf.edu/apps/stuwork/index.cfm?fuseaction=Search).

I cannot imagine it being too difficult to scrape the data from the dropdowns (job type, office, hours per week) and then to take input from the user for job title. Constructing the search query from this, we could perfrom a POST query, fetch the results, and display them in a ListView.

This would be a nice future addition to the proposed TES component.

drewvolz:
This PR displays the [student work listings](https://www.stolaf.edu/apps/stuwork/index.cfm?fuseaction=SearchResults) in list/detail format in the SIS section of the app.

Closes #135

~ | List | Detail
---|---|---
iOS | screen shot 2017-05-20 at 2 24 36 pm | screen shot 2017-05-20 at 2 24 24 pm
Android | screen shot 2017-05-20 at 3 34 04 pm | screen shot 2017-05-20 at 3 33 41 pm


hawkrives:
Prettier wants you to reformat it

drewvolz:
I would consider this ready for review.

source/views/sis/student-work/types.js
```diff
@@ -0,0 +1,18 @@
...
+  contactEmail: string,
+  contactFirstName: string,
+  contactLastName: string,
+  contactPhone: number,
+  description: string,
+  hoursPerWeek: string,
+  lastModified: string,
+  office: string,
+  skills: string,
+  timeOfHours: any,
```
hawkrives: Why is this `any`?

drewvolz: I've seen it as a string and a number

source/views/sis/student-work/detail.js
```diff
@@ -0,0 +1,206 @@
...
+function parseLinks(data: string) {
+  const allLinks = data.split(' ')
+  if (!allLinks.length) {
+    return []
+  }
+  return allLinks.filter(w => /^https?:\/\//.test(w))
+}
+
+export default function JobDetailView({job}: {job: JobType}) {
+  const title = job.title.trim() || ''
```
hawkrives: Please remove the `|| ''` from each of these that call `.trim()` – as it stands, you'll either crash (because there's no string to trim) or you'll get back a string (because you called trim) but you'll never use the right-hand value.

source/views/sis/student-work/detail.js
```diff
@@ -0,0 +1,206 @@
...
+  const description = job.description.trim() || ''
+  const office = job.office.trim() || ''
+  const type = job.type.trim() || ''
+  const comments = job.comments || ''
+  const skills = job.skills || ''
+  const hoursPerWeek = job.hoursPerWeek || ''
```
hawkrives: Why aren't you trimming this value?

source/views/sis/student-work.js
```diff
@@ -0,0 +1,144 @@
...
+  componentWillMount() {
+    this.refresh()
+  }
+
+  fetchData = async () => {
+    try {
+      const data: {[key: string]: JobType[]} = await fetchJson(jobsUrl)
+
+      const sorted = sortBy(data, [
+        j => jobSort.get(j.type) || jobSort.get('Unknown'),
```
hawkrives: I've been thinking about this – instead of doing the lookup for `Unknown`, what if we just said `|| Infinity` directly?

source/views/sis/student-work/job-row.js
```diff
@@ -0,0 +1,30 @@
...
+
+type JobRowPropsType = {
+  onPress: () => any,
+  job: JobType,
+}
+
+export function JobRow({onPress, job}: JobRowPropsType) {
+  const title = fastGetTrimmedText(job.title || '')
+  const office = fastGetTrimmedText(job.office || '')
+  const hours = fastGetTrimmedText(job.hoursPerWeek || '')
```
hawkrives: These are guaranteed to be strings already – don't do the `|| ''` dance.

hawkrives (changes requested):

source/views/sis/student-work/detail.js
drewvolz: Thanks

source/views/sis/student-work/detail.js
drewvolz: Good catch.

source/views/sis/student-work.js
drewvolz: Great idea.

source/views/sis/student-work/job-row.js
drewvolz: If they're blank, we will run into an issue with fastGetTrimmedText?

source/views/sis/student-work/types.js
hawkrives: Then mark it as `string | number`?

source/views/sis/student-work/job-row.js
hawkrives: No, we won't. We checked that yesterday.

source/views/sis/student-work.js
```diff
@@ -0,0 +1,144 @@
...
+  'https://www.stolaf.edu/apps/stuwork/index.cfm?fuseaction=getall&nostructure=1'
+
+const headerHeight = Platform.OS === 'ios' ? 33 : 41
+
+const jobSort = new Map([
+  ['On-campus Work Study', 1],
+  ['Off-campus Community Service Work Study', 2],
+  ['On-campus Summer Employment', 3],
+  ['Off-campus Summer Employment', 4],
+  ['Unknown', Infinity],
```
hawkrives: Please remove this pair, now that we're not using it

source/views/sis/student-work/detail.js
```diff
@@ -0,0 +1,206 @@
...
+
+function renderContact(
+  title: string,
+  office: string,
+  contact: string,
+  email: string,
+) {
+  const trimmedOffice = fastGetTrimmedText(office)
+  const trimmedContact =
+    fastGetTrimmedText(contact) || fastGetTrimmedText(email)
```
hawkrives: You don't need the full fastGet function for the email. The email won't have HTML in it.

source/views/sis/student-work/detail.js
```diff
@@ -0,0 +1,206 @@
...
+          onPress={() => openEmail(email, title)}
+        />
+      </Section>
+    : null
+}
+
+function renderHours(timeOfHours: string, hoursPerWeek: string) {
+  const time = fastGetTrimmedText(timeOfHours)
+  const hours = fastGetTrimmedText(hoursPerWeek)
+  const ending = hours == 'Full-time' ? '' : ' hrs/week'
```
hawkrives: Is this the only string you've seen in `hours`?

source/views/sis/student-work/detail.js
```diff
@@ -0,0 +1,206 @@
...
+          cellStyle="Subtitle"
+          title={time}
+          detail={hoursPerWeek + ending}
+        />
+      </Section>
+    : null
+}
+
+function renderDescription(description: string) {
+  const trimmedDescription = description.replace(/\t/g, ' ')
```
hawkrives: Instead of replacing tabs, what about replacing any consecutive whitespace with a single space? This is just a random idea.

source/views/sis/student-work/detail.js
```diff
@@ -0,0 +1,206 @@
...
+  return allLinks.filter(w => /^https?:\/\//.test(w))
+}
+
+export default function JobDetailView({job}: {job: JobType}) {
+  const title = job.title.trim()
+  const description = job.description.trim()
+  const office = job.office.trim()
+  const type = job.type.trim()
+  const comments = job.comments || ''
+  const skills = job.skills || ''
```
hawkrives: Remove the `|| ''` here as well.

source/views/sis/student-work/detail.js
```diff
@@ -0,0 +1,206 @@
...
+  const hoursPerWeek = job.hoursPerWeek.trim()
+  const timeOfHours = job.timeOfHours.toString() || ''
+  const lastModified = job.lastModified.trim()
+  const firstName = job.contactFirstName.trim()
+  const lastName = job.contactLastName.trim()
+  const name = `${firstName} ${lastName}` || ''
```
hawkrives: Remove the `|| ''` here as well.

source/views/sis/student-work/detail.js
```diff
@@ -0,0 +1,206 @@
...
+  }
+  return allLinks.filter(w => /^https?:\/\//.test(w))
+}
+
+export default function JobDetailView({job}: {job: JobType}) {
+  const title = job.title.trim()
+  const description = job.description.trim()
+  const office = job.office.trim()
+  const type = job.type.trim()
+  const comments = job.comments || ''
```
hawkrives: Remove the `|| ''` here as well.

source/views/sis/student-work/types.js
drewvolz: Sure.

source/views/sis/student-work/detail.js
```diff
@@ -0,0 +1,206 @@
...
+  const skills = job.skills || ''
+  const hoursPerWeek = job.hoursPerWeek.trim()
+  const timeOfHours = job.timeOfHours.toString() || ''
```
hawkrives: If this is either a string or a number, don't bother with the manual string coercion here. Just remove the .toString

hawkrives (changes requested):

source/views/sis/student-work/detail.js
drewvolz: * Flexible
* Full-time
* x-y

hawkrives: Ah, okay. Thanks.

source/views/sis/student-work/types.js
hawkrives: In general, you almost never want to annotate with `any`. You should either use a narrower type (like "string | number"), or a generic type if you're doing something like Array#map (`map = (Array<T>) => Array<U>` or whatever). Almost never `any` itself.

source/views/sis/student-work/detail.js
drewvolz: Maybe something similar to `replace(/ {2,}/g, ' ')`

drewvolz: We'll drop this... we want to leave in our newlines.

source/views/sis/student-work/detail.js
drewvolz: We need to in order to run our trimming on it?

source/views/sis/student-work/detail.js
hawkrives: Yep. I don't know how to dismiss this chain, but consider it dismissed.

source/views/sis/student-work/detail.js
hawkrives: Do we? Do we trim it?

source/views/sis/student-work.js
```diff
@@ -0,0 +1,143 @@
...
+  componentWillMount() {
+    this.refresh()
+  }
+
+  fetchData = async () => {
+    try {
+      const data: {[key: string]: JobType[]} = await fetchJson(jobsUrl)
+
+      const sorted = sortBy(data, [
+        j => jobSort.get(j.type) || Infinity,
```
hawkrives: Add a comment, like `// We have predefined orders for some job types, but we want all unknown types to show up at the end of the list` (but wrapped to 80 columns)

source/views/sis/student-work/detail.js
drewvolz: https://github.com/StoDevX/AAO-React-Native/pull/1093/files#diff-b070c2f61221250daa96b484cae400b8R90

hawkrives (changes requested):
I have a bigger request, too. In https://github.com/StoDevX/AAO-React-Native/pull/991/commits/23f87c061ef417bc930dc586f4a70318d30b3cbc, I pulled the trimming and website validation of things into another function, so the two detail views can share that logic.

I would like something similar here, please.

source/views/sis/student-work/detail.js
hawkrives: Please remove the getFastTrimmedText there – it's never going to be an html string.

Please :)

source/views/sis/student-work/detail.android.js
```diff
@@ -0,0 +1,225 @@
...
+          {'\n'}
+          {trimmedOffice}
+        </Text>
+      </Card>
+    : null
+}
+
+function renderHours(timeOfHours: string, hoursPerWeek: string) {
+  const time = fastGetTrimmedText(timeOfHours)
+  const hours = fastGetTrimmedText(hoursPerWeek)
```
hawkrives: I'm going to remove both these fastGetTrimmedTexts

hawkrives (approved):

hawkrives (approved):

drewvolz (commented):
Testing if I can hack my way to self approval.

hawkrives:
Oops. I didn't mean to delete that branch.

PR #1094, opened 2017-05-20
Add Google Play betas

hawkrives:
First off, I've finally figured out what versionCode does – it's a monotonically increasing value, so it should be incremented for each and every beta and whatnot.

Now! This allows us to upload betas to Google Play.

It also signs and aligns the apks.

Let's see if everything works on Travis, when it's not time to release a beta.


drewvolz (approved):
Wow. Great work, @hawkrives

hawkrives:
It passed, so I'm going to merge it. 

It may fail the next time we tag a commit - we'll have to see.

PR #1098, opened 2017-05-20
Add package-lock.json

hawkrives:
This is part of [npm@5](https://github.com/npm/npm/issues/16510).

It is completely backwards-compatible. 

I'm adding it now so that I can start watching how it interacts with the cli when we install updated packages.


elijahverdoorn (approved):
Interested to see how this works.

drewvolz (approved):

PR #1102, opened 2017-05-22
Add a Detail view to Calendar

hawkrives:
@drewvolz, I'm going to open this PR so we remember work on it.

What do we have / what do we need?

Do we want to do Android styles before merging this?


drewvolz:
> What do we have / what do we need?

What we _have_ is a detail view that spits out details in our simple tableview. What we _need_ is something more styled per platform, and maybe we need to rethink this view. If we want to get it right, I think that involves looking at how iOS and Android native calendar apps present the detail summary.

> Do we want to do Android styles before merging this?

~~Yes~~ No

drewvolz:
I'm fine with merging this as-is.

hawkrives:
I'll merge it as soon as it passes and we can clean it up before we release next.

PR #1134, opened 2017-06-03
Remove codepush entirely

### LINKED ISSUE #1131: Remove codepush
@hawkrives has a few thoughts on this:

>  - We should not change the app that people have installed. Codepush is a cool idea, but if we are going to think along these lines, we do not need to have the functionality.
> - It has become a nuisance to figure out the difference between js and binary releases
> - The iOS app store is at about a 24hr review time which is quick enough

drewvolz:
This removes the codepush package from the repository. The app builds and runs on iOS and Android still -- so everything works as it did before. 

We will need to add a way to refresh the app state (something a codepush reset previously did) which will be tracked in https://github.com/StoDevX/AAO-React-Native/issues/1135.

Closes #1131.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -71,7 +71,6 @@ project.ext.react = [
 ]
 
 apply from: "../../node_modules/react-native/react.gradle"
-apply from: "../../node_modules/react-native-code-push/android/codepush.gradle"
 
 /**
  * Set this to true to create two separate APKs instead of one:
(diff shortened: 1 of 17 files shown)


hawkrives:
I do want to note that with this PR, it becomes trivial to re-add codepush in the future. We can "just" `git revert <merge commit>` and we'll have codepush back.

hawkrives (approved):

elijahverdoorn (approved):
Given that we're having technical issues with it, I think that this is necessary. I still don't have a problem with using the service to push updates to users, and I think that ignoring the opportunity to use this in the future would be a mistake, but for now I defer to @hawkrives and @drewvolz and approve.

hawkrives:
I'd be down for re-evaluating whether or not to use it in the future, for sure.

PR #1151, opened 2017-06-08
clean up calendar times

hawkrives:
This PR does several things:

1. It hides the "location" row when it's blank; this fixes the case where it looks like there's a bunch of pointless padding under some rows.
2. If the "times" column is taller than the "info" column, `master` currently doesn't align the "location" with the bottom of the row. Now it should.
3. If a time ends or starts at midnight, the text will now read "Midnight".
4. If a multi-day event ends at midnight, it will not report the ending time at all.

### DIFF
--- a/source/views/calendar/event-row.js
+++ b/source/views/calendar/event-row.js
@@ -88,8 +88,8 @@ function CalendarTimes({event, style}: {event: EventType, style: any}) {
     )
   }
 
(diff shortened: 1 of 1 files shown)


hawkrives:
I’ll fix those two problems later, then merge this, and do another beta in the morning.

PR #1161, opened 2017-06-11
Reduce duplication between the Building Hours details views

hawkrives:
Before, we had… 100% duplication between the two platform's BuildingHoursDetail views.

Now, we have almost none.

detail/index is the entry point; it has the only use of `state` in the folder.

The different parts of the screen (the header, open/closed badge, schedules table) were extracted into appropriate files.

details/building orchestrates the structure of the detail view and performs the main data manipulations.

Note that even with the additional lines needed by having the imports in each file, this PR only adds about 40 lines of code; in fact, if I remove the top-of-file comments, I get `387 insertions(+), 387 deletions(-)` – wow.

Alright. That's it.

There should be no UI changes due to this PR.


drewvolz (approved):
👍 :shipit: great work!

PR #1173, opened 2017-06-12
Whitelist ListRow's props

hawkrives:
I noticed while doing some cleanup on BusView that I was misusing ListRow.

Flow provides an "exact object type", which acts as a property key whitelist.

I changed ListRow's props to use the exact object type, which caught two spots where I was misusing ListRow.

### DIFF
--- a/source/views/transportation/bus/bus-line.js
+++ b/source/views/transportation/bus/bus-line.js
@@ -1,6 +1,6 @@
 // @flow
 import React from 'react'
-import {View, StyleSheet, Platform} from 'react-native'
+import {View, StyleSheet, Platform, Text} from 'react-native'
 import type {BusLineType, FancyBusTimeListType} from './types'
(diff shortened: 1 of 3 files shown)


source/views/transportation/bus/bus-line.js
```diff
@@ -61,7 +60,9 @@ export function BusLine({line, now}: {line: BusLineType, now: moment}) {
     return (
       <View>
         <ListSectionHeader title={line.line} titleStyle={androidColor} />
-        <ListRow title="This line is not running today." />
+        <ListRow>
+          <Title><Text>This line is not running today.</Text></Title>
```
drewvolz, line 64: ~This syntax looks odd?~

drewvolz: Nevermind. This is fine.

drewvolz (approved):

PR #1175, opened 2017-06-12
Add basic transit maps

### LINKED ISSUE #1156: Add map view of Northfield Express bus + colored routes
We already have coordinates for the express in our data (https://github.com/StoDevX/AAO-React-Native/blob/master/data/bus-times/express.yaml).

The other two don't, but it wouldn't be too much extra work to collect the data points.

I envision a UITableCell / android list equivalent at the top/bottom of the times list that opens the map view.

Thoughts?

---

- [x] Northfield Express
- [ ] Red Route
- [x] Blue Route (https://github.com/StoDevX/AAO-React-Native/pull/1385)

hawkrives:
I added a basic Transit Map implementation.

Still to do: 
- should we add the route information to the map?
- test on Android
- add any coordinate info for the Red and Blue lines

Closes #1156


drewvolz (approved):

source/views/transportation/bus/map.js
```diff
@@ -0,0 +1,127 @@
...
+
+    const schedule = getScheduleForNow(activeBusLine.schedules, now)
+    if (!schedule) {
+      const notice = `No schedule was found for today, ${now.format('dddd')}`
+      return <NoticeView text={notice} />
+    }
+
+    const coords = schedule.coordinates || []
+    if (!coords.length) {
+      const notice = `No coordinates have been provided for today's (${now.format(
```
drewvolz, line 97: Maybe we could tweak the language a little to have less jargon. Something without "coordinates" ... 

- "No routes to display"
- "No schedule to map"
- "No route to map for today"
- "Nothing to map for today"

drewvolz:
Overall: wow! I really like where this is headed. We could poke at making the bus' path clear and maybe where the bus' estimated location is.

hawkrives:
I like both of those ideas; I think that locating the bus would probably want to happen after we have the route mapped out?

hannesmcman:
@hawkrives @drewvolz What happened to this?

hawkrives:
I got tired of having to update the old AirBnB react-native-maps package, and neither of us ever got around to adding the rest of the data for the other two bus routes IIRC

drewvolz:
The files are [still around](https://github.com/StoDevX/AAO-React-Native/pull/1175/files). I would still like to see this happen, I am betting @hawkrives is, too.

hannesmcman:
@hawkrives Could be build this using the mapbox map you used for CARLS?

drewvolz:
There should be no issues with using mapbox. Related to this is also https://github.com/StoDevX/AAO-React-Native/issues/1155

hawkrives:
Yeah, Mapbox can Dow whatever we want

hannesmcman:
@hawkrives @drewvolz If I were to take a stab at this, do you think you would prefer the same type of layout from the screenshots above (where the map is a separate view), or something more like the map view from CARLS (where the bus schedules would be slid up from the bottom over the map)?

drewvolz:
I kind of like the latter idea where it slides while the map is the central focus.

hawkrives:
I think I'd like to rethink the transit view, honestly; I'm not a huge fan of apps that start with a map of the bus route, because I'm usually opening the map to see when the next stop is. 

What I could see, though, might be something that has a little "notification" at the top of the view, telling you where the bus's next couple stops are, maybe something like [image], which I just ripped off of the internet. 

It could then have a square map, zoomed to show the entire route with each stop having a small but distinct dot, and ideally marking the current expected spot of the bus on the route. 

Below the square route could then be the normal row-based transit timetable, or something, because I think it's valuable to have an at-a-glance timetable of the bus route. 

You could tap on the square map to open an interactive, zoomable version, which would be full app screen, as well. 

---

Or something like that. 

If you're just looking to reimplement the map, I'd vote for just putting a link to the map at the bottom of the timetable again, please.

hannesmcman:
What you just described is almost exactly what I was envisioning! :) I'm all for it!

hannesmcman:
@hawkrives Could you draw a diagram maybe? Especially for what the "notification" thingy would look like?

drewvolz:
> I'd vote for just putting a link to the map at the bottom of the timetable again, please.

I’d vote for top

hawkrives:
I don't want the map at the top :p

hawkrives:
@hannesmcman Something of a combo between

img_5109

and

aao hours - 2017-09-20 07 50 13

drewvolz:
> I don't want the map at the top :p
 
I don’t think it’s easily discoverable at the bottom.

hawkrives:
I think it's too visible at the top.

drewvolz:
Visible over invisible 🤷‍♂️

PR #1209, opened 2017-06-16
expand calendar detail views to multiple lines

### LINKED ISSUE #1192: Calendar detail view's description should be multi line
It's currently restricted to a single line. 

image

drewvolz:
The calendar detail view's descriptions did not expand to multiple lines.

Closes #1192 

Before | After
---|---
screen shot 2017-06-16 at 4 29 28 pm | screen shot 2017-06-16 at 4 28 59 pm


hawkrives:
Why is the font size smaller now?

drewvolz:
A title has a different font size than a text element. Without having changed the font size explicitly, this also has the benefit of bringing it into line with the rest of our detail views.

Issue #1158, opened 2017-06-10
Add “report a problem” button / functionality to BuildingHours


hawkrives:
So that we can crowdsource the info. 

---

- [ ] styling (namely the colors of the day picker buttons) (#1509)
- [x] moving the "submit a problem" button to the bottom of the detail view
- [x] ~~implementing Github PR abilities~~ removed
- [x] implementing serialize-to-nice-YAML ability

hawkrives:
After we merge #1171, the remaining issues are:

- styling (namely the colors of the day picker buttons)
- moving the "submit a problem" button to the bottom of the detail view
- implementing Github PR abilities
- implementing serialize-to-nice-YAML ability

PR #1220, opened 2017-06-17
Format the header in hours reporting

drewvolz:
This throws a few styles on the header for the building hours report view. I've also nixed the footer on the submit button.

A piece of #1158

Before | After
--|--
unformatted | formatted

### DIFF
--- a/source/views/building-hours/report/overview.js
+++ b/source/views/building-hours/report/overview.js
@@ -5,12 +5,13 @@
  */
 
 import React from 'react'
-import {ScrollView, View, Text} from 'react-native'
+import {ScrollView, View, StyleSheet, Text} from 'react-native'
 import moment from 'moment-timezone'
 import {CellTextField} from '../../components/cells/textfield'
 import {CellToggle} from '../../components/cells/toggle'
 import {DeleteButtonCell} from '../../components/cells/delete-button'
 import {ButtonCell} from '../../components/cells/button'
+import * as c from '../../components/colors'
 import {TableView, Section, Cell} from 'react-native-tableview-simple'
 import type {
   BuildingType,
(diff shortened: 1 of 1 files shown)


source/views/building-hours/report/overview.js
```diff
@@ -20,6 +21,27 @@ import type {
 import type {TopLevelViewPropsType} from '../../types'
 import {summarizeDays, formatBuildingTimes, blankSchedule} from '../lib'
 
+const styles = StyleSheet.create({
+  helpWrapper: {
+    backgroundColor: c.white,
+    borderWidth: StyleSheet.hairlineWidth,
+    borderColor: '#ebebeb',
```
hawkrives: This is in colors somewhere.

source/views/building-hours/report/overview.js
```diff
@@ -192,7 +217,7 @@ export class BuildingHoursProblemReportView extends React.PureComponent {
             />
           </Section>
 
-          <Section footer="Thanks for reporting!">
+          <Section>
```
hawkrives: Why?

hawkrives (changes requested):

source/views/building-hours/report/overview.js
drewvolz: It's not.

source/views/building-hours/report/overview.js
drewvolz: It looked out of place.

drewvolz: If you'd prefer it back in, I'll add it.

hawkrives: I would like it for now, at least. I do want to rethink our buttons in this view, though, but that's a matter for a different PR.

hawkrives (approved):

PR #1250, opened 2017-06-27
Remove the lockfile

hawkrives:
My attempt at getting Travis to update our lockfiles is evidently not working.

Therefore, I'm disabling lockfiles, so that we actually receive package updates.


hawkrives:
Which breaks Danger. Of course. 

I'm going to have to downgrade it again. Luckily we don't use anything new since 0.18.

hawkrives:
Oh hey, Danger passed this time (I downgraded it to 0.18.0.)

I'll be merging this so that I can do the Prettier upgrade today.

PR #1280, opened 2017-07-12
[Test] poke at fixing failing build deps

drewvolz:
Do not review or merge this. PRing to get CI to build.


drewvolz:
Closing because this was a test. We do not need to rely on `react-dom` this way.

PR #1281, opened 2017-07-12
Remove Enzyme

hawkrives:
Enzyme (and react-native-mock) are causing our test failures.

https://github.com/airbnb/enzyme/issues/614#issuecomment-254139719 mentions that enzyme isn't quite intended for react-native testing (yet?), which I knew, but it was "mostly" working until last week.

This PR removes enzyme and related tools.

It also disables the two interaction tests I had written, since I haven't figured out how I want to approach interaction simulation without enzyme present yet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -116,16 +116,12 @@ install:
 
 before_script:
   - bash scripts/travis/init_android_emulator.sh
-  - greenkeeper-lockfile-update
 
 script:
   - bash scripts/travis/script.sh
   # we run `danger` out here to ensure that it is run,
   # even if another task in script.sh fails.
   - if [[ $JS ]]; then npm run danger; fi
-
-after_script:
-  - greenkeeper-lockfile-upload
 
 after_success:
   - bash scripts/travis/after_success.sh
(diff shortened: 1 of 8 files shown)


hawkrives:
So the current build failure... is that Travis isn't ending the JS build on a successful finish, for some reason.

drewvolz:
https://docs.travis-ci.com/user/common-build-problems/#My-builds-are-timing-out

drewvolz:
So we might have some processes still running. 

@hawkrives suggested a way to debug this:
> 1. Make a new branch called `travis-stalls` or something
> 2. Disable the non-js builds on it (that's the first commit)
> 3. Then comment out the `after_success` section, wait for it to finish
> 4. Then comment out danger
> 5. Then comment out each part of the js' scripts section, waiting for each run to finish (none of the js scripts need any other of the scripts to have run)

drewvolz (approved):
:+1: :white_check_mark: :shipit: thanks for fixing the build!

hawkrives:
I disabled the call to `ssh-add`.

I guess the ssh daemon was still running or something?

We don't use this key anymore; it used to be used for pushing changes back to github, but that's currently disabled.

PR #1292, opened 2017-07-17
[WIP] Android home screen widget

elijahverdoorn:
Here's my long-promised android home screen widget. This will read a JSON file put in the filesystem, and provide the values to the widget.

There is also some code in place using the `react-native-fs` package to store a JSON file in the file tree, but for some reason that file is not getting placed properly, and is not giving an error. I haven't put together any services, so the updating is currently only going to happen whenever the balances are pulled from the server. This should be updated in the future, along with fixing the file placement in the first place.

screen shot 2017-07-16 at 5 23 55 pm


hawkrives:
(I fixed the merge conflict so that Travis can try building it.)

source/flux/parts/sis.js
```diff
@@ -20,6 +20,7 @@ export function updateBalances(forceFromServer: boolean = false) {
   return async (dispatch: () => {}, getState: any) => {
     const state = getState()
     const balances = await getBalances(state.app.isConnected, forceFromServer)
+    updateBalancesInFileSystem(balances)
```
hawkrives: This should be `await`ed, so that it's completed

hawkrives:
So for flow's two comments:

```
source/lib/financials/balances.js:54
54: export function updateBalancesInFileSystem(balances): Promise {
                                               ^^^^^^^^ parameter `balances`. Missing annotation
```

Any `export`ed function must have it's arguments and return type annotated. In this case, the argument has no annotation. 

```
source/lib/financials/balances.js:54
54: export function updateBalancesInFileSystem(balances): Promise {
                                                          ^^^^^^^ Promise. Application of polymorphic type needs <list of 1 argument>. (Can use `*` for inferrable ones)
```

For this one, flow's like "a promise of what?" It wants to know what the promise will contain. In this case, I'd make two changes: one, annotate it as `Promise<any>`, and two, return the rnfs.whatever promise so that the rest of the system can use it if need be.

hawkrives:
@elijahverdoorn Is there some other license we need to accept for the new constraintsolver package?

Checking the license for package Solver for ConstraintLayout 1.0.2 in /usr/local/android-sdk/licenses
Warning: License for package Solver for ConstraintLayout 1.0.2 not accepted.
Checking the license for package ConstraintLayout for Android 1.0.2 in /usr/local/android-sdk/licenses
Warning: License for package ConstraintLayout for Android 1.0.2 not accepted.
FAILURE: Build failed with an exception.
* What went wrong:
A problem occurred configuring project ':app'.
> You have not accepted the license agreements of the following SDK components:
  [ConstraintLayout for Android 1.0.2, Solver for ConstraintLayout 1.0.2].
  Before building your project, you need to accept the license agreements and complete the installation of the missing components using the Android Studio SDK Manager.
  Alternatively, to learn how to transfer the license agreements from one workstation to another, go to http://d.android.com/r/studio-ui/export-licenses.html

elijahverdoorn:
I'll be fixing these issues on Thursday night when I have more time. The JS doesn't work, but I'll fix the flow annotations and take another stab at getting it to write the file.

ConstraintLayout is super cool (https://developer.android.com/training/constraint-layout/index.html) because it eliminates the ultra-nested nature of older android views. Think of it as flexbox for android. I'm not sure what they're talking about with the license, the library was released by Google in February and I played with it then, so if there is a license I'm sure I accepted it then and just didn't have to do so again since I was working from the same computer. I'll uninstall and reinstall and see if it asks me to accept something.

hawkrives:
We can potentially do this to accept the ConstraintLayout license: https://github.com/travis-ci/travis-ci/issues/6617

hawkrives:
We'll have to figure out where the sdkmanager binary lives

elijahverdoorn:
Putting this here so I can find it later.

https://docs.travis-ci.com/user/languages/android/#Dealing-with-Licenses

android/app/build.gradle
```diff
@@ -201,6 +201,7 @@ dependencies {
     compile "com.android.support:appcompat-v7:23.0.1"
     compile "com.facebook.react:react-native:+"  // From node_modules
     compile 'com.android.support.constraint:constraint-layout:1.0.2' // For android home screen widget
+    compile 'com.squareup.okhttp3:okhttp:3.8.1' // Make requests from android code
```
elijahverdoorn: OkHttp is widely used in native android to make HTTP requests easily off the main thread.

android/app/src/main/java/com/allaboutolaf/BalancesIntentService.java
```diff
@@ -0,0 +1,60 @@
...
+import com.squareup.okhttp.Response;
+
+import java.io.IOException;
+import java.util.logging.Logger;
+
+/**
+ * Created by everdoorn on 8/8/17.
+ */
+
+public class BalancesIntentService extends IntentService {
```
elijahverdoorn, line 22: I've been thinking about this, and really doing stuff from the JS code and passing to the native via a file is not a good solution. We want the widget to update without the user having to start the app, so we have to go native. This is the start of a native service implementation that should allow the widget to update in the background. I'm a tad stuck with the authentication, though. Do either @hawkrives or @drewvolz know how we might pass the user's info that we store in the keychain to the native system? I haven't poked around the source of that package yet.

elijahverdoorn:
Had to push unfinished code since I've been developing this on my work computer, which I have to give back at the start of the day tomorrow and I didn't want to throw away what I've got locally. Sorry for the mess of commits.

android/app/src/main/java/com/allaboutolaf/BalancesIntentService.java
hawkrives: All valid points. I was actually just thinking about this last night, while I was wishing I had a watch app for balances.

AFAIK, react-native-keychain delegates to a secure store on the native side. On iOS, that's the system keychain; I don't know what it uses on Android.

hawkrives: … but you should be able to just read the credentials from that.

hawkrives:
I'm going to go ahead and close this; we can revisit it when we next approach home screen widgets.

PR #1293, opened 2017-07-17
Add Title-IX related contact information

elijahverdoorn:
I was contacted by Kari Hohn, the college Title-IX Coordinator about putting some more information into the app. We've been emailing about it, and settled on what you see here. I'm starting to wonder if there is a need for a dedicated Resources section of the app, as important contacts seems to be a little cluttered (Pause Kitchen feels even more out of place now!). Maybe adding tabs to the Contacts section would fix the problem and let us add a few more in the future without it becoming a massive list?

@erichkauffman I'd encourage you to reach out to the It's On Us chairs for this upcoming year and see what they think about this data change.

### DIFF
--- a/data/contact-info/5-college-pastors.yaml
+++ b/data/contact-info/5-college-pastors.yaml
@@ -1,0 +1,8 @@
+title: College Pastors
+phoneNumber: '5077863092'
+buttonText: Call Pastor's Office
+image: safe ride.jpg
+text: >-
+  College Pastors are confidential resources trained in pastoral care. They
+  are able to speak with students of any faith, or who don’t practice a
+  religious faith, about all aspects of life.
--- a/data/contact-info/6-health-services.yaml
+++ b/data/contact-info/6-health-services.yaml
@@ -1,0 +1,13 @@
+title: Health Services
+phoneNumber: '5077863063'
+buttonText: Call Health Servies
+image: safe ride.jpg
+text: >-
+  St. Olaf Health Services provides free, confidential, on-campus healthcare
+  visits for students. Medications and testing are available to students during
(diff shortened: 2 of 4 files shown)


hawkrives:
I do like the idea of splitting them into tabs within the contacts screen. I'll go ahead and merge this, and we can open an issue about reorganizing them?

data/contact-info/5-college-pastors.yaml
```diff
@@ -0,0 +1,8 @@
+title: College Pastors
+phoneNumber: '5077863092'
+buttonText: Call Pastor's Office
+image: safe ride.jpg
```
hawkrives, line 4: Do we even use these images in the app?

hawkrives (approved):

drewvolz:
> I'm starting to wonder if there is a need for a dedicated Resources section of the app, as important contacts seems to be a little cluttered (Pause Kitchen feels even more out of place now!).

Adding tabs is a decent stab at fixing the problem. I also would propose a `master > detail` listview so that more info can be found in a child view and the name/phone are listed up front... 

numbers-important

drewvolz (approved):

hawkrives:
@drewvolz Those look nothing like a Master > Detail view 😝

drewvolz:
Think of it as master > detail because we keep getting rid of the detail chevron

hawkrives:
A plus icon doesn't make sense, either…

drewvolz:
I'm not advocating for the + sign. I'm showing one example of a limited amount of info that we can get across, and then I am asking for you to imagine a detail view with more info.

hawkrives:
pshhh imagination

PR #1298, opened 2017-07-19
upgrade react-native to 0.45

hawkrives:
> Latest RN is 0.46, but it doesn't play nice with react-navigation, and they don't seem to have any active plans to release an update, so I'm sticking with 0.45 for now.

I went to make a new beta tonight, and it started crashing on launch. After some research, I realized some of our recent dependency updates switched out references to `View.propTypes` with `ViewPropTypes`, in accordance with React's 0.44+ guidance.

Since we were on 0.42, however, we didn't have ViewPropTypes available for those packages to use, and they crashed on launch.

So I bit the bullet and upgraded us to a newer react-native, and an alpha react release…

This wouldn't have been possible before I removed Enzyme last week, either, so what a perfect storm this was.

---

This PR also includes a few dependency upgrades:

- `"babel-plugin-flow-react-proptypes": "3.4.2"`
- `"react-native-tableview-simple": "0.16.8"`
- `"react-native-datepicker": "1.6.0"`
- `"react-native-button": "2.1.0"`
- `"flow-bin": "0.45.0"`

---

This PR also includes a few flow fixes, namely the fact that flow 0.47 (which we aren't on quite yet, but I tried it for a bit and am including the fixes here) validates the number of arguments you pass to a function, so I had to fix several occurrences of function types taht didn't claim to accept any arguments.

Flow 0.45, which we are now on, also caught several `style` references that didn't exist, so HOORAY for flow.


hawkrives:
Once this finishes building, I'm going to merge, and (if I'm still awake then) I'll make another beta, so it'll work.

PR #1300, opened 2017-07-19
fix :CFBundleIdentifier Not Found error

hawkrives:
It turns out that the React Native CLI was looking for the app bundle under the name "AllAboutOlaf", but the bundle's name was "All About Olaf" (with spaces.)

What I did was to change the bundle identifier to the no-space version, then add the spaced version as the Display Name for the app target. Now the CLI is happy, and our users are happy because the app still has spaces in its homescreen name.


drewvolz (approved):
:+1: :shipit: Thanks for finding and fixing that.

Issue #1189, opened 2017-06-15
Report Hours screen doesn't actually submit anything


hawkrives:
I'm envisioning it directly submitting a PR to our repo, by way of a github token embedded in the app.

To do this, we'll need to be able to format YAML in a way that we like from a data structure, and we'll need to hook up the PRing bit too.

elijahverdoorn:
I still think that we should run a backend server for this stuff instead of directing things to Github, but if something was built to direct this information to GitHub, it would be better than nothing.

hawkrives:
I still don’t see what a server gets us over static files in github. We’d still need to verify submissions before accepting them, which PRs give us for free. We’d have to _host_ the server and maintain uptime, which github gives us for free. 

What else is there?

Kinda a moot point, since I haven’t built the PRing function yet, but I’m curious.

hawkrives:
<details><summary>data</summary>

```json
{
      "name": "Stav Hall",
      "image": "stav",
      "category": "Food",
      "schedule": [
        {
          "title": "Breakfast",
          "hours": [
            {
              "days": [
                "Mo",
                "Tu",
                "We",
                "Th",
                "Fr",
                "Sa"
              ],
              "from": "7:00am",
              "to": "9:45am"
            },
            {
              "days": [
                "Su"
              ],
              "from": "8:30am",
              "to": "10:15am"
            }
          ]
        },
        {
          "title": "Lunch",
          "hours": [
            {
              "days": [
                "Mo",
                "Tu",
                "We",
                "Th",
                "Fr"
              ],
              "from": "10:30am",
              "to": "2:00pm"
            },
            {
              "days": [
                "Sa",
                "Su"
              ],
              "from": "11:00am",
              "to": "1:30pm"
            }
          ]
        },
        {
          "title": "Dinner",
          "hours": [
            {
              "days": [
                "Mo",
                "Tu",
                "We",
                "Th",
                "Fr",
                "Sa",
                "Su"
              ],
              "from": "4:30pm",
              "to": "7:30pm"
            }
          ]
        }
      ],
      "breakSchedule": {
        "fall": [],
        "thanksgiving": [],
        "winter": [],
        "interim": [],
        "spring": [],
        "easter": [],
        "summer": []
      }
    }
```

</details>

```js
stringified = jsYaml.safeDump(data, {flowLevel: 4})
    .split('\n')
[...]

PR #1304, opened 2017-07-21
Activate "submit building hour update" button

hawkrives:
So submitting a GitHub issue, while nice, isn't really workable – we'd either need to make our users set up and log in to a github account, or ship our app's private token with the app somehow, which would allow someone to extract it from the app and pretend to be the app (not the end of the world, but not ideal).

So I went with email. `allaboutolaf@stolaf.edu` currently emails @drewvolz, me, and @rye; lmk if you want to be added (or removed).

Closes https://github.com/StoDevX/AAO-React-Native/issues/1189.

### DIFF
--- a/source/views/building-hours/report/submit.js
+++ b/source/views/building-hours/report/submit.js
@@ -20,7 +20,6 @@ export function submitReport(current: BuildingType, suggestion: BuildingType) {
   )
 }
 
-
 function makeEmailBody(before: string, after: string, title: string): string {
   return dedent`
     Hi! Thanks for letting us know about an hour change.
--- a/source/views/building-hours/report/submit.js
+++ b/source/views/building-hours/report/submit.js
@@ -4,6 +4,7 @@ import jsYaml from 'js-yaml'
 import type {BuildingType} from '../types'
 import {email} from 'react-native-communications'
 import dedent from 'dedent'
+import querystring from 'querystring'
 
 export function submitReport(current: BuildingType, suggestion: BuildingType) {
   const before = stringifyBuilding(current)
--- a/source/views/building-hours/report/submit.js
+++ b/source/views/building-hours/report/submit.js
@@ -1,0 +1,54 @@
+// @flow
+
+import jsYaml from 'js-yaml'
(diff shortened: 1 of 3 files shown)


hawkrives:
I stole like a quarter of this code from https://github.com/StoDevX/AAO-React-Native/issues/1189.

rye:
@hawkrives I'd be down to be added.  Always nice to keep my ear to the ground, if nothing else.  It seems I've already been added.

From my own experience, it seems there is no really easy way to ship a private GitHub API key to the world as suggested besides doing just that; from GitHub's perspective I'm sure having many devices using the token for direct API access without any rate limiting or filtering in the middle is not ideal.  Hence, either you write a little API yourself to handle it (take JSON payloads from clients, process and send to GitHub) or you just use email—the latter is far easier.

hawkrives:
I want to add a link to the email that creates a github issue for us in one click: https://eric.blog/2016/01/08/prefilling-github-issues/

For example, https://github.com/StoDevX/AAO-React-Native/issues/new?&labels=data/hours&title=[building]%20Suggestion%20for%20&body=This%20is%20a%20prefilled%20issue

That way we can still easily collect reports into our issue tracker.

drewvolz:
Including a prefilled link for quickly creating the issue sounds like a great idea. Thanks for finding that.

elijahverdoorn:
I'd like to be added to the alias.

elijahverdoorn (approved):

source/views/building-hours/report/submit.js
```diff
@@ -20,22 +20,44 @@ export function submitReport(current: BuildingType, suggestion: BuildingType) {
   )
 }
 
-function makeIssueBody(before: string, after: string): string {
+
+function makeEmailBody(before: string, after: string, title: string): string {
   return dedent`
+    Hi! Thanks for letting us know about an hour change.
```
drewvolz: Maybe substitute "hour" with "schedule"? Because they could change more than just hours.

drewvolz: Make this relevant as it is now shown as outdated...

PR #1306, opened 2017-07-21
Enable Nightlies

hawkrives:
This _should_ allow @drewvolz and I (and others at their behest) to use new builds of master on a nightly basis.

That's … the primary way that we get annoyed by things and fix them. And it's been turned off ever since we switched away from Hockeyapp.

This should turn it back on; we'll know tonight.

I think I have it set up not to distribute these builds to the normal beta teams, hopefully, so you all won't get spammed with notifications.


drewvolz (approved):
Thanks @hawkrives!

PR #1312, opened 2017-07-23
remove dedent from building-hours/submit

hawkrives:
Since I can't get it to behave, and since it's the reason we keep losing important whitespace… it's got to go.

My attempts to make it work: https://runkit.com/hawkrives/building-stringification


drewvolz (approved):

hawkrives:
And we mananged to miss the merge window for today, so we'll have to wait for tomorrow's build.

PR #1336, opened 2017-08-02
Clean up fastlane and travis a bit

hawkrives:
- I removed the rest of the old "commit and push back to the branch" experiment
- Fastlane just added a `setup_travis` action that replaces our `ci_keychains` lane, so I switched to that
- I moved the shell script that sets up an Android emulator into fastlane
- I updated Fastlane to 2.50.1
- I simplified the `:build` lanes for both iOS and Android, deduplicating the structure a bit

---

And then, the biggest thing I did: I added a new `propagate_version` lane that's automatically run as part of the build process.

It takes the version from package.json and writes it into the gradle and Xcode config files, thereby allowing us to never need to update those version numbers.

This will allow us to release a new version entirely from GitHub, as we'll be able to just update the version in package.json and make a new release, without needing any command-line tools.

So, our workflow goes from:

```shell
$ npm version 2.4.0-beta.11
> changing package.json
> changing Info.plist
> changing app/build.gradle
> git commit
> git tag v2.4.0-beta.11
$ git push --follow-tags
```

(where `$` indicates a prompt, and `>` indicates a command that's run automatically.)

which isn't bad, but it needs a computer. It also doesn't lend itself to letting us set the info dynamically at build time, which we need for the Google Play nightlies.

The new workflow (on a computer) is this:

```shell
$ npm version 2.4.0-beta.11
> changing package.json
> git commit
> git tag v2.4.0-beta.11
$ git push --follow-tags
```

Then, Fastlane will automatically update the versions in build.gradle and Info.plist at build time for us.

This also allows us to release a new beta entirely from GitHub easily:

1. Update the package.json version
2. Make the commit message "v2.4.0-beta.11"
3. Make a new release, called "2.4.0-beta.11"

---

TL;DR: no necessary workflow changes; just +79/-202 lines in the build config.


drewvolz (approved):
Excellent

Issue #1343, opened 2017-08-03
Prevent duplicate sections in student work list


drewvolz:
We need to group items into case-agnostic sections. So we either need to toUpper or toLower the data and the section header comparison. This is currently visible in the student work list with one item which has a lowercase header.

The problem is illustrated below:

**o**n-campus non-work study
**O**n-campus non-work study

This would create two different sections for the same group.

hawkrives:
so we actually do that

[`views/student-orgs/list.js#L104`](https://github.com/StoDevX/AAO-React-Native/blob/master/source/views/student-orgs/list.js#L104):
```js
          $groupableName: head(startCase(sortableName)),
```

https://lodash.com/docs/#startCase

hawkrives:
but good catch

drewvolz:
You are referencing Student Orgs, and I am referencing Student Work.

hawkrives:
oops

PR #1344, opened 2017-08-03
Title-cased job types

hawkrives:
Turns 'on-campus non-work study' into 'On-Campus Non-Work Study'

I also updated the job-type sorting algorithm – it turns out that, for our data, we really just want to sort the categories _backwards_ - that is, "On-Campus Work Study" should come before "Off-Campus Work Study", and the "Work Studies" should come before the "Summer Employments".

Closes https://github.com/StoDevX/AAO-React-Native/issues/1343.

### DIFF
--- a/source/views/contacts/data.js
+++ b/source/views/contacts/data.js
@@ -1,7 +1,6 @@
 // @flow
 import {data} from '../../../docs/contact-info.json'
 import type {CardType} from './types'
-
 ;(data: Array<CardType>)
 
 export {data}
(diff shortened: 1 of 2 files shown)


hawkrives:
So what we've learned here is that Hawken can't read. 

I'll fix those problems tomorrow.

drewvolz:
I think I fixed them up for you. Let me know if I made any wild assumptions.

drewvolz (approved):

hawkrives:
I'm going to have to look through it to figure out what was assuming that data would be an object instead of an array, but that looks like you got them.

hawkrives:
Soooo the type of `data` was annotated wrong - the api returns an array of jobs, not an object.

drewvolz:
Great... my bad with assuming the type annotation was correct. Thanks for all the work!

PR #1347, opened 2017-08-04
Update webcam URLs for new backend

hawkrives:
Broadcast Media just finished a switchover from StreamGuys to AWS. This will save them a buttload of money, but it causes a bit of inconvenience for us.

I'll be cherry-picking this commit to the v2.2 branch and submitting v2.2.4 later today.

Closes https://github.com/StoDevX/AAO-React-Native/issues/1345

PR #1348, opened 2017-08-04
update ksto stream url

hawkrives:
related to https://github.com/StoDevX/AAO-React-Native/pull/1347

I'm not sure where the uplink status moved to.


drewvolz (changes requested):
The uplink checking might be doomed. It's not working right now, so we should take that out.

hawkrives:
okay, but that should be another PR, no?

drewvolz (approved):
I feel that it's related enough? But if you insist...

PR #1351, opened 2017-08-04
Remove uplink status check from KSTO

drewvolz:
This removes the uplink checking that was provided by streamguys. I haven't tested it.


hawkrives (approved):
LGTM on skimming it

PR #1354, opened 2017-08-05
Add special-case tel: and mailto: handling

hawkrives:
With this patch, when someone taps a `tel:` or `mailto:` protocol link, SafariWebView won't crash because the RN Linking library will handle it.

This will let us use phone numbers and email addresses in the new Markdown components.

### DIFF
--- a/source/views/components/__tests__/open-url.test.js
+++ b/source/views/components/__tests__/open-url.test.js
@@ -1,0 +1,25 @@
+/* eslint-env jest */
+
+import {canOpenUrl} from '../open-url'
+
+describe('canOpenUrl', () => {
+    test('opens http:// links', () => {
+        expect(canOpenUrl('http://google.com')).toBe(true)
+    })
+    test('opens https:// links', () => {
+        expect(canOpenUrl('https://google.com')).toBe(true)
+    })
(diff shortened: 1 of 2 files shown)


drewvolz:
I am testing this out. I'm seeing that your protocol-relative regex url fails to catch `mailto` and `tel` within the string. I've made changes locally to remove the trailing `//` which catches protocol-relative urls, which gets us a littler farther.

`tel` will prompt twice (as we expect it to, I think) but `mailto` has a different issue where it blows up due to "NSMutableDictionary cannot be converted to NSString"

hawkrives:
I will double-check this both locally and on device and accommodate different protocols properly - I forgot that // is part of http

hawkrives:
My new approach doesn't assume that protocols have a `//` – instead, it requires that they have a `:`.

While this seems like it would prevent you from opening protocol-relative URLs, remember that this regex is just used to _bypass_ the normal WebView opening procedure, so if someone linked to `//google.com`, it would still be passed to SFWebView/CustomTabs. I don't know if they handle those, but that's no different than before.

drewvolz (approved):
👍 :shipit: confirmed, this is fixed.

PR #1357, opened 2017-08-05
Change Contacts over to Markdown

hawkrives:
Before | After
--- | ---
screen shot 2017-08-05 at 4 02 22 pm | screen shot 2017-08-05 at 3 56 00 pm

I mostly have the red bold text in here to demonstrate how to add per-view styles to the Markdown view.

Closes #1297.

PR #1358, opened 2017-08-05
Move the Contacts data to markdown

hawkrives:
As the title says.

Part of https://github.com/StoDevX/AAO-React-Native/pull/1357.


hawkrives:
actually this isn't too bad. I'll split out the dictionary PRs, but I'll just stick this into the view update one.

PR #1360, opened 2017-08-05
Port Dictionary's detail view to Markdown

hawkrives:
Before | After
--- | ---
screen shot 2017-08-05 at 5 11 21 pm | screen shot 2017-08-05 at 5 10 59 pm

PR #1361, opened 2017-08-05
Dictionary data update

hawkrives:
As part of the Markdown view, I realized that we needed the ability to determine linebreaks.

Before, we just kinda ignored them, but now I've switched to a Github-like style, where a linebreak in the source = a linebreak on screen.

And, if you want a new paragraph, just add a blank line.

Part of #1360.


drewvolz (approved):
👍 :shipit:

PR #1362, opened 2017-08-06
port Calendar to SectionList

hawkrives:
No visual differences on the ListView; empty calendars need some work.

screen shot 2017-08-05 at 9 30 31 pm

---

You will notice two comments along these lines: `// the proper return type is $ReadOnlyArray<{title: string, data: $ReadOnlyArray<EventType>}>`.

That's because I cannot for the _life_ of me convince Flow and SectionList to play nice together.

It's complaining about things like that I didn't annotate the whole object that's being passed in, but I would have expected that to validate as a subset – "I just need something that conforms to this shape" – but whatever.

Except, I went to the [Flow docs](https://flow.org/en/docs/lang/width-subtyping/): 

> "It’s safe to use an object with “extra” properties in a position that is annotated with a specific set of properties."

OK, thanks, flow docs.

For now, I've just annotated those as `any`.

### DIFF
--- a/source/views/calendar/event-list.js
+++ b/source/views/calendar/event-list.js
@@ -5,19 +5,21 @@
  */
 
 import React from 'react'
-import {StyleSheet} from 'react-native'
+import {StyleSheet, SectionList} from 'react-native'
 import * as c from '../components/colors'
-import SimpleListView from '../components/listview'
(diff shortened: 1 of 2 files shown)


drewvolz (approved):
👍 :shipit:

PR #1364, opened 2017-08-06
Port Menus to SectionList

hawkrives:
> this fixes the conditional-star-showing! with ListView, we had to always show the stars on the menu, even when we were only showing specials, because weird things would happen when we toggled them conditionally. But it works now!

Before | After
---- | ----
screen shot 2017-08-05 at 10 03 08 pm  | screen shot 2017-08-05 at 10 02 57 pm


drewvolz (approved):
👍 :shipit:

Issue #1299, opened 2017-07-19
Transportation "Open Map" is broken (Android, 2.4β11)


rye:
See [the error in BugSnag](https://app.bugsnag.com/gobbldygook/all-about-olaf-1/errors/596edfe4a83b40001b21319b) for crash logs.  Seems to be something up with `react-native-maps`.

hawkrives:
![img_0408](https://user-images.githubusercontent.com/464441/28367497-ccfced56-6c56-11e7-8acc-d0ad79519abc.PNG)

Another view of the error.

drewvolz:
Something something update google play services in `build.gradle`? 

https://github.com/airbnb/react-native-maps/issues/1051
https://github.com/airbnb/react-native-maps/issues/1408
https://github.com/airbnb/react-native-maps/issues/1028

hawkrives:
The maps are disabled, yeah? (Yeah.)

I'm going to push this back to v2.6; it's a bug, but I'm fine shipping without fixing it since there's no way to trigger it that I know of.

hawkrives:
I'm going to close this; we'll run into it again when we actually enable a map somewhere.

PR #1371, opened 2017-08-06
Pin Google Play Services version for rn-maps

hawkrives:
Part of #1299 

I want to run this locally before we merge it (or someone else can, not picky).


hawkrives:
Hmm. So I took the solution from https://github.com/airbnb/react-native-maps/issues/1028, but the build doesn't like it.

PR #1374, opened 2017-08-07
Trim the log that's sent to testflight to a max of 4,000 chars

hawkrives:
TestFlight will reject changelogs over 4,000 chars, so this PR will truncate our new git graph to a max of 4,000 chars.

I used https://stackoverflow.com/a/39907261/2347774 as a reference, so I think it's actually fancy and won't truncate within a word.


drewvolz (approved):
👍 :shipit: we'll see if it works...

Issue #1376, opened 2017-08-07
KSTO view crashes on iOS


hawkrives:
To reproduce: open the Ksto view on iOS, wait about 10 seconds.

Something in here is requesting data over the network, as I can see the network spinner spinning.

drewvolz:
I'd say the network request is a correlation... just because we ripped out the uplink checking. I'm clueless as to what else could be causing the crash... something in `react-native-video`?

hawkrives:
… can we just embed a WebView that iframes the player? would that work? would it play in the background?

drewvolz:
`react-native-video` merged the fix for this a minute ago. Let's hope for a version bump, or else we'll fork it until then.

PR #1377, opened 2017-08-07
only "render" the Video element when playing

hawkrives:
Bandaid for https://github.com/StoDevX/AAO-React-Native/issues/1376

to be tested


drewvolz:
Testing this shows that the app will not crash if the view is loaded, just when played... which inserts the `Video` element... so the original problem.

hawkrives:
Yeah but this works nicely as a stopgap

And it'd let me actually test the webcams :)

hawkrives:
I merged this as a stopgap while we fix the actual issue.

PR #1392, opened 2017-08-08
Add initial region logic to transportation map

### LINKED ISSUE #1187: Bus map sometimes opens completely zoomed out
I thought I'd fixed it, but I'm seeing it on my iPad at least.

Screenshot to come later.

img_0003

drewvolz:
This PR closes #1187 where the bus map would zoom out unnecessarily.

### DIFF
--- a/source/views/transportation/bus/map.js
+++ b/source/views/transportation/bus/map.js
@@ -32,7 +32,7 @@ export class BusMapView extends React.PureComponent {
   state = {
     intervalId: 0,
     now: moment.tz(TIMEZONE),
-    initialRegion: {
+    region: {
       latitude: 44.44946671480875,
       latitudeDelta: 0.06175530810822494,
       longitude: -93.17014753996669,
--- a/source/views/transportation/bus/map.js
+++ b/source/views/transportation/bus/map.js
@@ -61,20 +61,18 @@ export class BusMapView extends React.PureComponent {
     },
   }
 
-  onRegionChangeComplete = (region: {
+  onRegionChangeComplete = (newRegion: {
     latitude: number,
     latitudeDelta: number,
     longitude: number,
     longitudeDelta: number,
   }) => {
     this.setState(state => {
-      const newRegion = state.region
-
-      if (newRegion && isEqual(newRegion, region)) {
-        return state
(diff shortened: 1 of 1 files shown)


hawkrives:
This works?!?!

source/views/transportation/bus/map.js
```diff
@@ -31,7 +32,7 @@ export class BusMapView extends React.PureComponent {
   state = {
     intervalId: 0,
     now: moment.tz(TIMEZONE),
-    region: {
+    initialRegion: {
```
hawkrives: Why not just leave this named as "region"?

drewvolz: Because we use a new region named the same thing in `onRegionChangeComplete`?

drewvolz: Oh. I see your point.

source/views/transportation/bus/map.js
```diff
@@ -60,13 +61,21 @@ export class BusMapView extends React.PureComponent {
...
-  onRegionChange = (region: {
+  onRegionChangeComplete = (region: {
     latitude: number,
     latitudeDelta: number,
     longitude: number,
     longitudeDelta: number,
   }) => {
-    this.setState(() => ({region}))
+    this.setState(state => {
+      const initialRegion = state.initialRegion
```
hawkrives: Don't forget to rename it in here too. I might call the parameter "newRegion" so that you can keep this block as "region".

source/views/transportation/bus/map.js
```diff
@@ -68,13 +68,13 @@ export class BusMapView extends React.PureComponent {
...
-      const initialRegion = state.initialRegion
+      const newRegion = state.region
 
-      if (initialRegion && isEqual(initialRegion, region)) {
+      if (newRegion && isEqual(newRegion, region)) {
         return state
       }
 
-      return {initialRegion: region}
+      return {newRegion: region}
```
hawkrives: Flip these, or else you'll be setting "this.state.newRegion". 

I think you want to call the input parameter to the onRegionChangeDomplete function "newRegion", and the one you get out of state "region".

hawkrives (approved):

PR #1401, opened 2017-08-08
Inline part of the `script` section back into .travis.yml

hawkrives:
This is the start of my push to do more analysis of our builds over time.

Somehow.

It's also that I didn't like having the single-line blocks of code in the script.sh file 😄


hawkrives:
Oh yes, I also removed the "bundle:ios" and "bundle:android" runs from the JS build – the Android and iOS builds already bundle as part of their build process, so it seems to me like it just takes extra time for no real gain.

hawkrives:
This brought the JS build back down to 6 minutes.

PR #1402, opened 2017-08-08
Add npm lockfile

hawkrives:
I know we've tried this a few times before, but I think it's finally ready!

npm@5.3.0 will respect package.json over the package-lock.json if they differ, which means that our greenkeeper updates will actually build with the new packages now.


hawkrives:
After the first builds finish, I want to make a second build to see how much faster it is.

hawkrives:
And then probably another build with caching of `node_modules/` enabled, so we can see what that does to the times, because with a lockfile I'm less concerned about it getting outdated.

hawkrives:
Down to around 6-7 minutes for JS builds. Now to test caching.

PR #1407, opened 2017-08-09
force node 8 on ios builds

hawkrives:
maybe?


hawkrives:
can't believe it worked on the first try

PR #1413, opened 2017-08-10
Loading messages in data

elijahverdoorn:
Moves the loading messages for menus out of code and into data files, where they belong.

### DIFF
--- a/data/menus/bonapp-loading.yaml
+++ b/data/menus/bonapp-loading.yaml
@@ -1,0 +1,8 @@
+loadingMessage:
+  - 'Hunting Ferndale Turkey…'
+  - 'Tracking wild vegan burgers…'
+  - '"Cooking" some lutefisk…'
+  - 'Finding more mugs…'
+  - 'Waiting for omlets…'
+  - 'Putting out more cookies…'
+
--- a/data/menus/bonapp-loading.yaml
+++ b/data/menus/bonapp-loading.yaml
@@ -1,4 +1,4 @@
-loadingMessage:
+bonappLoading:
   - 'Hunting Ferndale Turkey…'
   - 'Tracking wild vegan burgers…'
   - '"Cooking" some lutefisk…'
--- a/data/menus/cage-loading.yaml
+++ b/data/menus/cage-loading.yaml
@@ -1,0 +1,6 @@
+loadingMessage:
+  - 'Checking for vegan cookies…'
+  - 'Serving up some shakes…'
+  - 'Waiting for menu screens to change…'
+  - 'Frying chicken…'
+  - 'Brewing coffee…'
--- a/data/menus/cage-loading.yaml
+++ b/data/menus/cage-loading.yaml
@@ -1,4 +1,4 @@
-loadingMessage:
+cageLoading:
   - 'Checking for vegan cookies…'
   - 'Serving up some shakes…'
   - 'Waiting for menu screens to change…'
--- a/data/menus/pause-loading.yaml
+++ b/data/menus/pause-loading.yaml
@@ -1,0 +1,6 @@
+loadingMessage:
+  - 'Mixing up a shake…'
+  - 'Spinning up pizzas…'
+  - 'Turning up the music…'
+  - 'Putting ice cream on the cookies…'
+  - 'Fixing the oven…'
--- a/data/menus/pause-loading.yaml
+++ b/data/menus/pause-loading.yaml
@@ -1,4 +1,4 @@
-loadingMessage:
+pauseLoading:
   - 'Mixing up a shake…'
   - 'Spinning up pizzas…'
   - 'Turning up the music…'
--- a/source/views/menus/index.js
+++ b/source/views/menus/index.js
@@ -3,7 +3,6 @@
  * All About Olaf
  * Menus page
  */
-
 import React from 'react'
 import {StackNavigator} from 'react-navigation'
 import {TabNavigator} from '../components/tabbed-view'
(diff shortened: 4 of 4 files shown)


.gitignore
```diff
@@ -69,3 +69,6 @@ docs/
 # android keystores
 android/app/signing.properties
 *.keystore
+
+# vim temp files
+*.swp
```
elijahverdoorn: I don't think I'm the only vim user who contributes to this project.

hawkrives: you think @drewvolz uses vim?

hawkrives:
I would rather not store the loading messages away from the components. They are tied to a particular instance of the component, in my mind – see also the Carleton menus, which have their own loading messages.

.gitignore
drewvolz: @hawkrives where's your evangelism for spacemacs

drewvolz: @hawkrives are you thinking I do not use vim?

drewvolz: I ask that this be cherry-picked into a separate PR, please.

elijahverdoorn: That's fine, I'll move it.

hawkrives: @drewvolz, 7/23/2017

> thanks though. I might ask out of curiosity sometime in the future but I really like text editing w/o all the shortcuts. It is less efficient, yes.

elijahverdoorn:
While I agree that they're tied to the component, I don't think that having lists of strings inline in the code is the right way to store them. They have a fallback "Loading..." message, which belongs inline, but I think it makes more sense to store separately since they don't have any impact on the logic of the view. I see the `docs/` folder as a proxy for android's `strings.xml`, which is a globally-accessible way to get a string resource, and this move brings us more in line with that.

If we really don't like this, and want to go the other way, maybe we could move the code-inline list into the individual `component.js` files, so that they're not all in the `index.js`. Doing that would  bring it closer to the semantic map that you're suggesting.

.gitignore
hawkrives: so yes, I am thinking that you do not use vim

hawkrives:
I agree that the messages a distinct from the _reusable_ component, but not from any specific instance of the component.

We have 7 instances of the MenuView, I believe, and each one can provide a custom list of loading messages to the underlying BonAppMenuView/GithubMenuView.

In the current architecture, the lists of strings are at the index.js level because that's where the individual components are created as tabs / list items. I don't see a need to make individual component files for each instance of the menu (StavHallMenu, TheCageMenu, SaylesHillMenu), which I read your last paragraph as proposing.

Apologies if I'm misrepresenting your argument.

.gitignore
drewvolz: @hawkrives that was in reference to you offering instructions on how you got spacemacs working. I make plenty use of emacs keybindings, and I do end up using vim for the majority of things.

elijahverdoorn:
That last paragraph didn't make much sense. Sorry, I was typing from phone while trying to remember what our code looked like. You can mostly ignore it.

I think the crux of my argument is that I think that static data doesn't belong in the JS code. The YAML is easier to edit and more modular, and fits my mental map of having all our data (strings included) outside the code. While it's not on my immediate list, if we were to ever add support for more languages than english, we'd want to move all the strings out of the code and into data files so we could easily port the whole app to the user's selected language. Keeping these principals in mind, moving all the data away from the code is a good move in my eyes; if everyone else feels differently it's not the end of the world.

hawkrives:
While I agree with the x-language idea in _principle_, uh, wait. I've never talked to you about CARLS, have I?

Anyway. I guess I have a mental divide between "significant" and "insignificant" data – I put the stuff that is in /data, in /data, so that it can be updated remotely. 

If we go down the route of <abbr title="internationalization">l18n</abbr>, then I would like to come up with a solution tailored to string changing, rather than shoe-horning it on top of our remote-data-loading system.

.gitignore
hawkrives: also spacemacs, being emacs, doesn't use vim swap files

elijahverdoorn:
No, I don't know anything about CARLS.

I agree that these strings are insignificant, but why wouldn't we give ourselves the option of updating remotely? We have the systems already built! I don't think we'll ever want to, but more options are better IMO.

Yes, proper localization would be best with a different system. Perhaps this is me crying for some global refactor and implementation of a string system similar to the `strings.xml` that I mentioned above. I'd be willing to work on something like that if there is interest. Discussion about such a system should move to an issue.

hawkrives:
> [...] why wouldn't we give ourselves the option of updating remotely?

Because our current infrastructure for updating remotely is ad-hoc, doesn't cache, and very WET.

> I'd be willing to work on something like that if there is interest.

We've got so many other things to work on, though! @drewvolz and I really need to be better about filing issues for them, too.

As of today, we have two new APIs on campus: /multimedia and /directory. There's that Papercut API I found recently. Migrating the dictionary to update remotely. Caching remote content locally so we don't fetch it on every page open. I'm building a new /apps/olecard API, slowly. There's an API for a live Northfield Lines bus tracker on their website. Investigating how to let people publish their own JS-only modules, and load them into the app. Make the app work for Carleton (all me). Pull transportation times from Github. Convince the registrar to let me build a read-only student data api for SIS. Make KSTO work. Pull Important Contacts from github. Make the app work better in landscape / on big screens. Allow people to submit Facilities work orders from their phones, if Facilities would like.

---

I won't tell you not to work on l18n, but given that our market is very specifically St. Olaf, … actually, translation could be nifty for language learners. hmm.

hmm.

If you want to work on l18n, go for it.

I think I got off track there somewhere. Anyway.

elijahverdoorn:
Haha yeah, there are a million good things that I could work on, it's hard to pick which one is most important/interesting. I don't want to dwell on this PR, it's a rather small change overall.

Vote by reacting to this comment. 👍 for merge, 👎 for deny. No hard feelings.

elijahverdoorn:
Deny has it 2-1. Closing PR.

PR #1427, opened 2017-08-14
Handle blank contacts in org listings

### LINKED ISSUE #1388: Render org member email if the name is missing
We should try to render the email address _if_ the position and name are blank.

What we render are member names and positions. It is possible to have members with no names and positions, but their email address is listed... so we end up with blank rows.

drewvolz:
This PR fixes issues in:

- **iOS**: the contact info for a Student Org member was completely blank if no first/last name.
- **Android**: the contact info for a Student Org member contained extra whitespace if no first/last name.

Closes #1388 

~ | Before | After |
-- | -- | -- |
iOS | truth-blank | truth-filled
Android | android-truth-blank | android-truth-filled

### DIFF
--- a/source/views/student-orgs/detail.android.js
+++ b/source/views/student-orgs/detail.android.js
@@ -5,11 +5,11 @@ import moment from 'moment'
 import {Card} from '../components/card'
 import * as c from '../components/colors'
 import type {StudentOrgType} from './types'
-import type {ContactPersonType} from './types'
 import type {TopLevelViewPropsType} from '../types'
 import Communications from 'react-native-communications'
 import openUrl from '../components/open-url'
-import cleanOrg from './clean-org'
+import {cleanOrg} from './util'
+import {showNameOrEmail} from './util'
 
 const styles = StyleSheet.create({
   name: {
@@ -67,14 +67,6 @@ export class StudentOrgsDetailView extends React.Component {
     navigation: {state: {params: {org: StudentOrgType}}},
   }
 
-  showNameOrEmail = ({c}: {c: ContactPersonType}) => {
-    if (!c.firstName.trim() && !c.lastName.trim()) {
-      return `${c.email}`
-    }
-
-    return `${c.firstName} ${c.lastName} (${c.email})`
-  }
-
   // Using Communications because `mailTo` complains about
   // the lack of an available Activity...
   openEmail = (email: string, org: string) => {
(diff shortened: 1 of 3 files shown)


hawkrives (changes requested):
- Please extract this helper function into the helper file. If there is no helper file here, please create one.
- When you extract the function, please change it to accept one parameter, as follows: `(c: ContactPersonType)`. That is, remove the object-as-first-parameter dance.

drewvolz:
> Please extract this helper function into the helper file. If there is no helper file here, please create one.

[`clean-org.js`](https://github.com/StoDevX/AAO-React-Native/blob/master/source/views/student-orgs/clean-org.js) exists but I don't think it would belong in there... do you? I think it might warrant another helper file.

hawkrives:
@drewvolz Instead of making another file, I might just rename `clean-org.js` to `util.js` or something, and add the helper function in there.

Since `clean-org.js` has only one function already.

drewvolz:
I think this is ready to go.

source/views/student-orgs/detail.android.js
```diff
@@ -5,11 +5,11 @@ import moment from 'moment'
 import {Card} from '../components/card'
 import * as c from '../components/colors'
 import type {StudentOrgType} from './types'
-import type {ContactPersonType} from './types'
 import type {TopLevelViewPropsType} from '../types'
 import Communications from 'react-native-communications'
 import openUrl from '../components/open-url'
-import cleanOrg from './clean-org'
+import {cleanOrg} from './util'
+import {showNameOrEmail} from './util'
```
hawkrives: Please import both of these at once.

source/views/student-orgs/detail.ios.js
```diff
@@ -5,10 +5,10 @@ import moment from 'moment'
 import {Cell, Section, TableView} from 'react-native-tableview-simple'
 import * as c from '../components/colors'
 import type {StudentOrgType} from './types'
-import type {ContactPersonType} from './types'
 import type {TopLevelViewPropsType} from '../types'
 import openUrl from '../components/open-url'
-import cleanOrg from './clean-org'
+import {cleanOrg} from './util'
+import {showNameOrEmail} from './util'
```
hawkrives: Please import both of these at once.

hawkrives (changes requested):

hawkrives:
One last thing! It looks good!

drewvolz:
Why don't we import from `types` like that?

hawkrives:
@drewvolz you mean here?

```js
import type {StudentOrgType} from './types'
import type {TopLevelViewPropsType} from '../types'
```

Because they're importing from two different files.

If you see anywhere where there are two import statements targeting the same file, please merge them into one :)

drewvolz:
Good catch. I’ve now merged the import statements you mentioned earlier.

hawkrives (approved):

PR #1438, opened 2017-08-18
Nix insecure streamguys + wp.stolaf exceptions

drewvolz:
We're no longer using streamguys and our links from wp.stolaf are now https


hawkrives:
Links to pages.stolaf will be mandatory https in about 10 minutes as well, but we don't have an exception for that.

hawkrives:
Hopefully the Mess will migrate week after next, and Oleville should become https the week of the 28th.

hawkrives:
Pages is now https-only.

hawkrives (approved):

Issue #1436, opened 2017-08-17
No bundle URL present in build 3752


drewvolz:
2.4.0 (3752) crashes from launch.

> RCTFatalException: No bundle URL present…jsbundle file in your application bundle in mh_execute_header

hawkrives:
I guess step one is to do a fresh build of master locally. [3752](https://travis-ci.org/StoDevX/AAO-React-Native/builds/265408087) was built from https://github.com/StoDevX/AAO-React-Native/commit/be61e191b0c2dccea1747d295c0088704d9de5db.

drewvolz:
I've now successfully built  [3752](https://travis-ci.org/StoDevX/AAO-React-Native/builds/265408087) from https://github.com/StoDevX/AAO-React-Native/commit/be61e191b0c2dccea1747d295c0088704d9de5db for iOS 10.3 on both Release and Development locally.

hawkrives:
My next step would be to check the Travis logs for 3752.3 and see if there's anything weird in it. 

Because there were no substantive changes from https://github.com/StoDevX/AAO-React-Native/compare/9679c0d8ee4cfa7030e83d9e47dcbab1ff615e67...be61e191b0c2dccea1747d295c0088704d9de5db.

hawkrives:
Here's a clue: 

This is a [build log for CARLS](https://travis-ci.org/carls-app/carls/jobs/266580604#L3275-L3282) (I truly believe this is the exact same problem that AAO is having).

Take a look at the timestamps:

```
[00:15:41]: ▸ Generating 'CARLS.app.dSYM'
[00:15:41]: ▸ Running script 'Bundle Data Files'
[00:15:45]: ▸ Running script 'Bundle React Native code and images'
[00:16:00]: ▸ the transform cache was reset.
[00:16:02]: ▸ Touching CARLS.app
[00:16:02]: ▸ Signing /Users/travis/Library/Developer/Xcode/DerivedData/CARLS-grsdkuczesfovtfkfgcwidnmnsid/Build/Intermediates/ArchiveIntermediates/CARLS/InstallationBuildProductsLocation/Applications/CARLS.app
[00:16:03]: ▸ Touching CARLS.app.dSYM
[00:16:04]: ▸ Archive Succeeded
```

[...]

PR #1449, opened 2017-08-21
Downgrade babel-preset-react-native

hawkrives:
reverts fdfc07e2b9ae829fa48505c150cd087a716edd26

an attempt at fixing #1436 

because I am really kinda at a loss, but it seems like an update to how js is compiled might silently break the bundling? maybe?


hawkrives:
yolo. I'm going to merge this and make a nightly – if it doesn't fix it, I'll revert this in the morning.

PR #1450, opened 2017-08-21
Update fastlane's login information

hawkrives:
I wonder how companies do this? Just switch out the one person's credentials when they leave?


hawkrives:
> The basic requirement is to have one code signing identity shared across your team. The easiest way to do that is to create a new Apple ID for the team (e.g. ios-dev@company.com) and use that from now on.

Ok. That's from <codesigning.guide>

PR #1469, opened 2017-08-26
Important contacts refresh

drewvolz:
This PR introduces a refresh to the important contacts view.

**Main:**
* Simplified list/detail views
* Group contacts by category

**Misc:**
* Whitespace/character changes to YAML
* Filename changes

~ | List | Detail
--|--|--
iOS | ios-list | ios-detail
Android | android-list | android-detail


drewvolz:
~What am I missing... `npm run validate-data -- --quiet` shows that "category" is an additional prop.~

Ah. Forgot to update the schema.

source/views/contacts/contact-list.js
```diff
@@ -0,0 +1,77 @@
...
+      contact,
+    })
+  }
+
+  renderSeparator = () => <ListSeparator />
+
+  renderSectionHeader = ({section: {title}}: any) =>
+    <ListSectionHeader title={title} spacing={{left: 10}} />
+
+  renderFooter = () =>
```
hawkrives: Please extract this into a constant outside of the class - it doesn't need to be called as a function. 

Just assign the ListFooter to a constant called "footer" or something. It should be in this file, as it's similar in intent to the styles object.

source/views/contacts/contact-list.js
```diff
@@ -0,0 +1,77 @@
...
+import {data} from './data'
+import groupBy from 'lodash/groupBy'
+import toPairs from 'lodash/toPairs'
+import * as c from '../components/colors'
+import type {ContactType} from './types'
+
+const AAO_URL = 'https://github.com/StoDevX/AAO-React-Native/issues/new'
+
+const groupContacts = (contacts: ContactType[]) => {
+  const grouped = groupBy(contacts, b => b.category || 'Other')
```
hawkrives: We don't need the `||` here, as we're guaranteed through the schema that our objects will have a category.

source/views/contacts/contact-list.js
```diff
@@ -0,0 +1,77 @@
+/**
+ * @flow
+ * All About Olaf
+ * iOS Contact page
```
hawkrives: Remove the word iOS?

hawkrives (changes requested):
Overall, nice work! Really nice!

drewvolz:
Changes addressed. Now to wait on the build again.

hawkrives (approved):

PR #1490, opened 2017-08-31
Remote news story CSS

drewvolz:
* Convert `css` located in `data` into `json` in `docs`
* Use local stylesheet if in development
* Use remote stylesheet if in production
* Rule for `figure` width
* Rule for `word-wrap` on long strings

### DIFF
--- a/source/views/news/news-item.js
+++ b/source/views/news/news-item.js
@@ -28,43 +28,18 @@ export default function NewsItem(props: {
 }) {
   const {story, embedFeaturedImage} = props.navigation.state.params
 
+  // Use local or remote styles
+  let newsCSS = ''
+  if (process.env.NODE_ENV === 'development') {
+    let css = require('../../../docs/news-styles.json')
+    newsCSS = `<Style>${css}</Style>`
+  } else {
+    newsCSS =
+      '<link rel="stylesheet" href="https://stodevx.github.io/aao-react-native/news-styles.css">'
+  }
+
   const content = `
-    <style>
-      body {
-        font-family: -apple-system, Roboto, sans-serif;
-        font-size: 16px;
-        margin: 0;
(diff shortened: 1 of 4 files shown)


drewvolz:
Question: is it worth pulling in the stylesheet on each story, or should the content be fetched once in the news list and passed as a prop to the story view?

... so load it for each view, or once on fetch?

hawkrives:
Once per view. Just insert the `<link>` element and let the WebView cache if it wants.

hawkrives (changes requested):
This should be a `.css` file within `data/`, and be transformed to a `json` file in `docs/`. You should never commit files to `docs/`.

source/views/news/news-item.js
```diff
@@ -28,43 +28,18 @@ export default function NewsItem(props: {
 }) {
   const {story, embedFeaturedImage} = props.navigation.state.params
 
+  // Use local or remote styles
+  let newsCSS = ''
+  if (process.env.NODE_ENV === 'development') {
+    let css = require('../../../docs/news-styles.json')
+    newsCSS = `<Style>${css}</Style>`
```
hawkrives: Lower-case `Style`. It's an html element.

drewvolz: oops

drewvolz:
> Once per view. Just insert the <link> element and let the WebView cache if it wants.

So the way it currently works is okay

drewvolz:
> This should be a .css file within `data/`, and be transformed to a `json` file in `docs/`

~Do you have any recommendations as to how we'd convert the `css` to `json`?~

I've found a way. I'll push shortly.

drewvolz:
Fixed!

hawkrives (approved):
I think this looks good to me!

PR #1491, opened 2017-09-01
Streaming events

### LINKED ISSUE #127: Streaming events view
It would be neat to have the feeds from [https://www.stolaf.edu/multimedia/](https://www.stolaf.edu/multimedia/) show up in the app. Luckily, there is an [RSS feed that we can make JSON](https://ajax.googleapis.com/ajax/services/feed/load?v=1.0&num=-1&q=https://www.stolaf.edu/multimedia/streams/rss/rss.cfm). Providing a list of these events which would load a view of the stream might make for a neat addition to the Streaming Media section of the app.

drewvolz:
This PR introduces a view to show a monthly list of upcoming streaming events. Tapping on a row reveals a detail web view that allows for the stream to be watched.

Closes #127.

~ | List | Detail
--|--|--
iOS | ios-list | ios-detail
Android  | android-list | android-detail


drewvolz:
Thoughts on the rows? I think they need some work on the title and detail information.

drewvolz:
@hawkrives -- lint is complaining, but I think that style _is_ being used.

hawkrives:
@drewvolz In list.js, yes, but not in row.js, where it's also defined.

hawkrives:
~@drewvolz I hereby formally request that you switch to the `recording` icon:~

@drewvolz What do you think about the `recording` icon, instead of the star?

screen shot 2017-08-31 at 8 45 12 pm

drewvolz:
The `recording` icon is interesting. I'll have to think about it.

hawkrives (approved):

hawkrives:
Doing the Travis dance due to #1506.

hawkrives:
I'm thinking that we'll merge this, play with it for a day or two on nightly / beta, then decide if we want to release it or let it bake a bit longer.

PR #1515, opened 2017-09-06
Add Holland Hall hours

hawkrives:
These will likely change in a week or two, as the building gets used.

I will update them when I get the new hours.

### DIFF
--- a/data/building-hours/8-hh.yaml
+++ b/data/building-hours/8-hh.yaml
@@ -1,5 +1,4 @@
 name: Holland Hall
-subtitle: Under Construction
 category: Academia
 
 schedule:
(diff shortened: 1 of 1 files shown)


drewvolz (approved):

hawkrives:
I couldn't come up with a better way to do the Closed-on-Saturday bit (because it's closed on Saturday for whatever reason), so I just left it out.

That led me to file #1516.

drewvolz:
Isn't adding something that's closed as trivial as omitting hours for that day?

drewvolz:
...and that's exactly what you did. So you want a way to _say_ it's closed instead of hiding that day.

hawkrives:
Yes. To be more explicit about a lack of hours on that day.

drewvolz (approved):

PR #1518, opened 2017-09-07
Add to device calendar

### LINKED ISSUE #1211: Create an "Add Event to [Device] Calendar" button
It would show on the event detail view, and add the event to your device's calendar.

drewvolz:
This PR introduces a button on the Calendar detail to add event details to the device calendar. It closes #1211.

Thanks to @hawkrives who linked `react-native-calendar-events` and helped work out the promise logic in `calendar-util`, so thank you very much for your help!

Here is a brief description of the states and steps of requesting device calendar permissions. We update the view in the card/cell footer to inform the user the result of pushing the button and allowing/denying permissions thru alerts.

Status | Description
--|--
Add | The user has the option to hit the "Add to calendar" button.
Push | The user pushes the button. Until we receive access, we'll request device calendar access. The status text will change to "Adding event to calendar..." until we receive further status information.
Done | The user gave permissions and the calendar event is now in their phone's calendar and the add button becomes disabled. The status text changes to "Event has been added to your calendar".
Denied | The user denied permission to access their calendar. The status text changes to "Could not add event to your calendar". If they hit "Add to calendar" again at this point, we will prompt them once again for device permissions.

~ | iOS | Android
--|--|--
Add | ios-1 | android-1
Push | ios-2 | android-2
Done | ios-3 | android-3
Denied | ios-4 | android-4

### DIFF
--- a/source/navigation.js
+++ b/source/navigation.js
@@ -50,12 +50,12 @@ const styles = StyleSheet.create({
 
 export const AppNavigator = StackNavigator(
   {
-    CalendarView: {screen: CalendarView},
     HomeView: {screen: HomeView},
     BuildingHoursDetailView: {screen: BuildingHoursDetailView},
     BuildingHoursView: {screen: BuildingHoursView},
     BuildingHoursProblemReportView: {screen: BuildingHoursProblemReportView},
     BuildingHoursScheduleEditorView: {screen: BuildingHoursScheduleEditorView},
+    CalendarView: {screen: CalendarView},
     ContactsView: {screen: ContactsView},
     ContactsDetailView: {screen: ContactsDetailView},
     CreditsView: {screen: CreditsView},
--- a/source/navigation.js
+++ b/source/navigation.js
@@ -50,12 +50,12 @@ const styles = StyleSheet.create({
 
 export const AppNavigator = StackNavigator(
   {
+    CalendarView: {screen: CalendarView},
     HomeView: {screen: HomeView},
     BuildingHoursDetailView: {screen: BuildingHoursDetailView},
     BuildingHoursView: {screen: BuildingHoursView},
     BuildingHoursProblemReportView: {screen: BuildingHoursProblemReportView},
     BuildingHoursScheduleEditorView: {screen: BuildingHoursScheduleEditorView},
-    CalendarView: {screen: CalendarView},
     ContactsView: {screen: ContactsView},
     ContactsDetailView: {screen: ContactsDetailView},
     CreditsView: {screen: CreditsView},
--- a/source/views/calendar/calendar-util.js
+++ b/source/views/calendar/calendar-util.js
@@ -2,11 +2,11 @@
 
 import {Platform, Alert, Linking} from 'react-native'
 import RNCalendarEvents from 'react-native-calendar-events'
-import type {EventType} from './types'
+import type {CleanedEventType} from './types'
 import bugsnag from '../../bugsnag'
 import {tracker} from '../../analytics'
 
-export function addToCalendar(event: EventType): Promise<boolean> {
+export function addToCalendar(event: CleanedEventType): Promise<boolean> {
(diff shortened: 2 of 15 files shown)


source/navigation.js
```diff
@@ -50,12 +50,12 @@ const styles = StyleSheet.create({
 
 export const AppNavigator = StackNavigator(
   {
+    CalendarView: {screen: CalendarView},
```
hawkrives: @drewvolz

source/views/calendar/calendar-util.js
```diff
@@ -0,0 +1,72 @@
...
+      return true
+    })
+    .then(status => {
+      if (!status) {
+        return false
+      }
+      return saveEventToCalendar(event)
+    })
+    .catch(err => {
+      alert(err)
```
hawkrives: This line won't work.

source/navigation.js
drewvolz: Whoops again.

source/views/calendar/calendar-util.js
drewvolz: Good point. Let's make that work.

source/views/calendar/event-detail.android.js
```diff
@@ -86,25 +90,76 @@ const shareItem = (event: EventType) => {
...
+      } else {
+        this.setState({
+          message: 'Could not add event to your calendar',
+          disabled: false,
+        })
+      }
+    })
+  }
+
+  CalendarButton = ({event}: {event: EventType}) => {
```
hawkrives: This should not be a class method; it should take the required state from this component as props instead

hawkrives: .. we should really cleanEvent before we open this screen. That way we have a clean event everywhere. I will assume here that we've done that. 

    const CalendarButton = ({message, disabled, onPress}) => {
    return (
      <Card footer={message} style={styles.card}>
        <ButtonCell
          title="Add to calendar"
          disabled={disabled}
          onPress={onPress}
        />
      </Card>
    )
    }

Then, in the parent, 

    onPressButton = () => this.addEvent(this.props.navigation.state.params.event)
    render() {
      …
      <CalendarButton onPress={this.onPressButton} message={this.state.message} disabled={this.state.disabled} />
      …
    }

source/views/calendar/event-detail.android.js
```diff
@@ -86,25 +90,76 @@ const shareItem = (event: EventType) => {
...
+
+  state: {
+    message: string,
+    disabled: boolean,
+  } = {
+    message: '',
+    disabled: false,
+  }
+
+  addPressed = async ({event}: {event: EventType}) => {
```
hawkrives: And this should be called "addEvent" imo

hawkrives (changes requested):

source/views/calendar/event-detail.android.js
hawkrives: This should also just take a single argument, not an object.

source/views/calendar/types.js
```diff
@@ -15,6 +15,7 @@ export type GoogleEventType = {
 type EmbeddedEventDetailType = {type: 'google', data: GoogleEventType}
 
 export type EventType = {
+  title?: string,
```
hawkrives: What is title? Isn't "summary" the title?

hawkrives: https://github.com/StoDevX/AAO-React-Native/blob/ea4871ac31ef6b67028f2d8151e3f1bafdb8ca2e/source/views/calendar/clean-event.js#L7

cleanEvent should return a CleanedEventType, which things should be updated to use. 

And this "title" property should be removed from EventType because it's not in EventType anymore.

drewvolz: Ah. Didn't understand it became a new type. I'll fix that up. Thanks.

hawkrives (approved):

PR #1536, opened 2017-09-10
update package-lock

hawkrives:
whee


hawkrives:
(I requested reviewers because since I turned on the "protect branches from admins" checkbox, I can't merge things without approval.) 

[… and I won't turn that off, since that resulted in about an hour of unintentional Carleton building hours in AAO last week.]

elijahverdoorn (approved):

PR #1537, opened 2017-09-10
Reorganize all app files

hawkrives:
This was originally a proof of concept that got a bit out of hand… and actually worked.

- Moves `components/` up under `source/`
- Makes a `source/init/` folder for "initialization files" – think bugsnag, GA, etc
- Dynamically loads homescreen buttons and screen navigation keys from the files

And then, the big change is that it changes what a "view" is.

We currently kinda abuse the notion of a "view" – we call the thing that is produced by tapping a homescreen button a view, but we have nowhere for the detail views (or the views housed under the tab bars) to live, so they just kinda wander around.

This branch introduces the concept of a "module," which is a discrete … view. Each tab in a tabbed view is a module (`OtherModesTransitView`, `StudentJobPostingsView`), as is each top-level view (those that are more than just a tab switcher, at least) – `BonAppMenuView`, `CalendarEventsListView`, `BusLineView`, for example.

These "modules" are then composed into an actual "view", which is somehow presented by the app (usually by the homescreen buttons, but also by the toolbar buttons.)

Here's an example "view":

https://github.com/StoDevX/AAO-React-Native/blob/94f60325fe3b12885676e13b3536b9455d4bcc4e/source/views/campus-dictionary.js#L1-L20

Every view must export two named exports: `navigation` and `view`. 

`navigation` is used by the navigation system to allow links and buttons to work, and the keys there must be globally unique. All exported "navigation" exports are joined together into one giant object to create the current [`navigation.js`](https://github.com/StoDevX/AAO-React-Native/blob/master/source/navigation.js).

`view` is used by the homescreen to render the buttons. (There's also a new type of view, `hidden`, which facilitates the toolbar buttons.)

I think that this new organization method really shines in its _consistency_ across all of our views – the "view" is now responsible for creating the tabs, and the content of each tab has to be created in a "module" and imported into the view.

-----

This PR is not intended to be merged anytime soon. I just wanted to open it so it didn't languish in GitHub purgatory. (… actually, looking it over again, it needs some cleanup even beyond the merge conflicts. Some of the restructuring I did for the modules resulted in some code duplication, so I haven't found the right balance of "module separation" yet.)

I'd be interested to hear your thoughts, either here or in person, @elijahverdoorn.


drewvolz:
@hawkrives I am a big fan of this reorg :+1:

drewvolz:
@hawkrives if you need help cleaning up merge conflicts, let me know

Issue #1540, opened 2017-09-10
Load bus data from github


hawkrives:
Currently, the bus data is still bundled with the app. We should pull it from github.

Issue #1534, opened 2017-09-10
Building hours list not refreshing


drewvolz:
The list of hours is not updating the open/closed and number of minutes. The detail view seems to be working.

Issue #1404, opened 2017-08-08
Transportation bus schedules are slow to use, sometimes


hawkrives:
Especially on slow phones. 

I want to speed them up. 

> Not necessarily part of 2.4.

hawkrives:
This needs to be profiled and sped up by doing less work, I think.

hawkrives:
This was not solved by #1542. 

It's still really horrendously slow. To the point of me restarting the app. 

This really needs some profiling - probably in Android Studio?

I also want to look at making the android tabs only render the views on either side of the active one, instead of keeping all the views you've opened in memory - I think our old tabs wrapper did that, and it should be possible to port it forward since we're still on the same underlying library.

rye:
I can troll around and run some profiles to see if I can come up with a cause.  It could just be janky code, who knows?

rye:
It seems that the animation is slow and so is the loading. (Though it was sped up quite a bit by the space reductions afforded by #1736.)

hawkrives:
I think this was solved by SwiftUI and dropping Android

PR #1542, opened 2017-09-10
Load bus data from Github

hawkrives:
Closes https://github.com/StoDevX/AAO-React-Native/issues/1540

- Loads data from Github for both the bus routes and the "other modes" list
- Allows the user to "pull to refresh" the bus route (but it really just updates the time)
- [more of a cleanup than anything] Passes data to BusMap instead of letting it load its own data

### DIFF
--- a/source/lib/report-network-problem.js
+++ b/source/lib/report-network-problem.js
@@ -4,7 +4,7 @@ import {tracker} from '../analytics'
 import bugsnag from '../bugsnag'
 
 export function reportNetworkProblem(err: Error) {
-	tracker.trackException(err.message)
-	bugsnag.notify(err)
-	console.warn(err)
+  tracker.trackException(err.message)
+  bugsnag.notify(err)
+  console.warn(err)
 }
--- a/source/lib/report-network-problem.js
+++ b/source/lib/report-network-problem.js
@@ -1,0 +1,10 @@
+// @flow
+
+import {tracker} from '../analytics'
+import bugsnag from '../bugsnag'
+
+export function reportNetworkProblem(err: Error) {
+	tracker.trackException(err.message)
+	bugsnag.notify(err)
+	console.warn(err)
+}
--- a/source/views/transportation/bus/bus-view.js
+++ b/source/views/transportation/bus/bus-view.js
@@ -48,9 +48,9 @@ export class BusView extends React.PureComponent {
       return defaultData
     })
 
-    // if (process.env.NODE_ENV === 'development') {
-    //   busLines = defaultData.data
-    // }
+    if (process.env.NODE_ENV === 'development') {
+      busLines = defaultData.data
+    }
 
     this.updateTime()
 
--- a/source/views/transportation/bus/bus-view.js
+++ b/source/views/transportation/bus/bus-view.js
@@ -26,6 +26,7 @@ export class BusView extends React.PureComponent {
     intervalId: 0,
     loading: false,
     now: moment.tz(TIMEZONE),
+    // now: moment.tz('Fri 8:13pm', 'ddd h:mma', true, TIMEZONE),
   }
 
   componentWillMount() {
@@ -91,7 +92,6 @@ export class BusView extends React.PureComponent {
 
   render() {
     let {now} = this.state
-    // now = moment.tz('Fri 8:13pm', 'ddd h:mma', true, TIMEZONE)
     const activeBusLine = this.findLine()
 
     if (!activeBusLine) {
--- a/source/views/transportation/bus/bus-view.js
+++ b/source/views/transportation/bus/bus-view.js
@@ -7,29 +7,30 @@ import {BusLine} from './bus-line'
 import moment from 'moment-timezone'
 import {NoticeView} from '../../components/notice'
 import type {TopLevelViewPropsType} from '../../types'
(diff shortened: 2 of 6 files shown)


source/views/transportation/bus/bus-view.js
```diff
@@ -1,66 +1,117 @@
...
+  fetchData = async () => {
+    const start = Date.now()
+    this.setState(() => ({loading: true}))
+
+    let {data: busLines} = await fetchJson(GITHUB_URL).catch(err => {
+      reportNetworkProblem(err)
+      return defaultData
+    })
+
+    // if (process.env.NODE_ENV === 'development') {
```
drewvolz: @hawkrives

hawkrives: oops

drewvolz (dismissed):
This PR suffers from the same issue as #1534. The list simply isn't updating any changes visually from the interval refreshing.

hawkrives:
Okay, @drewvolz, but that's not due to this PR, is it?

hawkrives:
So here, updateTime is being called, but nothing's updating.

edit: nope, I'm silly. the view is updating fine… if I don't override `state.now` to always be the same value inside of render.

hawkrives:
okay, this is bizarre – if I do a `setInterval` for 1 second, it all works perfectly. If I do 5 seconds, it never updates.

source/views/transportation/bus/bus-view.js
```diff
@@ -1,66 +1,117 @@
...
+    }
+
+    this.setState(() => ({busLines, loading: false}))
+  }
+
   updateTime = () => {
     this.setState(() => ({now: moment.tz(TIMEZONE)}))
   }
 
+  refreshTime = async () => {
```
rye, line 73: This function smells a lot like `fetchData`, can any common behavior be extracted out in a convenient way?

source/views/transportation/bus/bus-view.js
```diff
@@ -1,66 +1,117 @@
...
-import {data as defaultBusLines} from '../../../../docs/bus-times.json'
+import * as defaultData from '../../../../docs/bus-times.json'
 
 const TIMEZONE = 'America/Winnipeg'
 
-export class BusView extends React.PureComponent {
-  static defaultProps = {
-    busLines: defaultBusLines,
-  }
+const GITHUB_URL = 'https://stodevx.github.io/AAO-React-Native/bus-times.json'
```
rye, line 17: Can the repo URL get turned into a constant somewhere that this is built off of?  (This is out-of-scope for this PR but an idea nonetheless)

source/views/transportation/other-modes.js
```diff
@@ -25,6 +27,9 @@ const Container = glamorous.view({
   borderColor: c.iosLightBackground,
 })
 
+const GITHUB_URL =
```
rye, line 30: Ditto to my comment about adding a central constant for this.  Also, does the body of this const need to be on a different line?

source/lib/report-network-problem.js
```diff
@@ -0,0 +1,10 @@
+// @flow
+
+import {tracker} from '../analytics'
+import bugsnag from '../bugsnag'
+
+export function reportNetworkProblem(err: Error) {
```
rye, line 6: This is a good thing to centralize.

rye (changes requested):
Just a couple of off-hand code style tweaks is all I can offer at this point, but it looks good to me otherwise.

source/views/transportation/bus/bus-view.js
hawkrives: I haven't come up with anything.

source/views/transportation/bus/bus-view.js
hawkrives: I keep pushing that back to when I want to centralize all data fetching in the redux store… this is, however, unlikely to change, and is easy to grep for.

source/views/transportation/other-modes.js
hawkrives: Say "hi" to prettier – the string + assignment is over 80 chars long, so prettier did what it does.

hawkrives:
Alright, new stuff!

I changed BusLine to calculate a bunch of stuff once, and then store it in state, instead of calculating it on every render.

This also only re-renders the list when the "now" minute changes, to further optimize it - see below.

https://github.com/StoDevX/AAO-React-Native/blob/1e85c97af2812e573973370b8eada193e9b1f259/source/views/transportation/bus/bus-line.js#L96-L104

I'm still unsure as to why `setInterval(fn, 1000)` works while `setInterval(fn, 5000)` doesn't, but with the rendering optimizations we're good to go with just rendering every second.

This PR now probably addresses https://github.com/StoDevX/AAO-React-Native/issues/1404, but that'd require testing to verify.

hawkrives:
hang on, I need to revise my moment comparison in here. don't merge yet.

hawkrives:
I've fixed up my shouldComponentUpdate checks – it now only re-calculates the data when it actually needs to, instead of calculating and then just not rendering.

@rye, @drewvolz, @elijahverdoorn – this is ready for review.

rye (approved):

drewvolz (approved):

Issue #1541, opened 2017-09-10
Add footnote about "bus schedules may change without notice" to bus view


hawkrives:
Similar to #1517

rye:
This is probably a good idea.  Insofar as we have humans poking data files, which is how things will be, something's bound to go wrong or get missed.

PR #1543, opened 2017-09-10
Add info footer to bus views

hawkrives:
Closes https://github.com/StoDevX/AAO-React-Native/issues/1541
Depends on https://github.com/StoDevX/AAO-React-Native/pull/1544

> Bus routes and times subject to change without notice
> Data collected by the humans of All About Olaf

screen shot 2017-09-10 at 3 28 34 pm

### DIFF
--- a/source/views/transportation/bus/bus-line.js
+++ b/source/views/transportation/bus/bus-line.js
@@ -11,7 +11,12 @@ import moment from 'moment-timezone'
 import * as c from '../../components/colors'
 import {Separator} from '../../components/separator'
 import {BusStopRow} from './bus-stop-row'
-import {ListRow, ListSectionHeader, Title, ListFooter} from '../../components/list'
+import {
+  ListRow,
+  ListSectionHeader,
+  Title,
+  ListFooter,
+} from '../../components/list'
 
 const TIME_FORMAT = 'h:mma'
 const TIMEZONE = 'America/Winnipeg'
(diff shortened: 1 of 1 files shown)


drewvolz (approved):

drewvolz:
Not sure if I like it as two separate chunks or as one chunk.

One | Two
--|--
screen shot 2017-09-10 at 4 43 55 pm | screen shot 2017-09-10 at 4 43 55 pm

rye (approved):

hawkrives:
reopening PR to rebuild with #1544 included

drewvolz:
@hawkrives so... thoughts? I like them closer together with one new line

hawkrives:
@drewvolz … I like them apart

rye:
@drewvolz @hawkrives well _I_ like them closer together but not right on top of each other—is there any way to turn this from `\n\n` to something less stringy and more elemental?  Then the `line-height` could get set and we could tweak that at will.  Or, the `line-height` could just get set on the footer as a whole.  That works, too.

hawkrives:
Y'all should see what it looks like if you just use two ListFooters - lots of breathing room. 

@drewvolz I think the reason I went for the two lines originally was to allow the user to tap on the second line, in an obvious fashion. However, this api's probably going to need to be rethought soon, so I guess I don't really care.  

@rye Possible, yes, sure. I don't remember how well react-native likes line-height. 

@drewvolz can you try setting `lineHeight: 1.5` on the single-linebreak one?

hawkrives:
I'd like to merge this one, like #1545; whatever we settle on should be shared between both. Objections? @elijahverdoorn @drewvolz @rye

rye:
I'd say go ahead in the same fashion as #1545.

Issue #1563, opened 2017-09-12
Bus list UI not auto refreshing 


drewvolz:
The bus lists are not updating the UI automatically.

hawkrives:
Define "not updating the UI". Just times, just the dot, anything?

hawkrives:
Well, darn. it's really not updating anything.

I did confirm last night that the building hours list does work.

hawkrives:
I think it's that I set the `setInterval` back to 5000 accidentally, instead of 1000.

https://github.com/StoDevX/AAO-React-Native/blob/cd024d356ae855d2c242891a6b58aa778594b267/source/views/transportation/bus/bus-view.js#L36

drewvolz:
Changing the interval did not fix this issue.

hawkrives:
Sigh.

drewvolz:
I believe the problem is within the return. Setting it to `true` fixes the UI update issue @hawkrives 

https://github.com/StoDevX/AAO-React-Native/blob/cd024d356ae855d2c242891a6b58aa778594b267/source/views/transportation/bus/bus-line.js#L103-L113

hawkrives:
Oh. Yes.

hawkrives:
Lets play a game - can you find the bug of what's not changing there?

drewvolz:
All three comparisons are constantly returning false, so I'm going to blame something in `nextProps` or `nextState`.

hawkrives:
`currentMoments` is the list of time strings. When the minutes change, the list of times won't always change – it'll only change when an iteration of the bus' route is complete.

I'll add another check to the busLine shouldComponentUpdate.

PR #1546, opened 2017-09-11
Optimize Building Hours views (and make them actually update)

hawkrives:
This PR uses `shouldComponentUpdate` to optimize the rendering and data calculations in the Building Hours rows to make it feasible to do a `setInterval` of 1000ms.

As an added benefit, the building hours view should be a good bit snappier on slow devices.

Closes #1534, I think
Depends on a function added in #1542 (that's the flow error)

PR #1565, opened 2017-09-12
set the bus view refresh rate to 1 second

hawkrives:
We discovered in https://github.com/StoDevX/AAO-React-Native/pull/1546 and https://github.com/StoDevX/AAO-React-Native/pull/1542 that our setInterval calls only seem to work if we run on an interval of around 1000ms.

When I was testing https://github.com/StoDevX/AAO-React-Native/pull/1542, I set it to 1000; however, as I was cleaning up my changes, I appear to have reverted that before I made the PR.

This PR sets the BusView refresh interval to 1000 ms.

Closes https://github.com/StoDevX/AAO-React-Native/issues/1563

### DIFF
--- a/source/views/transportation/bus/bus-view.js
+++ b/source/views/transportation/bus/bus-view.js
@@ -31,9 +31,9 @@ export class BusView extends React.PureComponent {
 
   componentWillMount() {
     this.fetchData()
-    // This updates the screen every five seconds, so that the "next bus"
-    // times are updated without needing to leave and come back.
-    this.setState(() => ({intervalId: setInterval(this.updateTime, 5000)}))
+    // This updates the screen every second, so that the "next bus" times
+    // are updated without needing to leave and come back.
+    this.setState(() => ({intervalId: setInterval(this.updateTime, 1000)}))
   }
 
   componentWillUnmount() {


drewvolz:
I feel weird saying that `1000ms` refresh works and that `5000ms` refresh does not.

drewvolz:
This is worth looking deeper into.

hawkrives:
Yes. However, we have nothing to go on, and 1000ms works, so I'm not going to hold anything up on this research.

hawkrives:
(Also, _you_ don't have to _say_ anything about it 😝)

drewvolz (approved):
Approving this PR is _saying_ that.

drewvolz:
You might as well update the bus map to `1000ms` if we're going to change them all.

hawkrives:
That will be another PR.

hawkrives:
@drewvolz I'd _say_ that approving this PR is merely approving _a_ fix to the problem. It's just _acknowledging_ that this seems to fix the problem. Our options are, don't fix it, or use this fix that I found.

Because we _really_ have nothing to go on, unless we want to write our own `setInterval` shim. Which we don't.

We'll just have to check back when we upgrade React Native next time.

PR #1573, opened 2017-09-13
Tweak important contacts

drewvolz:
- Removes the phone number from the detail text of important contacts
- Adjust rows to only render the non-md description snippet
- Move format number function back into detail view
- Delete helper file

Mirrors https://github.com/carls-app/carls/pull/72 and https://github.com/carls-app/carls/pull/74

~ | Before | After
--|--|--
iOS | ios-before | ios-after
Android | android-before | android-after

### DIFF
--- a/source/views/contacts/contact-detail.js
+++ b/source/views/contacts/contact-detail.js
@@ -6,7 +6,7 @@ import glamorous from 'glamorous-native'
 import {phonecall} from 'react-native-communications'
 import {tracker} from '../../analytics'
 import {Button} from '../components/button'
-import {formatNumber} from './contact-helper'
+import openUrl from '../components/open-url'
 import type {ContactType} from './types'
 
 const Title = glamorous.text({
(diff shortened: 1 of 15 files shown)


source/views/contacts/contact-detail.js
```diff
@@ -41,14 +46,17 @@ export class ContactsDetailView extends React.PureComponent {
...
-  navigation: {state: {params: {contact: ContactType}}}
+  props: {navigation: {state: {params: {contact: ContactType}}}}
 
   onPress = () => {
     const item = this.props.navigation.state.params.contact
-    tracker.trackScreenView(
-      `ImportantContacts_${item.title.replace(' ', '')}View`,
-    )
-    promptCall(item.buttonText, item.phoneNumber)
+    const {title, phoneNumber, buttonText, buttonLink} = item
```
hawkrives: Just combine this line with the previous one, please.

hawkrives (changes requested):
One request, and good work!

source/views/contacts/contact-detail.js
drewvolz: Done.

hawkrives (approved):

PR #1577, opened 2017-09-13
Pull-to-refresh on streams list

drewvolz:
This PR changes what `refreshing` and `onRefresh` listen to. This enables the nice pull-to-refresh logic to get new stream data.

### DIFF
--- a/source/views/streaming/streams/list.js
+++ b/source/views/streaming/streams/list.js
@@ -97,7 +97,7 @@ export class StreamListView extends React.PureComponent {
       const mapped = toPairs(grouped).map(([title, data]) => ({title, data}))
 
       // wait 0.5 seconds – if we let it go at normal speed, it feels broken.
-      const elapsed = start - Date.now()
+      const elapsed = Date.now() - start
       if (elapsed < 500) {
         await delay(500 - elapsed)
       }
--- a/source/views/streaming/streams/list.js
+++ b/source/views/streaming/streams/list.js
@@ -62,8 +62,8 @@ export class StreamListView extends React.PureComponent {
 
     // wait 0.5 seconds – if we let it go at normal speed, it feels broken.
     const elapsed = Date.now() - start
-    if (elapsed < 1500) {
-      await delay(1500 - elapsed)
+    if (elapsed < 500) {
+      await delay(500 - elapsed)
     }
 
     this.setState(() => ({refreshing: false}))
--- a/source/views/streaming/streams/list.js
+++ b/source/views/streaming/streams/list.js
@@ -19,6 +19,7 @@ import moment from 'moment-timezone'
 import qs from 'querystring'
 import {toLaxTitleCase as titleCase} from 'titlecase'
 import type {StreamType} from './types'
+import delay from 'delay'
 
 const CENTRAL_TZ = 'America/Winnipeg'
 const url = 'https://www.stolaf.edu/multimedia/api/collection'
@@ -45,7 +46,7 @@ export class StreamListView extends React.PureComponent {
     error: null,
     loaded: false,
     noStreams: false,
-    refreshing: true,
+    refreshing: false,
     streams: [],
   }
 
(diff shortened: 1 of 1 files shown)


source/views/streaming/streams/list.js
```diff
@@ -45,7 +46,7 @@ export class StreamListView extends React.PureComponent {
     error: null,
     loaded: false,
```
hawkrives, line 47: What is "loaded"? It doesn't seem to be used anywhere.

source/views/streaming/streams/list.js
```diff
@@ -128,8 +138,8 @@ export class StreamListView extends React.PureComponent {
         ItemSeparatorComponent={ListSeparator}
         keyExtractor={this.keyExtractor}
         style={styles.listContainer}
-        refreshing={this.props.loading}
```
hawkrives, line 131: Where are the props to this component defined? I don't see them, and it's hampering my ability to understand this component… 😜

source/views/streaming/streams/list.js
```diff
@@ -55,6 +56,9 @@ export class StreamListView extends React.PureComponent {
 
   getData = async () => {
     try {
+      const start = Date.now()
+      this.setState(() => ({refreshing: true}))
```
hawkrives: Either this shouldn't default to `false`, or it shouldn't be called here, because this function is called as soon as the component is loaded, so refreshing effectively starts out set to true.

source/views/streaming/streams/list.js
```diff
@@ -92,6 +96,12 @@ export class StreamListView extends React.PureComponent {
       const grouped = groupBy(processed, j => j.$groupBy)
       const mapped = toPairs(grouped).map(([title, data]) => ({title, data}))
 
+      // wait 0.5 seconds – if we let it go at normal speed, it feels broken.
+      const elapsed = start - Date.now()
```
hawkrives: This is the incorrect code that I noticed over the weekend… it should be `Date.now() - start`.

source/views/streaming/streams/list.js
```diff
@@ -128,8 +138,8 @@ export class StreamListView extends React.PureComponent {
         ItemSeparatorComponent={ListSeparator}
         keyExtractor={this.keyExtractor}
         style={styles.listContainer}
-        refreshing={this.props.loading}
-        onRefresh={this.props.onRefresh}
+        refreshing={this.state.refreshing}
+        onRefresh={this.getData}
```
hawkrives: I like to define a second function that calls my `getData` function, but does the timing things separately. That way, my initial component mounting isn't delayed.

I usually call that function `refresh`, and it does the `setState{refreshing:true}`, `getData()`, `setState{refreshing:false}`. For example, the one in Calendar: 

https://github.com/StoDevX/AAO-React-Native/blob/e04aaea8810ce1ffaa1c7693a4e401add7bc5f9c/source/views/calendar/calendar-google.js#L97-L110

hawkrives (changes requested):

source/views/streaming/streams/list.js
drewvolz: This removes a reference to `props` that are not used. This PR fixes anything relating to `props`.

source/views/streaming/streams/list.js
drewvolz: Ah... thanks.

source/views/streaming/streams/list.js
drewvolz: I'll nix it.

source/views/streaming/streams/list.js
drewvolz: What will set refreshing to `true`, then, once it's set back to `false`?

source/views/streaming/streams/list.js
drewvolz: Nevermind. It's used -- it is for showing the `LoadingView` in render.

source/views/streaming/streams/list.js
hawkrives: Actually, scratch this. If you do what I recommended in my fourth comment, this is irrelevant and correct.

source/views/streaming/streams/list.js
drewvolz: Great idea on making it easier to cleaner. Thanks.

hawkrives (approved):

hawkrives (approved):
Lol nice.

rye (approved):

PR #1578, opened 2017-09-13
Check the current minute in BusLine's shouldComponentUpdate call

hawkrives:
Closes https://github.com/StoDevX/AAO-React-Native/issues/1563

I was trying to be smart and not cause another render unless we really needed one, so I checked the minimal set of properties that might change.

I checked two of the `props` – "line" and "openMap" – because they're simple pointer comparisons.

I only checked "currentMoments" from `state`, because it's the simplest data structure that will be generated as part of the state. You really don't want to replace a render bottleneck with a giant datastructure comparison bottleneck. 

What I forgot is that "currentMoments" will only change when the bus wraps around its route, because "currentMoments" is what's used to render the next stop times.

So, I added the third and final prop to the sCU check: `now.minute`. We definitely want to re-render every minute, because the bus dot can move once a minute. I also removed the "currentMoments" check, because currentMoments is a (mathematical) function of `this.props.now` and `this.props.line` – given the same line info, and the same time, it will always produce the same data.

I'm not checking any of the state things, because in order for them to change, the schedule stored in `this.props.line` will change (it's immutable, so any changes will result in a new object, and it'll get a new memory address) and because all of our functions are pure, this component's state only depends on the input props.

---

I've actually made a pretty big change in this PR – I essentially moved all the data generation back into `render`, and I removed all `state` storage. I did this because we'll only call render when the data might have changed, and we're not checking any of the cached data in sCU anymore anyway, so "caching" it's just going to lead to bugs and obscured data flows.

---

I'm not going to make a similar PR to remove data from Building Hours' row state, because it has a lot more rows to potentially re-render and thus re-generate data for, so it makes sense to check the building status and hours shown in sCU there and thus we have to cache them.

### DIFF
--- a/source/views/transportation/bus/bus-line.js
+++ b/source/views/transportation/bus/bus-line.js
@@ -6,7 +6,6 @@ import {getScheduleForNow, getSetOfStopsForNow} from './lib'
 import get from 'lodash/get'
 import zip from 'lodash/zip'
 import head from 'lodash/head'
-import isEqual from 'lodash/isEqual'
 import last from 'lodash/last'
 import moment from 'moment-timezone'
 import * as c from '../../components/colors'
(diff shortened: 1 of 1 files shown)


drewvolz (approved):
Good work! Thank you for the explanation.

PR #1590, opened 2017-09-14
Use functional setState and don't call refresh in orgs

hawkrives:
We want to call fetchData because it 
* (a) doesn't incur a 500ms delay penalty on first load, and

### DIFF
--- a/source/views/student-orgs/list.js
+++ b/source/views/student-orgs/list.js
@@ -89,7 +89,7 @@ export class StudentOrgsView extends React.Component {
   }
 
   componentWillMount() {
-    this.refresh()
+    this.fetchData()
   }
 
   fetchData = async () => {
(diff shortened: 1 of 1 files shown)


drewvolz:
Is (b) relevant? I don't think I've seen anything other than the `LoadingView` here. I think you're referring to the refreshable spinner at the top of the list... which may not be enabled here.

hawkrives:
It is enabled here. I can't say _why_ you haven't seen it, but what I listed are the two things that calling getData instead of refresh will do.

drewvolz:
StudentOrgs has a refreshing spinner? This is `StyledAlphabetListView` we're talking about.

hawkrives:
It appears to not have a refresh indicator. Or do pull on refresh.

drewvolz:
Now. I would _like_ for that to be there. Pull-to-refresh. That will have to wait until a custom `AlphabetList` component is written.

hawkrives:
Correct.

drewvolz (approved):

PR #1591, opened 2017-09-14
Clean up news-container

hawkrives:
- Use functional setState
- Use reportNetworkProblem
- Remove a promise .then from componentWillMount for consistency


drewvolz (approved):
Good work

PR #1594, opened 2017-09-14
Sort data files in natural sort order when bundling

hawkrives:
We want 10 to come after 9, when we name files to maintain a sort order.

Without this, that 10th category will sort right after the 1st.


drewvolz (approved):
Good catch

PR #1598, opened 2017-09-15
Fix KSTO

### LINKED ISSUE #1376: KSTO view crashes on iOS

drewvolz:
Closes #1376 

* Forked react-native-video to fix iOS crash
* Lib function to timeout on a promise (thanks @hawkrives)
* Check to see if m3u8 chunklist is returned within 6 seconds
* Pause stream state if we do not see data coming in from stream
* Run this check every time we unpause the stream

---

Edit from  @hawkrives:

I've added a simple state machine to this component.

The Button can be in one of three states: playing, paused, or checking. It can only ever be in one of those states, which simplifies the logic around checking the status.

Screenshots:

On Start | Checking | Playing
--- | --- | --- 
screen shot 2017-09-16 at 12 21 23 pm | screen shot 2017-09-16 at 12 21 25 pm |  screen shot 2017-09-16 at 12 21 26 pm


source/views/streaming/radio.js
```diff
@@ -34,31 +36,61 @@ export default class KSTOView extends React.PureComponent {
...
+    if (this.state.uplinkError) {
+      this.setState(() => ({paused: true}))
+    }
+  }
+
   changeControl = () => {
     this.setState(state => ({paused: !state.paused}))
+
+    // If we try to play...
+    if (this.state.paused) {
```
hawkrives: This is not guaranteed to have the new state. I almost think that this block should go into the setState callback.

source/views/streaming/radio.js
```diff
@@ -34,31 +36,61 @@ export default class KSTOView extends React.PureComponent {
...
+      this.setState(() => ({uplinkError: false, message: ''}))
+    } catch (err) {
+      this.setState(() => ({
+        uplinkError: true,
+        message: 'The KSTO stream is down. Sorry!',
+      }))
+    }
+
+    // If the stream is down or we had an error, pause the player
+    if (this.state.uplinkError) {
```
hawkrives: UplinkError is not guaranteed to have been set by this point.

source/views/streaming/radio.js
hawkrives: I think I know what I'd like this to do; let me take a stab at it in a bit.

hawkrives:
oops.

hawkrives:
fixed?

rye (approved):

hawkrives:
I'm still plotting to work on this before merging, fyi.

hawkrives (approved):
I shall now approve my own work lol

rye (approved):

drewvolz:
Your state machine is much nicer. Good work!

hawkrives:
Oh munglfraijs.

Fixing prettier.

PR #1599, opened 2017-09-15
Don't set the calendar to refresh until you refresh

hawkrives:
This is fallout from (that other PR where I switched to call getEvents instead of refresh).

It never stops refreshing, today.


drewvolz (approved):
Good catch!

PR #1603, opened 2017-09-15
Add a Moodle shortcut

hawkrives:
&nbsp; | &nbsp;
--- | ---
screen shot 2017-09-15 at 10 19 58 am | screen shot 2017-09-15 at 10 20 12 am

---

I could go either way on this.

One the one hand, I want to build a somewhat-limited native Moodle view, using the Moodle apis – they exist, and they're enabled. This leads to the question: given that said native view would be more limited than the web moodle, should we ship the web version if we'll ship a less featured native one later?

On another hand, will anyone actually use this view?

---

Thoughts? Especially @rye and @OmarShehata and @elijahverdoorn.

### DIFF
--- a/source/views/views.js
+++ b/source/views/views.js
@@ -120,6 +120,15 @@ export const allViews: ViewType[] = [
     tint: c.periwinkle,
     gradient: c.lightBlueToBlueDark,
   },
+  {
+    type: 'url',
+    url: 'https://moodle.stolaf.edu/',
+    view: 'MoodleView',
+    title: 'Moodle',
+    icon: 'graduation-cap',
+    tint: c.cantaloupe,
+    gradient: c.yellowToGoldDark,
+  },
   //   {
   //     type: 'view',
   //     view: 'HelpView',


rye:
@hawkrives I would totally use something like this, and this is an exciting idea.  I think a lot of people would like to use Moodle within AAO.

Insofar as the web interface for Moodle feels kinda janky, I think that building a "native" interface using the API's would be far more extensible and flexible.  Even if we can only facilitate simple actions, (like listing courses, contact information for profs, showing assignments/grades) it's far better than expecting Moodle to look nice on mobile.  If, of course, Moodle looked nice and was well-designed on mobile, that'd be a different story, but I'm skeptical that that will happen.

*If* this webview shortcut gets shipped, I still think it'd be very useful.  And I think it's totally okay to jump ship and trade a poorly-designed, fully-featured webview for a well-designed, less-featureful native API-based view as we move forward.  If we could track (via analytics) what users _actually use_ the Moodle view for, (my guess would be checking grades) that would be the most optimal as we could prioritize those views.

But, for now, adding a webview shortcut is totally fine by me.

hawkrives:
(@rye: as a side project, you should totally work on figuring out how google analytics works. We track a few things, but I don't really know how to view them in the dashboard.)

elijahverdoorn:
I'm all for this one as is, with a potential to edit it in the future to be a native view. Doing that follows our pattern anyway - we did it with the caf views about a year ago. I say merge it.

hawkrives:
@rye / @elijahverdoorn one of you's going to have to approve it then :)

source/views/views.js
```diff
@@ -120,6 +120,15 @@ export const allViews: ViewType[] = [
     tint: c.periwinkle,
     gradient: c.lightBlueToBlueDark,
   },
+  {
+    type: 'url',
+    url: 'https://moodle.stolaf.edu/',
```
rye, line 125: Good that we're using the dynamic URL.

rye (approved):
Looks fair enough to me.  I just want us to keep an API-based native-y solution on the wishlist.

source/views/views.js
hawkrives: Heck yeah. I'm not going to try and remember to update this one every year. Not when I can just do this. 

Side note: how long will it take for a google search for "st. Olaf moodle" to return 17-18 instead of 16-17.

rye: Fair point.  FWIW, if we didn't care about keeping the old Moodles accessible, one could feasibly use Google Search Console's [Change of Address](https://support.google.com/webmasters/answer/83106) feature.  (As long as they've verified ownership and stuff)

Issue #700, opened 2017-02-14
Display status message if SIS login fails


drewvolz:
Ideally there would be a message indicated in the form that makes it known that the login info provided from a fresh attempt was incorrect.

Current status: 
**Release**: incorrectly reports successful sign in
**Beta**: gives no notice of incorrect login beyond changing back to Sign In
**Ideal**: give notice of incorrect login

PR #1613, opened 2017-09-16
Add error messages to St. Olaf login panel

hawkrives:
Closes https://github.com/StoDevX/AAO-React-Native/issues/700

… yeah. That's about it.

I added some more flow types to the reducer, too, and removed the old Token login stuff.

screen shot 2017-09-16 at 3 07 34 pm


drewvolz (approved):
Good work!

PR #1621, opened 2017-09-17
Upgrade Travis' Android builds to Trusty Tahr

hawkrives:
and do some little gradle script cleanups, too.

I've confirmed that the built app launches and runs on my emulator.


rye (approved):
Wheeeee

PR #1623, opened 2017-09-17
Remove use of react-native-video from KSTO screen

hawkrives:
I'm really _really_ tired of `react-native-video`.

We've been abusing what it's intended to do anyway, IMO 😸

So… let's abuse another component!

`<WebView/>`.

---

This PR adds an invisible WebView to the KSTO screen. As the button changes the playing state in the KstoView, the WebView wrapper (called StreamPlayer) sends the appropriate events to the `<audio>` tag in the WebView.

Inside the WebView, we have some javascript that listens for and dispatches events from the `<audio>` tag that actually plays this.

By never removing and remounting the WebView, we can respond to events sent from Control Center. The WebView will then bubble the new playing state back up to the button, so everything's still in sync.

---

I just realized that this isn't working on Android. Grr. However, after some testing, the `react-native-video` implementation isn't playing any static on Android, either. Additionally, the live stream at https://pages.stolaf.edu/ksto/listen _also_ does not work in Chrome on my emulator.

What do we want to do about Android?

### DIFF
--- a/source/views/streaming/radio.js
+++ b/source/views/streaming/radio.js
@@ -22,7 +22,7 @@ import {TabBarIcon} from '../components/tabbar-icon'
 const kstoStream = 'https://cdn.stobcm.com/radio/ksto1.stream/master.m3u8'
 const image = require('../../../images/streaming/ksto/ksto-logo.png')
 
-type Viewport = {width: number, height: number, }
+type Viewport = {width: number, height: number}
 
 type HtmlAudioError = {code: number, message: string}
 
--- a/source/views/streaming/radio.js
+++ b/source/views/streaming/radio.js
@@ -6,36 +6,33 @@
 
 import React from 'react'
 import {
-  StyleSheet,
-  View,
-  ScrollView,
-  Text,
   Dimensions,
   Image,
+  ScrollView,
+  StyleSheet,
+  Text,
+  View,
+  WebView,
 } from 'react-native'
 import * as c from '../components/colors'
 import Icon from 'react-native-vector-icons/Ionicons'
-import Video from 'react-native-video'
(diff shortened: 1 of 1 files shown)


source/views/streaming/radio.js
```diff
@@ -190,11 +175,200 @@ const Title = () =>
...
+    <title>KSTO Stream</title>
+
+    <audio id="player" webkit-playsinline>
+      <source src="${url}" />
+    </audio>
+
+    <script>
+      var player = document.getElementById('player')
+
+      /////
```
drewvolz, line 279: What are all of these “/////“ blocks?

drewvolz (commented):
I get you’re trying to solve an issue here. The issue is a component that doesn’t update and has been broken for some time. So I say we go with your implementation. Is this allowed? I thought something in the iOS dev guidelines prohibited invisible webviews. I’d have to look it up.

source/views/streaming/radio.js
hawkrives: Dividers.

hawkrives:
> Is this allowed? I thought something in the iOS dev guidelines prohibited invisible webviews. I’d have to look it up.

We also have _no freaking idea_ how rn-video works, nor how to fix it when it breaks, and it sometimes just crashes!

At the very least, this PR won't crash the app when it fails.

(Also, how is an invisible webview more of an abuse than an invisible video player?)

---

Yes, I am sure that I want to go this way.

---

Additionally, I can neither confirm nor deny that Android works with this PR… I don't get any static with either Chrome or the app, but they're making both making network requests, so I _think_ it's actually working.

drewvolz (approved):

PR #1627, opened 2017-09-17
AlphabetListView cleanups

hawkrives:
- extract the "searchable" stuff into a `SearchableAlphabetListView` component
- show included data in Dictionary while data loads in the background – it now shows the refreshing spinner when it opens
- code cleanups


drewvolz (changes requested):
A comment on strange behavior

1. Typing a space character seems to clear the search
2. Typing too quickly seems to clear the search

drewvolz:
Great work otherwise extracting this into what it is now.

hawkrives:
@drewvolz Do those not happen on master?

drewvolz:
@hawkrives they do not happen on master.

hawkrives:
@drewvolz darn.

drewvolz:
After playing with this, sticking the `SearchBar` into the `header` gives us the scrolling behavior and list look we want, but not the functionality.

The `SearchBar` is then at the mercy of the list. The list tends to refresh the header at odd times... like when we try to search for multiple words with a space (which we don't support), or when we type too quickly... @hawkrives said he still wants to find a way to make this work with the list `header`.

I have another solution which is to keep it _outside_ of the `AlphabetList` (as it is presently) and instead show/hide the component when we reach or leave the top of the list.

This applies to `DictionaryView` and certainly can be applied to `StudentOrgs`

```diff
diff --git a/source/views/dictionary/list.js b/source/views/dictionary/list.js
index ffd9566e..2b76b13a 100644
--- a/source/views/dictionary/list.js
+++ b/source/views/dictionary/list.js
@@ -57,8 +57,18 @@ export class DictionaryView extends React.Component {
 
   state: {
     results: {[key: string]: Array<WordType>},
+    atTop: boolean,
   } = {
     results: terms,
+    atTop: true,
+  }
+
+  handleScroll = (event: any) => {
+    let currentOffset = event.nativeEvent.contentOffset.y
+    this.setState(() => {
+      let atTop = currentOffset < 5
+      return {atTop: atTop}
+    })
   }
 
   onPressRow = (data: WordType) => {
@@ -129,12 +139,14 @@ export class DictionaryView extends React.Component {
 
     return (
       <View style={styles.wrapper}>
-        <SearchBar
-          getRef={ref => (this.searchBar = ref)}
-          onChangeText={this.performSearch}
-          // if we don't use the arrow function here, searchBar ref is null...
-          onSearchButtonPress={() => this.searchBar.unFocus()}
-        />
+        {this.state.atTop
+          ? <SearchBar
+              getRef={ref => (this.searchBar = ref)}
+              onChangeText={this.performSearch}
+              // if we don't use the arrow function here, searchBar ref is null...
+              onSearchButtonPress={() => this.searchBar.unFocus()}
+            />
+          : null}
         <StyledAlphabetListView
           data={groupBy(this.state.results, item => head(item.word))}
           cell={this.renderRow}
@@ -149,6 +161,7 @@ export class DictionaryView extends React.Component {
           renderSeparator={this.renderSeparator}
           keyboardDismissMode="on-drag"
           keyboardShouldPersistTaps="never"
+          onScroll={this.handleScroll}
         />
       </View>
     )
```

hawkrives:
@drewvolz what I meant to say was, the correct solution is still to use the header prop, but _for this next version_, we should stick with the always-visible search bar.

drewvolz:
Sure. I was suggesting a solution if you wanted a patch that shows/hides it.

hawkrives:
@drewvolz This is ready for review again. I've tested, and both of your notes above are fixed.

rye (approved):

drewvolz (approved):

PR #1644, opened 2017-09-18
Remove auto-play of ksto audio

hawkrives:
This is a remnant of a previous iteration of the KSTO webview player, when I would unmount and remount the WebView when the audio paused.

Closes https://github.com/StoDevX/AAO-React-Native/issues/1643


drewvolz (approved):
Thanks

PR #1664, opened 2017-09-21
Add CodeClimate Configuration

rye:
I'm opening this PR just for kicks and also to see what CodeClimate has to say with a good configuration file.  Since we're using ESLint 4, we need to have custom configurations to use that channel of the ESLint engine. (It's not the default just yet)

Open to feedback, but I'd like to see how this builds and looks before saying we should add it.


rye:
Welp.  The CodeClimate build on this PR was using the (nonexistent) `master` version of `.codeclimate.yml` and thus didn't use `eslint-4` for the build. 🙃

CodeClimate will be whining until we include these new config files, so I guess it makes sense to merge this once Travis gets done.

drewvolz (approved):

rye:
Yeah, so I'm just going to merge this and see if it works nicely. (We can always turn this off later and revert this PR.)  This isn't actually touching anything related to the app's codebase itself so it should be fine.

The only change from this PR that affects our existing CI flow is the addition of the `.eslintignore` file, which ignores all `min.js` files.  We might consider also adding an explicit `node_modules` ignore so we don't get errors for our dependencies.

But for now, I'll just merge this.

PR #1673, opened 2017-09-24
Replace sortable listview?

drewvolz:
#### Background
The sortable list view that we use from [gitim/react-native-sortable-list](https://github.com/gitim/react-native-sortable-list) works _okay_. We have some frustrations with it due to Android. This PR explores replacing it with [deanmcpherson/react-native-sortable-listview](https://github.com/deanmcpherson/react-native-sortable-listview).

#### Summary
I **do not** think this PR serves as a good replacement. If anything, we will lose some functionality by switching to this component. I don't think switching will get us what we want.
<hr> 

**What we have**

Action | iOS | Android
--|--|--
Press and drag | ✅ |  ✅
Reload new order | ✅ |  ✅
Scrolling on large list | ✅ |  ✅
Rotation resizing | ✅ |  ✅

**What this proposes**

Action | iOS | Android
--|--|--
Press and drag | ✅ |  ✅
Reload new order | ✅ |  ✅
Scrolling on large list | ❌ | ❌
Rotation resizing | ❌ | ❌

<hr>

#### Issue with rotation

Rotating in the `proposed` switch doesn't seem to work correctly. Glancing at the internals, resizing for rotation doesn't seem to be taken into consideration. It appears to resize when a row is pressed.

problems

### DIFF
--- a/source/views/home/edit.js
+++ b/source/views/home/edit.js
@@ -89,6 +89,7 @@ type RowProps = {
   data: ViewType,
   active: boolean,
   width: number,
+  sortHandlers?: any,
 }
 
 type RowState = {
@@ -209,7 +210,7 @@ class Row extends React.Component<void, RowProps, RowState> {
 
 type Props = {
   onSaveOrder: (ViewType[]) => any,
-  order: string[],
+  order: ViewType[],
 }
 type State = {
   width: number,
--- a/source/views/home/edit.js
+++ b/source/views/home/edit.js
@@ -21,7 +21,8 @@ import fromPairs from 'lodash/fromPairs'
 
 import EntypoIcon from 'react-native-vector-icons/Entypo'
 import IonIcon from 'react-native-vector-icons/Ionicons'
-import SortableList from 'react-native-sortable-list'
+import {Touchable} from '../components/touchable'
+import SortableListView from 'react-native-sortable-listview'
 
 import type {ViewType} from '../views'
 import {allViews} from '../views'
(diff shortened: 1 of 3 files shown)


drewvolz:
This PR serves more as evidence that we _could_ switch, but I **do not** think we should because of issues with rotation and scrolling.

source/views/home/edit.js
```diff
@@ -235,18 +239,39 @@ class EditHomeView extends React.PureComponent<void, Props, State> {
...
+  }
+
+  onRowMoved = event => {
+    let order = [...this.props.order]
+    order.splice(event.to, 0, order.splice(event.from, 1)[0])
+    this.props.onSaveOrder(order)
+  }
+
+  onMoveEnd = () => {
+    this.setState(() => ({activeRowView: ''}))
```
drewvolz, line 253: If this line is moved to `onRowMoved`, there's a chance that a single selection will leave that row with `active` styles. It behaves appropriately as-is in `onMoveEnd` which does not care if the row was moved or not. This is visible in the gif I've included... which was before I found this callback.

source/views/home/edit.js
```diff
@@ -200,18 +200,21 @@ class Row extends React.Component<void, RowProps, RowState> {
         <Text style={[styles.text, {color: this.props.data.tint}]}>
           {this.props.data.title}
         </Text>
-        {reorderIcon}
+        <Touchable highlight={false} {...this.props.sortHandlers}>
+          {reorderIcon}
```
drewvolz, line 204: This at least makes just the reorder icon the only part of the cell that enables the drag... solving an issue we had on Android, I think.

hawkrives:
Good work. Let's stick with our current impl.

PR #1680, opened 2017-09-24
Other modes list

drewvolz:
This PR introduces a refresh to the Other Modes transportation view.

* Simplified format to list/detail views
* Group modes by category (bus, car, bike)

~ | List | Detail
--|--|--
Android | android-list | android-detail
iOS | ios-list | ios-detail


hawkrives:
I didn't add a refresh thing to this when I made it use remote data because I don't think that pull-to-refresh gains you anything on this view. 

I think that we should approach some of these views as though they are local, and not present a pull-to-refresh for them.

drewvolz:
I'm trying to think of a case where pull-to-refresh is beneficial. I think I want it in there for the case where a network request fails the first time. Could be poor signal, could be no signal. I don't want to force anyone to leave the view and come back to make a second network request just to reload the data.

hawkrives:
I don't want to user to ever need to think about loading this data. The app should load the data for this type of view in the background and cache it off-line correctly so the user never needs to reload the view.

drewvolz:
We don't ever indicate to the user if the refresh failed. We have no messaging in-place to say the refresh didn't succeed. All we do is fallback to old data. Chances are that if someone sees outdated data without a message about it, I'll bet they'll try to refresh it via pull-to-refresh.

drewvolz:
If you want it removed, we can remove it. Do you still want it nixed, @hawkrives?

hawkrives (changes requested):
I would like the pull-to-refresh gesture removed. 

Other than that, I like this a lot!

drewvolz:
Removed.

source/views/transportation/other-modes/other-modes-detail.js
```diff
@@ -0,0 +1,71 @@
+// @flow
```
hawkrives: Could you please rename these three files to remove the `other-modes-` prefix?

source/views/transportation/other-modes/other-modes-detail.js
```diff
@@ -0,0 +1,71 @@
...
+    return {
+      title: navigation.state.params.mode.name,
+    }
+  }
+
+  props: {navigation: {state: {params: {mode: OtherModeType}}}}
+
+  onPress = () => {
+    const {name, url} = this.props.navigation.state.params.mode
+    tracker.trackScreenView(
```
hawkrives: What does this do if there's no URL to open? Should this be moved into the If-block?

hawkrives: Actually, why didn't you use trackedOpenUrl?

source/views/transportation/index.js
```diff
@@ -9,7 +9,7 @@ import React from 'react'
 import {TabNavigator} from '../components/tabbed-view'
 import {TabBarIcon} from '../components/tabbar-icon'
 
-import OtherModesView from './other-modes'
+import {OtherModesView} from './other-modes'
 import BusView from './bus'
 
 export {BusMapView} from './bus/map'
```
hawkrives, line 15: Please export {OtherModesDetailView} from this file.

source/navigation.js
```diff
@@ -22,6 +22,7 @@ import {
   BuildingHoursScheduleEditorView,
 } from './views/building-hours'
 import TransportationView, {BusMapView} from './views/transportation'
+import {OtherModesDetailView} from './views/transportation/other-modes'
```
hawkrives: Please import this in the same way that BusMapView is imported.

hawkrives (changes requested):
Looks good overall!

source/views/transportation/other-modes/other-modes-detail.js
drewvolz: Can we pick a convention for everything? I think this differs _per component_ these days.

hawkrives: Yes. It does. One of the reasons for my reorganization. 

I don't see a need to duplicate the folder hierarchy in the file name. That's why I requested that you remove it.

rye (approved):
Besides anything @hawkrives may have mentioned, looks good to me.

hawkrives (approved):

PR #1682, opened 2017-09-25
Condense student work detail view

drewvolz:
This PR introduces a more readable version of the iOS student work detail view.

By grouping the text together and using only `title` and `detail` rather than `subtitle`, we get a much sleeker view. I certainly feel it's much easier to parse. This was inspired by @hawkrives' work on the CARLs app to make this readable.

Before | After
--|--
before | screen shot 2017-09-24 at 10 42 25 pm

### DIFF
--- a/source/views/sis/student-work/detail.ios.js
+++ b/source/views/sis/student-work/detail.ios.js
@@ -21,40 +21,67 @@ const styles = StyleSheet.create({
   },
 })
 
-function Title({job}: {job: JobType}) {
-  return job.title || job.type
-    ? <Section header="JOB">
-        <Cell cellStyle="Subtitle" title={job.title} detail={job.type} />
-      </Section>
-    : null
-}
+function Information({job}: {job: JobType}) {
+  const title = job.title ? <Cell cellStyle="Basic" title={job.title} /> : null
 
-function Contact({job}: {job: JobType}) {
-  const contactName = getContactName(job).trim() || job.contactEmail
-  return job.office || contactName
-    ? <Section header="CONTACT">
-        <Cell
-          cellStyle="Subtitle"
-          title={contactName}
-          detail={job.office}
-          accessory="DisclosureIndicator"
-          onPress={() => email([job.contactEmail], null, null, job.title, '')}
-        />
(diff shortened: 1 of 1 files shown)


hawkrives (commented):
Did we lose the category?

I don't like having two of the rows only be on the left… isn't this a good spot to use the blue-tinted LeftDetail? When we know all of the words and they're all short?

I'm not sold on using the LeftDetail, but … hmm. What do you think about pulling out the job title into an area above the tables, like Building Hours?

Also, where did the link to oleville.com come from?

drewvolz:
> Did we lose the category?

Huh. I'll add that back in 👍 

> I don't like having two of the rows only be on the left… isn't this a good spot to use the blue-tinted LeftDetail? When we know all of the words and they're all short?

Okay, let's try out `LeftDetail`'s text. Looks like we can't control how wide it goes...

hmm

> I'm not sold on using the LeftDetail, but … hmm. What do you think about pulling out the job title into an area above the tables, like Building Hours?

Here's what that looks like.

title

> Also, where did the link to oleville.com come from?

It is embedded in the `comments` section of the posting. I figured showing the links up-front rather than at the bottom might work well.

hawkrives:
Links: ah. 

For the LeftDetail items:

Office | Foo
Link | oleville.com
etc

Not putting the link in the detail :p

drewvolz:
> Not putting the link in the detail :p

I have a better reason than "not wanting to." It's because links get really long, and they'll cut-off the title. Same goes for the job title. I tried "Title | Job" and "Link | stolaf.edu" but you risk the left label being truncated to one or two characters and an ellipses. Here are a few screenshots to make this clear...

Thoughts? I'm okay with moving the `title` to the top outside of the table.

Ellipse link label | Ellipse title label
--|--
screen shot 2017-09-24 at 9 39 58 pm | screen shot 2017-09-24 at 9 38 24 pm

hawkrives:
I would very much like to move the title out of the tables.

hawkrives (approved):

elijahverdoorn (approved):

PR #1696, opened 2017-09-26
Call "refresh" on Balances componentWillMount

### LINKED ISSUE #1667: Fetch balances when the view opens
I believe that the Balances tab currently only fetches data when you pull-to-refresh, when the app opens, and when the user logs in.

We should also make the view refresh the data in `componentWillMount`.

hawkrives:
Closes #1667.

This makes me much happier. I'll still want to do an analysis on the server code, but this might be all that's needed.

### DIFF
--- a/source/views/sis/balances.js
+++ b/source/views/sis/balances.js
@@ -37,14 +37,14 @@ class BalancesView extends React.Component {
     updateBalances: boolean => any,
   }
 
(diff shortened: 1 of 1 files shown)


hawkrives:
Yeah, ok, eslint

rye (approved):

PR #1715, opened 2017-09-27
Clean up CodeClimate Build Flow

### LINKED ISSUE #1681: CodeClimate's ESLint pass is complaining about undefined flowtype-related rules
This is a strange quirk.  I want to mark these issues as Wontfix, but I'd rather figure out what the root of the problem is.  I've self-assigned as a way of keeping track of this.

rye:
Now that `eslint-plugin-flowtype` is added back, I can clean up the CodeClimate build stuff and make it fit what we need more precisely.  I think this is a good first start.

Affected parts:
 - Disabled PHP checking. (we don't have any PHP in this repository, thank goodness)
 - Turned on a bunch of RuboCop cops pertaining to style, which will only hit the fastlane stuff
    - I might turn those off, haven't decided yet.
 - Tried out making CSS linting more talkative but decided against it—it's somewhat pedantic.  I'm open to having it on, though.

*Closes #1681.*

### DIFF
--- a/fastlane/actions/activate_rogue_team.rb
+++ b/fastlane/actions/activate_rogue_team.rb
@@ -53,7 +53,7 @@ module Fastlane
           FastlaneCore::ConfigItem.new(key: :bundle_id,
                                        default_value: 'com.volz.drew.aao.rogue'),
           FastlaneCore::ConfigItem.new(key: :profile_name,
-                                       default_value: 'match Development com.volz.drew.aao.rogue'),
+                                       default_value: 'match Development com.volz.drew.aao.rogue')
         ]
       end
 
--- a/fastlane/actions/activate_rogue_team.rb
+++ b/fastlane/actions/activate_rogue_team.rb
@@ -53,7 +53,7 @@ module Fastlane
           FastlaneCore::ConfigItem.new(key: :bundle_id,
                                        default_value: 'com.volz.drew.aao.rogue'),
           FastlaneCore::ConfigItem.new(key: :profile_name,
-                                       default_value: 'match Development com.volz.drew.aao.rogue')
+                                       default_value: 'match Development com.volz.drew.aao.rogue'),
         ]
       end
 
--- a/fastlane/lib/util.rb
+++ b/fastlane/lib/util.rb
@@ -1,3 +1,4 @@
+# coding: utf-8
 # Adds the github token for stodevx-bot to the CI machine
 def authorize_ci_for_keys
   token = ENV['CI_USER_TOKEN']
--- a/fastlane/lib/util.rb
+++ b/fastlane/lib/util.rb
@@ -23,7 +23,7 @@ def current_build_number(**args)
     when :ios
       (latest_testflight_build_number + 1).to_s
     else
-      UI.input "Please enter a build number: "
+      UI.input 'Please enter a build number: '
     end
   rescue
     '1'
@@ -49,7 +49,7 @@ def propagate_version(**args)
   build = current_build_number(track: args[:track] || nil)
 
   UI.message "Propagating version: #{version}"
-  UI.message "into the Info.plist and build.gradle files"
+  UI.message 'into the Info.plist and build.gradle files'
 
   # encode build number into js-land
   set_package_data(data: {version: "#{version}+#{build}"})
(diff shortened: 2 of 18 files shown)


rye:
The pushed changes were just some changes to the Ruby files and the specs associated with them.  Since JS stuff has already been being checked by ESLint, there wasn't anything to fix there.  (We can add checks back later, if we want.)

fastlane/actions/activate_rogue_team.rb
```diff
@@ -53,7 +53,7 @@ def self.available_options
           FastlaneCore::ConfigItem.new(key: :bundle_id,
                                        default_value: 'com.volz.drew.aao.rogue'),
           FastlaneCore::ConfigItem.new(key: :profile_name,
-                                       default_value: 'match Development com.volz.drew.aao.rogue'),
+                                       default_value: 'match Development com.volz.drew.aao.rogue')
```
hawkrives: So I really like trailing commas. Can we make rubocop add them for us?

fastlane/platforms/android.rb
```diff
@@ -63,7 +63,7 @@
   lane :matchesque do
     match_dir = clone_match
 
-    # don't forget – lanes run inside of ./fastlane
+    # don't forget --- lanes run inside of ./fastlane
```
hawkrives: Uh

fastlane/lib/util.rb
```diff
@@ -11,9 +12,7 @@ def authorize_ci_for_keys
 
 # Gets the version, either from Travis or from Hockey
 def current_build_number(**args)
-  if ENV.key?('TRAVIS_BUILD_NUMBER')
-    return ENV['TRAVIS_BUILD_NUMBER']
-  end
+  return ENV['TRAVIS_BUILD_NUMBER'] if ENV.key?('TRAVIS_BUILD_NUMBER')
```
hawkrives, line 15: I get what this does, but I think the expanded form is more understandable to non-ruby devs, which is what most of us are. 

And it's not a big deal to go either way.

fastlane/platforms/android.rb
hawkrives: I like my utf-8 chars. Why shouldn't I use them in comments?

hawkrives:
What css does csslint even look at?

fastlane/actions/activate_rogue_team.rb
rye: Does ESLint allow this?

hawkrives: Yes, and prettier adds them all for us in our JS code.

rye: [This entry in the style guide](https://github.com/bbatsov/ruby-style-guide#no-trailing-array-commas) is why I did all of these.

fastlane/platforms/android.rb
rye: You can use them in comments, but the file should probably have UTF-8 set in its variables if you do.  There's a Cop for this that says you should only use ASCII comments.

fastlane/actions/activate_rogue_team.rb
hawkrives: Yeah, I think I remember reading that back when I was trying to learn ruby. I just disagree :p 

I like trailing commas, because they let me add new items to the collection without highlighting extra lines in the diff.

fastlane/platforms/android.rb
hawkrives: Then we're going to need to add the magic comment to all the files. I guarantee you that I'll want to put in more Unicode over time. Blame the Mac keyboard.

rye: (So, like, `Rølvaag` is cool, but `—` or `–` is not, at least, according to RuboCop.)

hawkrives: … how does that make any sense?

hawkrives: Silly rubocop.

hawkrives:
Could you please disable:

- no trailing commas
- no Unicode in comments

rye:
Sure @hawkrives.  Also, csslint points at the stuff in `data/` (news-y stuff, I think?) and grunts angrily.

hawkrives:
… alright. Can I see the css problems anywhere, before I make a decision there?

rye:
Sure, I can turn on the angry grunting again and you can take a peek when I push.

hawkrives (approved):

PR #1718, opened 2017-09-28
Move greenkeeper-lockfile-upload to after_script, not after_success

rye:
This is in accordance with their spec, and also will allow us to see if there are any errors with how we're doing things.


hawkrives (approved):

hawkrives:
We'll have to merge this and wait for a greenkeeper PR to see if this actually worked

PR #1721, opened 2017-09-29
Group live streaming events into separate section

drewvolz:
Live events will form their own grouped section at the top if any exist. This helps take the thinking out of comparing the listed time to your time... especially when you're in a different timezone.

### DIFF
--- a/source/views/streaming/streams/list.js
+++ b/source/views/streaming/streams/list.js
@@ -71,8 +71,11 @@ export class StreamListView extends React.PureComponent {
 
   getStreams = async (date: moment = moment.tz(CENTRAL_TZ)) => {
     try {
-      const dateFrom = date.format('YYYY-MM-DD')
-      const dateTo = date.add(1, 'month').format('YYYY-MM-DD')
+      const dateFrom = date.clone().format('YYYY-MM-DD')
+      const dateTo = date
+        .clone()
+        .add(1, 'month')
+        .format('YYYY-MM-DD')
 
       let params = {
         class: 'upcoming',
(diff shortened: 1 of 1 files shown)


source/views/streaming/streams/list.js
```diff
@@ -71,8 +71,11 @@ export class StreamListView extends React.PureComponent {
 
   getStreams = async (date: moment = moment.tz(CENTRAL_TZ)) => {
     try {
-      const dateFrom = date.format('YYYY-MM-DD')
-      const dateTo = date.add(1, 'month').format('YYYY-MM-DD')
+      const dateFrom = date.clone().format('YYYY-MM-DD')
```
hawkrives: You don't need the clone here - you're not mutating the date, just making a string from it.

hawkrives (changes requested):
👍

source/views/streaming/streams/list.js
drewvolz: Not cloning it means the date is mutated because it's passed by reference. Anything that eventually uses it below the first few lines of `getStreams` would have to make sense of a date that is one month in the future. I want to prevent that.

drewvolz: Wait, I think you're pointing to line 74 instead of 75. Your comment is fair.

hawkrives: I am indeed pointing to L74. 😄

rye (approved):

source/views/streaming/streams/list.js
drewvolz: Fixed.

elijahverdoorn (approved):

hawkrives (approved):

hawkrives:
Spurious build failure,

```
[13:44:41]: ▸ curl: (56) SSLRead() return error -9806
[13:44:41]: ▸ curl: Saved to filename 'boost_1_63_0.tar.gz'
[13:44:41]: ▸ Command /bin/sh failed with exit code 56
```

PR #1729, opened 2017-10-01
Reconfigure the innards of FancyMenu

hawkrives:
Firstly, I'd like to apologize for this PR: this whole component over here is a bit of a mess. I hope that it's a bit easier to follow after this.

So. Things that this PR does:

- Uses `ListHeaderComponent` on `<SectionList />` to render the Filter toolbar, thus allowing it to move with the list
  - Removes a level of nesting, and removes a now-unneeded stylesheet
- Adds Pull-To-Refresh capabilities on the menus, which, when coupled with the aforementioned ListHeaderComponent, works like you would expect
  - on iOS, the PTR is above the filter bar, and on Android, it comes over the top (because the filter bar is part of the list)
- [bugfix] It properly calls updateFilters on componentWillMount in FancyMenu
- It pulls out and names a number of routines that were hiding inside of `render()` or some other area, hopefully making the code easier to follow overall
  - I still can't shake the feeling that this can be dissected into separate components somehow (this = the whole menu structure) but I'll keep letting my subconscious stew on that
- It renders the Café's message inside of the list, allowing you to switch menus (and eventually days)
  - This actually resolves a TODO that was in the code: 
      ```
      // TODO: Figure out how to pass this down to FancyMenu, so we can render
      //       the filterbar to let people change meals if a meal/day isn't available.
      ```
  - I switched FancyMenu to use ListEmptyComponent to render the messages pretty recently (https://github.com/StoDevX/AAO-React-Native/commit/1adf7dab27e36c7eead5d44c5327becd3a086299)
  - Then I just took advantage of that, and of the ListHeaderComponent, to be able to pass a message down from BonAppMenu to FancyMenu

I think that's about it.

### DIFF
--- a/source/views/menus/components/fancy-menu.js
+++ b/source/views/menus/components/fancy-menu.js
@@ -1,7 +1,7 @@
 // @flow
 
 import React from 'react'
-import {View, StyleSheet, SectionList} from 'react-native'
+import {StyleSheet, SectionList} from 'react-native'
 import * as c from '../../components/colors'
 import {connect} from 'react-redux'
 import {updateMenuFilters} from '../../../flux'
@@ -12,6 +12,7 @@ import type {
   MasterCorIconMapType,
   ProcessedMealType,
   MenuItemContainerType,
+  StationMenuType,
 } from '../types'
 import size from 'lodash/size'
 import values from 'lodash/values'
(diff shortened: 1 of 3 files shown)


drewvolz (approved):
That is quite the rework. Looking forward to testing it out.

PR #1734, opened 2017-10-01
Some README tweaks

rye:
These are just some simple tiny tweaks, but I think they improve the flow.

- Change title of README to "All About Olaf"
  - We don't really need this to be the repo name, and it's just confusing.
- Flatten the `npm run` instructions
- Some other stylistic changes, like changing `…` (which I can't type with my keyboard) to `...` (which will render practically the same, anyways)


hawkrives (changes requested):
… can I please have my unicode ellipses?

hawkrives (approved):
🐻 🤗

PR #1736, opened 2017-10-02
Minify data files

rye:
This PR simply minifies our JSON files when we bundle our data.  This will drastically decrease the in-flight time of our data files, as we would no longer use whitespace here.  But, this will make fetching much faster.


hawkrives (changes requested):
Just one request - the usual way to generate minified JSON is to just omit the second two arguments to stringify. Would you mind omitting the final two arguments in each of these, please?

rye:
Requested changes made.

hawkrives (approved):
:shipit:

PR #1756, opened 2017-10-06
Bump version after release

hawkrives:
iTC wont let us submit new builds to the 2.4.0 release anymore.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "all-about-olaf",
-  "version": "2.4.1-alpha.1",
+  "version": "2.4.1-pre",
   "private": true,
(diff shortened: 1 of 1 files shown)


rye (commented):
I tend to prefer `X.Y.Z-pre` for in-development versions, but I suppose since we're shipping nightlies this makes sense.

hawkrives:
I could certainly to -pre until we tag the first alpha or beta. I actually prefer that.

rye (approved):
Yeah, keeping `-pre` on the version until we tag it is a great way to distinguish between unreleased and released versions, not to mention requiring less maintenance in the long-run.

PR #1772, opened 2017-10-12
Add KSTO schedule view + "call" button

drewvolz:
This PR introduces a button on the KSTO component to launch a Calendar view with the upcoming show schedule.

~ | Android | iOS
--|--|--
Play | android-ksto | ios-ksto
List | android-schedule | ios-ksto-list
Detail | android-schedule-detail | ios-ksto-detail


drewvolz:
... looks like we'll have to add a way to pass-in the "powered by" message.

drewvolz:
We need to confirm KSTO has an accurate schedule before this is used.

rye (approved):
Looks great! That "listen" button felt a little lonely to me.

source/views/streaming/radio/index.js
```diff
@@ -156,7 +160,18 @@ export default class KSTOView extends React.PureComponent<void, Props, State> {
             {error}
           </View>
 
-          {button}
+          <View style={buttonStyles.buttonWrapper}>
+            {playButton}
+
+            <View style={buttonStyles.space} />
```
drewvolz: I was trying to think of a way to avoid wrapping both the `play` and `calendar` (and now `call`) buttons in wrapping views with duplicate `marginRight` stylings. I settled on a simple `<View />` which has a small width. If we try to directly apply the `marginRight` style to the buttons themselves, we end up with the issue we had on Webcams. We do not want to have a larger area be selected because we shifted an element a few pixels.

source/views/streaming/radio/schedule.js
```diff
@@ -0,0 +1,19 @@
...
+export default class KSTOSchedule extends React.PureComponent {
+  static navigationOptions = {
+    title: 'Show Schedule',
+  }
+
+  render() {
+    return (
+      <GoogleCalendarView
+        navigation={this.props.navigation}
+        calendarId="kstonarwhal@gmail.com"
```
drewvolz: I'd like for one of us to confirm KSTO still uses this account.

source/views/streaming/radio/schedule.js
```diff
@@ -0,0 +1,19 @@
...
+import {GoogleCalendarView} from '../../calendar/calendar-google'
+
+export default class KSTOSchedule extends React.PureComponent {
+  static navigationOptions = {
+    title: 'Show Schedule',
+  }
+
+  render() {
+    return (
+      <GoogleCalendarView
```
drewvolz, line 16: We need a way  to provide a customized `powered by` message to the calendar. This is now a logged issue.

source/views/streaming/radio/index.js
```diff
@@ -362,6 +377,12 @@ type ActionButtonProps = {
   onPress: () => any,
 }
 
+type ScheduleActionButtonProps = {
```
hawkrives: This should move down to right above the ScheduleButton definition

source/views/streaming/radio/index.js
hawkrives: Applying a margin _to_ the Touchable shouldn't extend the touchable area - the problem with Webcams was that the margin was being added within the Touchable, which expands the area that it covers. Keep in mind that the CSS box model goes "margin" | "border" | "padding" | "content" | "padding" | "border" | "margin", and the "margin" is considered to be outside of anything that concerns the box. 

This is a fine solution anyway, however.

source/views/streaming/radio/schedule.js
```diff
@@ -0,0 +1,19 @@
+// @flow
+
+import React from 'react'
+import {GoogleCalendarView} from '../../calendar/calendar-google'
+
+export default class KSTOSchedule extends React.PureComponent {
```
hawkrives: To go with the rest of the app, this should be called "KSTOScheduleView".

source/views/streaming/radio/index.js
```diff
@@ -14,29 +14,29 @@ import {
   View,
   WebView,
 } from 'react-native'
-import * as c from '../components/colors'
```
hawkrives: This file should be named something other than "index", and "index" should then only export these two components (I think Dictionary behaves in a way that I would like you to emulate.)

These two components should then be exported again from "views/streaming/index", which is the only place where other files should import from.

source/navigation.js
```diff
@@ -35,6 +35,7 @@ import SettingsView from './views/settings'
 import CreditsView from './views/settings/credits'
 import PrivacyView from './views/settings/privacy'
 import LegalView from './views/settings/legal'
+import KSTOSchedule from './views/streaming/radio/schedule'
```
hawkrives: This should be reorganized so that it's exported from the same file (views/streaming) as the radio view itself. And ideally all three tabs and associated views would be imported from views/streaming.

hawkrives (changes requested):
Like the idea! Nitpicks.

source/views/streaming/radio/schedule.js
drewvolz: @hawkrives fixed this in #1904

source/views/streaming/radio/schedule.js
drewvolz: @hawkrives gave KSTO access to the account.

rye (approved):

hawkrives:
I will review this again later.

hawkrives (approved):

PR #1773, opened 2017-10-12
KSTO call button

drewvolz:
This is an addition to #1772.

This PR introduces another action button on the KSTO view to allow listeners to call the station.

Android | iOS
--|--
screen shot 2017-10-11 at 11 25 55 pm | screen shot 2017-10-11 at 11 25 43 pm

### DIFF
--- a/source/views/streaming/radio/index.js
+++ b/source/views/streaming/radio/index.js
@@ -23,7 +23,7 @@ import type {TopLevelViewPropsType} from '../../types'
 
 const kstoStream = 'https://cdn.stobcm.com/radio/ksto1.stream/master.m3u8'
 const image = require('../../../../images/streaming/ksto/ksto-logo.png')
-const stationNumber = '5077863602'
+const stationNumber = '+15077863602'
 
 type Viewport = {width: number, height: number}
 
--- a/source/views/streaming/radio/index.js
+++ b/source/views/streaming/radio/index.js
@@ -13,6 +13,7 @@ import {
   Text,
   View,
   WebView,
+  Platform,
 } from 'react-native'
 import * as c from '../../components/colors'
 import Icon from 'react-native-vector-icons/Ionicons'
@@ -140,8 +141,6 @@ export default class KSTOView extends React.PureComponent<void, Props, State> {
       </Text>
     ) : null
 
-    const playButton = this.renderPlayButton(this.state.playState)
-
     return (
       <ScrollView
         contentContainerStyle={[styles.root, sideways && landscape.root]}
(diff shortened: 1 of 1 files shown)


source/views/streaming/radio/index.js
```diff
@@ -165,12 +171,22 @@ export default class KSTOView extends React.PureComponent<void, Props, State> {
...
 
-            <ScheduleButton
+            <SmallActionButton
+              icon="ios-call"
+              text=""
+              style={buttonStyles.smallButton}
+              onPress={this.callStation}
+            />
+
+            <View style={buttonStyles.space} />
```
drewvolz: There has to be a better way. See https://github.com/StoDevX/AAO-React-Native/pull/1772#discussion_r144191128 as to why.

source/views/streaming/radio/index.js
```diff
@@ -165,12 +171,22 @@ export default class KSTOView extends React.PureComponent<void, Props, State> {
 
             <View style={buttonStyles.space} />
 
-            <ScheduleButton
+            <SmallActionButton
+              icon="ios-call"
```
hawkrives: This icon should be changed per-platform.

source/views/streaming/radio/index.js
```diff
@@ -165,12 +171,22 @@ export default class KSTOView extends React.PureComponent<void, Props, State> {
...
+              text=""
+              style={buttonStyles.smallButton}
+              onPress={this.callStation}
+            />
+
+            <View style={buttonStyles.space} />
+
+            <SmallActionButton
               icon="ios-calendar"
```
hawkrives: This icon should be changed per-platform.

source/views/streaming/radio/index.js
```diff
@@ -377,7 +393,7 @@ type ActionButtonProps = {
   onPress: () => any,
 }
 
-type ScheduleActionButtonProps = {
+type SmallActionButtonProps = {
   icon: string,
   text: string,
```
hawkrives: This should be an optional property, since it's optional, and it should then be omitted from the two calls in this file.

hawkrives (changes requested):
Also, how does this look in landscape on a 4s?

source/views/streaming/radio/index.js
```diff
@@ -18,10 +18,12 @@ import * as c from '../../components/colors'
 import Icon from 'react-native-vector-icons/Ionicons'
 import {Touchable} from '../../components/touchable'
 import {TabBarIcon} from '../../components/tabbar-icon'
+import {phonecall} from 'react-native-communications'
 import type {TopLevelViewPropsType} from '../../types'
 
 const kstoStream = 'https://cdn.stobcm.com/radio/ksto1.stream/master.m3u8'
 const image = require('../../../../images/streaming/ksto/ksto-logo.png')
+const stationNumber = '5077863602'
```
rye: Please change to `+15077863602`. (Add prefix `+1` country code for international users.)  Doubt it'll actually make a difference for these users due to system-wide settings, but it's nice to have.

source/views/streaming/radio/index.js
rye: Agreed.  It works in my Android emulator, but it'd be nice to be platform-specific or general.

rye (changes requested):
Just a couple of additions to @hawkrives' feedback.

hawkrives:
I'm going to merge this into the downstream branch, as I've addressed the feedback.

PR #1775, opened 2017-10-12
Add User-Agent Header to fetch and fetchJson calls

### LINKED ISSUE #1757: Set our own custom User-Agent string
We should do this.  This is a good way of telling what versions our users are using, among many other things.

rye:
This is an exported constant that can be used anytime we want to make requests.

Being that this is my first functional (i.e. non-data, non-metadata) change, I'm completely open to feedback.  Also, I want to send this with all network requests, preferably.

If someone has additional information we should send, too, I can do that.

@hawkrives, if you want and are able, you could just "fix" the SIS integration server-side for all phones with the `AllAboutOlaf/` header set, since previous versions won't have it.

*Closes #1757.*

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "all-about-olaf",
-  "version": "2.4.0",
+  "version": "2.4.1-pre",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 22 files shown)


drewvolz (approved):
:+1: :shipit:

hawkrives:
@rye My main thought is that it seems like this'd be better set in https://github.com/StoDevX/AAO-React-Native/blob/master/source/globalize-fetch.js, by removing the `...args` in global.fetch and replacing it with `url, opts`, then merging the opts together and setting User-agent (if it's not already set). Does that make sense?

rye:
Yeah, that makes way more sense to me than the way I did it. I was just hesitant to change that method. We could merge this and then make that change, or I can make the change on this PR. (I'd favor the former just so it gets merged ASAP and I'd then do the latter refactoring step tomorrow, but I'm fine with whatever is decided.)

hawkrives:
Is merging ASAP a priority? Even when we do merge this, it's not going live until 2.4.1, right? If merging ASAP is a priority, then we can do it this way and then generalize it, but I'd rather not do this and then immediately undo it unless we need to.

rye:
Well, it depends. I might not be able to get the refactor done and fully tested to the level the other one was for the nightly build tonight, which is why I lean more towards merging sooner than later so we know everything works "in production."  There's also the possibility of doing more with fetching feature-wise, like consolidating error handling. But, I can hold off.

The other side of it is, I'd like to make sure build numbering works and that we can get real results from this change—a nightly would be a good place to do this.  It also allows IT's side of SIS balances to be enabled based on UA sometime tomorrow, rather than after or during the break.

Also, generally, I'm all for shipping this tonight and a rewrite tomorrow—I'll happily do both.  It's just a matter of style, and PR's are free so I don't see why shipping this and doing an about-face later is necessarily bad besides additional CI cycles.  I'm basically completely 50–50 on this one, so whatever other people think will work for me.

hawkrives:
I don't know that I'll have bandwidth to do anything to /olecard tomorrow; I'll be basically fighting fires on the new site launch all day. (Especially since we haven't actually launched it yet, and it's still going to go live today.)

I would prefer to centralize the headers into globalize-fetch before we merge; I know the format that the user agent will take, and so sometime between now and whenever we merge this I will take a few minutes and update my regex to handle our new UA string.

hawkrives (approved):

PR #1777, opened 2017-10-13
Add "College" after the colleges on the Express schedule

rye:
I don't think there's a reason why we shouldn't do this.  It's this way on the Blue schedule, so I figured I'd make this somewhat consistent.  I could also go the other way. (drop "College" from both)


hawkrives (approved):
I'm good with adding the word

rye:
Alright, I'll merge.

PR #1787, opened 2017-10-17
Update St. Olaf calendar URL

hawkrives:
We have changed calendar backend systems, so the old URL will go away soon. This is, I believe, the new ID for the master St. Olaf calendar.

It has the (dis)advantage of including Alumni and Athletics events, but we're not sure how to exclude them yet.

### DIFF
--- a/source/views/calendar/index.js
+++ b/source/views/calendar/index.js
@@ -18,7 +18,7 @@ export default TabNavigator(
       screen: ({navigation}) => (
         <GoogleCalendarView
           navigation={navigation}
-          calendarId="le6tdd9i38vgb7fcmha0hu66u9gjus2e@import.calendar.google.com"
+          calendarId=" 5g91il39n0sv4c2bjdv1jrvcpq4ulm4r@import.calendar.google.com"
         />
       ),
       navigationOptions: {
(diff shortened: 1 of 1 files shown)


drewvolz:
> the old URL will go away soon

Do you have a date in mind?

rye (approved):

drewvolz:
> It has the (dis)advantage of including Alumni and Athletics events, but we're not sure how to exclude them yet.

The `summary` prop contains `St. Olaf College Women's` or `St. Olaf College Men's` as the first bit of text on all athletic events. Also, the `description` prop contains  `athletics.stolaf.edu`. That can get us a pretty far with filtering, but it would be great if that were server-side.

drewvolz:
Is the next logical step to write a parser for tribe events?

hawkrives:
By "go away soon" I mean that the old calendar won't be updated anymore starting next week. The URL will still work, but event updates and new events won't show up.

I'd rather not write anything for Tribe's events plugin, tbh. GCal works, and it works well, and Google doesn't randomly die for 30 second intervals.

We're hoping to eventually get a proper feed without Alumni and Athletics events, but for now this is what we've got.

drewvolz:
I hear you. I understand that GCal works well. I feel we would gain flexibility to separate out event categories thru Tribe's feed and access to a few more pieces of event info. But I understand if we do not do that.

hawkrives:
I'm just currently of the opinion that I'd prefer the stability of GCal over any added features of TEC.

hawkrives:
That opinion might change as time goes on.

rye:
My review here stands—filtering could always come later.

PR #1791, opened 2017-10-17
Update package-lock.json

rye:
_Mumble grumble_.

Seems like `greenkeeper-lockfile-upload` really doesn't like our scheme for builds, so we get to keep doing this manually for the time being.


hawkrives:
Lol.

We'll want to figure this out at some point; npm, Inc is working on a thing tentatively called `cipm`, for use in CI environments, which will have a _major_ speedup when installing fresh.

https://twitter.com/maybekatz/status/905833870142078976

https://twitter.com/maybekatz/status/919306327331139584

It completely ignores the package.json, installing only from package-lock.json, so if we want to use it (which we do) we'll have to make the CI update the package-lock for us.

But, for now, we can continue to do it manually! :)

hawkrives (approved):

Issue #1792, opened 2017-10-17
New OleCard balances data fetching API


hawkrives:
I just made a JSON endpoint for the OleCard check balances screen.

No longer will we suffer under the tyrrany of parsing HTML on it!

*cough*

Anyway.

Hitting https://www.stolaf.edu/apps/olecard/checkbalance/, logging in, and then `GET`ing https://www.stolaf.edu/apps/olecard/checkbalance/api.cfm will return a response that matches the following typedef:

```typescript
type MealPlanInfo = {
	plan: ?string,
	leftDaily: ?string,
	leftWeekly: ?string,
}

type AccountBalance = {
	account: string,
	numeric: number,
	formatted: string,
}

type OleCardBalances = {
	data: {
		accounts: Array<AccountBalance>,
		meals: ?MealPlanInfo,
	},
	error: ?string,
}
```

As an example, here is what I get when I open that page:

```json
{
  "data": {
    "accounts": [
      {
        "account": "STO Ole Dollars",
        "numeric": 3150,
        "formatted": "$31.50"
      }
    ],
    "meals": null
  },
  "error": null
}
```

hawkrives:
I'd really like to get this in to 2.next, but I'll completely understand if we want to push it to 2.next+1.

If someone wants to do this this afternoon, I can cut an RC this evening, we can test, and then we can submit the build later tonight and hopefully have it approved tomorrow.

rye:
I feel like this could go to 2.5 or 2.6, seems suitable.  Along with platform changes like RN upgrades, this'll give us a hearty amount of stuff to do.  But yeah, a JSON API beats HTML parsing any day.

PR #1796, opened 2017-10-18
Olecard json endpoint

hawkrives:
Supersedes https://github.com/StoDevX/AAO-React-Native/pull/1794
Closes https://github.com/StoDevX/AAO-React-Native/issues/1792

I … haven't actually tested this; I need to do some actual work now.

But I think all the logic is good now.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5370,7 +5370,7 @@
       "requires": {
         "hoek": "2.16.3",
         "isemail": "1.2.0",
-        "moment": "2.18.1",
+        "moment": "2.19.1",
         "topo": "1.1.0"
       }
     },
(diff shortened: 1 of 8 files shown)


rye (commented):
| Logged out | Logged in |
|---|---|
| screen shot 2017-10-18 at 11 38 53 am | screen shot 2017-10-18 at 11 37 58 am |

Is there any way we can change `17 meals/week-$130 Flex$'s/semester` to something more pleasant?

rye:
I realize that they come this way from the API, but I think we might be able to do parsing to make it better.

hawkrives:
I'd like to do that parsing server-side, I think?

hawkrives:
I'll find a list of the options and clean them up in the responses there

rye:
Doing that server-side would be very helpful.

rye (approved):

PR #1798, opened 2017-10-18
Make iOS open the webcams page (like Android)

hawkrives:
Broadcast Media is hoping to launch their new mobile-friendly theme next week, so let's prepare for that!

BM would like to have the analytics for the page, and the main reason (IIRC) that iOS opens teh stream directly is that the page was hard to navigate on mobile. With a mobile-friendly website, that concern goes away.

### DIFF
--- a/source/views/streaming/webcams.js
+++ b/source/views/streaming/webcams.js
@@ -132,14 +132,8 @@ type ThumbnailProps = {
 
 class StreamThumbnail extends React.PureComponent<void, ThumbnailProps, void> {
   handlePress = () => {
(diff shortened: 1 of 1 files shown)


hawkrives:
I don't know what Travis is doing.

Right now, I've got Danger and CodeClimate reports, and no travis status at all.

Weird.

drewvolz:
I saw that, too. Travis seems to be back.

rye (approved):

hawkrives:
yay travis

drewvolz (approved):

PR #1818, opened 2017-10-26
Build `greenkeeper/` branch pushes

rye:
This PR enables branch builds for any branches matching `/^greenkeeper\//`.

This could be counteracted by disabling PR builds through a similar maneuver on `/^greenkeeper\//` PR's instead, thereby forcing builds for only pushes on these branches.

Considering that our use of this strategy differs from the sterile configuration of Travis, I think this might actually solve things and get Greenkeeper to update the lockfile from time to time.  At least, I saw promising results for [this branch's JS build](https://travis-ci.org/StoDevX/AAO-React-Native/jobs/293184175#L719):

```console
No dependency changed.
```

And, Greenkeeper actually attempted to log in, too! I'd call that at least a slight win.


rye:
Yeah, it looks like Travis now dutifully builds at least twice on `greenkeeper/` branches, `n`-ce for pushes and once for the PR.

Here are the relevant lines from Travis' [Requests](https://travis-ci.org/StoDevX/AAO-React-Native/requests) panel for this PR:

screen shot 2017-10-26 at 9 54 47 am

rye:
If building twice on Greenkeeper branches is inconvenient, I'd argue that we could easily tweak our build scripts and tell them not to run _everything_ on _all builds_.  (I.e. don't _actually_ go through the full platform build steps for Greenkeeper pushes.)

hawkrives:
For _now_, I'd say let's merge this and let Travis build twice on Greenkeeper PRs.

They're rarely urgent anyway.

---

Longer-term, this seems like a good time to move to Travis' [Build Stages](https://docs.travis-ci.com/user/build-stages/). Not because of the stage abilities, since most of our time is spent either booting or running one long process, but because you can specify [`if:` conditions for jobs](https://docs.travis-ci.com/user/conditional-builds-stages-jobs/#Conditional-Jobs).

---

`greenkeeper-lockfile` appears to push a new commit with the updated package-lock.json file.

Given that, I'd propose that we do two things:

1. Continue to run both PR and push builds for Greenkeeper PRs
2. Take advantage of the new [`if:` conditional jobs](https://docs.travis-ci.com/user/conditional-builds-stages-jobs/#Conditional-Jobs) feature to differentiate between PR and push builds

Our current builds look something like this:

```yaml
jobs:
  include:
    - os: linux
      env: ["JS=true"]

    - os: linux
      env: ["ANDROID=true"]

    - os: macos
      env: ["IOS=true"]
```

I'm proposing something like the following:


```yaml
jobs:
  include:
    - os: linux
      env: ["JS=true"]
      if: (NOT branch =~ ^greenkeeper\/) OR (type = pull_request)
    
    - os: linux
      env: ["ANDROID=true"]
      if: (NOT branch =~ ^greenkeeper\/) OR (type = pull_request)
    
    - os: macos
      env: ["IOS=true"]
      if: (NOT branch =~ ^greenkeeper\/) OR (type = pull_request)

    - os: linux
      env: ["GK=true"]
      if: branch =~ ^greenkeeper\/ AND type = push
```

So we'd wind up with, I think,

- Greenkeeper pushes new commit
- Travis starts running the GK=true job, skipping the others because it's a "push" build
- Travis starts running the JS/IOS/ANDROID builds, because it's a "pull_request" build
- The GK job finishes and pushes a new commit to the branch
- Travis starts running the GK=true job, skipping the others because it's a "push" build
	- `greenkeeper-lockfile` notices that it's already run, so it ends quickly
- Travis starts running the JS/IOS/ANDROID builds, because it's a "pull_request" build
	- It cancels any of the previous PR jobs that have not yet started (probably just iOS)
	- They now build with the updated package-lock.json
	- and with `cipm`, it will completely ignore package.json, so we'll know that it's a reproducible build

My conditional syntax probably needs some tweaking, but how's this sound overall?

rye:
> [...] this seems like a good time to move to Travis' [Build Stages](https://docs.travis-ci.com/user/build-stages/).

Yes, please.  This is an analog to a feature in GitLab CI that I have sorely missed in our Travis flow.

Cue long-running digression containing me extolling the virtues of build stages and pipelining:

- **Pipelining is something that I think that is perfect for our use-case** — we could run tests, CodeClimate, linting, flow, etc. all in one stage, and in our next stage (since we have already proven that the code successfully builds once) we could run the "deployment"-type things — bundling data, building app files, etc. (not sure where that fits into our pipeline but I'll investigate our current flow)
  - The chief problem this introduces lies within dependency management, since NPM and the like are notoriously slow and scary when it comes to these things.  Luckily, `node_modules` is cacheable.
  - **I think this would simplify things in the long run**; more ephemeral and focused builds may lead to quicker build failures and diagnoses.  This way, iOS builds and Android builds, the notorious time consumers, would only run in situations where all prior stages had successfully passed.  This would prevent us from waiting 30 minutes when we _already know_ that the build will fail.
    - Embracing ephemerality in CI would involve us cutting things apart and making our build scripts more focused and less monolithic.
  - I've already been using workflows like this in GitLab CI and it's a life-saver.  Docker builds are not nearly as bad as Android/iOS builds are, but they get the same treatment in my private API workflow — they come last, after everything else has succeeded.

Back to the thing at hand:

> Not because of the stage abilities, since most of our time is spent either booting or running one long process, but because you can specify [`if:` conditions for jobs](https://docs.travis-ci.com/user/conditional-builds-stages-jobs/#Conditional-Jobs).

`if` conditions are great, though I think these could be simplified by more heavily embracing the magic of build stages.

My ideal would be to have: (i) a `Test` stage to contain anything like Jest, Flow, ESLint, CodeClimate, etc., running in parallel, and then (ii) a `Build + Deploy` stage to contain everything like bundling data, building + deploying Android, building + deploying iOS, etc.

The advantage this would afford us is mentioned above:

> iOS builds and Android builds, the notorious time consumers, would only run in situations where all prior stages had successfully passed.

And this is huge.  Cutting things up into separate jobs would allow the JS build tools to run in parallel and provide an earlier cancellation for the post-success stage if something breaks downstream.  Like I say, I've already seen huge speed-up for getting failed build reports by switching to this kind of structure, and while caching might be a bit of a pain to implement, I think it's worth it.  Plus, using caching and the like may help to quell fears about boot times.

 - Another thought I had was about integration tests in iOS and Android, if we're doing that.  Those would be post-tests after the app had built, so they'd go with the Build + Deploy stage.  As we add more and more tests, I think we'll see further slowdown earlier on, and getting the same failure three times is redundant.

- - -

**In general, I like @hawkrives' suggestion, and I'd say it should be implemented first**, but I'd also argue that it might be time to look into using two distinct build stages, each with their own matrix of things to check.

hawkrives:
Just one response real quick:

> The chief problem this introduces lies within dependency management, since NPM and the like are notoriously slow and scary when it comes to these things. Luckily, node_modules is cacheable.

image

That's what `cipm` (from npm inc) is designed to help with. (from https://twitter.com/maybekatz/status/905833870142078976)

hawkrives (approved):
aproooooove

hawkrives:
This PR probably has the highest comment-word-count to lines-changed ratio in here

rye:
Yeah, `cipm` would make my argument for splitting things up into more jobs even stronger, as install times would be even shorter.

PR #1820, opened 2017-10-26
Use Travis build stages

hawkrives:
This lets us conditionally run certain stages.

Currently, we only run the greenkeeper stage on branches matching `^greenkeeper/`, and we only run the "build gh-pages data" stage on master.

I also properly centralized the definition of the node version (upon realization that Travis just passes whatever you give as `node_js: ???` directly as `nvm install ???`, so I can use env variables), and I moved the Ruby version def next to it.

I removed the Android emulator job for the time being. We can set it up again when/if we ever want an android emulator on Travis. Travis documents how to do it.

I moved the js-build/check job to occur _before_ the Android/iOS jobs, so that if the JS fails we don't queue up for a macOS machine.

I moved the "deploy" task into a its own stage, which runs only on `master` and between the js build and the iOS/Android builds. I could also move it to run simultaneously with the iOS/Android builds (@rye, thoughts?)

Because I removed the Android emulator, I removed the `CAN_DEPLOY` variable, because we no longer have iOS/Android jobs that must never deploy. (`CAN_DEPLOY` used to prevent both Android jobs from attempting to deploy a new build; it was part of the input to the `run_deploy` variable.)

I removed `npm config set spin=false progress=false`, figuring that npm's probably disabled those automatically on CI envs now.

I moved `npm run danger` to `after_script` on the js build, instead of being the last item in `script`, which should have no differences in behavior.

I removed the apparently-not-working my-heroku-based build time tracker that I don't have the energy to investigate, as well.

I removed the travisci_deploy_key* files, as we no longer serve up the keys needed to unlock them.

I changed `npm install -g npm@5.4` to just `@5`.

I print the before/after `ruby` versions, like we do `node`.

I think that's all.


hawkrives:
https://travis-ci.org/StoDevX/AAO-React-Native/builds/293312621?utm_source=github_status&utm_medium=notification

Whoops. That shouldn't be deploying data.

EDIT: Fixed

```diff
   - name: Deploy Data
-    if: branch = master
+    if: branch = master AND type != pull_request
   - name: Build + Deploy App
```

rye:
Unless I comment on it below, the comment above sounds good!

> I moved the "deploy" task into a its own stage, which runs only on `master` and between the js build and the iOS/Android builds. I could also move it to run simultaneously with the iOS/Android builds (@rye, thoughts?)

"Deploy" in this case refers to data bundling and updating?  If so, that makes sense to me.  I'd say it should be deployed whenever the JS tests were successful, whether or not that takes place before or during the time when the apps build is a triviality.

> Because I removed the Android emulator, I removed the `CAN_DEPLOY` variable, because we no longer have iOS/Android jobs that must never deploy. (`CAN_DEPLOY` used to prevent both Android jobs from attempting to deploy a new build; it was part of the input to the `run_deploy` variable.)

Very nice.  If we add the Android emulator back, we can making this more modular and making it not require its own global variable.

> I removed `npm config set spin=false progress=false`, figuring that npm's probably disabled those automatically on CI envs now.

Has this been verified?  We can do `npm install --no-progress` if it is still there.

I'll submit a review shortly.

hawkrives:
> > I removed npm config set spin=false progress=false, figuring that npm's probably disabled those automatically on CI envs now.
>
> Has this been verified? We can do npm install --no-progress if it is still there.

I have just verified that there is no terminal junk from running `npm install`. We no longer need those two commands.

hawkrives:
> "Deploy" in this case refers to data bundling and updating? If so, that makes sense to me. I'd say it should be deployed whenever the JS tests were successful, whether or not that takes place before or during the time when the apps build is a triviality.

That's what I've aimed for with the separate stage between JS and native, so we're in agreement there.

.travis.yml
```diff
@@ -8,29 +6,109 @@ env:
...
     - PATH=${PATH/\.\/node_modules\/\.bin/}
-    # variables that are used later on
+    # sets our node version in one nice place
     - TRAVIS_NODE_VERSION=8
-    - PUSH_BRANCH=0
-    - CAN_DEPLOY=0
-    - USE_EMULATOR=no
-    - run_deploy=0
+    # set the ruby version up top
+    - TRAVIS_RUBY_VERSION=ruby-2.4.1
```
rye, line 12: MRI 2.4.2 is the latest version, includes some security fixes.  Let's use that?

rye (approved):
Seems good to me.

.travis.yml
hawkrives: Can we do that in another PR? I like running latest. I just don't want to wait for yet another build to complete here :)

rye: PR into this PR?  Or into `master`?

hawkrives: `master` after we merge this, methinks.

PR #1823, opened 2017-10-27
Remove "npm i -g npm@5" steps

hawkrives:
Now that node bundles up to date versions on npm, we don't need to update them. 

Node 8.8 is currently bundling npm 5.4.something, which is plenty recent for my taste. 

Eventually, we'll install cipm, but not until it's released ;)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,7 +57,7 @@ jobs:
       node_js: $TRAVIS_NODE_VERSION
       env: [JS=yes]
       before_install:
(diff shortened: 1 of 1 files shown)


hawkrives:
The npm-install step on the JS job just dropped from ~15s to ~3. Me like.

rye (approved):
Yeah, and I'd expect Node and npm to keep synchrony pretty well.  Making things faster is nice.

PR #1828, opened 2017-10-28
Print the fastlane lane picked by auto_beta

hawkrives:


rye (approved):

rye:
It appears that `run_deploy` is `nil`, which produces [this line in Travis](https://travis-ci.org/StoDevX/AAO-React-Native/jobs/294348464#L3456).

As a result, the build goes straight to `just build`, and does not get deployed.

hawkrives:
Hmm. It should have been set by https://travis-ci.org/StoDevX/AAO-React-Native/jobs/294348464#L3188, though… 

The only think that should have changed there was that I removed the default run_deploy value when we went to stages. It shouldn't have affected the conditional one…

PR #1832, opened 2017-10-31
Put run_deploy logic into fastlane/lib/util

### LINKED ISSUE #1830: Missing nightly builds
As of the recent transition to build stages, we no longer get the cron-triggered nightly builds.

I propose that we remove this variable from `.travis.yml` and bake it directly into the Fastlane util.  Or, we figure out why `run_deploy` isn't getting properly found all of a sudden.

rye:
I don't know how to feel about this.  For one, it consolidates the logic into the same place it's used, but it does introduce concern about Travis into Fastlane, and if we want the two to be decoupled, (I can't think of a reason why) that might work.

However, we know that this should work, unless something is wrong in the logic itself.

I'd say this is PR is mutually exclusive with #1831 unless that fails to fix #1830, in which case I'd want that reverted.  I'll be offline for the next ~6 hours so if someone decides they want to merge this instead of #1831, feel free, just keep in mind that both shouldn't be there.

*Closes #1830.*

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,6 @@ env:
     - TRAVIS_NODE_VERSION=8
     # set the ruby version up top
     - TRAVIS_RUBY_VERSION=ruby-2.4.2
-    # always set run_deploy
-    - run_deploy=0
     # GH_TOKEN:
     # - used by greenkeeper to update package-lock.json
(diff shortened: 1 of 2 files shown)


hawkrives:
@rye I'm going to merge #1831, because now I'm curious as to if it'll fix the problem. (It shouldn't…)

I'd like to leave this open, though, because I would like to move the logic into Fastlane later. 

(I swear I had a reason for writing it in Bash before, but I can't remember why – maybe back when I controlled the deploy stuff from bash too? Idk.)

rye:
@hawkrives should we wait for another cron to run through with the changes in #1831?  If not, I can merge with `master` and basically revert the changes in #1831.  This would move the logic into the Fastlane thingy.

hawkrives:
@rye Yes, I would like to wait for another cron to run before we merge this.

hawkrives (approved):

hawkrives:
(I re-added the `pr/don't merge` label so that people [read: so that I] don't come through and merge this early. 😄)

drewvolz (approved):

rye:
Okay, so if the cron job does not successfully deploy tonight, we will merge this tomorrow morning.

hawkrives:
@rye Yup, sounds good.

drewvolz:
The cron succeeded.

hawkrives (approved):

Issue #1508, opened 2017-09-04
Store bus route colors in bus route files


hawkrives:
instead of separately, in the interface itself.

This would warrant a change to the schema, as well.

PR #1834, opened 2017-11-02
Update Bus view for flow@0.53

hawkrives:
TODO:

- [ ] Update bus route schema for new keys
- [ ] Test this on a simulator
- [ ] Fix tests to use the same time zone (https://travis-ci.org/StoDevX/AAO-React-Native/jobs/296101642#L2632-L2635)

---

Closes https://github.com/StoDevX/AAO-React-Native/issues/1508

This PR targets the React-Native 0.49 upgrade branch, not `master`.

I did a bit of tweaking to the Bus data files; they now define the route colors internally, instead of looking them up in the JS (what).

They also now store the coordinates as a key/value pair, so that we don't have to duplicate the coordinates for out/back stops.

---

As to the rest of the PR… well. I reworked the internals of the Bus view pretty significantly, and added a few more tests.

Our data is formatted for humans, which means that the stops are defined separately from the stop _times_, so that you can look down the arrays to line them up visually.

However, this view works on a stop-by-stop basis.

Before, what we'd do is pick out the times that went with each stop as we rendered that stop.

With this patch, the first thing we do is to process the two arrays into a "timetable" array of objects that contains all the data needed to render the rows.

This colocation helps to reduce the amount of data we need to pass around and recalculate, and I think that it really simplifies the internals of this view.

… more of a description to come?


hawkrives:
I should note that this PR won't be green when it merges into the update branch; there should be no flow errors in /transportation, but there will still be a number of errors outside of the scope of this PR.

hawkrives:
I'm going to merge this, continue work in the other branches, and maybe point back here when it's finally ready.

PR #1836, opened 2017-11-02
Alternative icon switching in iOS Settings

drewvolz:
Alternative icons allow for us to embed images within the Xcode project and switch them to be the app's icon at runtime.

I first started creating [my own library](https://github.com/drewvolz/react-native-dynamic-icon) to accomplish this but later discovered a [fully functional RN library](https://github.com/kamsteegsoftware/react-native-alternate-icons) after I had achieved the basic goal. I think we should opt to use and improve the other library as we continue to make use of it.

This feature is available in iOS 10.3+. There is a check on the Settings view to hide the row if the OS requirements are not met.


source/views/settings/icon.js
```diff
@@ -0,0 +1,95 @@
+// @flow
+import React from 'react'
+import {ScrollView, Image, StyleSheet, Text} from 'react-native'
+import Icons from 'react-native-alternate-icons'
```
hawkrives: This should be "import * as" (yes, that's not what their docs say, but their docs are wrong)

source/views/settings/icon.js
```diff
@@ -0,0 +1,95 @@
...
+  } = {
+    iconType: '',
+  }
+
+  componentWillMount() {
+    this.getIcon()
+  }
+
+  setIcon(iconType: string) {
+    iconType == 'default' ? Icons.reset() : Icons.setIconName(iconType)
```
hawkrives: ===

hawkrives: Actually, I think I'd like this to go into a full-fledged if-else statement – something about executing side-effects in a ternary feels odd

source/views/settings/sections/odds-and-ends.js
```diff
@@ -9,13 +9,27 @@ import {connect} from 'react-redux'
 import {CellToggle} from '../../components/cells/toggle'
 import {PushButtonCell} from '../../components/cells/push-button'
 import {trackedOpenUrl} from '../../components/open-url'
+import Icons from 'react-native-alternate-icons'
```
hawkrives: `import * as`

source/views/settings/sections/odds-and-ends.js
```diff
@@ -29,11 +43,24 @@ class OddsAndEndsSection extends React.Component {
...
 
   render() {
+    if (this.state.loading) {
+      return <View />
+    }
+
     return (
       <View>
         <Section header="MISCELLANY">
+          {this.state.supported ? (
```
hawkrives: `!this.state.loading && this.state.supported` IMHO

source/views/settings/sections/odds-and-ends.js
```diff
@@ -29,11 +43,24 @@ class OddsAndEndsSection extends React.Component {
       url: 'https://github.com/StoDevX/AAO-React-Native',
       id: 'ContributingView',
     })
+  onAppIconButton = () => this.onPressButton('IconSettingsView')
 
   render() {
+    if (this.state.loading) {
+      return <View />
+    }
```
hawkrives: Remove because of next comment

source/views/settings/icon.js
```diff
@@ -0,0 +1,95 @@
...
+  },
+  icon: {
+    width: 16,
+    height: 16,
+    borderRadius: 5,
+    borderWidth: 0.5,
+  },
+})
+
+const icons = [
```
hawkrives: I almost feel like this should be defined in /images, as we have other JS/image mappings there…

source/views/settings/icon.js
```diff
@@ -0,0 +1,95 @@
...
+    paddingVertical: 10,
+  },
+  title: {
+    fontSize: 16,
+  },
+  icon: {
+    width: 16,
+    height: 16,
+    borderRadius: 5,
+    borderWidth: 0.5,
```
hawkrives: Add an explicit borderColor plz

hawkrives: Also use Stylesheet.hairlineWidth

source/views/settings/icon.js
```diff
@@ -0,0 +1,95 @@
...
+                )
+              }
+              accessory={
+                includes(this.state.iconType, val.type)
+                  ? 'Checkmark'
+                  : undefined
+              }
+              cellStyle="RightDetail"
+              cellContentView={
+                <Column style={styles.content}>
```
hawkrives: Do we actually need this custom cell-ness? Can't we just use the stock cell view with "image" and "accessory"?

source/views/settings/icon.js
hawkrives: To follow up on that, this logic has a race condition. Because the library uses callbacks, not promises, the only way to guarantee that come code will be executed after other code is to run it _in_ the callback. You'll either need to wrap the library in promises (use pify, works well) or just call this.getIcon in the callback of each logic branch here.

ios/AllAboutOlaf.xcodeproj/project.pbxproj
```diff
@@ -753,6 +759,7 @@
 			children = (
 				742B362B86E443C09A7F9FD9 /* Entypo.ttf */,
 				E6FBE71A05CC43E2B4885180 /* Ionicons.ttf */,
+				3A8C0CF21F9EF65C0064B66C /* windmill.png */,
```
hawkrives: Do we really only need one icon file?

Should/could the icon go into images.xcassets as an app icon, instead?

hawkrives (dismissed):
Looks good overall!

source/views/settings/icon.js
hawkrives: Wait, do these not accept callbacks?

hawkrives: Despite using completion handlers (thus callbacks) in the objc code, these two functions do not take a callback. 

 We'll need to PR that upstream before I'm comfortable accepting this, sadly.

hawkrives: https://github.com/kamsteegsoftware/react-native-alternate-icons/blob/122be1d1296efb87dcaf0075467b879a1be25c0d/ReactNativeAlternateIcons.m#L18

source/views/settings/icon.js
drewvolz: Agreed

ios/AllAboutOlaf.xcodeproj/project.pbxproj
drewvolz: I was trying to limit duplication of files. I would be fine creating a separate space for them. Also I am unsure how to reference the file if it’s in xcassets.

On another note, should we update the icon used in the credits view based off of the selected iOS icon?

source/views/settings/icon.js
drewvolz: Okay. We’ll PR for callbacks.

source/views/settings/icon.js
drewvolz: Probably?

source/views/settings/sections/odds-and-ends.js
drewvolz: I didn’t like this to begin with. Thanks for a better idea.

drewvolz:
The build will fail due to flow issues caused by the package we are depending on. They need their prop types annotated.

ios/AllAboutOlaf.xcodeproj/project.pbxproj
hawkrives: >On another note, should we update the icon used in the credits view based off of the selected iOS icon?

huh, we could do that, couldn't we. That'd be neat.

> Do we really only need one icon file?

I was mainly asking if we really only needed one file, because the main icon has to come in 24, 44, 96, 48, 1024, and some other sizes, for things like iPhone homescreen (@{1,2,3}x), iPad homescreen (@{1,2}x), Spotlight (@{1,2,3}x), Settings (@{1,2,3}x) and maybe others.

source/views/settings/icon.js
drewvolz: I've changed `setIcon` to an if/else block to avoid executing side-effects in a ternary. Although this github block collapses, we'll still need to PR upstream to get the objc code to take a callback.

source/views/settings/icon.js
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  setIcon(iconType: string) {
+    if (iconType === 'default') {
+      Icons.reset()
+    } else {
+      Icons.setIconName(iconType)
+    }
+
+    this.getIcon()
```
hawkrives: We need to PR upstream to allow reset and setIconName to take callbacks

drewvolz: Progress has been made at https://github.com/hawkrives/react-native-alternate-icons/tree/hawken

hawkrives: You should update this to use my fork maybe so we can continue to prototype; we can eventually switch back once we PR it and it gets approved / released.

hawkrives: Mine is published on npm as `@hawkrives/react-native-alternate-icons`

rye (commented):
I have nothing to add, really, to @hawkrives' review.  As a bit of commentary, in general, I'd like us to add new features like this for both platforms rather than just iOS.  Android API 26 introduces [Adaptive Icons](https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive.html) which could be used, but this request is not specifically relevant to this project's discussion and should rather be implemented "upstream." (in the rn package)

But anyhow, this is a cool feature, and looks fine to me.

hawkrives:
@rye I’m pretty sure that adaptive icons are completely different - unless I’m misreading the description, you only get one piece of iconography, and the system uses the extra data you provide to alter it slightly - reframe it, or wiggle, or whatever. 

The iOS feature is for providing entirely separate pieces of iconography. If that feature exists on android too, then I’d be happy to implement support for it. (Well, maybe not me personally, but maybe me personally.)

hawkrives:
@drewvolz I have confirmed that this feature expects you to provide alternative icons in every size. 

<https://developer.apple.com/ios/human-interface-guidelines/icons-and-images/app-icon/>:
> To ensure that alternate icons appear consistently throughout the system—the user shouldn't see one version of your icon on the Home screen and a completely different version in Settings, for example—provide them in the same sizes you provide for your primary app icon (with the exception of the App Store icon).

drewvolz:
@hawkrives, drat... thanks for letting me know.

hawkrives:
@drewvolz I think I have a partial script in the repo somewhere to generate a bunch of sizes scaled down from the 1024px one.

drewvolz:
Maybe this? https://github.com/StoDevX/AAO-React-Native/blob/master/scripts/make-icons.js

hawkrives:
Probably that. Looks like it takes an input filepath, but then it always overwrites the main app icons.

hawkrives:
I think I forgot to relink the library after merging master.

hawkrives (commented):
What about the 1x icon for the non-retina iPad mini (runs iOS 9)?

drewvolz:
Every app example I've seen so far provides a 2x and 3x version of alternative icons. Although Apple's docs says to have all the sizes except for the iTunes size, I'm not sure what we should do. This seems to work, fwiw. We can provide 1x, but we might as well just provide files that are named with the `~ipad` scheme in that case. Again, I'm not sure what the right way to do this is.

hawkrives:
I’d say we should experiment. 

Make very distinct icons in photoshop that say exactly what they are, or are blocks of solid color or something, then save them in all the ways we can think of, then run the app on different simulators and see what icon gets used.

source/views/settings/sections/odds-and-ends.js
```diff
@@ -28,11 +46,21 @@ class OddsAndEndsSection extends React.PureComponent<Props> {
...
       id: 'ContributingView',
     })
+  onAppIconButton = () => this.onPressButton('IconSettingsView')
 
   render() {
     return (
       <View>
         <Section header="MISCELLANY">
+          {this.state.supported ? (
+            <Cell
```
hawkrives: Why use `<Cell/>` here instead of `<PushButtonCell/>`?

hawkrives (commented):
One question, then I think we're good to go.

~~Well, I also want a @1x version of the icon – the app still runs on the non-retina iPad Mini 1 and iPad 2 😅.~~ (addressed by https://github.com/StoDevX/AAO-React-Native/pull/1836#issuecomment-345591507)

Anyway, I think we can start out with these icons and play around with figuring out how the system actually uses icons in a future issue / PR. I'd like to get this in to 2.5.

source/views/settings/sections/odds-and-ends.js
hawkrives: Ah, because you have `cellStyle="RightDetail"`.

hawkrives: But you don't have a `detail=` prop, so I'd like to request that this change to `PushButtonCell`.

hawkrives: Please change this to `PushButtonCell` and sort the remaining props alphabetically.

hawkrives (changes requested):
One change (just to consolidate my feedback.)

drewvolz:
Those last changes should do it, I think.

hawkrives (commented):
I think it wants `key` at the top, because it's a React-level prop, not a component one

drewvolz:
:crossed_fingers:

drewvolz:
Thank you. I wasn't looking within the prop itself to see if those were sorted or not.

drewvolz:
That should be it! (_I say again_)

hawkrives (approved):

rye (approved):

PR #1851, opened 2017-11-08
Mark flow-typed/npm as vendored in .gitattributes

rye:
This helps GitHub's language comprehension engine, [linguist](https://github.com/github/linguist), understand what's going on.

Read more about this feature on the [Linguist README](https://github.com/github/linguist/blob/4166f2e89d3e0d0d21e67ac5677a56a4efaa8221/README.md#vendored-code).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 *.pbxproj -text
+flow-typed/npm/* linguist-vendored


hawkrives:
Uh

What is special about marking it as vendored for syntax highlighting?

It’s still JS?

rye:
`linguist` is also used for generating repository-wide language statistics.  Yes, it's still JS, but it's to my understanding that this stuff is auto-generated and installed via `flow-typed`.  I'd rather these not inflate the actual language counts, similar to ignoring this in codeclimate, too.

hawkrives (approved):

drewvolz (approved):
Can either of you answer why I do not see a percentage change with `js` on `mark--flow-typed-npm--as-vendored` vs `master`?

hawkrives:
I don't think that bar changes on the branch

hawkrives:
For example, https://github.com/StoDevX/AAO-React-Native/tree/gh-pages (gh-pages branch) shows the same bar contents

rye:
It should update after I merge, i.e. momentarily.

rye:
FWIW, I also think that it takes a while for them to re-calculate.

PR #1852, opened 2017-11-08
Leave BugSnag breadcrumb on navigation

### LINKED ISSUE #1800: Send screen navigations to BugSnag as "breadcrumbs"?
Because our breadcrumb trail looks a little empty most of the time right now.

rye:
This PR causes `trackScreenChanges` to send a breadcrumb of type `navigation` (the only reserved key in the metadata object) and then send the previous screen as well.

*Closes #1800.*

I'm open to tweaking the syntax and stuff or offloading this to a different location.

### DIFF
--- a/source/app.js
+++ b/source/app.js
@@ -70,14 +70,15 @@ export default class App extends React.Component {
     const currentScreen = getCurrentRouteName(currentState)
     const prevScreen = getCurrentRouteName(prevState)
 
+    if (!currentScreen) return
+    
     if (currentScreen !== prevScreen) {
       tracker.trackScreenView(currentScreen)
+      bugsnag.leaveBreadcrumb(currentScreen, {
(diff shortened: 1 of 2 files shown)


source/app.js
```diff
@@ -72,6 +73,11 @@ export default class App extends React.Component {
     if (currentScreen !== prevScreen) {
       tracker.trackScreenView(currentScreen)
     }
+
+    bugsnag.leaveBreadcrumb(`Navigated from to ${currentScreen}.`, {
```
drewvolz: One nitpick: "Navigated from to" ... having both "from" and "to" is tripping me up.

rye: Oops, I left that in. I'm just logging the current screen now.

hawkrives (approved):

PR #1840, opened 2017-11-04
Upgrade to React Native 0.50 / Flow 0.56

hawkrives:
**WIP!**

To do:

- [x] Actually run the app and see what's broken
- [x] Fix home screen reducer test(?) (https://github.com/StoDevX/AAO-React-Native/pull/1840#issuecomment-342716180)
- [x] Make bus tests always run in UTC
- [x] Update bus route schema for new properties
- [x] Fix react-native internal flow error
- [x] "Fix" react-navigation flow errors
- [x] "Fix" `this.props.navigation` flow errors
- [x] Fix FilterType flow errors
- [x] Make codeclimate ignore /flow-typed
- [ ] There's a dark line at the bottom of our toolbar now
- [ ] "Error: the `<Image>` component cannot contain children. If you want to render content on top of the image, consider using absolute positioning." (https://github.com/facebook/react-native/commit/c55fae1e2697520a55d23a0816552bec09920142)
- [x] StatusBar doesn't go back to white text after closing SFViewController
- [x] Transportation "null is not an object (evaluating _state.schedule)"
- [x] make busview look good (https://github.com/StoDevX/AAO-React-Native/pull/1879)

---

~~Some of the bus tests are failing for some reason.~~

~~I still need to update the schema for the bus routes to accommodate the added keys.~~

~~That pretty much gets us running again. Then I need to actually launch the app and test it, because I revamped both the bus line views and the types around all the redux stores.~~

~~I should be able to do the bus refactoring on master, so I’ll want to extract that from this PR.~~

~~I should be able to do the new redux types on master, so I’ll try to pull that out as well.~~

That should leave the main changes in a future react-native@0.50 PR as going from `React.Component<DefaultProps, Props, State>` to `React.Component<Props, State>`, which is pretty much just mechanical and should be a much smaller diff. 

I should also pull out some of the other type fixes I made (eg, turns out that `_.head` wants an array, not a string, but we keep giving it strings, so I switched to `str[0]`). 

---

## Eventual PRs

- [x] Bus view refactoring (https://github.com/StoDevX/AAO-React-Native/pull/1879)
- [x] Typed Redux stores (https://github.com/StoDevX/AAO-React-Native/pull/1880)
- [x] Misc. fixes
- [ ] React-Native 0.50 / Flow 0.56 (+codeclimate) / flow typedefs for a bunch of dependencies
- [x] Remove react-native-view-shot
- [x] Remove SIS/courses tab
[...]

PR #1853, opened 2017-11-08
Fork react-native-search-bar

hawkrives:
It hasn't been updated to use the `prop-types` package, even though there's been a PR open since August (https://github.com/umhan35/react-native-search-bar/pulls).

It also hasn't been committed to since January.

I have forked it and updated it to work with React 16+.

I have tested that the searching still works.

Blocks https://github.com/StoDevX/AAO-React-Native/pull/1840


rye (approved):
Seems good to me.

PR #1859, opened 2017-11-10
Remove react-native-view-shot and the Storybook view

hawkrives:
I never finished implementing what I wanted with it, and I just got tired of updating it to call things like the rest of the app does.

If I ever find the time / motivation to do something with this again, I can always pull this out of the history, but it really needs a reworking, both to fix the errors that it has with RN50 and to upgrade it to v2.0 of the library.


rye (approved):
Aww. I wasn't too attached to it anyways.

PR #1860, opened 2017-11-10
Remove four unused packages (and add one that was missing package.json)

hawkrives:
## ① `babel-plugin-flow-react-proptypes`
I added this once, long ago, when flow was a new addition to the project. Then I forgot to enable it.

We could probably enable this without much issue, but I don't know that we need it. Given that it's never been enabled, we also don't know how well it works. (Probably fairly well, and it seems pretty actively maintained, since we get all those package updates for it.)

## ② `react-addons-shallow-compare`
I don't even remember why I added this, but none of our tests use it now. Probably was added for Enzyme or something.

## ③ `marked`
At one point, we compiled the markdown in our source files into HTML, and rendered that directly into a WebView.

Now, however, we send down the raw markdown as a JSON payload and render that with react-markdown, so marked never gets used.

## ④ `dedent`
I used it at one point to strip out leading indentation in the Danger strings, but it had some issues so I quit using it. Now it's not used anywhere.

## ❺ `mkdirp`
I use `mkdirp` in bundle-data.js… can't believe that I forgot to add it to our devdependencies. Anyway, I've added it now.

## flow-typed typedef removals
I removed these four typedefs as we no longer use these packages.

1. `frisbee`
2. `react-native-animatable`
3. `react-native-cookies`
4. `react-native-tab-view`


rye (approved):
Looks fine, pending passing builds.

PR #1862, opened 2017-11-10
Remove courses and course search tab from SIS module

hawkrives:
I don't know when I'll be able to build a data endpoint for SIS course data, so there's no timeline on us re-enabling the Courses tab.

The Search tab… has never ever done anything, and if we wanted to do it we'd essentially just scrap the file and start over, so I'm just clearing it out.


rye (approved):
This should be revisited at a later date, but having commented code there isn't necessary.  

<sub>_something something SIS API?_</sub>

hawkrives:
Something something data endpoint, which I noted in my initial comment 😉

rye:
Yeah, I still maintain that it'd be nice if SIS was not old and creepy.  It seems like a lot of data (and usage) could be served from SIS if it had a nice API, and it'd serve as a great central place to poll against, but that's not under my control so I'll just grumble in the corner. 😁

hawkrives:
Something something registrar says "no go away that's my data you can't have it"

We're actually pretty lucky to have the course data that we do, tbh. A lot of the schools I've looked at since I started Gobbldygook don't have structured data for the available courses at all; I count myself lucky if they have a publicly accessible catalog in the first place.

Not saying that SIS couldn't be better, but I'm going to have to keep an ear out as we get closer (~5 years) to looking for a SIS replacement to make sure that it doesn't lock down the data any more than this one does.

PR #1863, opened 2017-11-10
Remove giant header comments from source files

hawkrives:
```diff
@@ -1,8 +1,4 @@
-/**
- * @flow
- * All About Olaf
- * Index view
- */
+// @flow
```

1. IMHO, they serve no purpose – you know what app you're working on.
2. They're out of date – we haven't kept the name of the view current, and some say "iOS" when they're cross-platform (*cough* home screen *cough*)

I believe that @drewvolz (hi) added them primarily because they're "what you do" in Cocoa-land. I disagree with them there, as well, but that's a different debate. These comments date back to some of the very first code that was written for this project.

I chatted with @drewvolz about removing these, and he seemed positive, so I'm PRing it.

### DIFF
--- a/source/app.js
+++ b/source/app.js
@@ -1,8 +1,4 @@
-/**
- * @flow
- * All About Olaf
- * Index view
- */
+// @flow
 
 import './globalize-fetch'
 import './setup-moment'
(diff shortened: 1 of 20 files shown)


rye (approved):
Yeah, I don't have much attachment to this and it seems a bit strange to have two different comment styles right next to each other.

PR #1865, opened 2017-11-10
Upload source maps to Bugsnag for actual line-based error numbers

### LINKED ISSUE #1864: Upload JS source maps to Bugsnag when releasing
This would help us to debug some of the stuff.

https://docs.bugsnag.com/platforms/react-native/showing-full-stacktraces/#uploading-source-maps is a doc about this.

I think that we should add another machine to the "build + deploy" stage to bundle these, as it takes ~3 minutes to bundle on Travis IIRC. It should bundle both Android and iOS in release mode, as that's the only type that we upload to Testflight/Play Store, and then we can just use `npx bugsnag-sourcemaps` to upload to Bugsnag.

hawkrives:
Closes https://github.com/StoDevX/AAO-React-Native/issues/1864

Docs: https://docs.bugsnag.com/platforms/react-native/showing-full-stacktraces/#uploading-source-maps

---

React-Native apps run off of minified JS bundles, to optimize for filesize and thus memory usage.

That's great for users, but not so great for debugging, especially through Bugsnag.

Luckily, Bugsnag lets us upload source maps that tell it what line/column the code actually came from, and they recently released a `bugsnag-sourcemaps` tool that uploads for us.

I know that in #1864 I asked for another build in parallel with the native machines, but I've walked back on that. If I generate and upload the source maps after the native machines, then I'll never try to upload a source map for a build that doesn't exist. I also don't have to install ruby on the third machine just to run `propagate_version`, and I won't have to worry about somehow generating the wrong version number to send to Bugsnag, because I'll be using the same version that we uploaded to Testflight/Play Store.

The source maps will only be built and uploaded on nightly / beta / release builds, not on every Travis build.

I also:
- updated fastlane to 2.56 (I think deppbot is broken again)
- fixed a ruby indentation issue
- removed some whitespace from .travis.yml
- cleaned up a bit of code in the JS-side bugsnag integration

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -164,7 +164,7 @@ before_install:
   - nvm install "$TRAVIS_NODE_VERSION"
   - nvm use "$TRAVIS_NODE_VERSION"
   - echo "node -v is now $(node -v)"
-  
+
   # travis doesn't install yarn automatically on osx
   - if [[ $TRAVIS_OS_NAME == "osx" ]]; then brew install yarn --without-node; fi
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -36,7 +36,7 @@ GEM
     faraday_middleware (0.12.2)
       faraday (>= 0.7.4, < 1.0)
     fastimage (2.1.0)
-    fastlane (2.62.0)
+    fastlane (2.64.0)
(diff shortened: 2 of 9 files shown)


rye (approved):
I'm down to test this out in the next beta.

PR #1870, opened 2017-11-10
Switch back to 2.4.2-pre

rye:
Yeah, this is getting annoying.  We should come up with something different for this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "all-about-olaf",
-  "version": "2.4.2-beta.2",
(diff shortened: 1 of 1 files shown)


hawkrives:
I would like for us to not reset the version while we are producing betas. I'm fine with going to `-pre` once we've released a final, but during the run up to release I'd like to just leave the number alone.

Thoughts?

rye:
Sounds good to me. During the beta phase we can just increment the beta number.

PR #1880, opened 2017-11-12
Add types to redux stores and connected components

hawkrives:
With this PR, we can know the actual shape of the redux state from whatever components need to use it.

Part of https://github.com/StoDevX/AAO-React-Native/pull/1840.

I also cleaned up the shape of some of the dark corners of the store, and I switched to using `combineReducers` so as to not generate a new object when none of the sub-reducers has changed.

I'm really happy with how I figured out to do the prop types for the connected components:

```js
type ReduxStateProps = {
   initialUsername: string,
   initialPassword: string,
   loginState: LoginStateType,
}
 
type ReduxDispatchProps = {
   logIn: (username: string, password: string) => any,
   logOut: () => any,
   validateCredentials: (username: string, password: string) => any,
   setCredentials: (username: string, password: string) => any,
 }
 
type Props = ReduxStateProps & ReduxDispatchProps
```

Now we can _really_ see what comes from where to the connected components! (okay tbh this was more of a problem in gobbldygook than in aao but this'll head it off early)


hawkrives:
Reopening to codeclimate picks up the new config file

hawkrives:
… ooor not.

hawkrives:
_Most_ of this diff is just adding type annotations to the existing code. 

The big changes I made were:

- Restructuring the init flow – I moved the initialization's start to root.js, instead of starting it when the file is imported. I also grouped the redux store loading code into two Promise.all segments, to enforce the prose comment in there.
- Removing the sub-reducers in SIS – we used to have both `courses` and `balances` as children of `sis`. I removed courses because the rest of the code is mostly removed now, which let me move balances up a level. Knowing more about how the app works now, I'd propose that we keep these flat – harking back to my reorganization PR, I'd propose that we have a reducer per … feature, I guess, which means that "sis" should be renamed "balances", and we'd add a new top-level reducer for "courses".

rye (approved):
Phew. Builds/runs on iOS and Android just fine and passes my cursory testing.

hawkrives:
Thanks for doing some extra testing @rye!

PR #1883, opened 2017-11-12
Misc fixes in preparation for Flow 56

hawkrives:
This PR is all of the miscellaneous changes I made while in the depths of the React-Native v0.49 / v0.50 upgrade.

- 9fe7e6e Like https://github.com/StoDevX/AAO-React-Native/pull/1881 for CodeClimate, this configures `danger` to not worry about `@flow` annotations in `flow-typed/`
- 3367beb There was a property missing in the `OtherModesType` flow type. I added it. (Flow 56 noticed this.)
- `several` Some of these commits are pulling the state/props types out of components and into external types – Flow 56 changes how you specify Props and State.
- 23fd5d9 / 28e65fd Removal of `_.head` from AlphabetListViews. Turns out that `_.head` expects an array, not a string, so I just switched to accessing `str[0]` instead.
- bff938a Hopefully the last time I need to correct an import of `react-native-communications`
- 3199761 I missed this in https://github.com/StoDevX/AAO-React-Native/pull/1859; this removes the "Open Storybook" button from Settings in dev mode
- e4f2241 / 93294f8267e1763300f3f52172344e8e5e1c57df Optimize the openUrl listeners on iOS, I guess
- 34c6d6a It makes more sense to check if something is an array than to check if it's an object – arrays are objects, but the inverse is not true
- c64fc4d rework the datepicker to pick the platform component at compile time (flow 56 doesn't like the original code)
- bd5f752 reduce promises in event-detail adding to calendar logic or something
- 4f2bd32 fix timezone bug in a parse-hours test (I think it's a bug)

That's more-or-less everything.

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -16,6 +16,8 @@ const jsFiles = danger.git.created_files.filter(path => path.endsWith('.js'))
 
 // new js files should have `@flow` at the top
 jsFiles
+  // except for those in /flow-typed
+  .filter(filepath => !filepath.includes('flow-typed'))
   .filter(filepath => {
     const content = readFile(filepath)
     return !content.includes('@flow')
--- a/source/app.js
+++ b/source/app.js
@@ -74,7 +74,9 @@ export default class App extends React.Component {
     const currentScreen = getCurrentRouteName(currentState)
     const prevScreen = getCurrentRouteName(prevState)
 
-    if (!currentScreen) return
+    if (!currentScreen) {
+      return
+    }
 
     if (currentScreen !== prevScreen) {
       tracker.trackScreenView(currentScreen)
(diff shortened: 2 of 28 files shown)


hawkrives:
The `examine-me` commit is because I'm thinking that might have broken changing colors back from light to dark.

Edit: this has been resolved as of 93294f8

rye (approved):

rye:
I don't have much grumbling to do because this is all mostly just flow stuff. Builds and runs fine on iOS and Android.

One point of note: when I switch quickly between components on iOS, (not saying whether or not this is related to the PR because I don't know exactly) I get a warning:
screen shot 2017-11-12 at 6 39 14 pm

hawkrives:
@rye That’s a known issue and not related to this PR. I think @drewvolz even has an issue in here somewhere for it. Or I might have closed it. Not sure.

PR #1903, opened 2017-11-18
Handle getting HTML from BonApp when we expected JSON

### LINKED ISSUE #1884: Handle unexpected HTML when requesting menus from BonApp
I don't know what HTML it's returning, but `JSON Parse error: Unrecognized token '<'` means that we're getting HTML instead of JSON.

We've gotten 233 of these errors in the last week, fwiw, across both wifi and cell connections.

hawkrives:
Closes https://github.com/StoDevX/AAO-React-Native/issues/1884 (or ameliorates it at least)

---

I made two main changes here:

1. I check to see if we're handling the "json parse error unexpected token <" error and special-case it if we are
2. I added [`p-retry`](https://www.npmjs.com/package/p-retry) to do exponential backoff and retry the menu fetch up to three times before giving up – given that we see this error right after network changes, I figure that giving it some more time to reconnect would probably fix this

I also don't report the error to bugsnag/google anymore. We could if we wanted to, I guess. I can't decide if I want to or not.

Screenshot of the new error screen:

screen shot 2017-11-18 at 6 31 29 pm

Edit: I updated the screenshot

### DIFF
--- a/source/views/menus/menu-bonapp.js
+++ b/source/views/menus/menu-bonapp.js
@@ -53,7 +53,7 @@ type Props = TopLevelViewPropsType & {
   name: string,
 }
 type State = {
-  error: ?Error,
+  errormsg: ?string,
   loading: boolean,
   refreshing: boolean,
   now: momentT,
@@ -63,7 +63,7 @@ type State = {
 
 export class BonAppHostedMenu extends React.PureComponent<Props, State> {
   state = {
-    error: null,
+    errormsg: null,
     loading: true,
     refreshing: false,
     now: moment.tz(CENTRAL_TZ),
@@ -95,7 +95,7 @@ export class BonAppHostedMenu extends React.PureComponent<Props, State> {
     } catch (error) {
       tracker.trackException(error.message)
       bugsnag.notify(error)
-      this.setState(() => ({error}))
+      this.setState(() => ({errormsg: error.message}))
     }
 
     this.setState(() => ({cafeMenu, cafeInfo, now: moment.tz(CENTRAL_TZ)}))
@@ -220,8 +220,8 @@ export class BonAppHostedMenu extends React.PureComponent<Props, State> {
       return <LoadingView text={sample(this.props.loadingMessage)} />
     }
 
-    if (this.state.error) {
-      return <NoticeView text={`Error: ${this.state.error.message}`} />
+    if (this.state.errormsg) {
+      return <NoticeView text={`Error: ${this.state.errormsg}`} />
     }
 
     if (!this.state.cafeMenu || !this.state.cafeInfo) {
--- a/source/views/menus/menu-bonapp.js
+++ b/source/views/menus/menu-bonapp.js
@@ -233,12 +233,10 @@ export class BonAppHostedMenu extends React.PureComponent<Props, State> {
     }
 
     if (this.state.errormsg) {
-      let msg = ''
+      let msg = `Error: ${this.state.errormsg}`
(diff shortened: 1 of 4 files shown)


source/views/menus/menu-bonapp.js
```diff
@@ -83,19 +86,28 @@ export class BonAppHostedMenu extends React.PureComponent<Props, State> {
...
-        fetchJsonQuery(bonappMenuBaseUrl, {cafe: props.cafeId}),
-        fetchJsonQuery(bonappCafeBaseUrl, {cafe: props.cafeId}),
+        retry(this.requestMenu(props.cafeId), {retries: 3}),
+        retry(this.requestCafe(props.cafeId), {retries: 3}),
       ])
     } catch (error) {
-      tracker.trackException(error.message)
-      bugsnag.notify(error)
-      this.setState(() => ({error}))
+      if (error.message === "JSON Parse error: Unrecognized token '<'") {
```
drewvolz, line 110: Could there ever be any other type of unrecognized token other than `<` we'd want to check for? Perhaps the comparison could be for "JSON Parse error:"

source/views/menus/menu-bonapp.js
```diff
@@ -83,19 +86,28 @@ export class BonAppHostedMenu extends React.PureComponent<Props, State> {
...
+
   fetchData = async (props: Props) => {
     let cafeMenu: ?MenuInfoType = null
     let cafeInfo: ?CafeInfoType = null
 
     try {
       ;[cafeMenu, cafeInfo] = await Promise.all([
-        fetchJsonQuery(bonappMenuBaseUrl, {cafe: props.cafeId}),
-        fetchJsonQuery(bonappCafeBaseUrl, {cafe: props.cafeId}),
+        retry(this.requestMenu(props.cafeId), {retries: 3}),
```
drewvolz, line 106: Is there a way for us to present a button of sorts to allow them to physically push "retry" instead of expecting them to switch the tab twice?

source/views/menus/menu-bonapp.js
hawkrives: I know what `<` signifies. Other errors I would like to handle separately.

source/views/menus/menu-bonapp.js
hawkrives: … I was thinking they'd just go back to the homescreen.

Anyway. I (a) don't want a button because that's a bigger undertaking than I wanted to do here, and (b) because if we've tried three times and it's still failing, then I think we'll have other problems (the backoff starts at 1000ms).

So I don't think a button would help, and I don't really want to put it in right now. 😸

drewvolz: My beef is that you've instructed them to try again.

source/views/menus/menu-bonapp.js
drewvolz: Ok.

source/views/menus/menu-bonapp.js
hawkrives: Fair. I still don't think that a button is beneficial here, given the error conditions in which I expect this screen to appear.

Perhaps I will come around later. Who knows.

hawkrives: so uh… `NoticeView` supports a button and callback already. I'll … just go use that.

drewvolz: I think that was what I had in mind and was hinting towards 😄

drewvolz (approved):
Nice work!

PR #1912, opened 2017-11-19
Greenkeeper/eslint plugin react 7.5.1

hawkrives:
includes a crash fix


hawkrives:
Poll: Does anyone want the `jsx-sort-props` rule to be enabled, or can I just turn it off?

/cc @drewvolz, @rye, @elijahverdoorn

rye:
Hum. It'd be nice if the props were sorted alphabetically, but for now we can turn the grumbling off.

hawkrives:
If you'd like us to enable it, I guess file an issue like https://github.com/StoDevX/AAO-React-Native/issues/1936? I'll turn it off for this PR.

hawkrives:
Well. I'll actually file an issue; that rule has been enabled for a very long time, and I don't see why we shouldn't enable it.

PR #1915, opened 2017-11-19
Webcam thumbnail highlight wrapper

### LINKED ISSUE #1905: Webcam tiles show highlight on touch (again)
Another introduction of #1648 reintroduced by #1901.

On press, the webcam thumbnails show a box around the rounded images. This is most likely from moving the padding and border-radius to the image instead of the touchable? Most likely fallout from some recent refactoring.

drewvolz:
Closes #1905

Adds a `<View>` wrapper with the `cell` styling. Fixes the highlight to be just around the image container.

### DIFF
--- a/source/views/streaming/webcams/thumbnail.js
+++ b/source/views/streaming/webcams/thumbnail.js
@@ -41,6 +41,7 @@ export class StreamThumbnail extends React.PureComponent<Props> {
         : transparentPixel
 
     return (
+      // do not remove this View; it is needed to prevent extra highlighting
       <View style={styles.cell}>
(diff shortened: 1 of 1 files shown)


hawkrives:
Would you mind adding a comment in there, too? I keep removing it because it seems unnecessary, and a comment would help prevent this from happening again.

hawkrives (commented):
Doesn’t that need to be a block comment? Or is that only within a jsx tag?

drewvolz:
It compiles and runs.

hawkrives (approved):

Issue #1092, opened 2017-05-19
Add data bundling to the gradle build process


hawkrives:
I moved the `npm run bundle-data` command into the Xcode build process in https://github.com/StoDevX/AAO-React-Native/commit/1ec3ff9c5a0046d6a6d7512163e9ee213f5e9379.

It'd be really nice to do that for Android.

@elijahverdoorn, if you want to tackle calling `npm run bundle-data` from somewhere in gradle, that'd be grand.

PR #1942, opened 2017-11-29
Integrate "npm run bundle-data" into the android build process

hawkrives:
This mimics what iOS builds do already, where the normal `xcodebuild` calls `npm run bundle-data` so that there's no way to build the app without the data being bundled.

Closes https://github.com/StoDevX/AAO-React-Native/issues/1092


rye (approved):
Very nice.

drewvolz:
Nice 👍

PR #1943, opened 2017-11-29
Rearrange the Fastlane scripts

hawkrives:
And do some cleanup, I guess.

- Removed the `hockeyapp` gem, which we haven't used since we switched to Testflight / Google Play
- Removed the silly "Pluginfile" thing from Fastlane, which let us remove the use of `Eval` in the Gemfile
- Moved the commands from `agnostic.rb` into a new `lib/commands.rb` file
- Split up `util.rb` into five files: "commands.rb", "git.rb", "sourcemaps.rb", "util.rb", and "versions.rb"
- Removed three unused fastlane actions
- Removed the `:ci-emulator` lane from the Android setup (I already removed the Android emulator Travis machine)


Gemfile
```diff
@@ -1,13 +1,11 @@
 source 'https://rubygems.org'
 
 gem 'fastlane'
-gem 'hockeyapp'
 gem 'json'
 gem 'xcodeproj'
 
-# rubocop:disable Eval
```
rye, line 8: Good move.  I was wondering why it was necessary to load this 'Pluginfile' from the Gemfile, but it seems as though this is just Fastlane's way of tracking plugins and getting its own file to manage.  ([docs](https://docs.fastlane.tools/plugins/create-plugin/#add-a-plugin-to-your-project))

fastlane/Fastfile
```diff
@@ -9,8 +9,11 @@ fastlane_version '2.50.1'
 require 'fileutils'
 
 import 'lib/before_all.rb'
+import 'lib/commands.rb'
```
rye, line 12: Fastlane's syntax is strange.  Their `import` is basically a `require`, but with some salty anger about third-party gems thrown in.  Oh, well.

rye (approved):
I'm all for this.  Not sure about Hockey or the `:ci-emulator` stuff, but if we're not using it I'd hesitate to keep it around, with all the bloat that it may add.

I have deliberately not added any commentary about code style here, as I expect to pore over these files once this PR is merged and provide more architectural feedback in the form of further refactoring.  Pending passing CI, this should work fine.

Gemfile
hawkrives: Yeah, they have a "fastlane add_plugin" command that expects the Pluginfile, but I think it's just as easy to use the Gemfile.

fastlane/Fastfile
hawkrives: Oh, is "import" not part of Ruby? I'm fine with switching away from it, especially where I import fileutils and the thing yells at me.

drewvolz (commented):
Great work!

PR #1946, opened 2017-11-29
Fix CodeClimate test exclusion

rye:
Apparently with #1945, I didn't properly set up the exclusion rule, so it was only looking in the top level?

CodeClimate should report fixed issues with this PR now.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -7,7 +7,7 @@ engines:
   duplication:
     enabled: true
     exclude_patterns:
-    - "__tests__/"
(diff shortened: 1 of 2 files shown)


hawkrives (commented):
I think I still see `__tests__` folders showing up in there?

rye:
My view shows all of those duplication issues in `__tests__` as resolved. (As they are no longer being tracked.)  For some reason CC thinks there were new issues, but I overrode it.

hawkrives (approved):

PR #1954, opened 2017-11-30
Christmasfest bus hours

### LINKED ISSUE #1951: Express Bus Route Change
The Express Bus is picking up at the Pump House during Christmasfest (30th-3rd). It will not approach Buntrock.

hawkrives:
Closes #1951 

This PR will need to be reverted after Sunday.


drewvolz (approved):
I guess the coordinates are fine not changing as long as we aren’t using the map.

rye (approved):

PR #1968, opened 2017-12-06
README Style Tweaks

rye:
This PR:

- Consolidates CI badges into the title
- Adds "Get it on Google Play" and "Download on the App Store" badges, linking to the appropriate places
- Re-orders and updates "Getting Started" instructions
- Turns "Note" section into a bulleted list
- Emboldens types of contributions in "Contributing" section.
- Adds legal notices for use of "Google Play", "iPhone", "iPad", and "the Apple logo" in documentation and the app itself


drewvolz (approved):
Sure, looks like a nicer readme. Thanks @rye

hawkrives:
My only thing is that I don't like having the badges on the same line as the title

drewvolz:
If we have the badges, should we remove the download links and keep the beta links?

rye:
Why not, @hawkrives? I think it looks fine, especially with &middot; between.

Once I'm done with lunch, I can yank the download links as they are now redundant.

hawkrives (approved):
Meh sure

PR #2005, opened 2017-12-25
Remove badges from README.md

hawkrives:
I'll be frank: I don't like having the badges on the first line of the README. 

I also don't think that they gain us anything; I don't think that any of the primary maintainers use the badges to check on the app's status, and our code coverage % isn't really something to be proud of IMHO. 

I'd also make the argument that having the badges on the first line of the README makes it harder to read, both on GitHub's mobile view and (more importantly) in the plain-text Markdown file (which should always be nice to read IMHO.) I see the readme in those two forms much much more often than I see the desktop-sized README.

All that is to say, I don't think they bring us really any benefit, and I don't like them, so I'm proposing to remove both of the badges from the readme.


rye (approved):
Yeah, I don't think I really use them for anything. In light of caching and propagation delays, they're usually wrong anyhow.

drewvolz (approved):

rye:
Yeah my argument would focus more on the fact that they don't have much use more than that they don't look good on mobile, because they look fine on my phone on mobile. (There might be a class we could apply that hides it on mobile in a div, idk.)

Anyhow, I'll go ahead and merge this.

PR #2007, opened 2017-12-26
Update react-markdown to the latest version 🚌

rye:
I'll poke at this.


rye:
`yarn run check` passes locally.

hawkrives:
IIRC, v3 was a pretty big change - I'm happy to take this on if you'd like. Otherwise I can try to provide guidance.

rye:
Yep, the Android build failed so I'd expect iOS to fail.  Feel free to provide commentary, I could work on this or you could go ahead and do it since you're more familiar.

hawkrives:
General commentary: I believe that we only interact with the library through the files in the components/markdown folder, which essentially wrap the ReactMarkdown component to output ReactNative nodes instead of DOM nodes, and to style the generated nodes as well. 

So I'd expect pretty much all of the changes to take place in there. 

I'll go re-read the release notes and try to give some more details.

hawkrives:
Changelog for v3: https://github.com/rexxars/react-markdown/blob/master/CHANGELOG.md

You'll need to handle the new node types, probably by just giving them a minimally-styled Text component, as well as the renamed nodes. 

We used the now-removed containerTagName prop to style the markdown component itself (the one that contained the actual text / precise nodes); you'll need to figure out how to use the "root" prop instead. 

We might want to use the transformLinkUri / transformImageUri function props to do something with URLs? I seem to recall some issues with links in Markdown, but I may be remembering the WebView wrapper we have for News instead. 

You should figure out what you want to do about the new Table nodes; either blacklist them, since we currently only invoke this component for markdown that we control, or figure out how to render a nice table I suppose. 

I think that's my info dump for this component.

hawkrives:
(The root of our markdown stuff: https://github.com/StoDevX/AAO-React-Native/blob/master/source/views/components/markdown/index.js)

hawkrives:
I started poking at this; whoof.

v3 uses [`remark`](https://github.com/remarkjs/remark) to parse instead of commonmark; remark is apparently part of https://unifiedjs.github.io of something (it's kinda hard to follow).

Long story short, something in this mess of stuff with _markdown text that doesn't ever live in a file_ depends on the `vfile` module, which … deals with locating files on the file system. It needs the `path` stdlib module… okay, so I'll install someone's copy of that, since `path` itself just deals with pathnames.

Now `vfile` wants to use `process.cwd()`. This is _also_ not available to react-native.

I just don't know if it's worth it for us to go further down this road… I think we'll need to open an issue over at react-markdown about react-native support. I believe that he wants to support it, eventually…

anyway. TL;DR: This is a much larger time investment than I anticipated. I'm going to close this; feel free to re-open if someone wants to work on it again.

I don't think we will upgrade in the foreseeable future.

rye:
Yeah, that's what I was coming to the conclusion of when I started working on this and I stopped shortly thereafter.  We could file an issue upstream, else just wait.

PR #2017, opened 2017-12-29
Move JS builds to Circle, and report a Danger status from each "task"

hawkrives:
TL;DR:

- Moves the old "JS" job from Travis to 6 distinct CircleCI jobs
- Drastically expands the number of things that Danger analyses
- Upgrades Danger to v3
- Downgrades the analyze-gym script to Python 2

---

On `master`, we have four "stages" within Travis: Greenkeeper, Test, Deploy, and Native.

The Test stage was comprised of one single machine, but it did six distinct tasks:

1. Check Flow types
2. Run Jest tests and upload to Coveralls
3. Check Prettier formatting
4. Run ESLint
5. Check validity of the data files
6. Collect results from all of those and upload to GitHub through Danger

At the beginning of this PR, I split that stage into six distinct machines within Travis, one for each task.

At the end of this PR, though, I realized that our Test stage's execution time had increased from ~4m to ~8.5m, which I didn't like.

After analyzing our build logs with @rye, I decided to try CircleCI to see if its infrastructure could cut down on the amount of startup time that we had, because almost all of our time was spent before running our actual task. (Each machine took ~2, 2.5m to run, of which ~10-50s was the actual script.)

This worked! Now, our average build time for the Test stage is ~1.5m.

---

The meat of the PR is the increased number of things that I make Danger analyze.

The new JS-General Danger check will check not only number of lines in the PR, or the results of the ESLint run, but also things like "did something add a duplicate header search path to the .pbxproj for xcode" or "are the imports in build.gradle sorted alphabetically".

I took Danger's motto to heart, essentially: 'Stop saying "you forgot to …"'

I tried to codify all the things I look for when `react-native link` has been run, so that Danger can check them, instead of me.

---

I also upgraded Danger to v3.

---

I had to downgrade the `analyze-gym` script to python2, in order to avoid having to install python3 on the Travis macOS builds. `3to2` worked pretty well for that.

---

Inside of the Fastlane scripts for building each project, I added some stuff to save out the log files / some information about where the built products live into files on the file system directly from ruby, instead of trying to orchestrate it in the shell.

---

That's about it. There's just a fair amount of code to check the things I look for in the Xcode files.

Issue #2016, opened 2017-12-29
Use Commit Statuses to communicate CI job/component statuses


rye:
This is an idea that I came up with: rather than getting a blind "push/pr passed or failed in Travis" bit of information, we could potentially tell Danger or some custom scripts to `POST /repos/StoDevX/AAO-React-Native/statuses/${COMMIT_SHA}`. (See the endpoint documentation at [this place](https://developer.github.com/v3/repos/statuses/#create-a-status))

This kind of behavior **_really does not need to rely on Danger_** and would be perfectly fine in a build script. We need to implement four different statuses: `error`, `failure`, `pending`, and `success`.  Our "Details" link could just be a link to the Travis build.  We can also put a `description` flag, (E.g. "ESLint had no complaints.") and use a `context` string of whatever we're running.

Questions to be asked/answered:

- [ ] Does this need to be part of our Danger run?  Or, can we just include a shell/JS/Ruby script that wraps the ephemeral behavior of updating commit statuses?  (I'm happy to write such a script, and I think it's more appropriate.)  We just need a GH token, which we already have, and some CI magic will ensue.

I'm going to start working on this in a `travis/` PR just to see what kinds of pretty things I can come up with.  Open to input.

hawkrives:
Seems like this might as well go into Danger, since the logs are generated explicitly for Danger… we'd just add a function that we call after we send the log to Github with `fileLog` (in the dangerfile we have now).

Not saying something couldn't be more generalized, but I've already got some logic in Danger-land for determining if a log file needs to be reported or not, so it seems easiest / future-proofest to just reuse that?

rye:
My big gripe is that we only run Danger once currently.  I'm planning on putting together a Proof-of-Concept so we can look at, stylistically, what it will do, which I will just run manually in Travis.  We can always port over.  (It'll be in JS, so it'll be easy to do.)

Ideally, I'd like to run this both before we run the business part of a job and after we finish that.  That way, we get a pending status and a "live" view into what's going on in CI.  If it passes, just report another status.  Then, we can just click on "Details" to look at the specific CI job that failed.

We should probably also only do this for non-cron builds.

hawkrives:
@rye Do you mind if I close this, since the associated PR was closed, and Circle gets us most of this now?

PR #2018, opened 2017-12-29
Start work on a commit status script

rye:
This script currently uses a combination of environment variables and fallback variables and is meant mostly as a proof-of-concept more than anything.  I'm totally open to feedback, I just hacked it together over the last hour or so.

It uses the `DANGER_GITHUB_API_TOKEN` variable (present in Travis) to post statuses as @stodevx-bot.  I'd use a GitHub App, but I don't exactly want to set up a webhook handler at the moment.  So, we should be wary of how much that token is being used—we only get 5000 requests per... hour?  I think that's what the interval is.

We could set up YAML files with canned responses!  That'd be cool.

Currently the only implementation of it is in the `before_script` block of the JS build, the `after_success` and `after_failure` blocks too.  No error handling yet.

I'm open to merging this with @hawkrives' work on #2017, <sub>(hey that was a fun year)</sub> so that we get more checks and we don't have a `js:<everything>` build.  I can go either way.

Closes #2017.


rye:
Some screenshots for your viewing pleasure:

|screen shot 2017-12-28 at 10 35 37 pm|
|-----|
| screen shot 2017-12-28 at 10 37 24 pm |

One shortcoming of this style is that the status only changes when the build starts, so in instances like this we get some weirdness when push builds ran but no PR builds have started yet.

rye:
Ideally, I'd like to have statuses for `js:lint`, `js:flow`, `js:<etc>`, and maybe even some for `android` and `ios`.

Just to point out again, one of the main reasons I'd like to implement these is so that we can get a heads-up view of what is running "currently" as well as what "failed" with a given commit.

hawkrives:
oh god you imported my other branch. You should probably undo the commit that touches the dangerfile… for now, we'll need to keep everything in a single file I believe. A liberal use of large comments to break things up should keep it understandable and maintainable, though.

rye:
Oh nice, it works.
screenshot_20171229-185331

@hawkrives, I have no idea what's going on with Danger. Feel free to take a stab at it. Also, this is still a script, for reasons I mentioned in #2016.

.travis.yml
```diff
@@ -130,41 +130,86 @@ jobs:
     - <<: *js
       stage: Test
       env: [{task: "js:flow"}]
+      before_script:
+        - echo "loglevel=silent" >> .npmrc
+        - mkdir -p logs/
+        - set -o pipefail
+        - npm run report $task "pending" "Checking the flow of things..."
```
hawkrives: If we move this into the `script` section itself (eww, I know), we could remove these other 4 lines that you had to add to each job… just an idle thought.

rye: Yeah, that's probably a good idea. Putting it as the first key will ensure it gets run

rye:
Now I get to play with reverting a merge... 🙃

scripts/github-status-report.js
```diff
@@ -9,121 +9,139 @@ const fetch = require('node-fetch')
...
+  let url = `${GITHUB_API}/rate_limit`
+  let options = {
+    mecthod: 'GET',
+    headers: GH_API_HEADERS(token),
+  }
+
+  let value = await fetch(url, options)
+    .then(response => response.json())
+    .then(response => {
+      return response
```
hawkrives, line 50: Drive-by comment: this `.then` clause serves no purpose.

rye: Yeah, yeah, yeah... I know

hawkrives:
Can't you just revert my commit from the merged branch?

rye:
Hmm. Let's see how this plays out.

rye:
There we go, Danger is cool again.

rye:
I am... going to close this one.  We can revisit at a later date if we want this, but I think the switch to Circle (for JS builds) at least got most of what I wanted done done.

Issue #1907, opened 2017-11-19
SIS login does not provide login feedback


drewvolz:
It appears to be stuck on “Logging in to St. Olaf” with incorrect credentials, not that I can test with anything other than that.

<img src="https://user-images.githubusercontent.com/5240843/32988762-f447a9ba-ccc7-11e7-926f-8bc510838a0c.jpeg" height=250>

hannesmcman:
Closed by #2019

PR #2019, opened 2017-12-29
Added feedback for failed SIS login, simplified logged in display

hannesmcman:
I started this PR with the intention of resolving #1907...I realized halfway through that this was already resolved when I checked the live version on my phone, but it wasn't working on master so I stuck it out. I modified the failed login feedback to be an alert and added a similar alert if the user is not connected to the internet. I also simplified the display of the login-credentials section on the settings view so that the inputs disappear when the user is logged in. I think the showNetworkFailureMessage() can be moved to another module and imported, but I wasn't sure where to put it so I thought I'd ask.

### DIFF
--- a/source/flux/parts/settings.js
+++ b/source/flux/parts/settings.js
@@ -16,6 +16,7 @@ import {
 import {type ReduxState} from '../index'
 import {type UpdateBalancesType} from './sis'
 import {updateBalances} from './sis'
+import {Alert} from 'react-native'
 
 export type LoginStateType = 'logged-out' | 'logged-in' | 'checking' | 'invalid'
 
@@ -83,9 +84,13 @@ export function logInViaCredentials(
   username: string,
   password: string,
 ): ThunkAction<LogInActions | UpdateBalancesType> {
-  return async dispatch => {
+  return async (dispatch, getState) => {
     dispatch({type: CREDENTIALS_LOGIN_START})
-
+    const state = getState()
+    if (!state.app.isConnected) {
+      dispatch({type: CREDENTIALS_LOGIN_FAILURE})
+      showNetworkFailureMessage()
+    }
     const result = await performLogin(username, password)
     if (result) {
       dispatch({type: CREDENTIALS_LOGIN_SUCCESS, payload: {username, password}})
@@ -93,8 +98,25 @@ export function logInViaCredentials(
       dispatch(updateBalances())
     } else {
       dispatch({type: CREDENTIALS_LOGIN_FAILURE})
+      showInvalidLoginMessage()
     }
   }
+}
+
+function showNetworkFailureMessage() {
+  Alert.alert(
+    'Network Failure',
+    'You are not connected to the internet. Please connect if you want to access this feature.',
+    [{text: 'OK'}],
+  )
+}
+
+function showInvalidLoginMessage() {
+  Alert.alert(
+    'Invalid Login',
+    'The username and password you provided do not match a valid account. Please try again.',
+    [{text: 'OK'}],
+  )
 }
 
 type LogOutAction = {|type: 'settings/CREDENTIALS_LOGOUT'|}
@@ -198,7 +220,7 @@ export function settings(state: State = initialState, action: Action) {
     }
 
     case CREDENTIALS_LOGIN_FAILURE:
-      return {...state, state: 'invalid'}
+      return {...state, loginState: 'invalid'}
 
     case CREDENTIALS_LOGOUT: {
       return {
--- a/source/flux/parts/settings.js
+++ b/source/flux/parts/settings.js
@@ -80,6 +80,21 @@ type LoginSuccessAction = {|
 |}
 type LoginFailureAction = {|type: 'settings/CREDENTIALS_LOGIN_FAILURE'|}
 type LogInActions = LoginStartAction | LoginSuccessAction | LoginFailureAction
+
+const showNetworkFailureMessage = () =>
+  Alert.alert(
+    'Network Failure',
+    'You are not connected to the internet. Please connect if you want to access this feature.',
+    [{text: 'OK'}],
+  )
+
+const showInvalidLoginMessage = () =>
+  Alert.alert(
+    'Invalid Login',
+    'The username and password you provided do not match a valid account. Please try again.',
+    [{text: 'OK'}],
+  )
+
 export function logInViaCredentials(
   username: string,
   password: string,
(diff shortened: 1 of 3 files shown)


source/flux/parts/settings.js
```diff
@@ -198,7 +221,7 @@ export function settings(state: State = initialState, action: Action) {
     }
 
     case CREDENTIALS_LOGIN_FAILURE:
-      return {...state, state: 'invalid'}
+      return {...state, loginState: 'invalid'}
```
hawkrives, line 223: Omg. Whoops. Good catch!

source/flux/parts/settings.js
```diff
@@ -83,20 +84,42 @@ export function logInViaCredentials(
...
       // since we logged in successfully, go ahead and fetch the meal info
       dispatch(updateBalances())
     } else {
       dispatch({type: CREDENTIALS_LOGIN_FAILURE})
+      showInvalidLoginMessage()
     }
   }
 }
 
+function showNetworkFailureMessage() {
```
hawkrives: I don't mind having these in this file; however, given that we only call them once, what about inlining them into the logInViCredentials function itself?

hawkrives (commented):
In general, wow! Thanks. 

I want to think about the wording of the alerts before I approve the PR, but that's all I'm holding off on. 

Thanks again!

source/flux/parts/settings.js
hannesmcman: Is this where they should be? Or should they be above the return function? https://github.com/StoDevX/AAO-React-Native/pull/2019/commits/d1f2b9d71a733dc4d391d66d51dbbee20e4e6e4f

source/flux/parts/settings.js
```diff
@@ -88,6 +88,20 @@ export function logInViaCredentials(
     dispatch({type: CREDENTIALS_LOGIN_START})
     const state = getState()
     const isConnected = state.app ? state.app.isConnected : false
+    function showNetworkFailureMessage() {
```
hawkrives: Ah. I was thinking just inline the `Alert.alert` calls directly into the spots they're called from:

```js
  if (!isConnected) {
    dispatch({ type: CREDENTIALS_LOGIN_FAILURE });
    Alert.alert(
      "Network Failure",
      "You are not connected to the internet. Please connect if you want to access this feature.",
      [{ text: "OK" }]
    );
  }
  const result = await performLogin(username, password);
  if (result) {
    dispatch({
      type: CREDENTIALS_LOGIN_SUCCESS,
      payload: { username, password }
    });
    // since we logged in successfully, go ahead and fetch the meal info
    dispatch(updateBalances());
  } else {
    dispatch({ type: CREDENTIALS_LOGIN_FAILURE });
    Alert.alert(
      "Invalid Login",
      "The username and password you provided do not match a valid account. Please try again.",
      [{ text: "OK" }]
    );
  }
```

But, I'm OK with having the shorter names to refer to them by.

---

Thus, I'll change my request:

Would you please move these two function definitions back to the top level, and define them above `logInViaCredentials`, please? 

And could you also change them to arrow functions, please? For example, 

```js
const showNetworkFailureMessage = () =>
  Alert.alert(
    "Network Failure",
    "You are not connected to the internet. Please connect if you want to access this feature.",
    [{ text: "OK" }]
  )
```

The extra line taken up by the closing brace just seems wasteful to me, since the function's only one statement long 😜

hannesmcman:
https://github.com/StoDevX/AAO-React-Native/pull/2019/commits/276fe370221fa3df8bafd9be0be4a4745b48edc8 resolves an issue where the Network Error alert would fire after an internet connection was established after an attempt to login while offline.

source/flux/parts/settings.js
```diff
@@ -17,6 +17,7 @@ import {type ReduxState} from '../index'
 import {type UpdateBalancesType} from './sis'
 import {updateBalances} from './sis'
 import {Alert} from 'react-native'
+import {updateOnlineStatus} from './app'
```
hawkrives: Is this import used?

source/lib/login.js
```diff
@@ -4,6 +4,7 @@ import * as Keychain from 'react-native-keychain'
 import * as storage from './storage'
 import buildFormData from './formdata'
 import {OLECARD_AUTH_URL} from './financials/urls'
+import {Alert} from 'react-native'
```
hawkrives: Is this import used?

source/flux/parts/settings.js
```diff
@@ -103,18 +104,18 @@ export function logInViaCredentials(
     dispatch({type: CREDENTIALS_LOGIN_START})
     const state = getState()
     const isConnected = state.app ? state.app.isConnected : false
-    if (!isConnected) {
-      dispatch({type: CREDENTIALS_LOGIN_FAILURE})
-      showNetworkFailureMessage()
-    }
```
hawkrives: I'd like to leave this here, just to short-circuit the logic when we know we're offline.

source/lib/login.js
```diff
@@ -45,10 +46,16 @@ export async function performLogin(
...
+  let loginFailed = false
   const loginResult = await fetch(OLECARD_AUTH_URL, {
     method: 'POST',
     body: form,
+  }).catch(() => {
+    loginFailed = true
   })
+  if (loginFailed) {
+    return false
+  }
```
hawkrives: Could you change this to something like 

```js
    const form = buildFormData({username, password})
    
    let loginResult = null
    try {
      loginResult = await fetch(OLECARD_AUTH_URL, {
        method: 'POST',
        body: form,
      })
    } catch (err) {
      return false
    }

    if (!loginResult) {
      return false
    }

    const page = await loginResult.text()

    if (page.includes('Password')) {
```

I don't like to assign to things outside of a Promise from a promise callback … it feels fragile. Instead, I used a normal try/catch block around the `await` keyword to catch the network error, and return false from the catch block.

hawkrives (changes requested):

source/flux/parts/settings.js
hannesmcman: Isn't this redundant with the if else clause below? If I put this back, it would trigger two alerts if the user wasn't connected to the internet..

hannesmcman: The reason I replaced this if statement with a combination of the try-catch in login.js and 
```javascript
else {
      dispatch({type: CREDENTIALS_LOGIN_FAILURE})
      if (isConnected) {
        showInvalidLoginMessage()
      } else {
        showNetworkFailureMessage()
      }
}
``` 
in settings.js was that sometimes the redux store wouldn't update the app.isConnected variable correctly...thus, even if the login was successful, the "network error" alert would still be shown. This way, the alert will only go off if both the login is unsuccessful and app.isConnected is false.

hawkrives:
You'll probably need to run `yarn prettier` on this branch. I'm going to make a quick PR to re-enable Danger so it reminds us about these things.

hawkrives (approved):
Approving pending re-prettification 😸

drewvolz (approved):
Great work, @hannesmcman

hawkrives:
:party:

PR #2028, opened 2018-01-03
Add <hr> and remove <sub> tag from README footer

hawkrives:
hawken no like

If someone like better, I'll deal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,5 +36,8 @@ You should [create one yourself](https://console.developers.google.com/projectse
 
 For full information, see [CONTRIBUTING.md](CONTRIBUTING.md).
 
-<sub>Google Play and the Google Play logo are trademarks of Google LLC.<br />
-Apple, the Apple logo, iPhone, and iPad are trademarks of Apple Inc., registered in the U.S. and other countries. App Store is a service mark of Apple Inc., registered in the U.S. and other countries.</sub>
+-----
+
+Google Play and the Google Play logo are trademarks of Google LLC.<br>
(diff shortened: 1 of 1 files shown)


drewvolz (approved):

rye:
If only there was a way to make the text smaller, that was the whole point.  I'd be okay with it, I suppose, but the legal things down there shouldn't be the same size as the rest of the text imo

hawkrives:
@rye Agreed. To me, the horizontal line serves much the same purpose, and I think the eye will ignore the disclaimer below the line (since it's at the bottom and we don't have any other thick lines in there).

rye:
My qualm is that unfortunately, GitHub Flavored Markdown doesn't support any font size that's smaller than regular text besides through the use of `<sub>`. (I've tried manually setting the font size in a custom `<p>`, using `<small>`, using `<font>`, to no avail...)  I'd be down to use such an option that keeps the baselines aligned across multiple sizes, and that would be most ideal, but alas, here we are without such an option because of sanitization. (Thanks @github.)

To that end, I should point out that I just don't really see why using `<sub>` is so bad here.  The intended purpose was to establish a "fine print" feel to that text, which I think was accomplished by `<sub>`.  To me, adding an `<hr>` and promoting only draws attention to that text, which is the opposite effect from what I wanted; not only would such an `<hr>` be the thickest line in this text region, but keeping the font size the same as the above text would make it seem like a continuation of it, which it's not.

Do I think it matters?  No, but I like it my way better. 😈  Anyhow, I would not be too upset if this got merged.

rye:
Hawken says `<h6>`s and I think that's a grand idea.

rye (approved):

PR #2051, opened 2018-01-06
Instructions for git lfs

drewvolz:
We added `git lfs` recently. It requires some setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,9 +16,11 @@ The St. Olaf community, now in pocket size… rewritten in React Native.
 
 - [Install React Native](http://facebook.github.io/react-native/docs/getting-started.html#content)
 - [Install Yarn](https://yarnpkg.com/en/docs/install)
+- [Install Git Large File Storage](https://help.github.com/articles/installing-git-large-file-storage/)
 - Clone the repository
 - `cd` into your clone
 - Run `yarn`
+- Run `git lfs pull`
 - For iOS: `npm run ios`.
 - For Android: launch your favorite Android emulator first, then run `npm run android`
 
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ The St. Olaf community, now in pocket size… rewritten in React Native.
 
 - [Install React Native](http://facebook.github.io/react-native/docs/getting-started.html#content)
 - [Install Yarn](https://yarnpkg.com/en/docs/install)
-- [Install Git Large File Storage](https://help.github.com/articles/installing-git-large-file-storage/)
+- [Git Large File Storage (LFS)](https://help.github.com/articles/installing-git-large-file-storage/)
 - Clone the repository
 - `cd` into your clone
 - Run `yarn`
--- a/README.md
+++ b/README.md
@@ -16,12 +16,13 @@ The St. Olaf community, now in pocket size… rewritten in React Native.
 
 - [Install React Native](http://facebook.github.io/react-native/docs/getting-started.html#content)
 - [Install Yarn](https://yarnpkg.com/en/docs/install)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,9 +16,11 @@ The St. Olaf community, now in pocket size… rewritten in React Native.
 
 - [Install React Native](http://facebook.github.io/react-native/docs/getting-started.html#content)
 - [Install Yarn](https://yarnpkg.com/en/docs/install)
+- [Install Git Large File Storage](https://help.github.com/articles/installing-git-large-file-storage/)
```
hawkrives: Nit: Don't expand the acronym here.

README.md
```diff
@@ -16,9 +16,11 @@ The St. Olaf community, now in pocket size… rewritten in React Native.
...
 - Clone the repository
 - `cd` into your clone
 - Run `yarn`
+- Run `git lfs pull`
```
hawkrives: Add `git lfs install` first. I managed to not install it system-wide, so let's be on the safe side.

hawkrives (changes requested):

README.md
rye: I think it only needs to be run once, but if this is peoples' first time using it, they should run that.  I think that since it adds LFS as a hook, cloning a LFS repository will probably also automatically pull them, but since these two operations are non-destructive, we don't need to worry about them.

drewvolz: That's part of the installation instructions in the link.

README.md
drewvolz: We define every term we use. I don't see why not?

README.md
drewvolz: @rye are you saying we should not include instructions for this at all?

rye (approved):
I'll approve, pending @hawkrives' approval.  I'm of the mind that these installation steps should include everything that is needed in order to get started, and I think that a proper installation of LFS will point to this solution.

Anyhow. Otherwise looks good to me.

README.md
hawkrives: Can we have the abbreviation here, too, then, please? "Git Large File Storage (LFS)" maybe?

README.md
hawkrives: I'm thinking that we could include the os x install instructions inline, since they're `brew install git-lfs` and `git lfs install`

hawkrives: and then link to the installation page for other options

README.md
```diff
@@ -16,9 +16,11 @@ The St. Olaf community, now in pocket size… rewritten in React Native.
 
 - [Install React Native](http://facebook.github.io/react-native/docs/getting-started.html#content)
 - [Install Yarn](https://yarnpkg.com/en/docs/install)
+- [Git Large File Storage (LFS)](https://help.github.com/articles/installing-git-large-file-storage/)
```
hawkrives: "Install Git Large File Storage (LFS)"

I guess that link's fine. It does more hand-holding than anything else I've found.

README.md
```diff
@@ -16,9 +16,11 @@ The St. Olaf community, now in pocket size… rewritten in React Native.
...
 - Clone the repository
 - `cd` into your clone
 - Run `yarn`
+- Run `git lfs pull`
```
hawkrives: I still want `git lfs install` before this. However I found my original instructions, I managed to not get it activated on this repo. It's an idempotent operation, more or less, though, so I'd still like to put it here. Please?

hawkrives (changes requested):
last tweakses

hawkrives (approved):

README.md
rye: @drewvolz No, I would say we should include this step to make sure that people have that.

Issue #2055, opened 2018-01-06
[Android] disable home screen buttons throws an error


drewvolz:
The error I saw depended upon if remote debugging was on or off.

Off | On
--|--
<img width="441" alt="screen shot 2018-01-06 at 8 42 33 am" src="https://user-images.githubusercontent.com/5240843/34641193-9107251c-f2bd-11e7-878e-dd241c821182.png"> | <img width="441" alt="screen shot 2018-01-06 at 9 11 26 am" src="https://user-images.githubusercontent.com/5240843/34641446-be2822b8-f2c1-11e7-859f-10c89257e606.png">

PR #2056, opened 2018-01-06
Fix android edit-home crashing on tap

hawkrives:
Closes https://github.com/StoDevX/AAO-React-Native/issues/2055


drewvolz (approved):
Thanks

Issue #2054, opened 2018-01-06
"Report a Problem" screen tweaks


hawkrives:
- rename to "Help"?
- add two more cards – maybe how to report something to the helpdesk, and to facilities?
- visual tweakses

suggestions welcome

rye:
Sounds good.

Maybe we could also put text on the card, above the button?  A short description of what we're doing, why we need location access, etc.?

drewvolz:
I'm addressing some of those things in a PR. One sec.

drewvolz:
Feel free to add-on to that PR or suggest changes.

drewvolz:
Would we consider what was done in #2061, #2063, and #2065 satisfactory? I think @rye was playing around with cards for helpdesk and facilities.

hawkrives:
I'm working on those, as well as as remotely disabling / configuring the tools.

PR #2061, opened 2018-01-06
Prettify the wi-fi reporting and add more description

drewvolz:
Part of #2054

pretty-wifi

### DIFF
--- a/source/views/help/wifi.js
+++ b/source/views/help/wifi.js
@@ -1,7 +1,7 @@
 // @flow
 
 import * as React from 'react'
-import {View} from 'react-native'
+import {StyleSheet} from 'react-native'
 import glamorous from 'glamorous-native'
 import {Card} from '../components/card'
 import {Button} from '../components/button'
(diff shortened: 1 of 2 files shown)


source/views/help/wifi.js
```diff
@@ -80,15 +81,45 @@ export class ReportWifiProblemView extends React.Component<Props, State> {
...
+        <Card footer={this.state.status}>
+          <Container>
+            <Title selectable={true}>Report a Wi-Fi Problem</Title>
+            <Description selectable={true}>
+              If you are having an issue connecting to any of the St. Olaf
+              College Wi-Fi networks, please tap the button below.
+            </Description>
+            <Description selectable={true}>
+              We will record your current location and what device you are
+              using, then send it to a server that IT maintains. They are
```
rye: >They are planning to look closely at where people are having issues!

We don't really speak on their behalf; I suppose I would be more neutral?  As in,

>The networking team can then use this information to identify where people are having issues!

rye: Also, we should be clear that this information is anonymous, and we do not collect usernames.

source/views/help/wifi.js
```diff
@@ -80,15 +81,45 @@ export class ReportWifiProblemView extends React.Component<Props, State> {
...
+      <View>
+        <Card footer={this.state.status}>
+          <Container>
+            <Title selectable={true}>Report a Wi-Fi Problem</Title>
+            <Description selectable={true}>
+              If you are having an issue connecting to any of the St. Olaf
+              College Wi-Fi networks, please tap the button below.
+            </Description>
+            <Description selectable={true}>
+              We will record your current location and what device you are
```
rye: >and what device you are using

could be something more like

>and some general information about the device you are using

source/views/help/wifi.js
rye: I think it's okay to be slightly more wordy here just to be clear about what we're suggesting.  Alternatively, we could add some of this to the legal section and point people that way.

source/views/help/wifi.js
```diff
@@ -80,15 +81,45 @@ export class ReportWifiProblemView extends React.Component<Props, State> {
...
   render() {
     return (
-      <ScrollView>
-        <Card footer={this.state.status} header="Report a WiFi Problem!">
-          <Button
-            disabled={this.state.status !== ''}
-            onPress={this.start}
-            title="Report"
-          />
+      <View>
```
hawkrives: We don't need this View anymore

hawkrives (changes requested):
Rather than use a Container inside the Card, I'd rather use the `style={}` prop of Card itself to add the padding.

source/views/help/wifi.js
drewvolz: Should we change that in the release notes, too?

source/views/help/wifi.js
drewvolz: Again, if we change this, let’s be sure the release notes read the same.

rye:
I have a riff on this that I will PR into this PR.

rye:
See #2063 for an idea.

drewvolz:
I've removed the outside view, the inner container, applied styles to the card, and tweaked the wording to be generic yet transparent.

hawkrives (approved):

PR #2063, opened 2018-01-06
Write the status out into the Button rather than a footer

rye:
Part of #2061.

This will be more familiar to most people, I think.  Since `status` is human-readable text, I don't see why we need to have a button that doesn't change and a footer below it.

| Before | During | After |
|---|---|---|
| screen shot 2018-01-06 at 12 50 58 pm | screen shot 2018-01-06 at 12 51 02 pm | screen shot 2018-01-06 at 12 52 27 pm |

Thoughts?

### DIFF
--- a/source/views/help/wifi.js
+++ b/source/views/help/wifi.js
@@ -82,7 +82,7 @@ export class ReportWifiProblemView extends React.Component<Props, State> {
   render() {
     return (
       <View>
-        <Card footer={this.state.status}>
+        <Card>
           <Container>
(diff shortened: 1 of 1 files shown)


rye:
Erm.  Okay, circleci?

drewvolz (approved):

hawkrives:
Mainly because in May, when I was testing this, I'd write out the entire payload to the footer. This is much nicer.

drewvolz:
This is the way the “add to calendar” currently works. Maybe we should change it over there, too.

PR #2064, opened 2018-01-06
Move JS builds to Circle, and report a Danger status from each "task"

hawkrives:
Closes https://github.com/StoDevX/AAO-React-Native/pull/2017

TL;DR:

- Moves the old "JS" job from Travis to 6 distinct CircleCI jobs
- Drastically expands the number of things that Danger analyses
- Upgrades Danger to v3
- Downgrades the analyze-gym script to Python 2

---

On `master`, we have four "stages" within Travis: Greenkeeper, Test, Deploy, and Native.

The Test stage was comprised of one single machine, but it did six distinct tasks:

1. Check Flow types
2. Run Jest tests and upload to Coveralls
3. Check Prettier formatting
4. Run ESLint
5. Check validity of the data files
6. Collect results from all of those and upload to GitHub through Danger

At the beginning of this PR, I split that stage into six distinct machines within Travis, one for each task.

At the end of this PR, though, I realized that our Test stage's execution time had increased from ~4m to ~8.5m, which I didn't like.

After analyzing our build logs with @rye, I decided to try CircleCI to see if its infrastructure could cut down on the amount of startup time that we had, because almost all of our time was spent before running our actual task. (Each machine took ~2, 2.5m to run, of which ~10-50s was the actual script.)

This worked! Now, our average build time for the Test stage is ~1.5m.

---

The meat of the PR is the increased number of things that I make Danger analyze.

The new JS-General Danger check will check not only number of lines in the PR, or the results of the ESLint run, but also things like "did something add a duplicate header search path to the .pbxproj for xcode" or "are the imports in build.gradle sorted alphabetically".

I took Danger's motto to heart, essentially: 'Stop saying "you forgot to …"'

I tried to codify all the things I look for when `react-native link` has been run, so that Danger can check them, instead of me.

---

I also upgraded Danger to v3.

---

I had to downgrade the `analyze-gym` script to python2, in order to avoid having to install python3 on the Travis macOS builds. `3to2` worked pretty well for that.

---

Inside of the Fastlane scripts for building each project, I added some stuff to save out the log files / some information about where the built products live into files on the file system directly from ruby, instead of trying to orchestrate it in the shell.

---

That's about it. There's just a fair amount of code to check the things I look for in the Xcode files.


rye:
@hawkrives, I forget, was migrating iOS/Android to Circle also planned?

hawkrives:
@rye Not as part of this PR, but we can revisit it in the future.

hawkrives:
Before, Travis was in charge of _everything_ related to our build process.

1. It handled updating `yarn.lock` in response to Greenkeeper PRs and pushing the updated files back to the PR branch
2. It ran every night around 1am to generate new "nightly" builds and upload them to Google Play and iTunes Connect
3. It compiled the data files and published them to the `gh-pages` branch
4. It linted, typechecked, and analyzed our JS code six ways to Sunday
5. It compiled the Android app enough to ensure that compilation still worked
6. It compiled the iOS app enough to ensure that compilation still worked

With this PR, of those responsibilities, only number 4 has moved to CircleCI.

1. This doesn't care where it lives, so I might as well leave it alone*
2. This runs off of `master`, which we keep always green, so it doesn't need the JS checks
3. This doesn't care where it lives, so I might as well leave it alone*
4. This moved, and became significantly faster
5. (and 6): I'm leaving these alone; given that the precursor checks still run, just in a different service, we still get the same coverage, and Travis is doing a better job recently about cancelling outdated PR builds

\*: plus, leaving it alone means that I don't have to set up the keys and whatnot again

rye (approved):
I think we should be fine to give this a try.  Travis is still enabled, and still alive and kicking—I don't think we should run into any issues, and if we do, we can solve them as they come up.

PR #2065, opened 2018-01-07
WiFi Issue Reporting Screen Improvements

hawkrives:
I'll take a screenshot in a bit.

"[…] make sure to use textAlign, not alignContents. They work the same for a ~single line of text, but alignContents shifts to left-align when the text goes multi-line."


hawkrives:
<details><summary>Collapsed b/c new changes</summary>

Screenshots:

Before | After
--- | ---
screen shot 2018-01-07 at 7 49 53 pm | screen shot 2018-01-07 at 7 50 13 pm

I think that I actually prefer it left-aligned. I am going to change this branch to do that, instead.

 </details>

hawkrives:
New screenshots:

Normal | Error | Success
--- | --- | ----
screen shot 2018-01-07 at 8 09 51 pm |screen shot 2018-01-07 at 8 09 39 pm | screen shot 2018-01-07 at 8 12 06 pm

source/views/help/wifi.js
```diff
@@ -54,35 +55,57 @@ function reportToServer(data) {
...
       }))
     }
   }
 
   render() {
+    const buttonEnabled =
+      this.state.status === 'init' || this.state.status === 'error'
+
+    let buttonMessage = 'Error'
+    if (this.state.status === 'init') {
```
rye: Can this `if`-`else if` chain be replaced by a switch-case?

hawkrives: It can, but then I have to add a `break` for each case since I can't return here.

rye: Hmm, what about a function above that converts `status` → `message`?

drewvolz: Case statement would be nice, but I can live with either.

rye: A function like I suggested could use switch-case and would also reduce the size of this render function. (Though, it would be a new call.)  Alternatively, and perhaps even better, this could be a map?  E.g.,
```js
const buttonMessages = {
  'init': 'Report',
  // ...
}

// ...

let buttonMessage = buttonMessages[this.state.status] || 'Error'
```

IDK if that's good coding style in JS, but that would handle the same cases and allow for a denser representation of the possible mappings.

hawkrives: I'm okay with doing an object. Flow even ensures that we don't use a property that doesn't exist on it.

drewvolz (approved):
Nice changes 👍

PR #2073, opened 2018-01-08
Reduce newlines in Card

hawkrives:
This is the one commit that I lost in the git-lfs removal.

This is the same as https://github.com/StoDevX/AAO-React-Native/pull/2062.


rye (approved):
Approved!  Again.

PR #2075, opened 2018-01-08
Circle config tweaks

hawkrives:
One main change here: 

When I was setting up the CircleCI cache, I couldn't decide if I wanted to store both the Yarn package tarball cache _and_ the local node_modules folder, or just one or the other.

I tested just caching the node_modules folder; Yarn decided to go download all of the packages again, every time it ran.

So I went with caching the yarn package cache, so that it would only need to extract the tarballs into the repository. This dropped around 30 seconds from each build.

I thought about storing both node_modules and the tarball cache, but there have been several times where I've needed to wipe out the local folder because of react-native. (such as it storing the third-party downloads in its folder. grr.)

But, since we're not using react-native on Circle (and if we did, I'd probably want each of them to have their own cache anyway), it should be fine to cache node_modules as well.

I'll push another commit to this branch so we can see if there's a speed improvement.


hawkrives:
`yarn` seems to be changing the yarn.lock on Circle's end.

I wonder if their copy is newer than the one we use on Travis for Greenkeeper, and is maybe doing the same change we see locally about `github:vm2`? Either way, this currently offers no speed improvement.

rye:
Yeah I've noticed the `vm2` weirdness cropping up again locally.

PR #2076, opened 2018-01-08
Use Globals and Global Functions for Constant URLs

rye:
This would make it feasibly easier to, for instance, change the repository URL, and also reduces the copying.

I wrote a `GH_PAGES_URL` function which takes a filename, (without a preceding `/`, though that wouldn't break anything) and spits out a url that is prefixed with our GitHub Pages URL.  We can add higher levels of sophistication and error handling to make that more robust, if we want.

Open to suggestions, and potentially other ideas about stuff that can be centralized like this.

### DIFF
--- a/source/views/menus/menu-github.js
+++ b/source/views/menus/menu-github.js
@@ -20,9 +20,9 @@ import {upgradeMenuItem, upgradeStation} from './lib/process-menu-shorthands'
 import {data as fallbackMenu} from '../../../docs/pause-menu.json'
 import {tracker} from '../../analytics'
 import bugsnag from '../../bugsnag'
-const CENTRAL_TZ = 'America/Winnipeg'
+import {GH_PAGES_URL} from '../../globals'
 
-const githubMenuBaseUrl = 'https://stodevx.github.io/AAO-React-Native'
+const CENTRAL_TZ = 'America/Winnipeg'
 
 type Props = TopLevelViewPropsType & {
   name: string,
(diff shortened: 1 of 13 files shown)


source/views/menus/menu-github.js
```diff
@@ -20,9 +20,9 @@ import {upgradeMenuItem, upgradeStation} from './lib/process-menu-shorthands'
 import {data as fallbackMenu} from '../../../docs/pause-menu.json'
 import {tracker} from '../../analytics'
 import bugsnag from '../../bugsnag'
-const CENTRAL_TZ = 'America/Winnipeg'
+import {GH_PAGES_URL} from '../../globals'
 
-const githubMenuBaseUrl = 'https://stodevx.github.io/AAO-React-Native'
+const CENTRAL_TZ = 'America/Winnipeg'
```
hawkrives, line 25: we should probably put `CENTRAL_TZ` in `globals.js` too…

hawkrives (approved):
These changes look good to me.

:grunts in painful anticipation of the next uplift to CARLS:

source/views/menus/menu-github.js
rye: Good idea.  WILCO.

hawkrives:
We, er, actually tried to change the repo name once.

That lasted just a few minutes before @drewvolz and I realized that all of the existing app copies use the old repo name as their base, so we'd need to automatically push data back to this repo name's gh-pages branch.

which sounds actually fairly feasible, now that I say it out loud again.

something to ponder.

rye:
I also notice that there are a few existing places where we have some constants (e.g. `lib/index.js`, which contains a bunch of such things.)

I think this is the moment when this PR pivots from being just a simple globals collection thing to more of a helper function refactoring PR. 😆

drewvolz (approved):
This has been a long time coming. Thanks @rye.

PR #2077, opened 2018-01-08
Use gradients on the homescreen buttons

hawkrives:
Before | After
--- | ---
screen shot 2018-01-07 at 11 16 11 pm | screen shot 2018-01-07 at 11 15 06 pm
Android Before | Android After

  

~I don't remember if this works on Android or not.~ EDIT (@rye): see above.

---

The original gradients here were extracted from Workflow.app for iOS by way of ImageMagick cropping.

There aren't enough anymore to cover all of the homescreen buttons, I don't think, so we'll want to change some of these around (hi Moodle / SIS).


rye:
| Before | After |
|---|---|
| Android Before | Android After |

hawkrives:
thanks @rye for the Android screenshots

rye (dismissed):
Looks good so far.  I don't think this needs to be an opt-in change/have a setting available for it, though I'm not sure if we want to enforce it on Android.  (It's not really too compliant with Android-y flat design style, but I could make the argument that that doesn't matter too much, it _does_ look pretty.)

hawkrives:
@rye I could also make the argument for picking one end of the gradient and using that as the non-gradient tint color, for some level of consistency between platforms.

hawkrives:
@drewvolz What do you think? I'm a bit nervous about this, just because I've been thinking about it for so long.

hawkrives:
@hannesmcman For that matter, what do you think about this change?

drewvolz:
@rye has an interesting point about flat style. I almost want to make it opt-in/opt-out if we're going to force it on Android. For that matter, if the work is there, make it possible to turn it on or off on both platforms.

hawkrives:
I distinctly do _not_ want to make it an option.

drewvolz:
Okay. Would you mind if we eliminated the duplicate gradients? 

purpleToIndigo, yellowToGoldDark, and lightBlueToBlueDark are repeated twice

hawkrives:
We should have no duplicated gradients when we ship, so yes we should remove them.

rye:
I'll check this out in under an hour.

hawkrives:
Assuming this is signed off, let's merge it and put it out in today's beta. We'll reserve the right to revert / disable this before release.

rye:
Currently building screenshots, and if it looks _acceptable_, we can merge.  Well, that is, if my Android emulator turns on nicely.

hawkrives:
@rye It probably looks exactly the same as it did in the OP. I believe the only change has been adding more spacing below the bottom row of buttons. 

But yeah, let's merge and tweak there.

rye (approved):
<details><summary>New Android screenshot</summary>screen shot 2018-01-15 at 07 58 48</details>

I think this is subtle enough that it shouldn't be much of an issue.  Let's give it a try.

PR #2078, opened 2018-01-08
Force gradle to use root package.json

rye:
I don't know what evidence it would have that it should do otherwise, but this PR explicitly tells it to do the right thing.  This was suggested by the other comments in the file.

I don't think this could cause breakage, but we'll see what CI says, now won't we. 😉

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -219,6 +219,7 @@ task copyDownloadableDepsToLibs(type: Copy) {
 
 // Bundles the app's data for the build process
 task bundleData(type: Exec) {
+    workingDir '../../'
     commandLine 'npm', 'run', 'bundle-data'
 }
 


hawkrives:
My danger heuristic is apparently buggy. 

I tested it for actually adding a dep, but not for doing nothing with them. 

Mumble.

hawkrives (approved):

rye:
Once this PR builds successfully, I will merge.  (Not to be overly optimistic...)

drewvolz (approved):

rye:
Travis, ¿why?

hawkrives:
I say ignore the iOS failure. 

Or wait until the morning and rebuild Travis/Mac.

hawkrives:
… did you delete the cache?
  
```
7m 05s: React/double-conversion [Debug]
  7m 02s: Running script 'Install Third Party'
```

rye:
Uh?  No.

hawkrives:
Huh. Third time's the charm, then, unless you also restarted the build… (I did twice this morning)

rye:
Yeah, I think it was a case of 3rd time's the charm.  I didn't restart it, but I watched it pass. 😄

PR #2084, opened 2018-01-08
Only reset the app icon if it has been changed

hawkrives:
iOS will show a notice any time you change the icon, even if you're changing it to what it's set to right now.

Let's avoid showing a dialog unless we need to.

(impetus provided by @drewvolz)


drewvolz (approved):
Thanks @hawkrives!

PR #2096, opened 2018-01-10
Prevent unneeded app icon change

### LINKED ISSUE #2094: Prevent selected icon from resetting
We should find a way to prevent the currently checkmarked row from resetting its icon if it’s already that icon. Pass this.state.iconType and see if it’s a checkmark like we do for accessory. 

https://github.com/StoDevX/AAO-React-Native/blob/master/source/views/settings/icon.js

drewvolz:
Closes #2094 

This prevents the currently selected icon from calling if it is already the app's icon.


rye (approved):

drewvolz:
Closing in favor of #2097

PR #2097, opened 2018-01-10
Refactor settings->choose icon to not prompt again when icon is already selected

hawkrives:
Closes https://github.com/StoDevX/AAO-React-Native/issues/2094

I pulled the Icon-picking table cells into their own component because… i wanted to, tbh.

I dislike defining functions at render time, because you get a new one every time you render. You can call it a premature optimization, but it's just kinda a habit. Also I like naming things.

Anyway, the meat of this change is in `IconCell#setIcon`, wherein we check if the icon you're picking is the selected icon.

@drewvolz, if you don't want to take this approach, we can do it another way.


drewvolz (approved):
This is cleaner. Thanks.

PR #2104, opened 2018-01-11
Make Danger always run

hawkrives:
CircleCI's [`run:` configuration](https://circleci.com/docs/2.0/configuration-reference/#run) allows you to control what certain jobs are run: `always`, `on_success` (the default), or `on_fail`.

I went with `on_success` figuring that I didn't want Danger to try and run if the `yarn install` failed, but I didn't count on the way that the other tasks exit with a non-zero exit code.

Now I'm figuring that even if Danger tries to run without `yarn` having finished properly, it'll just die too, and we'll know something went really wrong because there won't be a message from Danger.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,12 +2,15 @@ version: 2
 
 x-config:
   - &restore-cache-yarn
-      keys:
-        - 'v2-dependencies-{{ checksum "yarn.lock" }}'
-        - 'v2-dependencies-'
+    keys:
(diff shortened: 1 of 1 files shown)


hawkrives:
As a followup to this, I want to enable the `--frozen-lockfile` flag when we run `yarn install` on Circle.

rye (approved):
Sure

drewvolz (approved):

PR #2109, opened 2018-01-12
Make iOS 8.3 our minimum deployment target

### LINKED ISSUE #2079: Bump minimum required iOS version to 8.3
> ld: object file (/Users/travis/Library/Developer/Xcode/DerivedData/AllAboutOlaf-axgziitukyyjlyfvuepevdfannhu/Build/Products/Debug-iphoneos/libRNKeychain.a(RNKeychainManager.o)) was built for newer iOS version (8.3) than being linked (8.2)

I'm okay with doing this because we've never had an 8.2 device show up in analytics. 

At some point, we'll drop 8.x support entirely, but that day has not yet come.

hawkrives:
I'm OK with doing this because in all of our analytics captured over the past ~340 days, we've had exactly 0 users on iOS with anything below iOS 8.3.

Closes #2079.


drewvolz (dismissed):

rye (dismissed):
Sounds good.  BTW, I'm totally fine with being more aggressive based on analytics, so if we find that nobody is using iOS 9.0 or whatnot, we can cut that base out.  It'd save on headaches, I think.

hawkrives:
Yeah, but we still have one person sporadically using iOS 8.3, which is why I bumped to that.

 I'm all for reducing the supported OSes, but I have a hard drive removing support from even 0.1% of our active users (1 of the ~1000 total user count). 

Also I figure that we can pretty easily support three/four major OS versions, and hopefully as students graduate the old devices will age out with them. Anyway. 

Yes for dropping support when we don't see something used for a semester, I think.

hawkrives:
hmm. I'm not positive how to resolve that conflict.

hawkrives:
@rye feel free to merge after approving if you approve. ci passed.

rye (approved):

PR #2114, opened 2018-01-12
Move native builds to Circle

hawkrives:
Travis is having a (few) bad day(s).

Let's try moving all of our PR build status jobs to Circle.

This leaves the following tasks on Travis:

- building tagged builds and uploading to testflight/google play
- building nightlies and uploading to testflight/google play
- running greenkeeper and pushing back to the branch
- building the data and pushing to github pages

All of these tasks have something in common: they need privileges.

It's much easier to move things that don't need encrypted environment variables and ssh keys added so they can clone another git repo… much easier.

With this PR, most of our PRs won't need to hit Travis' infrastructure at all.

---

On Circle, our Android builds run in around 5 minutes, uncached, and around 3, cached. Our iOS builds run in around 6 minutes, uncached, or 5 ½, cached.

On Travis, our Android builds are nearly 8 minutes, and the iOS builds are roughly 20 minutes.

This is more speed than I dared dream of.

Heck, let's compare our iOS build times notice:

<details><summary>Analysis of slow build times (>20s) [Travis]</summary>

```
0m 40s: React/double-conversion [Debug]
  0m 38s: Running script 'Install Third Party'
0m 19s: RCTText/RCTText [Debug]
0m 24s: React/third-party [Debug]
0m 17s: Bugsnag/BugsnagStatic [Debug]
3m 39s: React/React [Debug]
  1m 22s: Compiling RCTModalManager.m
0m 18s: React/jschelpers [Debug]
0m 19s: React/yoga [Debug]
1m 16s: React/cxxreact [Debug]
  0m 25s: Compiling SampleCxxModule.cpp

```
</details>

<details><summary>Analysis of slow build times (>20s) [circle]</summary>

```
0m 25s: React/double-conversion [Debug]
  0m 24s: Running script 'Install Third Party'
0m 10s: React/third-party [Debug]
1m 14s: React/React [Debug]
  0m 30s: Compiling RCTModalManager.m
0m 32s: React/cxxreact [Debug]
0m 11s: React/yoga [Debug]

```
</details>

---

I think there are two Circle config changes to note:

1. I added "junit" test reporting.

    Circle will now say things like "Your build ran 292 tests in eslint with 0 failures. Slowest test: /home/circleci/project/source/analytics.js (took 0.00 seconds)." when you look at the eslint or jest jobs.

2. I removed the fallback caching keys ([see docs on caching](https://circleci.com/docs/2.0/caching/))
    > Because the second key is less specific than the first, it is more likely that there will be differences between the current state and the most recent cache. When a dependency tool runs, it would discover outdated dependencies and update them. This is referred to as a _partial cache restore_.

    We have three caches, currently: yarn, bundler, and gradle. Each of them only takes around 30 seconds to fully repopulate their cache from scratch.

    I'd rather invalidate the caches than deal with future potential (probable?) bugs around stale caches that something failed to bring up to date.

### DIFF
--- a/fastlane/lib/before_all.rb
+++ b/fastlane/lib/before_all.rb
@@ -1,7 +1,8 @@
 before_all do
   case lane_context[:PLATFORM_NAME]
   when :ios
-    setup_travis
+    setup_travis if travis?
+    setup_circle_ci if circle?
   end
 
   # set up global info for `gym`
--- a/fastlane/lib/util.rb
+++ b/fastlane/lib/util.rb
@@ -1,15 +1,31 @@
 # coding: utf-8
 
 def should_deploy?
-  cron? ||
-    !ENV['TRAVIS_TAG'].empty? ||
-    ENV['TRAVIS_COMMIT_MESSAGE'] =~ /\[ci run beta\]/
+  cron? || tagged? || forced_deploy?
 end
 
 def cron?
   ENV['TRAVIS_EVENT_TYPE'] == 'cron'
 end
 
+def tagged?
+  !ENV['TRAVIS_TAG'].empty?
+end
+
+def forced_deploy?
+  ENV['TRAVIS_COMMIT_MESSAGE'] =~ /\[ci run beta\]/
+end
+
 def pr?
-  ENV['TRAVIS_PULL_REQUEST'] != 'false'
+  # todo: figure out how to handle forked-pr/non-forked, or if we even want to
+  # ENV['TRAVIS_PULL_REQUEST'] != 'false' || ENV['CIRCLE_PR_NUMBER']
+  ENV['TRAVIS_PULL_REQUEST'] != 'false' || false
+end
+
+def travis?
+  ENV['TRAVIS'] == 'true'
+end
+
+def circle?
+  ENV['CIRCLECI'] == 'true'
 end
--- a/fastlane/lib/util.rb
+++ b/fastlane/lib/util.rb
@@ -9,7 +9,7 @@ def cron?
 end
 
 def tagged?
-  !ENV['TRAVIS_TAG'].empty?
+  ENV['TRAVIS_TAG'] && !ENV['TRAVIS_TAG'].empty?
 end
 
 def forced_deploy?
--- a/fastlane/lib/versions.rb
+++ b/fastlane/lib/versions.rb
@@ -19,8 +19,8 @@ def current_build_number(**args)
 end
 
 # get the current build number from the environment
-def build_number()
-  return ENV['TRAVIS_BUILD_NUMBER'] || ENV['CIRCLE_BUILD_NUM']
+def build_number
+  ENV['TRAVIS_BUILD_NUMBER'] || ENV['CIRCLE_BUILD_NUM']
 end
 
 # Get the current "app bundle" version
--- a/fastlane/lib/versions.rb
+++ b/fastlane/lib/versions.rb
@@ -2,7 +2,7 @@
 
 # Gets the version, be it from Travis, Testflight, or Google Play
 def current_build_number(**args)
-  return ENV['TRAVIS_BUILD_NUMBER'] if ENV.key?('TRAVIS_BUILD_NUMBER')
+  return build_number if build_number != nil
 
   begin
     case lane_context[:PLATFORM_NAME]
@@ -16,6 +16,11 @@ def current_build_number(**args)
   rescue
     '1'
   end
+end
+
+# get the current build number from the environment
+def build_number()
+  return ENV['TRAVIS_BUILD_NUMBER'] || ENV['CIRCLE_BUILD_NUM']
 end
 
 # Get the current "app bundle" version
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,17 +20,17 @@ x-config:
     paths: [~/Library/Caches/Yarn]
   - &restore-cache-bundler
     keys:
-      - 'v1-ruby-dependencies-{{ checksum "Gemfile.lock" }}'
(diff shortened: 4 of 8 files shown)


rye:
>On Circle, our Android builds run in around 5 minutes, uncached, and around 3, cached. Our iOS builds run in around 6 minutes, uncached, or 5 ½, cached.

This is a **_huge_** win.  My only concern is the limit that Circle has for us; @hawkrives, what does the pricing look like for any overages?  I'd be willing to check this PR out in "real life" as long as we keep an eye on quota or figure out what our solution will be.  (We're currently sitting at 12h11m = 731 minutes of our 1500 minutes of monthly Linux quota?  I think)  From what it sounds like, Circle are friendly to OSS projects, so we might want to get in contact with them and explain the situation.  We could ask for any suggestions they have about what they like from projects—since our jobs are <10m, (most <5m) we should be able to reduce our impact on their build infrastructure as a whole.  I'm just curious where we should plan on standing, and I definitely don't want a service cut-off if we suddenly run into issues.

I wonder if we should push for greater use of the `[skip ci]` tags to prevent usage by things that needn't be built fully?  I'm open to ideas.

I'll review officially shortly.

fastlane/lib/versions.rb
```diff
@@ -18,6 +18,11 @@ def current_build_number(**args)
   end
 end
 
+# get the current build number from the environment
+def build_number()
```
rye: No empty `()`'s please

fastlane/lib/versions.rb
```diff
@@ -18,6 +18,11 @@ def current_build_number(**args)
...
+  return ENV['TRAVIS_BUILD_NUMBER'] || ENV['CIRCLE_BUILD_NUM']
```
rye: Don't `return`, just put everything that follows after that.  (`return` is not required in Ruby, provided you are at the end of a block or method.)

```ruby
def build_number
  ENV['TRAVIS_BUILD_NUMBER'] || ENV['CIRCLE_BUILD_NUM']
end
```
should be sufficient.

hawkrives:
@rye I found a [post on their forums](https://discuss.circleci.com/t/free-plan-build-minutes-limit-on-circleci/17012) recently:

> If you’re building an Open Source project on MacOS you can contact us to request your project be covered by our Open Source plan which will remove the minutes limitation. OSS projects built on Linux do not count towards your usage limit.

First, we are on their open-source plan – if we weren't, we wouldn't be able to use the macOS builds at all. (Plus I emailed them in the fall asking to be put on it and they said we already were.)

Second, I almost read that message to say that there are no limits for projects on the open-source plan on the Linux VMs, either.

We can either reach out and contact them, or wait and see what happens when we go "over".

---

Other reasons I think my interpretation of the message above is right:

- "0 container is paid. 1 container is free." says the Org Settings page, yet we have 4 containers as we're on the open-source plan (and we can run four jobs in parallel, so we definitely have them)
- "No macOS plan selected" says the Org Settings page, yet we can run macOS builds

so what I'm getting at here is that I think that parts of their system aren't designed to show the status of open-source projects.

---

Or we could just email their support; that works too.

fastlane/lib/util.rb
```diff
@@ -1,15 +1,31 @@
 # coding: utf-8
 
 def should_deploy?
-  cron? ||
-    !ENV['TRAVIS_TAG'].empty? ||
-    ENV['TRAVIS_COMMIT_MESSAGE'] =~ /\[ci run beta\]/
+  cron? || tagged? || forced_deploy?
```
rye, line 4: This is good Ruby. 👌

fastlane/lib/before_all.rb
```diff
@@ -1,7 +1,8 @@
 before_all do
   case lane_context[:PLATFORM_NAME]
   when :ios
-    setup_travis
+    setup_travis if travis?
```
rye, line 4: Weird edge case here if both `travis?` and `circle?` return true, but I don't think we need to worry about that.  Open to other suggestions, but this certainly... looks better?

rye (changes requested):
This looks mostly good to me, just a couple of tweaks to the Fastlane changes.

hawkrives:
To make things more confusing, the Linux 4x parallelism plan has ∞ minutes per month.

---

If we assume that we're actually limited on our build minutes:

- If we go over on the macOS builds, it's 8¢/minute.
- If we go over on the Linux builds… we pay $50? maybe?

fastlane/lib/before_all.rb
hawkrives: I don't think we need to worry about that, unless Circle decides to try and pretend to be Travis.

Plus, from skimming the source ([`setup_travis`][t], [`setup_circle_ci`][c]), it looks like they might not even conflict with each other. Maybe the keychain setup would, IDK. 

[t]: https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/setup_travis.rb
[c]: https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/setup_circle_ci.rb

tl;dr I don't think we need to _ensure_ that we only run one of them.

fastlane/lib/versions.rb
hawkrives: Kk

rye (approved):
I think this was my only real gripe here.  Pending another review, maybe, and a passing CI run, :shipit:.

Also, I like watching the circle slowly fill in from empty to full.

hawkrives:
@rye Then let's wait for CI to pass and :shipit:!

hawkrives:
and CI is done!

rye:
:shipit: 🎉

PR #2118, opened 2018-01-13
Ensure that lockfiles don't change on CI

### LINKED ISSUE #2117: Pass --frozen-lockfile to "yarn install" on Circle
From #2081, by me:
> I'd like to combine the upgrading of yarn with adding the --frozen-lockfile argument, which sounds like it will fail the build if the lockfile changes. A bit drastic, maybe, but that lockfile should only ever change as a result of the Greenkeeper build, so I'd like to enforce that assertion.

hawkrives:
Closes #2117 

Let's see what happens!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,7 +46,7 @@ jobs:
     steps:
       - checkout
       - restore_cache: *restore-cache-yarn
-      - run: yarn install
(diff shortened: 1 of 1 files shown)


hawkrives:
Well, everything passed! Which is what I expected, tbh.

rye (approved):
Yay, reproducibility!

PR #2121, opened 2018-01-13
Remove codeclimate / csslint / appveyor

hawkrives:
Codeclimate and Appveyor, I'm removing because we don't use them anymore.

CSSLint I'm removing because we have a single CSS file and I want fewer top-level files 😸


rye (approved):
For the better, I'd say.

Issue #2120, opened 2018-01-13
Create non-2fa user for iTC access


hawkrives:
as seen in https://docs.fastlane.tools/best-practices/continuous-integration/

and give it the longest password that iTunes Connect will let us use

hawkrives:
Looks like I already have one. Next step: add it to iTC and configure Travis / fastlane to use _it_ to log in, instead of me.

PR #2123, opened 2018-01-13
Use an Apple ID without two-factor auth enabled for Fastlane / Testflight

hawkrives:
Closes https://github.com/StoDevX/AAO-React-Native/issues/2120.

I also did a bit of cleanup of the Travis env variables, which I've documented in the commits attached.

I don't think there's a good way to test this without just rolling it out and seeing if the nightly fails. _Especially_ not this weekend, since we're facing Travis queues of ≥ 24 hrs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,18 @@ env:
   # To update the secure tokens, run `travis encrypt NAME='VALUE' --add` after
   # installing the travis gem with `gem install travis`.
   global:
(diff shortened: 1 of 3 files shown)


rye (approved):
Hum, well.  This is a good thing.

hawkrives:
I'm going to merge this, because I don't have a good way to test it.

Issue #1799, opened 2017-10-18
Catch "cannot open mailto: urls" errors and alert the user


hawkrives:
## Error in all-about-olaf

**Error** in **main.jsbundle:17**
Unable to open URL: mailto:***@stolaf.edu

[View on Bugsnag](https://app.bugsnag.com/gobbldygook/all-about-olaf-1/errors/59e7c555be71a6001a70fef1?event_id=59e7c5af29fa37001a743db9&i=gh&m=ci)

## Stacktrace

    main.jsbundle:17 - l

[View full stacktrace](https://app.bugsnag.com/gobbldygook/all-about-olaf-1/errors/59e7c555be71a6001a70fef1?event_id=59e7c5af29fa37001a743db9&i=gh&m=ci)

*Created by Frog Pond via Bugsnag*

hawkrives:
where ***@stolaf.edu == "rives@stolaf.edu", for example

This in on 2.4.0 as well.

hawkrives:
I think we should add a try/catch somewhere (in openUrl, I guess) to check for these errors and ignore mailto:-caused errors.

drewvolz:
Does the action happen, or does the app crash? What is the outcome of a failed `openURL` with a `mailto`?

rye:
To me, it seems like this is something that we don't need to worry about; a failed `openURL` would result in the user not... getting any result from their action?  We could throw an alert in the UI.

It seems as though the result is a crash:

```js
{
  "code": 0,
  "exception_name": "EXC_CRASH",
  "subcode": 0,
  "exception": 10
}

// [...]

{
  "name": "SIGABRT",
  "signal": 6,
  "code": 0
}
```

My guess is that this crash is just because this exception bubbles up and ends up unhandled.

hawkrives:
We should handle it and alert that the mailto link couldn't be opened, IMHO

hawkrives:
I believe that our next step is to create a "sendEmail" function like our "openUrl" function; it should handle our common use cases (namely, a single author), and provide a central place to handle errors (such as showing an alert when the user can't handle `mailto:` URLs, or copying the email address to the clipboard, or something.)

Here are the places where we call the `email` function of `react-native-communications`. I don't think that we need to build our generic solution around that library.

```
source/views/settings/sections/support.js
29:    email(

source/views/student-orgs/detail.android.js
116:                onPress={() => this.openEmail(c.email, orgName)}
[...]

PR #2126, opened 2018-01-14
Allow remote configuration of the Report Issue / Help module

### LINKED ISSUE #2054: "Report a Problem" screen tweaks

hawkrives:
<details><summary>screenshots</summary>

Note | Screenshot
--- | ---
Not scrolled | screen shot 2018-01-13 at 11 48 01 pm
Scrolled | screen shot 2018-01-13 at 11 48 04 pm
Disabled tool, but not hidden | screen shot 2018-01-13 at 11 54 00 pm
A single button disabled | screen shot 2018-01-13 at 11 54 34 pm
Disabled with custom message | screen shot 2018-01-13 at 11 59 07 pm
</details>

Whee!

I think this feature is just about ready to go.

I've built out the ability to configure simpler "tools" like the Helpdesk (open url, send email, call phone) completely from the remote json file, while we can still enable/disable/hide/edit the title/description of the wifi reporter.

We can also configure the order of the rendered tools by reordering the YAML files that make up the data file.

I think the current files I've included to configure the app are fairly explanatory, but I will note that the Facilities entry includes an idea-based remote form configuration. For now, I'm just opening the (hideous) facilities form in an SFWebView – I don't want to add an http exception, and I don't want to build a native form right now. If you want to report it badly enough, you'll suffer through the form lol.

We can disable individual buttons on the tools, too. IDK why.

EDIT: We can also restrict tools to only certain versions of the app; there's a key that takes a server-compatible version query, like `>=2.5.0 <2.6`, so we could build several tools in case there's something backwards incompatible in here somewhere, and only the one that worked on your app version would show.

I put a message in the prose for the Helpdesk to explain that you need to be on-campus for the website link to work. I don't really want to do anything more at this time.

I tweaked the styles of the cards again; I removed the big bold Title style, because I remembered that there was a `header={}` prop for Card itself. If we don't like that style, we should think about improving it everywhere, not just changing it here. I also removed the left/right edge margin, and fixed the bottom margin on the list to match the top margin and the space between the cards.

---

Should we add a button in here for getting help with the app, too? Just a random late-night thought.

TL;DR: After this PR, this feature is no longer a shipping hold.

Closes #2054. We can open a new issue if we need more tweaks. 😸


hawkrives:
The Flow errors are because this depends on the `<Card/>` tweaks in #2125

hawkrives:
I think that some of the communications-esque methods in the new tools file could be pulled out into something to solve https://github.com/StoDevX/AAO-React-Native/issues/1799.

rye:
I just approved #2125, I'll take a look once the thing rebuilds after that gets merged.

data/_schemas/help.yaml
```diff
@@ -0,0 +1,90 @@
...
+  enabled: {type: boolean}
+  hidden: {type: boolean}
+  message: {type: string}
+  versionRange: {type: string}
+  title: {type: string}
+  body: {type: string}
+  buttons:
+    type: array
+    items:
+      # TODO: convert this to "enum" instead of "oneOf"
```
rye, line 16: Does this TODO need resolving in this PR?

rye (approved):
Looks pretty good to me.  I have no idea how well/if it will work in the wild, but we'll see, now won't we?  I'm fine with having this go into tomorrow's nightly.

data/_schemas/help.yaml
hawkrives: … no

I've tested that `oneOf` properly rejects items if they don't match. I believe that `enum` would provide nicer errors, but I couldn't get it to work.

So, no, this doesn't need to be resolved for a merge.

PR #2128, opened 2018-01-14
move refresh_dsyms call into the main fastlane invocation

hawkrives:
Travis "failed" tonight because the refresh_dsyms lane failed, because I think it tried to run `setup_travis` a second time, and something was nasty about setting up the keychain again.

https://travis-ci.org/StoDevX/AAO-React-Native/builds/328644162

This should fix that problem.


rye (dismissed):
Huh. Well, give it a try.  We can run a forced beta, I think?

hawkrives:
Oh dang it, I need to restrict it to just Travis. Tomorrow!

rye (approved):

PR #2130, opened 2018-01-14
Use generic phone call and email functions

hawkrives:
I pulled the new callPhone and sendEmail functions that I made for the Help views out into generic, reusable files.

These are nicer than the `react-native-communications` ones not only because I made the arguments nicer, but also because they automatically handle alerting the user when they can't send an email or call a phone number.

Closes https://github.com/StoDevX/AAO-React-Native/issues/1799


rye (dismissed):
Nice!

rye (approved):

PR #2132, opened 2018-01-14
Center bullets vertically in Markdown text

hawkrives:
Before | After
--- | ---
 | screen shot 2018-01-14 at 4 51 09 pm

I added some right-hand spacing to the paragraphs / list items, and configured the bullets to be centered.

What I'd _really_ like to do would be to align the bullet with the baseline of the first line of text, but I couldn't figure out how to manage that.

<details><summary>Example of what I'd like</summary>

from [https://stackoverflow.com/a/46891740/2347774](https://stackoverflow.com/a/46891740/2347774)

7vwgm
</details>


rye (approved):
I like this; though it would be better to align to the baseline on the top, this looks almost as appealing.

PR #2135, opened 2018-01-15
Forbid unnecessary curly brackets in jsx attrs

hawkrives:
Sometimes, we get code written like this: `<Element attr={'foo'} />`, instead of `<Element attr="foo" />`. The curly braces provide no benefit.

I found an ESLint rule to forbid them, [`react/jsx-curly-brace-presence`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-curly-brace-presence.md), and this PR enables it.

If you find yourself running afoul of this rule often, know that it's autofixable with `yarn lint --fix`.

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -71,6 +71,7 @@ rules:
 
   # react rules – https://github.com/yannickcr/eslint-plugin-react
   react/display-name: off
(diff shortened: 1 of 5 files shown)


rye (approved):
Looks good.

PR #2141, opened 2018-01-15
Add TouchID/FaceID protection of SIS balances information

### LINKED ISSUE #2124: Add TouchID/Face ID for login to SIS and SIS module
This would make the login process easier and add an additional layer of protection for the SIS module for those who desire it. Check out https://github.com/naoufal/react-native-touch-id

hannesmcman:
Users will be able to enable Touch ID or Face ID protection of the SIS balances module.

A diff without whitespace checks is available at https://github.com/StoDevX/AAO-React-Native/pull/2141/files?w=1.

Closes #2124


hannesmcman:
I'm having an issue with FaceID: a non-matching face is followed by a prompt to try again, but the try again button doesn't do anything...I have also been unable to test this on android so if someone could do that that would be great!

hawkrives:
I believe that https://facebook.github.io/react-native/docs/alertios.html will do everything you're trying to make `react-native-modal` do, and we'll want to find a non-modal-dialog solution for Android anyway, because I don't think modals with text inputs are common there.

ios/AllAboutOlaf.xcodeproj/project.pbxproj
```diff
@@ -49,6 +48,16 @@
 		3AE408501E1E280800F0FD83 /* LaunchScreen.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 3AE4084F1E1E280800F0FD83 /* LaunchScreen.storyboard */; };
 		832341BD1AAA6AB300B99B32 /* libRCTText.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 832341B51AAA6A8300B99B32 /* libRCTText.a */; };
 		BBDFD91C684046CB9FD0CD55 /* Entypo.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 742B362B86E443C09A7F9FD9 /* Entypo.ttf */; };
+		A3EA8C86662742928E156E45 /* libTouchID.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 905F5155B95042B89DE3FADA /* libTouchID.a */; };
+		2FE5DD35D6C74CBDAA79E5E9 /* EvilIcons.ttf in Resources */ = {isa = PBXBuildFile; fileRef = E7F4C82AC8DF4ADEBE2ED75C /* EvilIcons.ttf */; };
```
hawkrives: We usually run `yarn react-native link $library`, where $library is (in this case) `react-native-touch-id`, so that we don't re-link all the stuff we've removed in the past.

ios/AllAboutOlaf.xcodeproj/project.pbxproj
```diff
@@ -49,6 +48,16 @@
...
+		911AFFD0515E4820870BD025 /* Feather.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 7578036A3F5A41D5ABC88414 /* Feather.ttf */; };
+		D314365E6D264BFC9EE8CFFE /* FontAwesome.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 0D4853F7133F4472B8EA09F6 /* FontAwesome.ttf */; };
+		46A7D3950A5A4F6193ED2E4E /* Foundation.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 4D07275C973147788D8776E7 /* Foundation.ttf */; };
+		A0198A761F4B421380FB751B /* MaterialCommunityIcons.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 93C62C74F0F34814A515C299 /* MaterialCommunityIcons.ttf */; };
+		5445064DAFEF4267891C7732 /* MaterialIcons.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 1A1797A4E0CD4A159B7E5221 /* MaterialIcons.ttf */; };
+		A9C2CF0EA1E54AEDA882B107 /* Octicons.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 120227DBD4FE4E44AE8B829F /* Octicons.ttf */; };
+		4B899B1338F2467FBC64A25D /* SimpleLineIcons.ttf in Resources */ = {isa = PBXBuildFile; fileRef = A236B5D5F384407EAC985D9B /* SimpleLineIcons.ttf */; };
```
hawkrives: I'll need you to remove the fonts that got added to the Android folders, in addition to any added lines that reference the .ttf files in this .pbxproj.

ios/AllAboutOlaf.xcodeproj/project.pbxproj
```diff
@@ -956,8 +987,8 @@
 		83CBB9F71A601CBA00E9B192 /* Project object */ = {
 			isa = PBXProject;
 			attributes = {
-				LastSwiftUpdateCheck = 0810;
-				LastUpgradeCheck = 0910;
+				LastSwiftUpdateCheck = 810;
+				LastUpgradeCheck = 910;
```
hawkrives: Xcode and react-native fight about whether to add the leading-0 here. Xcode adds it, so if you could revert this change that'd be great.

ios/AllAboutOlaf.xcodeproj/project.pbxproj
```diff
@@ -967,7 +998,7 @@
 					};
 					13B07F861A680F5B00A75B9A = {
 						DevelopmentTeam = TMK6S7TPX2;
-						LastSwiftMigration = 0810;
+						LastSwiftMigration = 810;
```
hawkrives: same thing about leading 0s

ios/AllAboutOlaf.xcodeproj/project.pbxproj
```diff
@@ -1591,6 +1631,7 @@
 				LIBRARY_SEARCH_PATHS = (
 					"$(inherited)",
 					"\"$(SRCROOT)/$(TARGET_NAME)\"",
+					"\"$(SRCROOT)/$(TARGET_NAME)\"",
```
hawkrives: `react-native link` also duplicates this search path. Please un-duplicate it.

ios/AllAboutOlaf/Info.plist
```diff
@@ -91,11 +91,20 @@
 	<key>NSCalendarsUsageDescription</key>
 	<string>We use your calendar to add events to your calendar so that you remember what you wanted to attend.</string>
 	<key>NSLocationWhenInUseUsageDescription</key>
-	<string>We only use your location when you would expect us to, and never when you aren&apos;t using the app.</string>
+	<string>We only use your location when you would expect us to, and never when you aren't using the app.</string>
```
hawkrives: Revert, please

ios/AllAboutOlaf/Info.plist
```diff
@@ -91,11 +91,20 @@
...
 	<key>UIAppFonts</key>
 	<array>
 		<string>Entypo.ttf</string>
 		<string>Ionicons.ttf</string>
+		<string>EvilIcons.ttf</string>
+		<string>Feather.ttf</string>
+		<string>FontAwesome.ttf</string>
+		<string>Foundation.ttf</string>
+		<string>MaterialCommunityIcons.ttf</string>
```
hawkrives: Remove all of these added references, please.

hawkrives (changes requested):
`react-native link` is annoying.

source/lib/storage.js
```diff
@@ -62,6 +62,22 @@ export function getAcknowledgementStatus(): Promise<boolean> {
   return getItemAsBoolean(acknowledgementStatusKey)
 }
 
+const touchIDEnabledKey = 'settings:touchIDEnabled'
```
hawkrives: I don't think we need to store the TouchID auth / enabled status in AsyncStorage.

source/flux/parts/settings.js
```diff
@@ -35,6 +39,9 @@ const CREDENTIALS_VALIDATE_FAILURE = 'settings/CREDENTIALS_VALIDATE_FAILURE'
 const SET_FEEDBACK = 'settings/SET_FEEDBACK'
 const CHANGE_THEME = 'settings/CHANGE_THEME'
 const SIS_ALERT_SEEN = 'settings/SIS_ALERT_SEEN'
+const SET_TOUCHID_STATUS = 'settings/SET_TOUCHID_STATUS'
+const LOAD_TOUCHID_STATUS = 'settings/LOAD_TOUCHID_STATUS'
```
hawkrives, line 43: I don't think we need to store the Touch ID status in Redux.

hawkrives (commented):
Two comments

hawkrives:
I'm going to bring this up to speed with `master`, because I added an ESLint warning.

source/flux/parts/settings.js
hannesmcman: @hawkrives If not, how will the app remember whether or not the user wants this feature enabled or not? Saving this in Redux allows the user to toggle whether or not they want the feature enabled in settings 
screen shot 2018-01-16 at 5 37 35 pm

rye:
This seems like a v2.6-type of thing, I'll add it to that milestone?  Feel free to drop it back into v2.5, but we might miss the train on that one.

hawkrives:
2.6 sounds fine to me.

hawkrives:
(unless you really want this out _now_, @hannesmcman 😄)

hawkrives:
I think that we should to store a "passcode" that we get from the user, and use that when Touch ID fails instead of using their stolaf password?

Maybe?

hannesmcman:
That could work...idk though I feel like it's rare that Touch ID won't work and it might be hard to remember what your passcode is..not to mention that we would have to implement a secure way to reset it..

hawkrives:
Ok

I don't want a secondary password to protect this screen

So what if we just forget the user's credentials from Keychain when FaceID fails, and make them log in again?

hannesmcman:
Yeah that would work! May I ask why you want to avoid using a modal? Overkill?

hawkrives:
@hannesmcman (and @drewvolz): because we just changed every file in the repository in #2153, here's what I just did to update the `stoprint-take-2` branch with `master`:

```bash
# be sure that your copy of master is up-to-date – checkout master and pull it to be sure
$ git stash
$ git checkout master
$ git pull origin master
$ git checkout mybranch
$ git stash pop
# add "useTabs": true, as the first line in the package.json "prettier" section
$ subl / vim / emacs package.json
# convert all the files to use tabs
$ yarn prettier
$ git add .
$ git commit -m 'convert indentation to tabs'
# now merge with master, preferring our current branch whenever there are changes.
# because we already replicated master's tab-indents, this should be fine.
$ git merge -s recursive -X ours master
```

source/flux/parts/settings.js
drewvolz: I think @hannesmcman is right. We need some way to give the user the option to opt-in or out.

hannesmcman:
So I've tested it a bit on my android emulator and there is a very strange bug...Touch ID works until you enter a non-matching fingerprint and enter your password. Once you have been granted access by entering your password (which works as expected), if you leave the SIS module and come back to it Touch ID no longer works at all...I think this is a bug in react-native-touch-id, but I'm looking into it. Otherwise, everything works fine on iOS as far as I can tell

hawkrives (changes requested):
I've been talking this over with Drew, because he wanted to change my mind on this and failed.

I'd like one of three things to happen here:

1. **Protect the entire app behind biometric authentication**

    I believe this would be simpler than protecting the content of a single tab from an implementation perspective, and the current biometrics hardware is fast enough that it wouldn't harm usability, especially when the feature is off by default.

2. **Hide the contents of the Balances tab on load, with a button to launch the authentication process**

    My problem with presenting the auth dialog on tab open is that I might want to go to Student Work, or now TES, and yet I have to wait for the auth to pass or fail (or cancel it) before I can open the tab I actually wanted to open. If there were a button launched the authentication process, I'd be much happier.

3. **Nix it.**

    The app is already protected behind your phone's biometric auth and/or passcode.

    Most financial apps require auth because you can _do things_ with the balances, IMO. We can't; we can't even add to the balances, much less spend from them, so I really don't see the point in adding auth at all…

As always, my mind can probably be convinced of another path beyond these three options if you have one.

I just wanted to give you an update of where my thinking is on this feature. I'd never quite thought it through while it was in the backlog.

hannesmcman:
I like your second option best. I think protecting the whole app is overkill and I understand the UX problem caused by launching TouchID upon opening the SIS view. And I think the feature is worth implementing since I've heard people ask for it and most of the work is already done.

hawkrives:
@hannesmcman I know you opted for option 2, but _if_ you wanted to do option 1 you might look into https://github.com/crazycodeboy/react-native-splash-screen to hide the app while it's authenticating. (That lib's not needed for option 2, since only the "please authenticate" screen will be shown behind the dialogs.)

hannesmcman:
Honestly I think I agree with @hawkrives that this is probably not worth fully implementing into the app, especially since it would only work on iOS. We could certainly revisit it if anyone disagrees. I would think it could be valuable if, at any point in the future, we allow users to view their grades, or other information that is more sensitive that their account balances and remaining meals. I'm closing this for now, along with #2124.

PR #2144, opened 2018-01-16
Remove push notifications from the app

hawkrives:
- We've never hooked them up
- The iOS "Push Notifications" entitlement is a pain for external developers
- We've never hooked them up

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -199,7 +199,6 @@ dependencies {
     compile project(':react-native-linear-gradient')
     compile project(':react-native-maps')
     compile project(':react-native-network-info')
-    compile project(':react-native-onesignal')
     compile project(':react-native-restart')
     compile project(':react-native-vector-icons')
     // this is for react-native itself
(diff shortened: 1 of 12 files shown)


hawkrives:
… apparently _none_ of my fancy Danger Xcode/Android checks work right. I swear I tested them…

drewvolz:
I have a hope that we’ll one day use them. But doesn’t mean anything.

drewvolz (approved):
If we do use them... I guess we’ll put the work back. So I guess nix them.

hawkrives:
@drewvolz My thinking is that none of the other partially-complete work we keep around causes trouble for people, and we tend to remove it when it does (e.g., rn-view-shot, for me.)

This has never even been functional past a "log to the console when I get a push maybe" state, so I don't think it's worth keeping around at _all_.

@rye?

hawkrives:
oops

drewvolz:
Yeah. One day we’ll find a decent use for push. But for now it is a headache for everyone. When this passes, I am fine with merging it. I would like to hear from other developers on this project, though.

rye (approved):
Some day we'll get there, but yeah.  Go on ahead.

PR #2148, opened 2018-01-16
WIP: Explore moving nightlies and tagged builds to Circle

hawkrives:
WIP WIP WIP

Ideas:

- Share a build context between two macOS / Linux VMs to share the built app, and use the first one to build and the second to deploy, allowing more workflow-level control?


hawkrives:
Current status: auth works

Something in the iOS build isn't writing files to logs, so Danger's failing.

hawkrives:
Turns out that I had only asked check_build to write, not build itself. 

Now I just need to clean up the config and make sure the rest of the logic transferred.

rye (commented):
We will want to remove all the extra commented gobbledygook later, but this looks very good so far.

hawkrives:
Closed in favor of #2172

PR #2152, opened 2018-01-17
Make the KSTO stream a bit less hacksy

### LINKED ISSUE #1805: KSTO for Android is stuck while starting the stream
It does not go past the button message of "Starting". Works on iOS.

rye:
This PR attempts to solve our issues on Android by switching to loading the actual player uri rather than implementing our own player.

Some advantages to doing this this way:

- If BMS/KSTO changes the player, those changes are given to us by default.
- If BMS/KSTO wants to change the way they do analytics, they are free to do as they wish.
- If BMS/KSTO feels like completely re-implementing from the ground up, as long as this player does not change, we are fine.

We're still running into issues with the Android player, though, so this is a WIP.

Closes #1805

### DIFF
--- a/source/views/streaming/radio/controller.js
+++ b/source/views/streaming/radio/controller.js
@@ -165,7 +165,7 @@ export class KSTOView extends React.PureComponent<Props, State> {
             <ShowCalendarButton onPress={this.openSchedule} />
           </Row>
 
-          <StreamPlayer
+          {Platform.OS !== 'android' ? <StreamPlayer
             onEnded={this.handleStreamEnd}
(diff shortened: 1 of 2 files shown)


rye:
Working hypothesis is that this is something due to the disparity between iOS and Android handling, which would explain why the prior working example (which was basically a stripped-down) copy of the web view was causing problems.

Anyhow, our solution for v2.5 is probably just going to be to drop trying to get something native-y without a custom package, and instead, moving to just opening the player on Android.

hawkrives (dismissed):

hawkrives (approved):

PR #2153, opened 2018-01-17
Switch Prettier to using tabs instead of spaces 😈

rye:
Prettier supports this now, and I like tabs more.  As long as Prettier is managing the indentation, (which it is) we should be fine.


hawkrives:
Whitespace diff: https://github.com/StoDevX/AAO-React-Native/pull/2153/files?w=1

hawkrives:
@rye that's a bit big

rye:
Would you rather one PR for each file? 😉

hawkrives (dismissed):

drewvolz:
Does this bump your contributor status waaaay up??

drewvolz (dismissed):

rye:
Hehehe yeah it does.

hawkrives:
That's his real goal…

rye:
I mean, you're not wrong, but also tabs are probably better and it doesn't matter since prettier is the source of truth.

hawkrives (approved):

PR #2166, opened 2018-01-19
tint the active switches with their row color

hawkrives:
screen shot 2018-01-18 at 6 37 54 pm

Do we want to color the Android arrow icons?

### DIFF
--- a/source/views/home/edit/row.ios.js
+++ b/source/views/home/edit/row.ios.js
@@ -175,6 +175,7 @@ export class EditHomeRow extends React.Component<Props, State> {
 				</Text>
 
 				<Switch
+					onTintColor={tint}
 					onValueChange={this.onToggleSwitch}
 					value={this.props.isEnabled}
 				/>


source/views/home/edit/row.ios.js
```diff
@@ -175,6 +175,7 @@ export class EditHomeRow extends React.Component<Props, State> {
 				</Text>
 
 				<Switch
+					onTintColor={tint}
```
drewvolz, line 178: Should the option be added as an optional to the cell in rn-simple-tableview, too? https://github.com/StoDevX/AAO-React-Native/blob/master/source/views/components/cells/toggle.js

hawkrives: Not until we want to use it there, I'd think

drewvolz: I only argue it out of consistency. But okay.

hawkrives: I argue against it because of consistency lol – I want us to use the default color everywhere; this is the exceptional place.

drewvolz (approved):

PR #2167, opened 2018-01-19
Bring iOS deployment target back down to 8.2

hawkrives:
because iTunes

ERROR ITMS-90682: "Invalid Bundle. Submitting asset catalogs targeting iOS 8.3 or iOS 8.4 is currently unsupported. You'll either need to change the deployment target back to iOS 8.2 (or lower), or move it forward to iOS 9.0 or later."

I believe this is why we were on 8.2 originally…

We'll need to update the release notes, too.


drewvolz:
Oof... yes that looks familiar

rye (approved):
WTF Apple.  Anyhow.

PR #2170, opened 2018-01-19
Be nice to Circle (enforce Gradle, React-Native Bundler parallelism limits)

hawkrives:
Hopefully this will stop the spurious Android build errors we've been seeing on Circle.

I say merge this, if the build passes. If this slows down our local dev too much, we can play with increasing it a bit.

- - -

Additional stuff by @rye:

- Further reduced gradle limits to `512m`.  This _should_, ya know, be enough.  But we all know how that goes.

- Noticed when running the failing builds over SSH that a bunch of `node` processes were spinning up when the bundler was running, which is why that task always fails.  I hypothesize that this is because we don't _actually_ have limits on CircleCI when the container runs, so Node sees _lots_ of CPUs. (like, 64 or 32 of them)  [Metro](https://github.com/facebook/metro/blob/0f7ad193c75183eeff1b356644ccf22b0813bb04/packages/metro/src/commands/build.js#L39) uses some stupid syntax to check what the CPU count is, and so happily kicks off 64 or whatever workers. (which probably frightens Circle which gets stabby and kills all of our workers, thus ending gradle with no notice)

  - This will reduce the overall possible parallelism but will also ensure that we keep well under our limit.  Currently the new maximum if the `CI` variable is set is `2` workers.


drewvolz:
What gave you this idea?

rye:
A couple more things we can try:

- Turn off `org.gradle.parallel=true` in the `gradle.properties` (it's incubating, and I don't know if it's actually running in parallel?

- Turn on heap dumps on OOM? (`-XX:+HeapDumpOnOutOfMemoryError`)

I'm going to try with both of the options locally to see if there is any speed hit doing both of these things.

hawkrives:
@drewvolz https://encrypted.google.com/search?hl=en&q=node%20137, mainly. 137 seems to be both node and gradle's "OOM" error exit code.

hawkrives:
@rye I'm going to add `-XX:+HeapDumpOnOutOfMemoryError` since it just failed again.

hawkrives:
@rye Oh, right, Gradle's not actually the one dumping right now. Node's running into the OOM killer, currently.

rye:
Yeah, that's what it seems like, @hawkrives.  The previous memory setting should be fine—I think Circle might have swap or something, otherwise Java wouldn't hit an OOM exception I don't think.

rye:
Could this have something to do with https://github.com/nodejs/node/issues/18225 and/or https://github.com/libuv/libuv/pull/1655#issuecomment-358405254?  They're certainly recent issues.

rye:
Okay, for what it's worth, Node says [this](https://github.com/nodejs/node/blob/master/doc/api/process.md#exit-codes).
>* `>128` **Signal Exits** - If Node.js receives a fatal signal such as
  `SIGKILL` or `SIGHUP`, then its exit code will be `128` plus the
  value of the signal code.  This is a standard POSIX practice, since
  exit codes are defined to be 7-bit integers, and signal exits set
  the high-order bit, and then contain the value of the signal code.
  For example, signal `SIGABRT` has value `6`, so the expected exit
  code will be `128` + `6`, or `134`.

This means that our builds are getting SIGKILL'ed (`137`-`128`=`9`)

hawkrives:
SO says that the OOM killer sends SIGKILL: https://stackoverflow.com/a/7181152/2347774

rye:
So yes, I'm led to believe, again, that this might still be an OOM issue.  Someone (maybe me) should try debugging with SSH, logging in and running top if possible.  I've never done that before.

hawkrives:
you are welcome to! It prints out the ssh command for you to use as the first build step when you rerun with SSH.

rye:
Surely this can't be a bad idea to have too.  I'll merge `master` into this branch.

rye:
screen shot 2018-01-18 at 21 36 09

android/app/react.gradle
```diff
@@ -68,7 +68,7 @@ task bundleReleaseJsAndAssets(type: Exec) {
             entryFile, "--bundle-output", jsBundleFileRelease, "--assets-dest", resourcesDirRelease
     } else {
         commandLine "react-native", "bundle", "--platform", "android", "--dev", "false", "--entry-file",
-            entryFile, "--bundle-output", jsBundleFileRelease, "--assets-dest", resourcesDirRelease
+            entryFile, "--max-workers", "2", "--bundle-output", jsBundleFileRelease, "--assets-dest", resourcesDirRelease
```
hawkrives: I suppose there's no real need for us to run at full parallelism locally, either.

hawkrives (commented):
Approooooove (I started the pr so I can't)

rye (dismissed):
I approve, bwahahaha 😈

android/app/react.gradle
```diff
@@ -68,7 +68,7 @@ task bundleReleaseJsAndAssets(type: Exec) {
             entryFile, "--bundle-output", jsBundleFileRelease, "--assets-dest", resourcesDirRelease
     } else {
         commandLine "react-native", "bundle", "--platform", "android", "--dev", "false", "--entry-file",
-            entryFile, "--max-workers", "2", "--bundle-output", jsBundleFileRelease, "--assets-dest", resourcesDirRelease
+            entryFile, "--bundle-output", jsBundleFileRelease, "--assets-dest", resourcesDirRelease, (System.getenv("CI") == "true") ? "-j2" : ""
```
hawkrives: Nitpick: `System.getenv("CI") ?: "" != ""`? To handle systems that don't set CI to true and to make sure that gradle doesn't throw when not defined?

rye: Is that what the condition should be or the entire ternary?

hawkrives: The entire ternary. ?: is the "Elvis" operator; if the lhs doesn't exist / is falsy, it uses the right side. 

Bah. That's not really what I want. I want System.hasEnv, but I don't think that exists.

rye (dismissed):
I approve (again)

rye:
We decided we'll also bump up the maximum for Gradle again to `1536M`, since it's complaining about that.  Also, we should set the max number of workers in the Gradle config, perhaps.

hawkrives (commented):
🚢

rye (approved):
Very good.  We can always revisit parallelism issues if they pop up later.

PR #2171, opened 2018-01-19
Set Bugsnag env var from environment variables

hawkrives:
I also uppercased the Google Maps key; its been bugging me for a while that it's lowercase.


drewvolz (dismissed):

rye (approved):
Looks good to me!

PR #2172, opened 2018-01-19
Build and deploy on CircleCI

hawkrives:
This PR leaves only two things in Travis-land: greenkeeper updates and `gh-pages` builds.

There's no technical reason for that, but I'm really liking this gradual move we're doing.

This PR supersedes https://github.com/StoDevX/AAO-React-Native/pull/2148.
This PR requires https://github.com/StoDevX/AAO-React-Native/pull/2168 (because Circle fails signed builds otherwise) and https://github.com/StoDevX/AAO-React-Native/pull/2167 (because we can't deploy without it).

---

Notable things: 

- I add 3250 to the CircleCI build number, because we were using the Travis build number to number our TestFlight builds. This magic number brings us up to the Travis number. It can eventually be removed.
- I renamed a few variables
- I cleaned up the logic in `auto_beta`

Honestly, the only outstanding questions I have are if the cron thing actually works, and if I've configured the tagged builds right. I think they're good, though, tbh.


rye (approved):
I'm down to try this tonight.  We can always mess around with it later.

rye:
The nightly that ran last night had  `CIRCLE_WORKFLOW_ID = 2dfa2278-735f-4c74-a341-029ea15bb5ba`.  I don't think comparing the ID will work, unless we have documentation to suggest that it will remain that way forever.

Perhaps we should just set the NIGHTLY ENV variable?  That'd be the most surefire way.

hawkrives:
1. I'd expect the uuid to stay put… it doesn't make sense to keep changing. My current hypothesis is that it's unique per branch, so I'd like to see tomorrow's nightly's uuid. 
2. I can't set variables at the workflow level, afaict, so the only way for me to tell the iOS and Android builds to _build_ would be to duplicate them entirely. 
    Well… I suppose we could use yaml references, since we'd just need to add a new env key… okay, that could potentially work.

hawkrives:
More discussion in #2174

rye:
>1. I'd expect the uuid to stay put… it doesn't make sense to keep changing. My current hypothesis is that it's unique per branch, so I'd like to see tomorrow's nightly's uuid.

Yeah, this is fair, but also to _uniquely identify_ a given workflow, one would need to have an ever-changing UUID for each workflow.  See the `analyze-and-build` workflows on `master`, for instance, which never share a UUID.

PR #2173, opened 2018-01-19
Try setting the `cmd` key on the Android build Docker spec

rye:
This was a suggestion from a Feb '17 response on the CircleCI discussion board: https://discuss.circleci.com/t/builds-getting-killed-with-vague-message-received-signal-killed/10214/9

This would override the value if it's getting set weirdly.  Let's see what happens.

Closes #2170.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -152,7 +152,7 @@ jobs:
       - run: *run-danger
 
   android:
-    docker: [{image: 'circleci/android:api-27-node8-alpha'}]
+    docker: [{image: 'circleci/android:api-27-node8-alpha', cmd: '/bin/bash'}]
     environment:
       task: ANDROID
       FASTLANE_SKIP_UPDATE_CHECK: '1'
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -152,6 +152,9 @@ jobs:
       - run: *run-danger
 
   android:
+    # cmd: key is passed to solve a weird issue (see GitHub PR's 2170 and 2173).
+    # this overrides the thing and forces the container to run /bin/bash as the "command"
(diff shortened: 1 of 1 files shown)


hawkrives:
Didn't they set it on containerInfo in the forum though?

rye:
Yes, but I'm not sure why that would be any different, necessarily.  I think the `docker` key is analogous to the `containerInfo` key, and to me it would make sense to have the `cmd` key be there too.

Also, it seems as though the increase in memory/CPU usage happens right when ~Flow~ the packager runs.

Looks like we just got a successful build! :tada: :tada: :tada:

rye:
If you want, @hawkrives, I could merge this into ~the existing Gradle-related PR~ #2170 that was failing mysteriously, and we could see if this fixes that one. (In which case I think this should be our solution, perhaps.)

hawkrives:
I'll take this over that PR!

rye:
Okay! We can merge this once approved.  If it doesn't solve it, we can always revert back.

hawkrives (dismissed):

hawkrives:
>Actually, do you mind adding a comment to the yaml file?

Yeah, I can do that.  I also added `Closes` to the OP if that's okay with you.

hawkrives:
Documenting why we have that?

hawkrives:
I'm still approving with or without.

hawkrives (approved):

rye:
Thanks, @hawkrives.  We'll see, if this build passes twice in a row I'll be very happy.

hawkrives:
Yeah; I just looked at the queue log for master… there's a lot of red there. Hopefully that goes much greener now.

rye:
It seems we have another successful Android build 😁.

rye:
Here we go! Merging.

PR #2189, opened 2018-01-21
Credits icon reflects custom icon

drewvolz:
This PR swaps out the app icon displayed in the Credits view for whichever icon the user has selected to represent the app within the icon preferences.

Feedback is welcomed and appreciated.

~I think there is a fair amount of code duplication relating to icon functions that should be colocated and reused within a single helper for all icon related functions.~ maybe not

(Edit &lt;Hawken&gt;: nicer diff at https://github.com/StoDevX/AAO-React-Native/pull/2189/files?w=1)

### DIFF
--- a/source/views/settings/credits.js
+++ b/source/views/settings/credits.js
@@ -61,7 +61,6 @@ const formatPeopleList = arr => arr.map(w => w.replace(' ', ' ')).join(' • ')
 type Props = TopLevelViewPropsType
 
 type State = {
-	supported: boolean,
 	iconType: null | string,
 }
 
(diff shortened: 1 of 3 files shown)


source/views/settings/credits.js
```diff
@@ -56,22 +58,49 @@ const Contributors = glamorous(About)({
...
+	}
+
+	getIcon = async () => {
+		const name = await Icons.getIconName()
+		this.setState(() => ({iconType: name}))
+	}
+
+	render() {
+		const image = this.state.supported
+			? this.state.iconType === 'default' ? appIcons.oldMain : appIcons.windmill
```
drewvolz: This will break as we add and remove icons in the future.

hawkrives (approved):
😄

drewvolz:
Is there any need to refactor [icon.js](https://github.com/StoDevX/AAO-React-Native/blob/master/source/views/settings/icon.js) to use parts of this?

hawkrives:
mmm

Not as such, but it might be nice to move the GUI icon names into the source file too. Let's do that in another PR?

hawkrives:
@drewvolz and I agreed to look at the app icon settings again later.

PR #2202, opened 2018-01-22
Add KRLX tab to Streaming Media

hawkrives:
> … can't shake the feeling that I'm betraying Olaf somehow by adding Carleton stuff, but I don't see why the two stations would be in competition …

This PR adds a [KRLX](http://www.krlx.org) tab to the Streaming Media tab, next to KSTO.

It also makes the KSTO player stuff configurable at tab-definition-time.

You'll want to view this diff without whitespace (https://github.com/StoDevX/AAO-React-Native/pull/2202/files?w=1) because I re-indented a template string.

---

I just updated the schedule views to use #2203 to show the event descriptions instead of the locations. This works today for KRLX, and will allow KSTO to switch to it on their own sweet time.

---

I'm going to wait for two 👍s before merging.


drewvolz:
~The tab definitions for KRLX and KSTO are big. Any way we can slim those down into individual files?~ This all looks fine. We've communicated out of band that `index` acts as a configuration file.

drewvolz (dismissed):

drewvolz (approved):

rye (approved):

hawkrives:
alrighty!

PR #2203, opened 2018-01-23
Allow editing events after fetching but before rendering

hawkrives:
This PR adds a new optional `eventMapper` prop to `<GoogleCalendar/>`.

Usage: 

```jsx
<GoogleCalendarView
	calendarId="kstonarwhal@gmail.com"
	eventMapper={event => ({
		...event,
		title: event.title.replace(/^SUMO: /, ''),
		config: {
			...event.config,
			endTime: false,
		},
	})}
/>
```

That's actually an example from CARLS, but it's demonstrative.

That configuration will remove the prefix `SUMO: ` from each event, and will tell the row to hide the end time of the events.

You can also tell the rows to show the "description" of the event instead of the "location".

### DIFF
--- a/source/views/calendar/calendar-google.js
+++ b/source/views/calendar/calendar-google.js
@@ -13,7 +13,11 @@ import qs from 'querystring'
 import {GOOGLE_CALENDAR_API_KEY} from '../../lib/config'
 const TIMEZONE = 'America/Winnipeg'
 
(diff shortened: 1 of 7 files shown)


drewvolz (approved):
This is a nice solution to editing individual fields. Thanks!

PR #2215, opened 2018-01-23
Calendar description to shared/saved events

### LINKED ISSUE #2214: Shared events are missing description
Would it be helpful to add description to the events we’re sharing from the share sheet?

drewvolz:
Edited without testing or running

Closes #2214

### DIFF
--- a/source/views/calendar/calendar-util.js
+++ b/source/views/calendar/calendar-util.js
@@ -14,7 +14,7 @@ export function shareEvent(event: EventType): Promise<any> {
 	const location = event.location
 	const description = event.description
 	const message = `${title}\n\n${times}\n\n${location}\n\n${description}`
-	return Share.share({message})
+	return Share.share({message.trim()})
 		.then(result => console.log(result))
 		.catch(error => console.log(error.message))
 }
--- a/source/views/calendar/calendar-util.js
+++ b/source/views/calendar/calendar-util.js
@@ -55,7 +55,7 @@ async function saveEventToCalendar(event: EventType): Promise<boolean> {
 			eventToAdd.description = event.description
 		}
 
-		await RNCalendarEvents.saveEvent(event.title, ...eventToAdd)
+		await RNCalendarEvents.saveEvent(event.title, eventToAdd)
 
 		return true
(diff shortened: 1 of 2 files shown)


hawkrives:
Description isn't present in the calendar even that was passed in

drewvolz:
It’s listed within the types

hawkrives:
Description isn't supported in the saveEvent call

hawkrives:
(That's a different idea)

drewvolz:
It says it is supported in [the docs](https://github.com/wmcmahan/react-native-calendar-events)

Property | Value | Description
-- | -- | --
description (android only) | String | The description associated with the calendar event.
notes (ios only) | String | The notes associated with the calendar event.

hawkrives:
Android only

drewvolz:
Yeah, and notes is iOS only

hawkrives:
Then you'll have to pass the appropriate field by platform, I guess

source/views/calendar/calendar-util.js
```diff
@@ -12,7 +12,8 @@ export function shareEvent(event: EventType): Promise<any> {
 	const title = event.title
 	const times = getTimes(event)
 	const location = event.location
-	const message = `${title}\n\n${times}\n\n${location}`
+	const description = event.description
+	const message = `${title}\n\n${times}\n\n${location}\n\n${description}`
```
hawkrives: We should trim this string, so that we don't wind up copying 6 blank lines on an event with no location nor description.

hawkrives:
You'll have to trim `message` before you pass it to that function…

hawkrives:
Sorry, I forgot something about flow. 

Start out with an empty object and add each property to the empty object. Otherwise it won't let you add unknown properties to the "sealed" object

hawkrives:
Hmm. I'll think about flow more.

drewvolz:
All green :+1:

hawkrives (approved):
Sorry about that, Drew…

drewvolz:
No worries. We got that all cleared up. I should remember to check `flow-typed` when flow blows up.

PR #2219, opened 2018-01-23
Force nightlies to always be version 1.0.0

hawkrives:
@rye and I think that this might be the cause of our Play build issues, pushing two versions with the same "version name".

Also, this will allow easier filtering of nightly builds from the analytics, because we'll never ship v1 again.


drewvolz (dismissed):

rye (changes requested):
Can you move these lines down so that they don't change the value that is displayed in the app?

rye (approved):
Much better. Approved.

drewvolz (approved):

Issue #2216, opened 2018-01-23
Investigate why Play Store is eating ARM builds


rye:
From my chat with support, it seems like it might be due to out-of-order Android build numbers.  Maybe we should change the ordering?

hawkrives:
I'd actually be down to remove the Intel builds for a while. We could see if it solves the problem.

hawkrives:
From my research last year when I set this up, it wasn't clear how else to run two APKs for different archs. This seemed to be best practice then; I believe I took it directly from some set of android docs.

hawkrives:
We have agreed to disable x86 builds on nightlies for now, to see if the ARM builds stop disabling themselves.

hawkrives:
Here's an interesting thing: I'm playing around the Android Studio 3.1-alpha something, and it says

> Error:Conflicting configuration : 'armeabi-v7a,x86' in ndk abiFilters cannot be present when splits abi filters are set : x86,armeabi-v7a

hawkrives:
For reference, this is the doc I was read to get our current versioning on Android: https://developer.android.com/studio/build/configure-apk-splits.html#configure-APK-versions

rye:
Got the nightly just fine this morning.  We can try doing a beta sometime today to see if it works.

rye:
Beta just installed fine, so if it was an issue with ISSUE_DOWNGRADE getting thrown on installation, that's definitely fixed.  I'll still wait for 24 hours + after 2.5 gets released and rolled out without Play Store gobbling it up to call this one resolved.

rye:
Okay, so it seems as though we are pretty close to confident that this thing is no longer happening.  Closing this issue.

PR #2224, opened 2018-01-24
Raise minimum android sdk version

hawkrives:
- Update gradle
- Raise `compileSdkVersion` and `targetSdkVersion`
- Gradle complained about ndk splits, so I removed it
- Force all dependencies to use specified versions
- Use `implementation` instead of `compile`
- Gradle complained about spaces in the `rootProject.name`
- I think that's all

Does it build?

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -150,7 +150,7 @@ android {
     android {
         dexOptions {
             // Skip pre-dexing when running on Travis CI or when disabled via -Dpre-dex=false.
-            preDexLibraries = preDexEnabled && !travisBuild
+            preDexLibraries = preDexEnabled && !ciBuild
         }
     }
 }
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -149,6 +149,12 @@ android {
             preDexLibraries = preDexEnabled && !ciBuild
         }
     }
+    configurations.all {
+        resolutionStrategy.force 'com.android.support:appcompat-v7:26.1.0'
+        resolutionStrategy.force 'com.android.support:customtabs:26.1.0'
+        resolutionStrategy.force 'com.google.android.gms:play-services-base:11.8.0'
+        resolutionStrategy.force 'com.google.android.gms:play-services-maps:11.8.0'
+    }
 }
 
 // borrowed from https://gist.github.com/gabrielemariotti/6856974
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -98,7 +98,6 @@ def enableProguardInReleaseBuilds = false
 
 android {
     compileSdkVersion 23
-    buildToolsVersion "26.0.1"
 
     defaultConfig {
         applicationId "com.allaboutolaf"
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -191,21 +191,21 @@ if (propFile.canRead()) {
 
 dependencies {
     // please keep this list sorted
-    compile project(':bugsnag-react-native')
-    compile project(':react-native-calendar-events')
-    compile project(':react-native-custom-tabs')
-    compile project(':react-native-device-info')
-    compile project(':react-native-google-analytics-bridge')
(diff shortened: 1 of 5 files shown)


rye:
I get a bunch of these
```console
WARNING: Configuration 'compile' is obsolete and has been replaced with 'implementation'.
It will be removed at the end of 2018
WARNING: The specified Android SDK Build Tools version (23.0.1) is ignored, as it is below the minimum supported version (27.0.1) for Android Gradle Plugin 3.1.0-alpha09.
Android SDK Build Tools 27.0.1 will be used.
To suppress this warning, remove "buildToolsVersion '23.0.1'" from your build.gradle file, as each version of the Android Gradle Plugin now has a default version of the build tools.
```
messages for the various dependencies when they configure.

Also, I can't install over the existing version (on my phone).  Local builds should always have a higher version number than the alphas, which should have higher numbers than the betas, which should have higher numbers than production?

rye:
>Also, I can't install over the existing version. Local builds should always have a higher version number than the alphas, which should have higher numbers than the betas, which should have higher numbers than production?

This leads me to wonder if the real reason our builds are getting eaten (#2216) is perhaps that Play Store is trying to push a new version out but it fails, which would explain why it takes a while for the new release to come in.  Not sure.

hawkrives:
@rye About installing – we always have increasing version numbers on the Play store, though, because we use the Circle build number. Locally, your build uses 1 as the build number, which is why it's not higher…

hawkrives:
Following https://github.com/react-navigation/react-navigation/issues/1976, adding `android.enableAapt2=false` to gradle.properties fixes the production build by using the old asset processor.

It won't work forever, but it'll do for a while?

hawkrives:
yay, android is fixed.

rye:
I think there's probably a better way to generate our app version code to prevent the downgrading error message I get when I install on my phone, but I think this is good as long as it builds.  We could try a nightly?  Then a beta tomorrow, see if it eats it and whatnot.

hawkrives:
Sure. I'll let you merge at your leisure. I'm going to go make a PR to disable Intel builds.

android/settings.gradle
```diff
@@ -1,4 +1,4 @@
-rootProject.name = 'All About Olaf'
+rootProject.name = 'AllAboutOlaf'
```
rye, line 1: Does this change our project on Play Store or anything human-readable by end users?  Or is this just internal to the build and APK itself?

android/app/build.gradle
```diff
@@ -97,18 +97,14 @@ def enableSeparateBuildPerCPUArchitecture = true
 def enableProguardInReleaseBuilds = false
 
 android {
-    compileSdkVersion 23
-    buildToolsVersion "26.0.1"
+    compileSdkVersion 26
```
rye, line 100: Not sure why not 27?  It's out, and we have it in our CI build image.

android/settings.gradle
hawkrives: AFAIK, it's internal to the build? Gradle yelled at me that spaces would no longer be supported in Gradle 5, so I can't imagine that it'd be user-facing.

android/app/build.gradle
hawkrives: Ah, because this is what Android Studio fixed it to for me.

android/settings.gradle
rye: Ah, okay.

android/app/build.gradle
rye: Alright.  Well, we can ship like this for now.

rye (approved):
Very F-it- :shipit: at the moment, but really we can't tell if anything will go wrong without trying this out.

android/app/build.gradle
hawkrives: I'm about to push and update to 27

hawkrives: Nevermind, not going to 27 tonight:

> GC overhead limit exceeded

PR #2226, opened 2018-01-24
Only build Android ARMv7 APKs (not Intel)

hawkrives:
as the title says

Part of #2216


rye (approved):
I'm worried about the requirement by the Play Store to move towards having x64, but I'm sure we'll hear about that in some future RN release.

PR #2233, opened 2018-01-24
Force nightlies to be -pre

hawkrives:
Should fix iTC's issue with us uploading versions below the released version to testflight…

### DIFF
--- a/fastlane/lib/versions.rb
+++ b/fastlane/lib/versions.rb
@@ -46,15 +46,15 @@ def propagate_version(**args)
 
   UI.message "Propagating version: #{version}"
   UI.message 'into the Info.plist and build.gradle files'
+  
+  version = "#{version.split('-')[0]}-pre" if should_nightly?
+  UI.message "Actually putting #{version} into the binaries (because we're doing a nightly)"
 
   # encode build number into js-land – we've already fetched it, so we'll
   # never set the "+" into the binaries
   unless version.include? '+'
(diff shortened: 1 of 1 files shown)


fastlane/lib/versions.rb
```diff
@@ -46,15 +46,15 @@ def propagate_version(**args)
 
   UI.message "Propagating version: #{version}"
   UI.message 'into the Info.plist and build.gradle files'
+  
+  version = "#{version.split('-')[0]}-pre" if should_nightly?
```
rye, line 50: What happens if there is not a `-` in the version for some reason?

rye: If this case is already handled by `should_nightly?`, then that's okay.  I suppose this split won't be performed if `should_nightly?` is false.

rye: I guess my question is really: is it possible that `should_nightly?` will be truthy in a situation where we have a tag without a `-`?  I'm thinking if we release a new production build right around midnight.

hawkrives:
```
irb(main):001:0> '2.0.0'.split('-')[0]
=> "2.0.0"
```

We'll still append `-pre`, so it'll be fine.

rye: Oh true, I had a lapse in judgement about how split is handled.  Split, finding no delimiters, just returns the string it was called on 😄.

rye (approved):
:shipit:

PR #2258, opened 2018-01-26
Prevent Android APKs from including x86_64 splits

hawkrives:
This should address this morning's problem of devices trying to run 64-bit code that doesn't work.

Before | After
--- | ---
screen shot 2018-01-25 at 11 43 26 pm | screen shot 2018-01-25 at 11 42 11 pm

screen shot 2018-01-25 at 11 44 50 pm

According to https://android.jlelse.eu/controlling-apk-size-when-using-native-libraries-45c6c0e5b70a, 

> - `armeabi` – a very old ARM based architecture. Since Android 4.4, the CDD (compatibility definition) strictly requires ARMv7.
> - `armeabi-v7a` – the most common architecture currently available. This is a must-support ABI for all the applications.
> - `arm64-v8a` – the next generation 64 bit ARM architecture. All the flagship phones uses this architecture. Support of this ABI is optional as the devices can use armeabi-v7a version of the native library. 
> - `x86` – Intel-based devices; the ASUS ZenFone2 being the most well-known device of the lot along with many Lenovo phones. Any application has to support this.
> - `x86_64` – There are no android devices with 64-bit Intel processors available in the market.
> - `mips` and `mips64` – There are no android devices with MIPS architecture in market. (There was one device which was supposed to be launched in 2016 but I barely found any information.)

So we should be good with ARMv7 and x86.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -115,7 +115,7 @@ android {
         ]
 
         ndk {
-            abiFilters "armeabi-v7a", "arm64-v8a", "x86"
+            abiFilters "armeabi-v7a", "x86"
         }
     }
 
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -222,6 +222,7 @@ dependencies {
     // this is for react-native itself
     implementation fileTree(dir: "libs", include: ["*.jar"])
     implementation "com.android.support:appcompat-v7:26.1.0"
+    //noinspection GradleDynamicVersion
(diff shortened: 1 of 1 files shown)


rye (approved):

hawkrives:
(fwiw: iOS is re-running right now.)

rye:
Feel free to merge as soon as the builds finish (hopefully before the cron trigger, maybe)

rye:
Good work @hawkrives!

PR #2259, opened 2018-01-26
cache artifacts on Circle

hawkrives:
This should allow us to monitor the size of each native app bundle, as well as the JS payloads we send with the native binaries.

Maybe.

Hopefully Buildsize will kick in and do something here.

We'll see.

Maybe Buildsize doesn't comment until it's got something on master. Who knows. 

We'll find out!

---

Note: this iOS build path only exists for PRs from this repository! Any forked PRs will not be able to upload artifacts as the path won't exist. I don't know how Circle handles that.


hawkrives:
buildsize:

```
buildsize/total — Build size: 18.51 MB
```

rye (approved):
💼

PR #2263, opened 2018-01-27
Move the wifi endpoint

hawkrives:
I wanted to move the wifi endpoint to /apps/all-about-olaf/wifi, so that more things could potentially be added in the future.

This is the best time to do that.

Also this should let us change it in the future, maybe.

### DIFF
--- a/data/help/0-wifi.yaml
+++ b/data/help/0-wifi.yaml
@@ -14,4 +14,5 @@ body: |
 buttons:
   - title: Report an Issue
     action: custom
-    params: {}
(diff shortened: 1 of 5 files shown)


rye (approved):
Sounds good to me. Is this change already made server-side?

hawkrives:
@rye Yep. I moved it a few minutes ago.

rye:
@hawkrives sounds good. Go ahead and merge after passing builds, unless there's anything else for me to review.

PR #2267, opened 2018-01-27
Allow tabs to scroll on Android

hawkrives:
Ignore the purple titlebar…

This has been a problem since we switched to react-navigation however long ago.

screenshot_1517025792


rye (approved):
😂 okay

Issue #2045, opened 2018-01-05
Add analytics for various actions throughout the app


rye:
Specifically, we discussed analytics for #2032, but I'm thinking having little bits of reporting sprinkled about elsewhere might be another good idea.

I will come up with a list of stuff we don't seem to have that we should have at some point.

## Things

- [x] Login state (logged in, logged out, emphatically _not_ username)
- [x] ~~App resume vs. launch (#2211)~~
- [x] Disabled homescreen buttons
- [x] Presses of the "play stream" button (#2209)

hawkrives:
https://github.com/StoDevX/AAO-React-Native/blob/dedae1e6864d347f6944edbda7e1ce0f5542a0fa/source/flux/parts/homescreen.js#L99-L117

for #2032 in particular

hawkrives:
Login state

drewvolz:
Is there anything still needed to be done here?

PR #2269, opened 2018-01-27
Add more analytics event tracking

### LINKED ISSUE #2209: Add analytics for "play stream"
I think we can just add an "event". 

I do want to distinguish between "user pushes play" and "stream started itself", however.

hawkrives:
Part of https://github.com/StoDevX/AAO-React-Native/issues/2045

Specifically, this closes #2209; it tracks login state (#2045), and it tracks the disabled homescreen buttons (also #2045).

I also cleaned up a fair amount of the analytics stuff, centralizing the functions into `analytics.js`.

- Debounces the homescreen sorting calls, so we'll only be notified after 3 seconds have passed since it was last modified
- Edits the labels and event categories of the Dictionary, Orgs, and Hours views
- Adds tracking for calendar events and student job postings


source/flux/parts/homescreen.js
```diff
@@ -2,20 +2,33 @@
...
+	trackHomescreenDisabledItem,
+	trackHomescreenReenabledItem,
+} from '../../analytics'
 import * as storage from '../../lib/storage'
 import {type ReduxState} from '../index'
+import isEqual from 'lodash/isEqual'
+import debounce from 'lodash/debounce'
+
+// see https://css-tricks.com/debouncing-throttling-explained-examples/
+// we only invoke trackHomescreenOrder once every three seconds, to avoid
```
drewvolz, line 16: Why invoke it on a timer when we can invoke it upon unmounting of the component? Wouldn’t it make sense to grab their final config instead?

hawkrives: Because… because… 

My best answer is that I'm trying to isolate the side effects to the Redux action creators.

drewvolz: Ok 👍

drewvolz (approved):

PR #2272, opened 2018-01-27
Make the homescreen buttons ripple on Android and transparent on iOS

### LINKED ISSUE #2220: Enable ripple effect on Android homescreen buttons
currently, the touchable renders behind the gradient, so you can't see the ripple.

if we move the touchable in front, and set it to absoluteFill, it'll fit, cover the whole button, and show the ripple

hawkrives:
closes https://github.com/StoDevX/AAO-React-Native/issues/2220

On master, the homescreen buttons don't "ripple" like they ought to on Android.

The root cause of this is that the gradient is being rendered "above" the ripple effect.

This PR flips that on Android, so that the gradient is beneath the semitransparent ripple.

On iOS, the ordering is fine, because the Touchable needs to wrap whatever is going to go semitransparent, since the entire button fades.

So what I've done here is break the button into smaller chunks and reorder them based on platform.

---

There should be no differences in these screenshots.

| | Before | After
--- | --- | ---
iOS | screen shot 2018-01-27 at 12 35 34 pm | screen shot 2018-01-27 at 12 39 16 pm
Android | screenshot_1517078143 | screenshot_1517078364


source/views/streaming/radio/buttons.js
```diff
@@ -13,7 +13,7 @@ type ActionButtonProps = {
 }
 
 export const ActionButton = ({icon, text, onPress}: ActionButtonProps) => (
-	<Touchable hightlight={false} onPress={onPress} style={styles.button}>
```
rye, line 16: 🤦‍♂️

rye (approved):

PR #2275, opened 2018-01-29
Remove black home button text

drewvolz:
As per @hawkrives request:

> We should remove the black buttons before shipping.

We'll evaluate button colors in the near future.


hawkrives (approved):

rye:
>This change will decrease the build size from 54 MB to 18.51 MB, a decrease of 35.48 MB (192%)

Wow! (I know this is a misleading result lol)

rye (approved):
Looks good to me.  We should investigate turning down the backgrounds to increase contrast the other way.

PR #2277, opened 2018-01-30
Combine meta if possible in movies

drewvolz:
**This is a PR into the WIP `weekly-movies-hawken-3` and not `master`.**

This PR changes ~two~ one thing about the WIP movie view that @hawkrives and I are working on.

1. Directors and Writers will be listed once if they are the same people
~2. Dev mode will always return the current posted `next.json` file instead of the cached file~

... and then prettier made some changes.

Before | After
--|--
before | after

### DIFF
--- a/source/lib/cache.js
+++ b/source/lib/cache.js
@@ -269,15 +269,17 @@ export async function fetchWeeklyMovie(
 ): Promise<MaybeError<WeeklyMovie>> {
 	const cachedValue = await getWeeklyMovie()
 
-	if (!isOnline) {
-		if (cachedValue.isCached && cachedValue.value) {
+	if (process.env.NODE_ENV !== 'development') {
+		if (!isOnline) {
+			if (cachedValue.isCached && cachedValue.value) {
+				return {error: false, data: cachedValue.value}
+			}
+			return {error: true, message: 'You are currently offline'}
+		}
+
+		if (!cachedValue.isExpired && cachedValue.value) {
 			return {error: false, data: cachedValue.value}
 		}
-		return {error: true, message: 'You are currently offline'}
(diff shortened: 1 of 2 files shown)


drewvolz:
Shhh, lint.

source/lib/cache.js
```diff
@@ -269,15 +269,17 @@ export async function fetchWeeklyMovie(
 ): Promise<MaybeError<WeeklyMovie>> {
 	const cachedValue = await getWeeklyMovie()
 
-	if (!isOnline) {
-		if (cachedValue.isCached && cachedValue.value) {
-			return {error: false, data: cachedValue.value}
+	if (process.env.NODE_ENV !== 'development') {
```
hawkrives: Invert this if condition; if it's "=== dev", the mini diet can remove it entirely for the production builds.

source/views/streaming/movie/view.js
```diff
@@ -151,6 +151,32 @@ export class PlainWeeklyMovieView extends React.Component<Props> {
 		const movieTint = makeRgb(movie.posterColors.dominant)
 		const headerHeight = Math.max(Dimensions.get('window').height / 3, 200)
 
+		const WritersDirectors = () => {
```
hawkrives: Make this it's own component

hawkrives (dismissed):

source/lib/cache.js
drewvolz: Done

source/views/streaming/movie/view.js
drewvolz: Done

hawkrives (approved):

PR #2280, opened 2018-01-30
disable facilities and helpdesk tools

hawkrives:
At least until we actually ask them if they want in.

### DIFF
--- a/data/help/1-it-helpdesk.yaml
+++ b/data/help/1-it-helpdesk.yaml
@@ -1,7 +1,7 @@
 ---
 key: it-helpdesk
-enabled: true
-hidden: false
+enabled: false
+hidden: true
 
 title: Open a Helpdesk Ticket
 body: |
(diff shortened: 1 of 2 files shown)


drewvolz:
I’d hesitate on advertising it as much as we’re doing in the release notes if it ships as disabled.

...and hidden.

drewvolz:
Actually, what is the harm in enabling / making these available? They're no different than heading directly to the Facilities WO form itself, and no different than opening an email to the helpdesk yourself.

drewvolz:
My concerns are with the subject line in the helpdesk email template.

>  subject: I Haz A Problem
>  body:  [...] Please include your name, phone number, and next of kin.

All jokes aside, if we want to ship this, it should probably be presented with a blank subject and blank body. Filling in the "to" field is fine as it saves someone from having to do that. But the user will have to first delete the subject and body, unlike a blank email. The template does more harm than good here, IMHO.

hawkrives:
I would like to check with Heather on the wording of the HelpDesk card, and with Facilities to make sure they're OK with exposing the form to a wider audience.

hawkrives:
AFAIK, we aren't advertising either of these tools in the release notes.

drewvolz:
Just checked. You are correct.

drewvolz (approved):

PR #2287, opened 2018-01-31
Android Wifi Reporting: Use Lazy Geolocation

rye:
This PR removes the third argument from the `Geolocation.getCurrentPosition` call which we use for collecting positions when doing Wifi reporting, but only on Android.  On iOS, this third argument is preserved, but is turned from 5 seconds on timeout up to 15 seconds.  This gives the phone more time to think about where it is.

In the event of a timeout or something, we'd also rather not submit an empty object as "yeah sure this is good."  I have instead changed that to be a Promise rejection, [edit: on both platforms] which is then caught in the tool's handler and shipped off to BugSnag.  I think (?) that catching it like that will keep the client happy?  We won't throw an error state or something like that, because I don't think users will want to hear from our app about being sad. 🙃

### DIFF
--- a/source/views/help/wifi-tools.js
+++ b/source/views/help/wifi-tools.js
@@ -15,7 +15,7 @@ export const getIpAddress = (): Promise<?string> =>
 
 export const getPosition = (): Promise<Object> =>
 	new Promise((resolve, reject) => {
-		navigator.geolocation.getCurrentPosition(resolve, err => reject(err))
+		navigator.geolocation.getCurrentPosition(resolve, reject)
 	})
 
 export const collectData = async () => ({
--- a/source/views/help/wifi-tools.js
+++ b/source/views/help/wifi-tools.js
@@ -2,6 +2,7 @@
 
 import deviceInfo from 'react-native-device-info'
 import networkInfo from 'react-native-network-info'
+import {Platform} from 'react-native'
 import pkg from '../../../package.json'
 
 export const getIpAddress = (): Promise<?string> =>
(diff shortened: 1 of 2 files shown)


hawkrives (dismissed):
I will note that we're still sending the thing off to the server; now the position field will be "null" instead of "{}". I'm OK with that, but I'm not sure if it's what you were expecting. 

I'd prefer to resolve with a {latitude: 0, …} object, instead, to make it clear that we're sending silly data to the server, instead of me doing that server side.

rye:
>I'd prefer to resolve with a {latitude: 0, …} object, instead, to make it clear that we're sending silly data to the server, instead of me doing that server side.

Well, that's one way of solving it.  I'd rather we send nulls over and have them be outright rejected/discarded server-side.  Or just be stored in the DB with null as their location.

hawkrives:
Upon further reflection, I'm ok with sending nulls- more semantically correct for representing lack of location. 

But, I'd like to explicitly resolve to null here, so that we explicitly know that we're sending nothing. 

I think you can return a value from catch to do that.

rye:
From my toying around in a JS ~thingy~ dev console, it seems like that happens—the rejection gets sent through to the catch, but returning from that catch seems to allow for further chaining onto the Promise with the return value from that `.catch`.

source/views/help/wifi-tools.js
```diff
@@ -14,13 +15,17 @@ export const getIpAddress = (): Promise<?string> =>
...
-		})
+	new Promise((resolve, reject) => {
+		if (Platform.OS === 'ios') {
+			navigator.geolocation.getCurrentPosition(resolve, reject, {
+				...args,
+				enableHighAccuracy: true,
+				maximumAge: 1000 /* ms */,
+				timeout: 15000 /* ms */,
+			})
+		} else {
```
hawkrives: Nit: Would you flip these conditions, please? Android's needing to not have the third argument is very much the exception, I hope.

hawkrives (approved):

PR #2311, opened 2018-02-08
Limit gradle memory to 512MB

hawkrives:
I also disabled the gradle daemon… maybe we only want to do that on CI? (Can we?)


rye (approved):
Sounds good to me.  I'll be interested to see what kind of a speed difference this makes.

PR #2327, opened 2018-02-12
rework the initial data loading process in CourseSearch

hawkrives:
> > > THIS BRANCH TARGETS @hannesmcman's COURSE SEARCH PR

@hannesmcman I reworked the "initial data loading" logic quite a bit here.

This is predicated around it being acceptable to search stale data while new stuff is loading. If that's not acceptable, then I'll have to go back to the drawing board again.

## Flow:

1. Student taps "Search" tab
2. React's initial state is set to `dataLoading: true`
3. React calls `componentDidMount`, which calls the (new) `loadCourseDataIntoMemory`
4. `loadCourseDataIntoMemory` checks if any courses are cached, because there's no point to returning early if there's nothing to search, and then loads the cached data into redux state
5. `CourseSearchView` is given new props, with `courseDataState = 'ready'`; however, the view remains in the loading state, because the component's local state hasn't updated yet
6. The promise returned from `loadCourseDataIntoMemory` resolves, and `updateCourseData` is dispatched; next, `this.state.dataLoading` is set to `false`
7. The component renders in the "data ready" state, and allows searching of the potentially stale data
8. `updateCourseData` does it's thing to make sure all the data is ready
9. At some point, the new data is loaded, and the next search executed will query the new data

I'm a bit fuzzy on the exact order of 5/6, but them being flipped doesn't change the overall point of this rework.

I also reduced the number of actions for loading course data to just `LOAD_CACHED_COURSES` and `COURSES_LOADED`. I could combine them both, since both action creators dispatch them sequentially, but it felt a tiny bit nicer to have them separate.

Thoughts?


source/views/sis/course-search/search.js
```diff
@@ -44,13 +49,17 @@ class CourseSearchView extends React.PureComponent<Props, State> {
...
 	state = {
+		dataLoading: true,
 		searchResults: [],
 		searchActive: false,
 		searchPerformed: false,
 	}
 
 	componentDidMount() {
-		this.props.updateCourseData()
+		this.props.loadCourseDataIntoMemory().then(() => {
```
hannesmcman: I think we need to make this callback async and add an `await` before `this.props.updateCourseData()`. Otherwise when no there is no data in the cache, `dataLoading` will be set to false right away and there will be no loading screen..I tested this and changing it to 
```javascript
this.props.loadCourseDataIntoMemory().then(async () => {
	await this.props.updateCourseData()
	this.setState(() => ({dataLoading: false}))
})
``` 
seems to do the trick.

hannesmcman (changes requested):
I only have one change, but otherwise I'm down with this.

source/views/sis/course-search/search.js
hawkrives: Half of my idea here was to avoid the need to wait for the course data to be checked against the server. 

The goal was to pull data from the cache as quickly as possible, then kick off the data update checks in the background - hence the not waiting for updateCourseData. 

---

Rereading your comment, you're right - we do need something to handle the initial empty cache case. Let me sleep on that.

hawkrives:
@hannesmcman I've fixed the issue where no-terms-cached meant that the spinner hid itself instantly; now, if there are no terms cached, the spinner won't hide, but the data update will start, and the spinner will be hidden after the data update finishes.

drewvolz (approved):

hannesmcman (approved):
Looks good!

PR #2330, opened 2018-02-13
added noticeview to course-search if data isn't cached

hannesmcman:
This should resolve any problems for users who haven't used the course-search feature or have cleared all app data. Users who have no cached course data will encounter a `NoticeView` upon entering the course search tab, prompting them to download the course data. Users will only have to take this step if they don't have any data cached on their phone yet.

### DIFF
--- a/source/views/sis/course-search/search.js
+++ b/source/views/sis/course-search/search.js
@@ -19,12 +19,19 @@ import toPairs from 'lodash/toPairs'
 import {CourseSearchResultsList} from './list'
 import LoadingView from '../../components/loading'
 import {deptNum} from './lib/format-dept-num'
+import {NoticeView} from '../../components/notice'
+
+const PROMPT_TEXT =
+	'We need to download the courses from the server. This will take a few seconds.'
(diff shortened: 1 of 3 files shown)


source/views/sis/course-search/search.js
```diff
@@ -19,12 +19,19 @@ import toPairs from 'lodash/toPairs'
 import {CourseSearchResultsList} from './list'
 import LoadingView from '../../components/loading'
 import {deptNum} from './lib/format-dept-num'
+import {NoticeView} from '../../components/notice'
+
+const PROMPT_TEXT =
+	'We need to download the courses from the server. This will take a few seconds.'
+const NETWORK_WARNING =
+	"You'll need an internet connection to download the courses."
```
rye: We could get creative with these messages.

rye (approved):
Seems good to me.  We can refine later.

drewvolz (approved):
LGTM 👍

rye:
Merge-y merge-y merge-y! :shipit:

PR #2364, opened 2018-02-19
Add margin back to help cards

drewvolz:
This PR adds the horizontal margin back to the help tool cards.

Before | After
--|--
screen shot 2018-02-18 at 5 15 49 pm | screen shot 2018-02-18 at 5 10 17 pm

### DIFF
--- a/source/views/help/tool.js
+++ b/source/views/help/tool.js
@@ -86,7 +86,7 @@ export const styles = StyleSheet.create({
 	card: {
 		paddingHorizontal: 20,
 		paddingVertical: 15,
-		marginHorizontal: 0,
+		marginHorizontal: 10,
 		marginBottom: 10,
 	},
 })


hawkrives:
I prefer it with no margin, FWIW. 

But if Drew and two others disagree with me, we can merge it. 

---

If Drew tries to sway you with the "It doesn't look like a card!" piece, I'll just say my rebuttal: our `<Card/>` doesn't currently look like a card on either iOS' HIG _or_ Material Design. This isn't a list of cards; it's a list of help tools. (Also, I offered to copy the styles that this uses out of `<Card/>`, so that it wouldn't be named a card, but that was deemed unsatisfactory 😜)

drewvolz:
I say we treat them as cards and try to tweak them to look more like cards.

drewvolz:
I'd like to design a much more native card. But we haven't done that yet. So if anyone else does not like this change, I'll close it.

hannesmcman:
I think it looks better... 👀 (as in with the margin)

hannesmcman (approved):

PR #2378, opened 2018-02-22
[WIP] Add filters to course search

### LINKED ISSUE #2342: Add filters to Course Search View
This would involve adding a filter feature similar to that in the menus section of the app that would allow users to filter by:

- Term
- GEs
- Department
- Lab
- Open
- Time blocks?

Any more?

hannesmcman:
Resolves #2342.

Desired filters:
- [x] Term
- [x] GEs
- [x] Department
- [x] Lab
- [x] Open
- [ ] Time blocks?


hannesmcman:
Here are some screenshots...I reused the filter-view component and most of the filter logic from the menus component..idk how I feel about the filtered terms showing up in the filter toolbar like that, especially when you filter by more than two, but that was the first idea that came to mind when I was thinking of what would be appropriate there..very open to dissent/other ideas!

• | •
--|--
screen shot 2018-02-22 at 1 40 28 pm | screen shot 2018-02-22 at 1 40 42 pm
screen shot 2018-02-22 at 1 41 05 pm |

hannesmcman:
I can also move the rest of the course search redux stuff into the course-search part in this PR... @hawkrives I know you mentioned wanting that to happen.

hawkrives:
Oh, I _do_ like showing a summary of the selected filters in the filter space. That's a great idea!

I'm sure we can come up with a way to represent them compactly – `18/19:Spr/Int, 19/20:All`, maybe?

source/lib/course-search/index.js
```diff
@@ -4,3 +4,4 @@ export {loadCachedCourses} from './load-cached-courses'
 export {updateStoredCourses, areAnyTermsCached} from './update-course-storage'
 export {CourseType, TermType} from './types'
 export {parseTerm} from './parse-term'
+export {loadGEs} from './load-ges'
```
hawkrives: Where is this called?

source/views/components/filter/types.js
```diff
@@ -6,7 +6,8 @@ export type ToggleSpecType = {
 }
 
 export type ListItemSpecType = {|
-	title: string,
+	title: string | number,
+	label?: string,
```
hawkrives, line 10: I'm not entirely clear on the difference between these two – can you summarize for me, please?

hawkrives (commented):
Wow! Nice work.

source/views/components/filter/types.js
drewvolz: Looks like it comes down to:
```js
title: term.term,
label: parseTerm(term.term.toString()),
```
* `term` would be a string or a number
* `label` is data massaged out of `parseTerm` which lives in [lib/course-search/parse-term.js](https://github.com/StoDevX/AAO-React-Native/blob/master/source/lib/course-search/parse-term.js) where it returns the human readable term name (Abroad current / next year, Fall current / next year, etc)

source/lib/course-search/index.js
drewvolz: Looks to have been introduced in https://github.com/StoDevX/AAO-React-Native/pull/2378/commits/d2fdcdbf6c3b4f2c15b449b8cad5f7df607f99e1#diff-df4b353eb39c9a68e6b0e3127a055bb4R69
and removed in https://github.com/StoDevX/AAO-React-Native/commit/da79336f7c854353c356e911d0e4650737a8c629#diff-df4b353eb39c9a68e6b0e3127a055bb4L69

hannesmcman: Yeah that was left over from my first attempt when I wasn't reusing the filter component..I kept it so that I could ask this question: For access to the acceptable list of GEs, departments, etc. should we load that data from the server and store it just like the course data, or should we try to somehow extract them from the course data itself? @hawkrives

source/views/components/filter/types.js
hannesmcman: Yep! I changed the spec type so that the applyFilters() function would work when the value of the key item specified isn't a string. The label allows the filter options to be displayed as "Fall 2016/17" instead of 20171, for example.

hannesmcman:
@hawkrives re:summary of selected filters 
Right now I have it so that this summary is cut off at one line: yay or nay?
screen shot 2018-02-22 at 4 17 25 pm

hannesmcman:
@hawkrives Ok so I got a filter up and running for "Open Courses," and realized that there are a whole bunch of courses listed as "Open" from past semesters...is this something that you want to fix for the course-data on the server itself or should that logic go in AAO?

source/lib/course-search/index.js
hawkrives: I'd recommend using https://github.com/StoDevX/course-data/blob/gh-pages/data-lists/valid_gereqs.json for a list of GE requirements.

It's generated automatically from the course data every night.

Of, you can do something to determine the gereqs from the downloaded course data; that's what I do in Gobbldygook, too.

hawkrives:
@hannesmcman Those courses are technically still "open", as far as SIS / the registrar are concerned.

I don't know that we want to change that?

I'd still question how useful an open/closed filter would be. I never ran into a situation where it was a useful distinction to me as a student, AFAIR?

drewvolz:
I say we shouldn't return past semester courses that are listed as open on the app's end.

hawkrives:
I say we shouldn't care about "open/closed" at all :p

drewvolz:
...and I see plenty of use for an open/closed filter for those trying to find open sections.

hawkrives:
I don't even know when a section gets "closed". I don't remember seeing any that were closed in anything that I wanted to take (cough CS /cough), even when they were full.

drewvolz:
>  I don't remember seeing any that were closed in anything that I wanted to take (cough CS /cough), even when they were full.

I was kicked out of a section that had closed during registration and had to find another section that was open

source/lib/course-search/index.js
drewvolz: This PR fetches https://stodevx.github.io/course-data/data-lists/valid_gereqs.json

hannesmcman: That's what it uses currently.

package.json
```diff
@@ -92,6 +92,7 @@
     "@hawkrives/react-native-alphabetlistview": "1.0.0",
     "@hawkrives/react-native-alternate-icons": "0.4.7",
     "@hawkrives/react-native-sortable-list": "1.0.1",
+    "bluebird": "3.5.1",
```
hawkrives: Can you use https://www.npmjs.com/package/p-props instead, please? It's a more focused dependency.

hawkrives (dismissed):

source/views/sis/course-search/lib/format-terms.js
```diff
@@ -0,0 +1,54 @@
...
+		case '3':
+			return {year: `${currentYearAbbrev}/${nextYear}`, semester: 'Spr'}
+		case '4':
+			return {year: `${currentYearAbbrev}/${nextYear}`, semester: 'Sum1'}
+		case '5':
+			return {year: `${currentYearAbbrev}/${nextYear}`, semester: 'Sum2'}
+		case '9':
+			return {year: `${currentYearAbbrev}/${nextYear}`, semester: 'Non-Sto'}
+		default:
+			return {year: `${currentYearAbbrev}/${nextYear}`, semester: 'Unk'}
```
drewvolz, line 52: N/A might work well here. @StoDevX/all-about-olaf anyone have an opinion?

hawkrives:
I did a fair amount of reading on Flow's "disjoint union types" tonight, trying to figure out what was going on, but I think it's actually this bug, instead: [Refine array types using `filter`](https://github.com/facebook/flow/issues/1414).

So, what I did – and I only did this because this is an exceptional case; we want to avoid `any` as much as possible, because flow can't check it – is to define a set of functions for filtering an array for just a specific type of spec. 

```js
export function filterListSpecs(specs: Array<FilterType>): Array<ListType> {
	const retval = specs.filter(f => f.type === 'list')
	return ((retval: any): Array<ListType>)
}
```

I filter the array, with hopefully bug-free code, then I apply a typecast to the result to tell Flow exactly what we're returning.

drewvolz (approved):

PR #2394, opened 2018-03-01
Improvements to open jobs detail cells

drewvolz:
This PR converts the contents of `<SelectableCell />` to a `<TextInput />`.

This PR picks up something we've been testing over on the carls app. I've swapped out our home-brewed link detecting on the Open Jobs detail views in favor of `<TextInput />` which, for iOS, has the ability to parse links, allow native selection of text, and applies data detectors inline.

### DIFF
--- a/source/views/sis/student-work/selectable.js
+++ b/source/views/sis/student-work/selectable.js
@@ -11,7 +11,6 @@ const styles = StyleSheet.create({
 		color: c.black,
 		backgroundColor: c.white,
 		paddingHorizontal: 15,
-		paddingVertical: 15,
 		paddingTop: 10,
 		paddingBottom: 10,
 		...Platform.select({
--- a/source/views/sis/student-work/selectable.js
+++ b/source/views/sis/student-work/selectable.js
@@ -1,20 +1,36 @@
 // @flow
 import * as React from 'react'
-import {Text, StyleSheet} from 'react-native'
-import {Cell} from 'react-native-tableview-simple'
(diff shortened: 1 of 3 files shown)


source/views/sis/student-work/selectable.js
```diff
@@ -1,20 +1,36 @@
...
 const styles = StyleSheet.create({
-	cell: {
-		paddingVertical: 10,
+	text: {
+		color: c.black,
+		backgroundColor: c.white,
+		paddingHorizontal: 15,
+		paddingVertical: 15,
+		paddingTop: 10,
+		paddingBottom: 10,
```
hawkrives: You're duplicating styles here.

hawkrives (dismissed):
You've modified both platforms here; unless you've figured out how to support https://developer.android.com/reference/android/widget/TextView.html#setAutoLinkMask(int) from Android, you'll need to make Android use the current components instead.

hawkrives:
I'm going to need to see either an Android screenshot or you're going to need to make `selectable-cell.android.js` so Android doesn't use TextInput here, because neither of us - to my knowledge - has ever looked at a readonly TextInput on Android.

drewvolz:
I developed on my android phone last night. This does not remove any functionality from android.

drewvolz:
Android also does not use the selectable cell file at all, FWIW.

hawkrives:
> Android also does not use the selectable cell file at all, FWIW.

In that case…

hawkrives (approved):

PR #2397, opened 2018-03-01
Trim news excerpts

drewvolz:
I'm noticing spaces at the beginning of some stories. Let's clean that up!

3 stories on the news feed with spaces at the start

### DIFF
--- a/source/views/news/news-row.js
+++ b/source/views/news/news-row.js
@@ -33,7 +33,7 @@ export class NewsRow extends React.PureComponent<Props> {
 					<Image source={thumb} style={styles.image} />
 					<Column flex={1}>
 						<Title lines={2}>{story.title}</Title>
-						<Detail lines={3}>{story.excerpt}</Detail>
+						<Detail lines={3}>{story.excerpt.trim()}</Detail>
 					</Column>
 				</Row>
 			</ListRow>


hawkrives:
Don't we generate the excerpt somewhere? If we use something the server provides, this is fine, but if we're making it let's `trim()` it there.

hawkrives:
tip: deleting the buildsize comment works better if we wait until the builds have all completed

drewvolz:
@hawkrives you're right, we do generate the excerpt inside [fetch-feed]( https://github.com/StoDevX/AAO-React-Native/blob/master/source/views/news/fetch-feed.js).

```js
excerpt: entities.decode(fastGetTrimmedText(item.excerpt.rendered)),
```

We're running [fastGetTrimmedText](https://github.com/StoDevX/AAO-React-Native/blob/master/source/lib/html.js#L41) on it, though?

```js
export function removeHtmlWithRegex(str: string): string {
	return str.replace(/<[^>]*>/g, ' ')
}

export function fastGetTrimmedText(str: string): string {
	return removeHtmlWithRegex(str)
		.replace(/\s+/g, ' ')
		.trim()
}
```

hawkrives:
Maybe those are whitespace-looking chars that aren't technically whitespace or something?

You'll probably need to check what it is on the simulator before we can get a good lead here :|

drewvolz:
Checked the feed. Those three excerpts contain `&nbsp; ` at the beginning.

hannesmcman (approved):

hawkrives:
Do we restrict the length of the excerpt (to, say, 150/300 characters)?

If so, we should run the html entities thing on that shortened excerpt before we display if. 

If we don't trim it down, we should do that, _then_ do the entities and finally remove whitespace.

drewvolz:
apathy.

Issue #2404, opened 2018-03-05
Enable IT help card


hawkrives:
We should fix the wording, but I just got Heather's approval to enable the widget.

drewvolz:
Closed by #2405

PR #2405, opened 2018-03-05
Enable IT Helpdesk tools

drewvolz:
Part of #2404

This enables the tool but does not change the wording as #2404 requests. I am unsure what Heather would like to change the wording to (@hawkrives)

### DIFF
--- a/data/help/1-it-helpdesk.yaml
+++ b/data/help/1-it-helpdesk.yaml
@@ -1,7 +1,7 @@
 ---
 key: it-helpdesk
-enabled: false
-hidden: true
+enabled: true
+hidden: false
 
 title: Open a Helpdesk Ticket
 body: |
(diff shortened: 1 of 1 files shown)


hawkrives:
Heather didn't have any opinions on the wording.

I like whimsy in my projects, which is why I wrote what I originally did in here.

If people start submitting tickets without editing the information, we can always change it remotely.

So… I care not. Ideas?

drewvolz:
I would encourage us to make it blank so that the email client pokes the user into setting a subject of their own.

hawkrives:
I don't think that will encourage people to set a subject, but whatevs.

hawkrives:
I would _strongly prefer_ to have a subject of some sort.

drewvolz:
We can change it if it is a problem in the future. Okay?

hawkrives:
😜

hannesmcman (approved):

PR #2417, opened 2018-03-10
Avoid ever logging redux course cache

### LINKED ISSUE #2416: Disable logging of course data on development
We would prefer not to log the entire course database twice for every action.

This should speed up the app in development like #2351 did for production.

@hawkrives made a few out of band comments:
> Use `stateTransformer` to remove the courses state from the log output. That should do it.
> 
> I don't want to modify the logged actions, since they happen much less often than the state, and if we don't modify them then we won't be scratching our heads while debugging that one. So just do the `stateTransformer` bit.

hawkrives:
Closes #2416 

This prevents `redux-logger` from serializing and logging the entire cached course data twice per redux action (before and after).

Instead, it prints `<omitted>`, so we know that it's been purposefully omitted.

This only modifies the logged state; it does not affect the actual in-memory state.

I also split the `sis` reducer into `balances` and `courses` reducers, to make it clearer where the associated actions / state should go, and I moved all of the course data into the `courses` reducer.

image


hawkrives:
~~Nevermind, the course data isn't stored there.~~

This comment is no longer accurate.

hawkrives:
I forgot to comment that I fixed this up, so disregard https://github.com/StoDevX/AAO-React-Native/pull/2417#issuecomment-372005197

drewvolz (approved):
Thank you!

PR #2419, opened 2018-03-10
We're unofficial (so let's tell the world so)

hawkrives:
Partially tongue-in-cheek, but I do think that we need something like this.

Wording of messages, layout, and behavior (currently there is none) is all open to discussion.


drewvolz:
As always, I’ll be the Debbie Downer. I’d like to request the wording to change.

hawkrives:
That's perfectly fine, but I'm going to come back at'cha with a request for suggestions ;-)

drewvolz:
- An unofficial St. Olaf app
- For students, by students
- For Oles, by Oles
- Not supported by St. Olaf

hannesmcman (dismissed):

source/views/home/notice.js
```diff
@@ -0,0 +1,31 @@
...
+import sample from 'lodash/sample'
+
+const messages = [
+	'☃️ An Unofficial App Project ☃️',
+	'For students, by students',
+	'By students, for students',
+	'SGA-owned and misunderstood',
+	'I AM AN UNOFFICIAL APP',
+	'we are (not) olaf',
+	'☃️',
```
drewvolz: Do we really want this

source/views/home/notice.js
```diff
@@ -0,0 +1,31 @@
...
+import * as c from '../components/colors'
+import sample from 'lodash/sample'
+
+const messages = [
+	'☃️ An Unofficial App Project ☃️',
+	'For students, by students',
+	'By students, for students',
+	'SGA-owned and misunderstood',
+	'I AM AN UNOFFICIAL APP',
+	'we are (not) olaf',
```
drewvolz: This one doesn't tell me much

source/views/home/notice.js
```diff
@@ -0,0 +1,31 @@
...
+import * as React from 'react'
+import * as glamorous from 'glamorous-native'
+import * as c from '../components/colors'
+import sample from 'lodash/sample'
+
+const messages = [
+	'☃️ An Unofficial App Project ☃️',
+	'For students, by students',
+	'By students, for students',
+	'SGA-owned and misunderstood',
```
drewvolz: Not necessarily true

source/views/home/notice.js
```diff
@@ -0,0 +1,31 @@
...
+	'I AM AN UNOFFICIAL APP',
```
drewvolz: Not a fan of the use of "I" here

drewvolz (changes requested):

drewvolz:
What we say on the front of the app is very important to me. This is what is forward facing and what the user will be confronted with _every single time they use the app_, and what new users will first read. I'd rather be generic than confusing. I'd like the messages to be straightforward, which is why I suggested what I said in this message https://github.com/StoDevX/AAO-React-Native/pull/2419#issuecomment-372043402.

drewvolz (approved):
👍 :shipit:

PR #2423, opened 2018-03-12
Course Search: Change departments filter to "OR"

### LINKED ISSUE #2422: Change Departments filter in Course Search to "OR"
I realized that it probably makes more sense for the departments filter to be an "OR" filter instead of "AND", so that you can select `MATH` and `CSCI` and get all the courses in either of those departments that match your query, rather than all the courses in both of those of those departments that match the query.

hannesmcman:
Closes #2422 .

I had to modify the logic in `applyOrListFilter` to work with filters like departments, where a course can have multiple departments.


source/views/components/filter/apply-filters.js
```diff
@@ -53,10 +54,16 @@ export function applyListFilter(filter: ListType, item: any): boolean {
...
-	itemValue: string,
+	itemValue: string | string[],
 ): boolean {
 	// An item passes, if its value is in the filter's selected items array
-	return filterValue.map(f => f.title).includes(itemValue)
+	if (!Array.isArray(itemValue)) {
+		return filterValue.map(f => f.title).includes(itemValue)
+	}
+
+	const valueToCheckAgainst = filterValue.map(f => f.title.toString())
```
hawkrives, line 64: I don't think we need toString here, do we?

source/views/sis/course-search/search.js
```diff
@@ -144,6 +144,7 @@ class CourseSearchView extends React.PureComponent<Props, State> {
 		const query = text.toLowerCase()
 		this.setState(() => ({query: query}))
 
+		console.log(this.props.allCourses[0])
```
hawkrives: Remove debugging :)

hawkrives (changes requested):

source/views/components/filter/apply-filters.js
hannesmcman: Flow gets mad without it, since title could be a number

hannesmcman:
Is that jest test failure a problem? Or is the test wrong?

hawkrives:
DST problems

hawkrives:
Just merge master in, push, and the build should go green.

hawkrives (approved):

PR #2424, opened 2018-03-12
Disable schedule-in-DST checking

drewvolz:
Disabling as a result of an out-of-band discussion with @hawkrives. This will at least get #2423 building.


hawkrives (approved):
Eventually either this code will be rewritten or we'll fix it.

Issue #2407, opened 2018-03-07
Ideas for bottom 3/4 of course search view


hannesmcman:
I have a few ideas for what to put under the search bar when there isn't a search in progress, but I can't decide which one is best/most useful...help me!

- [x] "Recent Searches" (similar to Apple Music type thing, which was the inspiration for the style of the search bar in the first place) (see #2427)
- [ ] "Saved Courses" section. This could be a list of saved courses or a way to navigate to a list of saved courses.
- [x] "Quick Search." This idea has been brought up before, and would basically entail the user selecting a few filters and pressing a button that would bring them to the search view and display all results matching those filters. (see #2459)
- Any combination of these
- More ideas?

drewvolz:
These sound good. I'd advocate for recent searches, or a top-level way of searching by department.

drewvolz:
I think we can close this?

hannesmcman:
Closing this until we get around to creating some sort of scheduling assistant.

hawkrives:
I think that such a thing belongs in a new issue, anyway. 👍

PR #2427, opened 2018-03-14
Add recent searches feature to course search

hannesmcman:
Stores the three most recent unique queries by the user, and allows them to search them easily again by tapping on them. 
screen shot 2018-03-13 at 8 30 45 pm

Currently relies on a fork of the react-native-searchbar repo, so hopefully this PR gets merged soon: https://github.com/localz/react-native-searchbar/pull/57. 
Addresses part of #2407


source/flux/init.js
```diff
@@ -56,6 +57,7 @@ export async function init(store: {dispatch: any => any}) {
 		store.dispatch(loadFeedbackStatus()),
 		store.dispatch(loadAcknowledgement()),
 		store.dispatch(loadFavoriteBuildings()),
+		store.dispatch(loadRecentSearches()),
```
hawkrives, line 60: I can't decide if I want you to move this to componentDidMount on the course search itself. 

Historically, I've tried to limit this list to things used across the app (I realize that some of these have already violated that principle) but this is a small piece of data to load. 

I'm going to think on it while I walk home and think. Pig me in a few hours if I don't respond.

source/flux/parts/courses.js
```diff
@@ -68,23 +69,70 @@ export function updateCourseData(): UpdateCourseDataActionType {
...
+	payload: string[],
+|}
+export function updateRecentSearches(
+	query: string,
+): ThunkAction<UpdateRecentSearchesAction> {
+	return (dispatch, getState) => {
+		const state = getState()
+
+		const oldRecentSearches = state.courses ? state.courses.recentSearches : []
+		if (oldRecentSearches.includes(query)) {
```
hawkrives: Should we lowercase these while comparing them, to avoid storing ones that only vary in case?

source/flux/parts/courses.js
```diff
@@ -68,23 +69,70 @@ export function updateCourseData(): UpdateCourseDataActionType {
...
+			return
+		}
+		let newRecentSearches = oldRecentSearches
+		if (newRecentSearches.length === 0) {
+			newRecentSearches = [query]
+		} else {
+			newRecentSearches.unshift(query)
+		}
+		if (newRecentSearches.length > 3) {
+			newRecentSearches.pop()
```
hawkrives: newRecentSearches.slice(0, 3) – to make it clearer that it'll always become three long.

source/flux/parts/courses.js
```diff
@@ -68,23 +69,70 @@ export function updateCourseData(): UpdateCourseDataActionType {
...
+	query: string,
+): ThunkAction<UpdateRecentSearchesAction> {
+	return (dispatch, getState) => {
+		const state = getState()
+
+		const oldRecentSearches = state.courses ? state.courses.recentSearches : []
+		if (oldRecentSearches.includes(query)) {
+			return
+		}
+		let newRecentSearches = oldRecentSearches
```
hawkrives: Why make a new variable here?

source/flux/parts/courses.js
```diff
@@ -68,23 +69,70 @@ export function updateCourseData(): UpdateCourseDataActionType {
...
+		if (newRecentSearches.length === 0) {
+			newRecentSearches = [query]
+		} else {
+			newRecentSearches.unshift(query)
```
hawkrives: newRecentSearches = [query, ...newRecentSearches]

I tend to avoid mutation methods except in performance sensitive code? You can push back if you wish.

source/views/sis/components/recent-search/list.js
```diff
@@ -0,0 +1,55 @@
...
+	queries: string[],
+}
+
+export class RecentSearchList extends React.PureComponent<Props> {
+	renderSeparator = () => <ListSeparator spacing={{left: 20}} />
+
+	renderItem = ({item}: {item: string}) => (
+		<ListRow
+			arrowPosition="none"
+			onPress={() => {
```
hawkrives: Inline this without the curly braces

hawkrives: Inline

source/views/sis/components/recent-search/list.js
```diff
@@ -0,0 +1,55 @@
...
+		this.props.onQueryPress(query)
+	}
+
+	keyExtractor = (item: string) => item
+
+	render() {
+		const {queries} = this.props
+		return (
+			<FlatList
+				ItemSeparatorComponent={this.renderSeparator}
```
hawkrives, line 32: Do we want these to be separated? Doesn't the App Store just use whitespace?

source/views/sis/course-search/search.js
```diff
@@ -173,7 +179,9 @@ class CourseSearchView extends React.PureComponent<Props, State> {
 			groupedCourses,
 			course => course.title,
 		).reverse()
-
+		if (text.toString().length !== 0) {
```
hawkrives: Remove the toString

source/views/sis/course-search/search.js
```diff
@@ -259,20 +275,30 @@ class CourseSearchView extends React.PureComponent<Props, State> {
...
+							<Separator />
 							{searchActive ? (
-								<CourseSearchResultsList
-									filters={this.props.filters}
-									navigation={this.props.navigation}
-									onFiltersChange={this.props.onFiltersChange}
-									searchPerformed={searchPerformed}
-									terms={searchResults}
-								/>
+								<View>
```
hawkrives: Remove wrapper view

hawkrives (changes requested):
Changes! And thanks!

source/flux/parts/courses.js
hannesmcman: I thought about that...just not long enough to actually do something about it! But yeah it's probably a good call.

source/flux/parts/courses.js
hannesmcman: If I change it to `newRecentSearches = [query, ...newRecentSearches]` it seems to break the searching...as in when I press return after typing in a query, it doesn't refresh the search results unless I do it again. Works fine as is..maybe the ... is too slow?

hawkrives: Meh. Ok.

source/views/sis/components/recent-search/list.js
hawkrives: then it's ok!

hannesmcman: sorry idk how to make that smaller

hawkrives: It's fine

hawkrives: Add a right: 20 to where you have left: 20 to add margin on the right too

source/flux/parts/courses.js
hannesmcman: Actually it's working now..? May have just been slow memory on my computer...I keep getting notifications that my memory is maxed out..

hawkrives: I would have a hard time believing that there was any performance difference between those two…

source/flux/init.js
hannesmcman: What do you think?

hannesmcman:
License Compliance?

source/flux/parts/courses.js
```diff
@@ -92,23 +92,18 @@ export function updateRecentSearches(
...
-		if (newRecentSearches.length === 0) {
-			newRecentSearches = [query]
-		} else {
-			newRecentSearches.unshift(query)
-		}
-		if (newRecentSearches.length > 3) {
-			newRecentSearches.pop()
-		}
+		recentSearches = [query, ...recentSearches]
+		recentSearches = recentSearches.slice(0, 3)
```
hawkrives: Join these together – `recentSearches = [query, ...recentSearches].slice(0, 3)`?

hawkrives (changes requested):
One more right now!

source/flux/init.js
hawkrives: This is fine for now

hawkrives:
dammit license thingy

edit: they now pass

hawkrives (approved):

PR #2429, opened 2018-03-14
Remove react-native-maps (and replace with Mapbox)

drewvolz:
@hawkrives did work to replace `react-native-maps` in favor of `react-native-map-box-gl`.

The bonus is -- #2306 compiles and runs once this PR is integrated. So we get to perform the RN upgrade.

If we keep using react-native-maps, otoh, https://github.com/react-community/react-native-maps/pull/2053 won't let us actually run the app.


hawkrives:
This will eventually be used by my fancy campus map implementation that's currently incubating in CARLS

drewvolz:
> that's currently incubating

marinating 🍲

rye (approved):
🐞

hawkrives:
@rye 👞

PR #2433, opened 2018-03-14
Course Search: debounce search function, fix spacing of empty recent search list

### LINKED ISSUE #2432: Course Search: results are sometimes not displayed on search
I think we just need to debounce `performSearch` like in the student org view

hannesmcman:
This resolves #2432.

### DIFF
--- a/source/views/sis/components/recent-search/list.js
+++ b/source/views/sis/components/recent-search/list.js
@@ -30,7 +30,13 @@ export class RecentSearchList extends React.PureComponent<Props> {
 		return (
 			<FlatList
 				ItemSeparatorComponent={this.renderSeparator}
-				ListEmptyComponent={<NoticeView text="No recent searches" />}
+				ListEmptyComponent={
+					<NoticeView
+						header="No recent searches"
+						style={styles.notice}
+						text="Your recent searches will appear here."
+					/>
+				}
 				data={queries}
 				keyExtractor={this.keyExtractor}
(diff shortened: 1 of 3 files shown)


hannesmcman:
I'm going to fix the spacing of the empty recent searches list too

hawkrives (dismissed):
sure

hannesmcman:
New empty recent searches look:
screen shot 2018-03-14 at 11 16 04 am

hawkrives:
What do you think about just not showing anything there if there aren't any recent searches?

hannesmcman:
Idk I kinda like it being there for the aesthetic...especially when thinking about what's going to go underneath it. I could probably be convinced otherwise though!

hawkrives:
What if we added another section beneath it, of "suggested searches" or something? Then if there were no recent ones, it'd just show the suggestions, and if there were recent ones, they'd be above the suggestions.


Ideas for suggestions: pick a random deptnum from our data and suggest it; pick a random professor; pick a random word from a random course's description.

hannesmcman:
I was thinking we could put a "Browse" section beneath it, where we'd have suggested filters, as well as an option to select as many filters as you want and give all the results that match those filters without having to type anything in the search bar..?

hawkrives:
That sounds fancy. We can merge what you have here for now.

hawkrives (approved):

PR #2437, opened 2018-03-15
Remove componentWill* methods

### LINKED ISSUE #2430: fix react@16.2.0 deprecations
thanks #2396

hawkrives:
Closes #2430

React is removing the `componentWill*` methods: https://github.com/reactjs/rfcs/blob/master/text/0006-static-lifecycle-methods.md

The impetus for this is a future React release which will switch to "async rendering", wherein they will do cool things like _interrupt the rendering of a component to render a higher-priority thing_. But, to do that, they can't guarantee that `componentWill*` will only be called once, so they're being deprecated.

Here's your migration chart:

```js
  UNSAFE_componentWillMount() {
    // New name for componentWillMount()
    // Indicates that this method can be unsafe for async rendering.
    // Prefer componentDidMount() instead.
  }

  UNSAFE_componentWillUpdate(nextProps, nextState) {
    // New name for componentWillUpdate()
    // Indicates that this method can be unsafe for async rendering.
    // Prefer componentDidUpdate() instead.
  }

  UNSAFE_componentWillReceiveProps(nextProps) {
    // New name for componentWillReceiveProps()
    // Indicates that this method can be unsafe for async rendering.
    // Prefer static getDerivedStateFromProps() instead.
  }
```

The trickiest one is `getDerivedStateFromProps`; here's an example migration I did:

```diff
-	componentWillReceiveProps(nextProps: Props) {
-		this.setState(state => ({
+	static getDerivedStateFromProps(nextProps: Props, prevState: State) {
+		return {
			buildings: groupBuildings(
-				state.allBuildings,
+				prevState.allBuildings,
				nextProps.favoriteBuildings,
			),
-		}))
+		}
	}
```

For some of the other `componentWillReceiveProps`-esque methods we have, like `_performSearch` in CourseSearch, I extracted the bulk of the logic into a function that lives outside of the class entirely and returns an object instead of calling `this.setState` directly. Then I call that function both from `static deriveStateFromProps` and from `this._performSearch`.

For example: 

https://github.com/StoDevX/AAO-React-Native/blob/9269c0149e8603132cd15b7e6ab8ac2a40eb55bd/source/views/sis/course-search/search.js#L205-L214

https://github.com/StoDevX/AAO-React-Native/blob/9269c0149e8603132cd15b7e6ab8ac2a40eb55bd/source/views/sis/course-search/search.js#L128-L138

I know there are contradictory labels on this PR, but they're accurate – it's ready for review, but it needs to actually be run on a sim before approving. I haven't yet reinstalled Xcode on my home machine, so I only think that this works.


hawkrives:
Man it feels good to have a net-negative LOC PR.

drewvolz (approved):
Builds and runs. :shipit: 👍

hawkrives:
Did you check the menu filters? Those (and the equivalent in course search, but they're nearly identical logic now) are what I'm most worried about.

drewvolz (dismissed):
I realize now that some things need tweaking

drewvolz:
Warning on Course Search

1 | 2 | 3
--|--|--
search warning  | screen shot 2018-03-14 at 10 54 40 pm


Warnings on Menu

1 | 2
--|--
screen shot 2018-03-14 at 10 52 34 pm | screen shot 2018-03-14 at 10 52 30 pm

source/views/menus/components/fancy-menu.js
```diff
@@ -66,26 +66,21 @@ class FancyMenu extends React.PureComponent<Props> {
...
-		const {foodItems, menuCorIcons, filters, meals, now} = props
-
-		// we want to reset the filters when the menu is re-opened,
-		// but we want to avoid updating ourselves recursively, and
-		// especially avoid creating new filters on every `newProps`.
-		if (!initial && filters.length) {
+	static getDerivedStateFromProps(nextProps: Props) {
+		// prevent ourselves from overwriting the filters from redux on mount
+		if (nextProps.filters.length) {
 			return
```
hawkrives: This needs to be "return null"

source/views/sis/course-search/search.js
```diff
@@ -76,7 +125,20 @@ class CourseSearchView extends React.PureComponent<Props, State> {
...
+	static getDerivedStateFromProps(nextProps: Props, prevState: State) {
+		if (prevState.query && nextProps.filters !== prevState.cachedFilters) {
+			return executeSearch({
+				text: prevState.query,
+				filters: nextProps.filters,
+				applyFilters: nextProps.applyFilters,
+				allCourses: nextProps.allCourses,
+				updateRecentSearches: nextProps.updateRecentSearches,
+			})
+		}
```
hawkrives: Add "return null" after this line, so the method returns null if the condition doesn't pass.

hawkrives:
There is one remaining issue on this PR; the radio doesn't properly dispatch events to the `<audio>` tag.

hawkrives:
@drewvolz / @hannesmcman / @rye This PR now works correctly and is ready for review / merge.

hannesmcman (approved):

hawkrives:
```
> Could not resolve com.google.android.gms:play-services-basement:12.0.1.
         > Could not get resource 'https://jitpack.io/com/google/android/gms/play-services-basement/12.0.1/play-services-basement-12.0.1.pom'.
            > Could not GET 'https://jitpack.io/com/google/android/gms/play-services-basement/12.0.1/play-services-basement-12.0.1.pom'.
               > Read timed out
```

I will restart the Android build.

rye (approved):

hawkrives:
I've now emptied Circle's caches and am rebuilding from the start.

PR #2459, opened 2018-03-21
Add Browse functionality to Course Search

hannesmcman:
This PR adds a browse functionality to the course search view in response to #2407 that allows the user to sift through all the courses offered by adding filters, without having to enter a search into the search bar. It also stores the three most recent filter combinations, similar to how recent searches are stored. 
Here are some screenshots: 
<table>
<tr><td>screen shot 2018-03-20 at 6 59 52 pm</td>
<td>screen shot 2018-03-20 at 7 03 17 pm<td></tr>
</table>


hannesmcman:
Here's the android build failure:
```
> Task :app:transformDexArchiveWithExternalLibsDexMergerForDebug FAILED
D8: Program type already present: com.google.android.gms.common.api.internal.zzda
```
Any ideas?
Edited: I just noticed that this build failure is not unique to this PR, and submitted an issue: #2465

hawkrives:
Overall, I'm fairly happy with this.

- I think we should brainstorm a new term for "No recent browses"
- Should the subtitle-area-thing ("Your recent filter combinations will appear here") end in a period or not?

source/views/sis/course-search/list.js
```diff
@@ -69,19 +71,23 @@ export class CourseSearchResultsList extends React.PureComponent<Props> {
 			title: 'Add Filters',
 			pathToFilters: ['courses', 'filters'],
 			onChange: filters => this.props.onFiltersChange(filters),
+			onLeave: filters => this.props.updateRecentFilters(filters),
```
hawkrives: Any particular reason you added `onLeave`, instead of just updating the recent filters at the same time we call `onFiltersChange`?

hannesmcman: Yeah I figured we only want to add a filter combination to the recent filter combinations list after its been completed. For example, if a complete filter combination were `18/19:Fall/Int/Spring, ALS-L, ENGL`, we only want to store that full combination, rather than storing a new filter combination each time we add a new filter. If we called `updateRecentFilters` during `onChange`, there would be three new recent filter combinations added to the list in the process: `18/19:Fall/Int/Spring`, `18/19:Fall/Int/Spring, ALS-L`, and `18/19:Fall/Int/Spring, ALS-L, ENGL`. Calling it during `onLeave`, we only add the desired complete combination: `18/19:Fall/Int/Spring, ALS-L, ENGL`.

hannesmcman:
How about "No recent filter combinations"?

As for the period, it seems that our other `NoticeView`'s have them!

drewvolz:
"No recent filters" ?

hawkrives:
That sounds even odder to me, @drewvolz. I'm happy with "no recent filter combinations".

drewvolz:
Will it fit on one line? Do we care if it fits on one line? I think "no recent filter combinations" sounds odd.

hawkrives:
I think "no recent filters" sounds odder, so I am in favor of "no recent filter combinations".

drewvolz:
I still am not in favor of "no recent filter combinations".

drewvolz (dismissed):
:shipit: :+1:

source/views/sis/course-search/search.js
```diff
@@ -114,6 +120,28 @@ function executeSearch(args: {
 	}
 }
 
+function executeBrowse(args: {
```
hawkrives: Does a name like "applyFiltersAndQuery" fit this function better than "executeBrowse"?

source/views/sis/course-search/list.js
```diff
@@ -62,19 +60,23 @@ export class CourseSearchResultsList extends React.PureComponent<Props> {
...
 	render() {
-		const {filters} = this.props
+		const {filters, browsing} = this.props
 
 		const header = (
 			<FilterToolbar filters={filters} onPress={this.onPressToolbar} />
 		)
 
-		const message = this.props.searchPerformed
+		let message = this.props.searchPerformed
```
hawkrives: Join this into a nested ternary?

source/views/sis/course-search/lib/format-filter-combo.js
```diff
@@ -0,0 +1,50 @@
...
+
+export type FilterComboType = {
+	filters: FilterType[],
+	description: string,
+}
+
+export function formatFilterCombo(filters: FilterType[]): FilterComboType {
+	const filterCombo = filters.filter(f => f.enabled)
+	const comboDescription = filterCombo
+		.map(f => {
```
hawkrives: Extract this into a separate function (probably still within this file)?

hawkrives: Then use it like `.map(f => describeFilter(f.key))`

source/views/sis/course-search/search.js
```diff
@@ -315,18 +423,31 @@ class CourseSearchView extends React.PureComponent<Props, State> {
...
+									updateRecentFilters={this.props.updateRecentFilters}
 								/>
 							) : (
-								<View style={styles.common}>
-									<Text style={styles.subHeader}>Recent</Text>
-									<RecentSearchList
-										onQueryPress={this.onRecentSearchPress}
-										queries={this.props.recentSearches}
+								<View style={[styles.common, styles.bottomContainer]}>
+									<RecentItemsList
```
hawkrives, line 441: I like the components here 👍

hawkrives (dismissed):

source/views/components/filter/section-list.js
```diff
@@ -22,10 +22,10 @@ export function ListSection({filter, onChange}: PropsType) {
 	function buttonPushed(tappedValue: ListItemSpecType) {
 		let result
 
-		if (includes(selected, tappedValue)) {
+		if (some(selected, tappedValue)) {
```
hawkrives: This should still be "includes".

source/views/components/filter/section-list.js
```diff
@@ -22,10 +22,10 @@ export function ListSection({filter, onChange}: PropsType) {
...
 			// if the user has tapped an item, and it's already in the list of
 			// things they've tapped, we want to _remove_ it from that list.
-			result = without(selected, tappedValue)
+			result = reject(selected, tappedValue)
```
hawkrives: This should still be "without".

source/views/components/filter/section-list.js
```diff
@@ -67,7 +67,7 @@ export function ListSection({filter, onChange}: PropsType) {
 	let buttons = options.map(val => (
 		<Cell
 			key={val.title}
-			accessory={includes(selected, val) ? 'Checkmark' : undefined}
+			accessory={some(selected, val) ? 'Checkmark' : undefined}
```
hawkrives: This should still be "includes".

source/views/sis/course-search/list.js
```diff
@@ -71,12 +73,17 @@ export class CourseSearchResultsList extends React.PureComponent<Props> {
...
-			: message
+			: this.props.searchPerformed
+				? 'There were no courses that matched your query. Please try again.'
+				: "You can search by Professor (e.g. 'Jill Dietz'), Course Name (e.g. 'Abstract Algebra'), Department/Number (e.g. MATH 252), or GE (e.g. WRI)"
+		// let message = this.props.searchPerformed
+		// 	? 'There were no courses that matched your query. Please try again.'
+		// 	: "You can search by Professor (e.g. 'Jill Dietz'), Course Name (e.g. 'Abstract Algebra'), Department/Number (e.g. MATH 252), or GE (e.g. WRI)"
+		// message = browsing
+		// 	? 'There were no courses that matched your selected filters. Try a different filter combination.'
+		// 	: message
```
hawkrives: Delete this version. I like the single ternary 👍

hawkrives (dismissed):
With Lodash, "includes" and "without" operate on static values - strings and numbers and things with an identity. "reject" and "some" operate with lambdas, where they pass each item to their second argument so you can check it dynamically. Since you're checking static values there, it makes sense to use includes and without. 

Also without mutates the input array, while reject doesn't, so your current code doesn't actually do anything in that branch.

source/views/components/filter/section-list.js
hawkrives: Okay actually

hawkrives: My initial comment here was wrong. Because you're assigning to a value, you should use the non-mutating "reject" here, _but_ it needs a callback as the second argument. 

`reject(arr, item => item === tappedValue)` should be what you're looking for.

hannesmcman: screen shot 2018-03-31 at 9 55 46 pm

I was using the `matches` shorthand from the lodash documentation, which I tested and it works just fine..if you want me to change it for readability purposes I can

source/views/components/filter/section-list.js
hannesmcman: The reason I changed this to `some` is that `includes` compares objects by reference using `===`, which has worked so far with our filter objects because we haven't been creating new filter objects...However, while I was testing this branch I noticed that when the filters were updated from a recent filter combination, the filter view wouldn't display the selected filters correctly, even though everything on the Redux side of things was accurate. After some debugging, I narrowed it down to the uses of `includes` and `without` in this file, and changing them to `some` and `reject`, respectively, fixed the issue. Let me know if that doesn't make sense.

source/views/components/filter/section-list.js
hannesmcman: See my comment above..

source/views/components/filter/section-list.js
hawkrives: Ah. Well. That makes sense. I would like you to change it away from the matches() shorthand, though; if we're going to use reject and some, let's use the Array.prototype versions instead - `selected.reject(item => _.isDeepEqual(item, tappedValue))`, and the same for the "some" function - "array.some". 

Lodash v4, when it came out, started moving away from the matches shorthands in some of its functions, and I feel like it'll keep removing them in v5. Plus, we don't use them anywhere else in the app, so it's more consistent to use the explicit callbacks here.

source/views/components/filter/section-list.js
hawkrives: I keep forgetting that these are objects, not primitives. Some is good in that case, but I'd like to use `selected.some(item => _.isDeepEqual(item, tappedValue))` instead of the matches shorthand, please.

hannesmcman:
Looks like there isn't an Array.prototype version of reject...I'll switch it back to the lodash version with an explicit callback for now, let me know if you think I should do something else. @hawkrives

hawkrives:
Ah sorry, Array.prototype just has filter, but reject is just a negated filter. 

We can use _.filter though. 

Does isEqual do a deep comparison or a shallow one?

hannesmcman:
Do you want me to switch `_.reject` then? (Assuming `_` is lodash) And if so, do you want it to be `_.filter` or `Array.prototype.filter`? `_.isEqual` does a deep comparison.

hawkrives:
I meant that leaving lodash's reject is fine, my apologies. I was typing too fast.

hawkrives (approved):

PR #2467, opened 2018-03-24
Fix android builds (¿by removing version resolutions?)

### LINKED ISSUE #2465: CircleCi Android Builds Failing
All of the android builds seem to be failing due to this error:
```
> Task :app:transformDexArchiveWithExternalLibsDexMergerForDebug FAILED
D8: Program type already present: com.google.android.gms.common.api.internal.zzda
```

hawkrives:
Closes https://github.com/StoDevX/AAO-React-Native/issues/2465

## Overview

This was my resource that made me look at the version pinning things: https://rnfirebase.io/docs/v3.1.x/faqs#[Android]-Duplicate-Dex-Files-error-(build-time-error)

> A common build time error when using libraries that require google play services is of the form:
'Failed on android with com.android.dex.DexException: Multiple dex files… '
>
> This error (https://github.com/invertase/react-native-firebase/issues/48) occurs because different versions of google play services or google play services modules are being required by different libraries.
>
> The process to fix this is fairly manual and depends on your specific combination of external libraries. Essentially what's required is to check the app level build.gradle file of each external library and establish which ones have a Google Play Services dependency.
>
> You then need to find the lowest common version of each G.P.S module dependency, require that in the app level build.gradle file of your own project, and exclude it from being required by the modules themselves. This will force the use of a consistent version of the G.P.S module.
>
> It's not a good idea to modify the version within the library's build.gradle, as this will be overwritten when you update the library, which will lead to the build breaking again.
>
> A good break down of this process can be found here:
https://medium.com/@suchydan/how-to-solve-google-play-services-version-collision-in-gradle-dependencies-ef086ae5c75f

## Changes

- I added a separate "cache" step for rubygems, to separate it from the Android build. 
    - None of the other builds use it, but this way it can run and cache while yarn is caching too.
- I removed a bunch of `save_cache` steps from jobs that shouldn't be editing the caches
- I added `--console=plain` to the cradle invocation, to clean up the terminal output
    - This disables the fancy progress bar things
- I removed the `google-play-services` manual version resolutions, as we no longer have multiple projects with conflicting versions since we switched to Mapbox
- I added more manual version resolutions for Android support libraries, until Android Studio quit yelling at me
- `react-native-device-info` and `react-native-linear-gradient` both support external configuration for `compileSdkVersion` and `targetSdkVersion`, so I'm taking advantage of that to make them match our config

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -127,7 +127,6 @@ jobs:
       - checkout
       - restore_cache: *restore-cache-yarn
       - run: yarn install --frozen-lockfile
-      - save_cache: *save-cache-yarn
       - run: *run-danger
 
   flow:
@@ -138,7 +137,6 @@ jobs:
       - checkout
       - restore_cache: *restore-cache-yarn
       - run: yarn install --frozen-lockfile
-      - save_cache: *save-cache-yarn
       - run: mkdir -p logs/
       - run: yarn run bundle-data
       - run: yarn run --silent flow check --quiet | tee logs/flow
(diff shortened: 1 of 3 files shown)


drewvolz (approved):
🤞

PR #2475, opened 2018-03-28
Replace fuzzysearch with levenshtein

hannesmcman:
Fuzzysearch gave wayyy too many results, especially for short queries..Using the Levenshtein algorithm returns results that are far more relevant. This change also solves the middle initial problem, where searching the full name of a professor, like `Olaf Hall-Holt` wouldn't return any results because of the missing middle initial. 

This is built on top of #2437


hawkrives:
@hannesmcman I can't merge this on top of that PR; it's not ready to be merged yet.

hannesmcman:
Yeah that's what I figured...I was thinking we would just merge this after #2437

hawkrives:
Okay

source/views/sis/course-search/search.js
```diff
@@ -82,10 +82,12 @@ function executeSearch(args: {
 
 	const results = filteredCourses.filter(
 		course =>
-			fuzzysearch(query, course.name.toLowerCase()) ||
-			fuzzysearch(query, (course.title || '').toLowerCase()) ||
-			(course.instructors || []).some(name =>
-				name.toLowerCase().includes(query),
+			leven(query, course.name.toLowerCase()) < 2 ||
```
rye, line 85: Could `2` be pulled out into a constant?  What makes `2` so special?

source/views/sis/course-search/search.js
```diff
@@ -82,10 +82,12 @@ function executeSearch(args: {
...
+			leven(query, (course.title || '').toLowerCase()) < 2 ||
```
rye, line 86: Ditto here.

source/views/sis/course-search/search.js
```diff
@@ -82,10 +82,12 @@ function executeSearch(args: {
...
+			(course.instructors || []).some(
+				name =>
+					name.toLowerCase().includes(query) ||
+					(leven(name.toLowerCase(), query) < 4 && query.length > 4),
```
rye, line 90: Ditto here, can `4` be a constant?

rye (commented):
Just some idle thoughts, looks good.

From [the Wikipedia Article](https://en.wikipedia.org/wiki/Levenshtein_distance), I am led to believe that
>- It [the Levenshtein distance] is at least the difference of the sizes of the two strings.

So, if I search for "cs", which is an incredibly vague search term, will I then not get any results?

hannesmcman:
I've changed my mind about levenshtein. I think we should move in the direction of #2487, so I am closing this PR.

PR #2487, opened 2018-04-02
Replace fuzzysearch with keywordSearch

hannesmcman:
After looking closer at #2475 I decided that Levenshtein distance gave too many unwanted results, just like fuzzysearch, so I set out on a quest for a better alternative. I couldn't find anything close enough to what I wanted on npm so I decided to make my own. The search algorithm I came up with goes as follows: 
1 . Split the user's query into the words that make it up and put them in an array
2. Do the same with the item you are checking against (i.e. course name, professor's name, etc.)
3. If every word in the user's query is contained within one of the words in the item being checked, allowing for a single-character error, then we have a match. Otherwise, not.

After some testing I found this algorithm to return far less fluff than fuzzysearch or levenshtein, while also being able to return relevant results in spite of minor typos.


source/views/sis/course-search/search.js
```diff
@@ -72,6 +72,18 @@ type State = {
...
 }
 
+function keywordSearch(query: string, item: string) {
+	const keywords = query.split(' ')
+	const itemWords = item.split(' ')
+	return keywords.every(keyword =>
+		itemWords.some(
+			word =>
+				hamming(keyword, word.slice(0, keyword.length)) < 2 &&
+				keyword.length <= word.length,
```
hawkrives: Just one optimization: let's move the length comparison to be first, so the easy check can short-circuit the more involved one.

source/views/sis/course-search/search.js
```diff
@@ -88,10 +100,9 @@ function executeSearch(args: {
 
 	const results = filteredCourses.filter(
 		course =>
-			fuzzysearch(query, course.name.toLowerCase()) ||
-			fuzzysearch(query, (course.title || '').toLowerCase()) ||
```
hawkrives, line 92: Did you mean to remove the Title search here?

hawkrives (dismissed):
One early optimization that stood out to me, and one question

source/views/sis/course-search/search.js
hannesmcman: Oops! Nope!

hannesmcman:
@hawkrives @rye 
screen shot 2018-04-02 at 11 34 45 am

hawkrives (approved):

hawkrives:
@hannesmcman I think we can just restart the iOS build. The actual error in that log is the 502 response it got from GitHub, so we ought to be able to just restart it and have it pass.

PR #2493, opened 2018-04-02
Course Search: Sort courses by DeptNum

### LINKED ISSUE #2491: Course search: sort results by department/number
Instead of sorting courses by name, let's do it by department/number (i.e. `MATH 252`)

### LINKED ISSUE #2492: Course Search: Search results sometimes don't update
After some testing, it seems that if the user presses enter too quickly after finishing typing their search, the results do not update. This probably has something to do with nesting the `executeSearch` function within the `setState` call. This only happens when there are active filters.

hannesmcman:
Resolves #2491 

In addition, I changed my mind about when filters should be reset. I think they should only be reset when the user presses 'cancel' instead of when the search bar is pressed. This way, you can search within the courses that you are browsing without the filters resetting on you.


rye (dismissed):

source/views/sis/course-search/search.js
```diff
@@ -258,16 +259,16 @@ class CourseSearchView extends React.PureComponent<Props, State> {
...
+	_performSearch = (query: string) => {
+		const newState = executeSearch({
+			text: query,
+			filters: this.props.filters,
+			applyFilters: this.props.applyFilters,
+			allCourses: this.props.allCourses,
+			updateRecentSearches: this.props.updateRecentSearches,
+		})
+		this.setState(() => newState)
+	}
```
hawkrives: There is no functional difference between these two versions. Try tweaking the debounce instead?

source/views/sis/course-search/search.js
```diff
@@ -141,16 +141,17 @@ function applyFiltersAndQuery(args: {
...
-	const grouped = groupBy(filteredCourses, r => r.term)
+	const sortedCourses = sortBy(filteredCourses, course => deptNum(course))
+	const grouped = groupBy(sortedCourses, r => r.term)
 	const groupedCourses = toPairs(grouped).map(([key, value]) => ({
 		title: key,
 		data: value,
 	}))
 
-	const sortedCourses = sortBy(groupedCourses, course => course.title).reverse()
+	const sortedTerms = sortBy(groupedCourses, course => course.title).reverse()
```
hawkrives: Make sortedCourses a "let" instead, and reassign to it here

source/views/sis/course-search/search.js
hannesmcman: I tried that and it didn't work...this was the only thing that worked!

source/views/sis/course-search/search.js
hannesmcman: sortedCourses and sortedTerms are pretty different though...sortedCourses is an array of courses, whereas sortedTerms is an array of terms (courses grouped into terms), do we want to store them in the same variable?

source/views/sis/course-search/search.js
hawkrives: Then this is relying on a timing thing that happens to be working right now but will randomly break in the future. We need to actually find the bug and fix it.

source/views/sis/course-search/search.js
hawkrives: Ah. I missed the grouping into terms, even though I read through to see if that's what you were doing. If the name's descriptive, let's keep it.

hawkrives:
@hannesmcman You may want to split this into two PRs since it solves two distinct issues. I'm happy to merge the sort-by-deptnum one, but the other one's going to take more thinking before I'm happy with it.

source/views/sis/course-search/search.js
hannesmcman: I do think this is where the bug is and I don't think it's a random timing thing. After some more testing, another alternative that works is 
```javascript
this.setState(
	executeSearch({
		text: query,
		filters: this.props.filters,
		applyFilters: this.props.applyFilters,
		allCourses: this.props.allCourses,
		updateRecentSearches: this.props.updateRecentSearches,
	})
)
```
This leads me to believe that there is a functional difference between passing an arrow function to `setState` and passing a static object to `setState` that is causing this difference in performance. I don't fully understand the nature of that difference, but might it be that `setState(() => executeSearch(...))` triggers a re-render before the results have been returned from `executeSearch`, whereas storing the result of `executeSearch` in a variable and calling `setState` synchronously after that result is known leads to the desired performance?

hawkrives: [React docs](https://reactjs.org/docs/react-component.html#setstate):

> This form of setState() is also asynchronous, and multiple calls during the same cycle may be batched together. For example, if you attempt to increment an item quantity more than once in the same cycle, that will result in the equivalent of:
>
> ```js 
> Object.assign(
>   previousState,
>   {quantity: state.quantity + 1},
>   {quantity: state.quantity + 1},
>   ...
> )
> ```
> Subsequent calls will override values from previous calls in the same cycle, so the quantity will only be incremented once.


[StackOverflow about "do re-renders happen between setState calls"](https://stackoverflow.com/questions/48563650/does-react-keep-the-order-for-state-updates/48610973#48610973): 

> The key to understanding this is that **no matter how many setState() calls in how many components you do inside a React event handler, they will produce only a single re-render at the end of the event**. This is crucial for good performance in large applications because if Child and Parent each call setState() when handling a click event, you don't want to re-render the Child twice.

-----

We've been avoiding the non-callback version of setState for a while, because remembering when things are queued and available vs. not queued and potentially skipped is annoying.

If the code here is somehow reliant on being executed in the same … execution cycle, I guess (although that's not the right term and I'm blanking on the correct one) – as the render, then that's a problem, because we cannot guarantee that that happens.

Because Javascript has what's called "run to completion" semantics, we're guaranteed that any function will be completed before something else starts running. This is why you have to mark the points where you'll give up control in async functions, with the `await` keyword. Or, if you use callbacks, the current function will continue executing once it's passed the function reference to the place that will invoke the callback (`readdir('./', (err, files) => console.log(files)); console.log('hi')` will log "hi" and then the list of files every time, because the outer scope has to finish before the inner one can start. This is also why "synchronous XHR" is deprecated and actually scheduled for removal from web browsers, because letting an HTTP request actually block the event loop is a recipe for angry users who can't interact with your page.)

So. Even though the user can keep typing in the search box, we're guaranteed by React's setState model and Javascript's semantics that every one of those `executeSearch`s that we pass to setState will be run, although not every intermediate state will be rendered. However, the final state will always be rendered.

---

Also, just to be clear – there's no such thing as a "synchronous setState". React just doesn't do that. Even the non-callback version is queued into an internal queue and batched up later.

---

Just to check, because I've failed to notice things several times today: you aren't passing anything from `this.state` to executeSearch, are you? If you are, you need to use the callback form and accept state as the first argument, instead of accessing `this.state`. (that is, `this.setState(state => executeSearch(state.foo))` instead of `this.setState(() => executeSearch(this.state.foo))`)

hannesmcman:
@hawkrives Ok! I'll get rid of the attempted bugfix and make a new PR for it.

hawkrives (approved):

hawkrives (approved):

PR #2495, opened 2018-04-02
enforce usage of "===" operator in JS

hawkrives:
from [ESLint's docs](https://eslint.org/docs/rules/eqeqeq):

> It is considered good practice to use the type-safe equality operators `===` and `!==` instead of their regular counterparts `==` and `!=`.
>
> The reason for this is that `==` and `!=` do type coercion which follows the rather obscure Abstract Equality Comparison Algorithm. For instance, the following statements are all considered true:
> 
> ```js
> [] == false
> [] == ![]
> 3 == "03"
> ```
> 
> If one of those occurs in an innocent-looking statement such as `a == b` the actual problem is very difficult to spot.

I thought I'd already blocked use of `==` and `!=`, but apparently not. This PR will block use of those two operators in our code.

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -34,6 +34,7 @@ rules:
   curly: [warn, multi-line]
   default-case: error
   guard-for-in: error
+  eqeqeq: [error, always, {'null': ignore}]
   linebreak-style: [error, unix]
   new-cap: off
   no-await-in-loop: warn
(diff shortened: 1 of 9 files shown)


drewvolz:
Been hoping we’d pick up on this for some time now. Thanks!

drewvolz (approved):

PR #2510, opened 2018-04-06
Update Tom Porter hours

hawkrives:
I've been informed that Tom Porter's hours are really weird (and sometimes PubSafe will wander by and lock the building even after the electronic locks are opened), but that setting it to 7-7 every day should be safe.

### DIFF
--- a/data/building-hours/6-4-tom-porter.yaml
+++ b/data/building-hours/6-4-tom-porter.yaml
@@ -5,10 +5,7 @@ image: tom-porter
 schedule:
   - title: Hours
     hours:
-      - {days: [Mo, Tu, We, Th], from: '10:00am', to: '8:00pm'}
-      - {days: [Fr], from: '7:00am', to: '5:00pm'}
-      - {days: [Sa], from: '8:00am', to: '9:00pm'}
-      - {days: [Su], from: '12:00pm', to: '9:00pm'}
+      - {days: [Mo, Tu, We, Th, Fr, Sa, Su], from: '7:00am', to: '7:00pm'}
 
 breakSchedule:
   fall: []


drewvolz:
I’m not so sure deviating from the published schedule is the right answer. I’d say the source should update if we are going to update. Otherwise we’ll be dealing with reports of incorrect hours and having to explain ourselves.

hawkrives:
I don't know that Tostrud has published hours…

drewvolz:
> I don't know that Tostrud has published hours…

If you're wondering whether "Tostrud" has published hours, those are [here](https://wp.stolaf.edu/recreation/facility-hours/). 

> Update Tom Porter hours

PR says "Tom Porter" but your comment above questions Tostrud. Porter Hall hours live [here](https://wp.stolaf.edu/recreation/).

hawkrives:
That page only has TPH hours for the weight room, which is more specific than what we list in our app.

hannesmcman:
screen shot 2018-04-06 at 2 11 00 pm

I think Porter is only open to the student body from 12-2pm, and is reserved by sports teams the rest of the day.

drewvolz:
Sounds as if the hours should be 12pm - 2pm according to the page @hawkrives

hawkrives:
… maybe it's time I break my streak and actually visit the building. TPH and the Art Barn are the two buildings on campus that I've never been inside.

hannesmcman:
@hawkrives you’ve never been to the art barn?? 😮

drewvolz:
@hawkrives the hours are interesting. I had a weight training course in TPH but it has special hours for athletes. Go and visit it to confirm, but I would advise us against deviating away from the list of published hours.

hawkrives:
I don't even know what all is in TPH, besides the weight room. 

I think we'd be having less of a discussion here if we listed TPH and the TPH weight room separately.

drewvolz:
Here's the [list](http://athletics.stolaf.edu/sports/2014/9/23/tomporterhall.aspx). I should have mentioned the class was weight lifting.

* Offices (Football)
* Meeting Rooms
* Locker Rooms
* Weight Room
* Equipment Room
* Athletic Training Room

hawkrives:
Weight Lifting apparently gets the building open at 5:45 this semester, too, fwiw.

PR #2519, opened 2018-04-10
Make course description selectable on iOS

drewvolz:
This PR converts the description field on the iOS course search detail view to `<SelectableCell />` (which is a `<TextInput />` with styling). This allows the description to take advantage of OS level data detectors and text selection.


hawkrives (approved):
@drewvolz merge when ready

PR #2520, opened 2018-04-10
Correct valid times data url

drewvolz:
The file names within the `course-data` repo are separated with underscores rather than hyphens


hawkrives (approved):

drewvolz:
@hawkrives we list valid times but we do not use that url anywhere. Should we bother to keep it up-to-date, or leave it? Additionally, why not add the [remaining files](https://github.com/StoDevX/course-data/tree/gh-pages/data-lists) as well?

PR #2383, opened 2018-02-25
Course search enhancements

drewvolz:
Closes #2341
* Fuzzy search
* Search titles

The titles are additional to the course names.
Does not solve searching for names where middle initials prevent a match in prof names

Issue #2502, opened 2018-04-03
Course search promise rejection


drewvolz:
Happens when the course search data for Android is loaded.

<img src="https://user-images.githubusercontent.com/5240843/38281133-5ba1e478-3766-11e8-9b8d-e79c63e4e6f0.png" width=420 />

hawkrives:
Can you give us the stack trace from the devtools? That'll have the actual file paths in it

drewvolz:
I could not. I spent an hour trying to get RN debugger to launch but it was freezing and stalling out. My devtools would not successfully launch.

drewvolz:
Weirdly enough, the error happens when I do not have a debugger enabled. I cannot replicate with the debugger, but I see the rejection when using as normal.

drewvolz:
Bugsnag has this to say:

<img width="770" alt="screen shot 2018-04-08 at 2 18 12 pm" src="https://user-images.githubusercontent.com/5240843/38472078-b311dea0-3b37-11e8-9f36-a60f0d643fa1.png">

drewvolz:
Relevant: https://github.com/facebook/react-native/issues/15902

hawkrives:
Try 

```js
// symbol polyfills
global.Symbol = require('core-js/es6/symbol');
require('core-js/fn/symbol/iterator');

// collection fn polyfills
require('core-js/fn/map');
require('core-js/fn/set');
```

The reason it works with a debugger is that remote debugging uses Chrome's js engine, instead of the bundled js engine.

drewvolz:
Getting a runtime error to the effect of `Incompatible receiver, Set required!` which that thread is hinting at `core-js` being the culprit.

hannesmcman:
So is there nothing we can do about this one besides wait?

hawkrives:
I'll try to dig into this. Assigning to me.

hawkrives:
Having looked at the code again, I think it's the Promise#finally:

https://github.com/StoDevX/AAO-React-Native/blob/f74db946946ef9dfa0eeb3c9b49d07c2430f2da9/source/views/sis/course-search/search.js#L233

Promise.prototype.finally is not yet standardized. I have now made a local branch that replaces the Promise chain here with an async function, which should have the same results.

hawkrives:
Not fixed yet. 

>So yeah, we still have the Symbol.iterator problem. Easy (temporary) solution: move away from p-props. 
>
[...]

PR #2523, opened 2018-04-12
Course Search: fix search bug

### LINKED ISSUE #2432: Course Search: results are sometimes not displayed on search

### LINKED ISSUE #2492: Course Search: Search results sometimes don't update

hannesmcman:
Resolves #2432.
Resolves #2492.

Hawken and I had a long comment thread about this in #2383 and so I've decided to bring this bug fix to another PR. For some reason, and I'll be honest I don't know why, the call to `this.setState` does not work properly when the `executeSearch` function is called within it, but it works fine if the results of `executeSearch` are stored in a variable, which is in turn passed to `this.setState`.

Doesn't work:
```javascript
this.setState(() => executeSearch({
	text: query,
	filters: this.props.filters,
	applyFilters: this.props.applyFilters,
	allCourses: this.props.allCourses,
	updateRecentSearches: this.props.updateRecentSearches,
}))
```

Works:
```javascript
const results = executeSearch({
	text: query,
	filters: this.props.filters,
	applyFilters: this.props.applyFilters,
	allCourses: this.props.allCourses,
	updateRecentSearches: this.props.updateRecentSearches,
})
this.setState(() => results)
```

I've tested this considerably and this change does eliminate the bug as far as I can tell. While it definitely seems at first glance that these two code snippets are functionally equivalent, my suspicion is that we're missing something subtly different between them that causes the latter to work. Thoughts?


rye (approved):

hawkrives (changes requested):
I still don't know what's going on, but I don't want to merge this as-is because it's only moving a race condition around IMO.

hannesmcman:
But if it there were a race condition in the second option, wouldn't it only work some of the time? I've tested it a whole bunch of times and it has never not worked... @hawkrives 

I know I'm not giving you a satisfying explanation for how it's functionally different from what we have now, but I just don't have the time right now to dive into the RN docs to search for the difference..I personally don't see the harm in merging this if it solves a bug that is the only major hold on shipping course search..

drewvolz:
#2502 is an additional blocker on course search fwiw

rye:
Needless to say, this works for me. I have no clue why we don't just merge this and try it and production?

rye (approved):

hawkrives:
It does work. 

I haven't wanted to merge it because it's just moving a race condition somewhere else, where it's harder to trigger, because the two implementations should be equivalent.

rye (commented):
Without knowing much about JS' internals, I have the two following observations:

- This seems to fix the underlying problem. I've tried it out a bunch and the issue we were having before no longer exists, or happens much less frequently at least.

- `() => [non-const return value from executeSearch]` should, in my opinion, be very different from `() => [const]`. I think that's what the underlying problem is, or…

I can't think of anything else, so I feel we need to either say this is okay for nightly testing (because it simply moves the evaluation and doesn't change the functional outcome of the method) or come up with a reason why this needs to change.

Personally, I'd rather test this in our nightly group where we can at least get real stack traces to indicate where the race condition is "moved to."

hawkrives (approved):
I have a WIP that solves whatever problem is here by avoiding the need to call executeSearch here entirely, so meh sure we can merge this.

hawkrives:
@rye do note that all "const" means in JS-land is that the variable name can not be rebound. That's all. (ie, no "const x= 2; x=3".)

hannesmcman:
@StoDevX/all-about-olaf can someone merge this?

rye:
Done.

PR #2536, opened 2018-04-23
Experimenting with fixing gradle and google analytics

drewvolz:
I'd like to find a better way. But let's see if this builds.


hawkrives (approved):
I'm OK with this if it works, which it seems to.

I'd just like to have another issue filed to figure out a different way to accomplish this, ideally centered around `app/build.gradle`.

PR #2547, opened 2018-04-24
Remove insecure http load exception for oleville

### LINKED ISSUE #2545: Remove entries and exceptions from Info.plist
Looking at iOS' `Info.plist`, I am wondering if we can remove two items:

- [x] KSTO scheme (app is no longer published, answer is _yes_)

https://github.com/StoDevX/AAO-React-Native/blob/89009eef5adc993ac50aad363ca9c0e856ab9e88/ios/AllAboutOlaf/Info.plist#L72-L75

- [x] Oleville insecure resource? ~The site is `https` now but do we have it listed because it loads insecure content?~

https://github.com/StoDevX/AAO-React-Native/blob/89009eef5adc993ac50aad363ca9c0e856ab9e88/ios/AllAboutOlaf/Info.plist#L94-L98

drewvolz:
This is part two of and closes #2545. We no longer need the insecure exception for oleville.

### DIFF
--- a/ios/AllAboutOlaf/Info.plist
+++ b/ios/AllAboutOlaf/Info.plist
@@ -87,11 +87,6 @@
 				<key>NSTemporaryExceptionAllowsInsecureHTTPLoads</key>
 				<true/>
 			</dict>
-			<key>oleville.com</key>
-			<dict>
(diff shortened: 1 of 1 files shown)


rye (approved):
Yes please. HTTPS only.

PR #2560, opened 2018-04-29
Report dictionary issue

drewvolz:
Report an issue with the dictionary words and definitions. This brings over a feature that was implemented in CARLS. This PR is a shared effort of work between @hawkrives and myself.


rye (dismissed):

drewvolz:
- [x] I have to fix the test snapshots folders

source/views/dictionary/report/stringify-entry.test.js
```diff
@@ -0,0 +1,40 @@
+/* eslint-env jest */
```
hawkrives: This file shouldn't exist

hawkrives (approved):
I don't think I've made any changes to this since merging it in CARLS so :shipit:

PR #2564, opened 2018-05-02
Modularize keywordSearch function

hannesmcman:
I published the keywordSearch function on npm today, so I figured I might as well modularize it here too!


source/views/sis/course-search/search.js
```diff
@@ -100,10 +88,10 @@ function executeSearch(args: {
 
 	const results = filteredCourses.filter(
 		course =>
-			keywordSearch(query, course.name.toLowerCase()) ||
-			keywordSearch(query, (course.title || '').toLowerCase()) ||
+			keywordSearch(query, course.name.toLowerCase(), 1) ||
+			keywordSearch(query, (course.title || '').toLowerCase(), 1) ||
 			(course.instructors || []).some(name =>
-				keywordSearch(query, name.toLowerCase()),
+				keywordSearch(query, name.toLowerCase(), 1),
```
hawkrives, line 94: Aren't these "1" parameters set to "2" in the current implementation? Does that matter?

hannesmcman: Yeah I changed the package to use <= instead of < so that that parameter is the permitted hamming distance.

hawkrives: Ah I missed that change

hawkrives (approved):

drewvolz (approved):

PR #2574, opened 2018-05-07
React Navigation 2.0

hawkrives:
wow

https://reactnavigation.org/blog/2018/05/07/react-navigation-2.0.html

- i cant even: "State persistence - automatically save state and reload it when the app restarts" https://github.com/react-navigation/react-navigation/pull/3716
- `XNavigator(...)` is now `createXNavigator(...)`
- `this.props.navigator.navigate()` now automatically re-uses previous views, which should fix the long-standing double-push bug (#155)

I kinda want to merge this and look at it tomorrow and file issues to fix things then

### DIFF
--- a/ios/AllAboutOlaf/Info.plist
+++ b/ios/AllAboutOlaf/Info.plist
@@ -108,7 +108,7 @@
 	<key>UIStatusBarHidden</key>
 	<false/>
 	<key>UIStatusBarStyle</key>
-	<string>UIStatusBarStyleLightContent</string>
+	<string>UIStatusBarStyleDarkContent</string>
 	<key>UISupportedInterfaceOrientations</key>
 	<array>
 		<string>UIInterfaceOrientationPortrait</string>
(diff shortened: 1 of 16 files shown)


drewvolz:
We should at least try to fix the breaking changes before we merge

hawkrives:
The rename isn't breaking; it's a deprecation, since they're not removing the old way until v3.

hawkrives:
That is to say, this should "just work" if we merge tonight with no further changes.

drewvolz (approved):

drewvolz:
@hawkrives has to say

> it doesn't work yet

drewvolz (dismissed):
it don't work yet

hawkrives:
Hokay.

Other changes, now: 

- Android now has the fancy new Material bottom tabs
- They're black on gold
- To match that, the titlebar is also black on gold. I ask that we sit with it for a day+ before looking for other colors (dark gray?)

---

Also, React Navigation changed how you define the title on a TabNavigator; it's now much closer to how you define it on an individual screen.

That is, instead of passing it as the second argument to TabNavigator, you just set `myTabNav.navigationOptions = {title: 'MyTitle'}`.

---

Oh, and Prettier reformatted the TabNavigators because there's only one argument now.

drewvolz (approved):

Issue #2580, opened 2018-05-09
Android Recent Apps color value not set


rye:
It looks kinda weird.

![Hawken's Screenshot](https://user-images.githubusercontent.com/1566689/39820791-3c191ca8-536c-11e8-8263-73c90e159f70.jpg)

We can apparently fix this by adding to a new file `android/app/src/main/res/values/colors.xml` the following XML, 

```xml
<?xml version="1.0" encoding="utf-8"?>
<resources>
    <!-- Change the color value for your need -->
    <color name="colorPrimary">#3F51B5</color>
    <color name="colorPrimaryDark">#303F9F</color>
    <color name="colorAccent">#FF4081</color>
</resources>
```

then also adding the following XML to `android/app/src/main/res/values/styles.xml`

```xml
<resources>
    <!-- Base application theme. -->
    <style name="AppTheme" parent="Theme.AppCompat.Light.NoActionBar">
        <!-- Customize your theme here. -->
        <item name="colorPrimary">@color/colorPrimary</item>
        <item name="colorPrimaryDark">@color/colorPrimaryDark</item>
        <item name="colorAccent">@color/colorAccent</item>
    </style>
</resources>
```

PR #2581, opened 2018-05-09
Android switcher color

hawkrives:
Closes https://github.com/StoDevX/AAO-React-Native/issues/2580

Also, we changed the accent color for check boxes and switches:

It's now purple, from the list of approved St. Olaf colors.

### DIFF
--- a/android/app/src/main/res/values/colors.xml
+++ b/android/app/src/main/res/values/colors.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <!-- Change the color value for your need -->
+    <color name="colorPrimary">#ffc107</color>
+    <color name="colorPrimaryDark">#c79100</color>
+    <color name="colorAccent">#6E3A5D</color>
+    <color name="textColorPrimary">#ffffff</color>
+</resources>
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -6,7 +6,7 @@ buildscript {
         google()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.1.0-alpha09'
+        classpath 'com.android.tools.build:gradle:3.1.2'
 
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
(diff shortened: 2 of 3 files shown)


drewvolz:
...purple?

drewvolz:
I don't know that our color scheme of gold, black, and purple looks the best. We should highly reconsider.

rye:
Purple doesn't seem that good in context.  It looked good on the theme editor, but kinda clashes with the gold title bar.

hawkrives:
I am open to suggestions.

rye:
I think I might like a dark grey or black, potentially.

How does Chapel Cranberry look on this?

android/app/src/main/res/values/colors.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <!-- Change the color value for your need -->
+    <color name="colorPrimary">#ffc107</color>
+    <color name="colorPrimaryDark">#c79100</color>
+    <color name="colorAccent">#6E3A5D</color>
```
rye, line 6: This should be downcased.

android/build.gradle
```diff
@@ -6,7 +6,7 @@ buildscript {
         google()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.1.0-alpha09'
+        classpath 'com.android.tools.build:gradle:3.1.2'
```
rye, line 9: Why was this change necessary?  (Or was this automatically performed?)

rye (commented):
I was thinking about #f98e2c as a possible accent color, too.  (Mellby Orange)

Or Chapel Cranberry. (#ac145a-ish)

rye:
The more I think about it, though, purple is actually a pretty cool color for these cases… it's roughly opposite the color wheel from yellow/gold.  I wish we had more screenshots. 😛

android/build.gradle
hawkrives: I think Android Studio was yelling at me / did it automatically

hawkrives:
I still like purple, as well 🐉

My android emulator isn't working atm, so I can't do more screenshots.

PR #2590, opened 2018-05-14
Fix dictionary updates

hawkrives:
~WIP; needs to be applied to Orgs as well, although the bug is invisible there. ~ This has been applied to Orgs as well.

This is a bug we found on CARLS; Drew diagnosed it. 

- Change the remote dictionary file
- Open the app and open Dictionary
- See the lack of the updated definition
- Search the list
- See the updated definition

This is caused by the list only rendering the result of performSearch, and performSearch not being called when the remote data is loaded. 

So, rather than call performSearch, I propose that we just do the filtering in render() itself. 

I think this is OK because it's a fast operation, for our scales, and render() won't be invoked unnecessarily for this top-level View.

### DIFF
--- a/source/views/dictionary/list.js
+++ b/source/views/dictionary/list.js
@@ -48,7 +48,7 @@ const styles = StyleSheet.create({
 type Props = TopLevelViewPropsType
 
 type State = {
-	results: Array<WordType>,
+	query: string,
 	allTerms: Array<WordType>,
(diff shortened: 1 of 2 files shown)


hawkrives:
This PR does not fix the bug in Building Hours, as I believe that to have a different root cause.

drewvolz (approved):

PR #2635, opened 2018-06-07
Next bottom tab bar iteration

rye:
TODO: Screenshots (later)

Changes to `c.black` active tint/`c.black25Percent` inactive tint on a `c.olevilleGold` background.

### DIFF
--- a/source/views/components/tabbed-view/index.js
+++ b/source/views/components/tabbed-view/index.js
@@ -25,9 +25,7 @@ export const TabNavigator: ComponentType = (screens, options = {}) =>
 	createTabNavigator(screens, {
 		backBehavior: 'none',
 		lazy: true,
-		activeTintColor: c.olevilleGold,
 		tabBarOptions: {
-			activeTintColor: c.black,
 			...(options.tabBarOptions || {}),
 			labelStyle: {
 				...Platform.select({
--- a/source/views/components/tabbed-view/index.js
+++ b/source/views/components/tabbed-view/index.js
@@ -25,6 +25,8 @@ export const TabNavigator: ComponentType = (screens, options = {}) =>
 	createTabNavigator(screens, {
 		backBehavior: 'none',
 		lazy: true,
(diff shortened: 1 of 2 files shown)


source/views/components/tabbed-view/index.js
```diff
@@ -25,9 +25,14 @@ export const TabNavigator: ComponentType = (screens, options = {}) =>
...
-		activeTintColor: c.olevilleGold,
+		activeTintColor: c.black,
+		inactiveTintColor: c.black25Percent,
+		theme: {
+			colors: {
+				primary: c.olevilleGold,
+			},
+		},
 		tabBarOptions: {
-			activeTintColor: c.black,
```
hawkrives: This one needs to stay around for iOS, or else we'll have blue tints.

hawkrives (changes requested):

source/views/components/tabbed-view/index.js
```diff
@@ -33,6 +33,8 @@ export const TabNavigator: ComponentType = (screens, options = {}) =>
 			},
 		},
 		tabBarOptions: {
+			activeTintColor: c.black,
+			inactiveTintColor: c.black25Percent,
```
hawkrives: Please remove inactiveTintColor for iOS. I don't know if it's supported or not, but I'd rather use the platform default convention for inactive tabs.

hawkrives (dismissed):

rye:
We should definitely think about ways to make it more clear that `tabBarOptions` applies to iOS and the rest applies to Android, because this is pretty confusing.

hawkrives:
It doesn't, though. Only those two options are iOS-only. 

The rest apply to both platforms. 

React Navigation's documentation is also confusing. 

We could just add a comment over it saying "// this prop is iOS only" – but at least we only have one place to change all of the tabs.

hawkrives (approved):

PR #2652, opened 2018-06-13
Ignore "ordinary" yellowbox warnings

hawkrives:
We keep having ~9 "Warning: isMounted(...) is deprecated" warnings pop up every time we launch the app in the simulator.

Turns out they're from React Native itself, and they'll be fixed in the next stable.

Until then, we can just ignore them.

We should revert this PR as part of the 0.56 upgrade.

### DIFF
--- a/source/root.js
+++ b/source/root.js
@@ -1,10 +1,17 @@
 // @flow
 
-import {AppRegistry} from 'react-native'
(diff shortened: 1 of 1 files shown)


hannesmcman:
Yes please!

hannesmcman (approved):
Yay!

PR #2705, opened 2018-07-26
Upgrade to React-Native v0.56

hawkrives:
… yeah

I've done cursory testing on iOS (I can open the app, and filter the menu, at least).

~I have successfully _built_ the Android app, but I have not attempted to run it.~ nope.
> Output:  error: resource android:style/TextAppearance.Material.Widget.Button.Borderless.Colored not found.

I also cleaned up a bunch of types across the codebase.

(I have not re-enabled react-navigation or redux; types for them have tended to be a giant pain, and we're careful-ish.)

---

I think that Danger doesn't like the Babel@7 upgrade.


hawkrives:
I'm going to close this; I don't think we can make the Android upgrade with 0.56.

We're already good for the Play Store's fall requirements, so no worries there.

There are a number of changes coming in 0.57 that ought to let us use AAPT2, which should let us actually upgrade.

I'd like to cherry-pick a lot of the flow changes from this branch, but I'll do that without having the PR open.

PR #2723, opened 2018-08-03
Fix cs lag

### LINKED ISSUE #2720: Course Search: Filter buttons lagging/causing app crash
Adding filters to course search is almost impossible, with how much it is lagging. It sometimes even causes the app to crash.

hannesmcman:
Resolves #2720.
Just needed to get rid of the `ScrollView` and decreased windowSize to prevent lag of navigation buttons.


hawkrives (approved):
LGTM

drewvolz (approved):

PR #2725, opened 2018-08-04
Remove Course filters from Redux and refactor

### LINKED ISSUE #2578: Search crashes when moving focus
This is on android. After a search on orgs or dictionary, we crash. It looks like the Android searchbar is trying to call methods that only exist on iOS?

hawkrives:
This builds on top of #2722.

(I've tested that searching works, in Courses, Orgs, and Dictionary, on iOS.)

Whoo boy. I did, at least, restrain myself from entirely removing Courses from redux, but … yeah.

Okay. Changes:

- I reworked `<Searchbar/>` to be a ["controlled component"](https://reactjs.org/docs/forms.html#controlled-components), and cleaned up the API across the components (for instance, it now always returns a string, which let me clean up `performSearch` in Dictionary.)
  - This involved forking the Android searchbar to make it take a `value` prop
- I removed course filters from Redux (the original impetus of this PR)
- <details><summary>I added a "no filters" NoticeView to FilterView</summary>
    screen shot 2018-08-03 at 8 02 07 pm</details>
- I extracted the AnimatedSearchbox from CourseSearch into its own component
  - this let me remove a good bit of nesting from CourseSearch, since AnimatedSearchbox was the reason it needed Viewport
- I moved the RecentsList component up a level on-disk
- I extracted the actual "searching" function from CourseSearch to `lib/execute-search`, and let it breath a little
- I moved the actual "searching" from CourseSearch to CourseResultsList
  - This lets me get rid of _so much code_, because I can just say "okay, given these arguments, what are the results?" instead of trying to handle all of the different change sources.
  - I also [memoized](https://www.npmjs.com/package/mem) the search function, so I didn't even need state to do the search!
- I put back in a ScrollView, being careful to not wrap the results in it, so that you can scroll the list of recent items

Yeah… I think that's everything.

Plz test? Thx.

## Questions for another PR
- How do you _exit_ browse mode? Or search mode, for that matter? If you activate the search bar, you can cancel it, but … yeah.

Closes #2578


hawkrives:
I have verified that Danger didn't send any actual comments.

rye (approved):

drewvolz (approved):

hannesmcman (approved):
Yes this is very good

PR #2732, opened 2018-08-04
Fix course search crash on iOS

hawkrives:
Turns out that iOS was crashing as soon as you tapped the search bar, if you didn't have the debugger attached.

`mem`'s internals were trying to copy an object property onto something that iOS doesn't like.

I replaced it with Lodash and a [WeakMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap). This should be fine; Drew tested it on iOS 9 and Android, and I tested on iOS 11.

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,6 @@
     "htmlparser2": "3.9.2",
     "keyword-search": "0.1.1",
     "lodash": "4.17.10",
-    "mem": "3.0.1",
     "moment": "2.22.2",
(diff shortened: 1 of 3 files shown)


drewvolz (dismissed):
👍 approval pending build passes; I did test this on both Android and iOS.

rye (approved):

drewvolz (approved):

PR #2733, opened 2018-08-04
Disable autocorrect in controlled searchbar

drewvolz:
I found that I had a better experience with autocorrect disabled.


hawkrives:
I feel like this should be a prop that we pass down, instead of forcing it to be disabled? So that we could enable it, if we wanted to, from the screens that needed it. It could be false by default.

hawkrives:
Whatever. We can file an issue for that change.

hawkrives (approved):

hawkrives:
Yeah, scratch my comment. I can't think of anywhere that searching is improved by autocorrect. I'm good with this as-is :shipit:

PR #2743, opened 2018-08-06
Support Radio theming

hawkrives:
This is also a test in the waters of app-wide theming. It uses [`@callstack/react-theme-provider`](https://github.com/callstack/react-theme-provider) to provide the theme object through React's Context api, so we don't have to pass the theme down ourselves.

&nbsp; | before | after
---- | --- | ---
ksto | ksto-before | ksto-after
krlx | krlx-before | krlx-after


drewvolz:
This needs another update from master.

hawkrives:
@drewvolz This PR has the dumpster; https://github.com/StoDevX/AAO-React-Native/blob/6bb2db24592a674159609f4cb8a3efe5e7e44b9d/images/streaming/ksto.png, where 6bb2db24592a674159609f4cb8a3efe5e7e44b9d is HEAD of this branch.

drewvolz (changes requested):
I don't think it's right to theme the buttons on this view differently than the accent colors of the entire theme and application.

drewvolz (approved):
But maybe I simply need to get used to it.

PR #2744, opened 2018-08-07
Replace FilterView with Interactive FilterToolbar

### LINKED ISSUE #2738: Sliding to pop navigation causes issue on willBlur
When I navigate in the app, I sometimes use the back button in the nav bar and sometimes pull on the side of the view to return back to the previous screen. They accomplish the same thing.

Within course filters, changes to the filter lists only apply if you tap the back button in the nav bar (on iOS only? Not sure). Do we want to have both back actions apply changes, or should they be considered to be different?

Response from Hawken:
> If they're behaving differently, that's a bug in React Navigation. I guess I used addListener for willBlur. https://github.com/react-navigation/react-navigation/issues/4637 is our bug, I think, but it's closed as WorkingAsIntended. The report says "willBlur is not always fired" and he says "user releases gesture, then willBlur fires, then didBlur fires".

> Here's a potentially easy fix: investigate if we can do screens that come in from the bottom. Because then you can't swipe back. I'd like to switch Settings and Filter back to modal, from-bottom transitions, if possible.

hannesmcman:
The goal of this PR is to make it easier to add filters in both the course search and menu views by allowing users to edit their filters right in the toolbar above the ListView of results. 

View | Toggle | List | Picker
-- | -- | -- | --
Course Search | screen shot 2018-08-06 at 10 14 19 pm | screen shot 2018-08-06 at 11 20 17 pm | N/A
Menus | screen shot 2018-08-07 at 12 34 33 am | screen shot 2018-08-07 at 12 34 52 am | screen shot 2018-08-07 at 12 34 42 am

Work remaining:
- [x]  Add another toolbar right below FilterToolbar with all the active filters in it, so that they can be easily disabled
- [x] ~~Remove FilterView component entirely (?)~~

Closes #2738


drewvolz:
> Add another toolbar right below FilterToolbar with all the active filters in it, so that they can be easily disabled

If there is some way we can prevent a separate toolbar, I would be all for it. I was thinking we could explore interactions? The filter buttons are buttons. If tapped on, it opens the modal view. If pressed on, perhaps that could enable or disable the entire button?

hannesmcman:
@drewvolz I was thinking something along these lines:



Is that what you were thinking or are you talking about using force touch on the toolbar buttons?

drewvolz:
I was suggesting a long press, but now that I see your example I think I’m okay with the second row.

hannesmcman:
Active Filter Toolbar added...updated Screenshots:

View | One | Two | Three
-- | -- | -- | -- 
Course Search | screen shot 2018-08-08 at 1 58 46 am | screen shot 2018-08-08 at 1 56 23 am | screen shot 2018-08-08 at 1 56 43 am
Menus | screen shot 2018-08-08 at 1 57 04 am | screen shot 2018-08-08 at 1 57 15 am |screen shot 2018-08-08 at 2 02 01 am

source/lib/course-search/types.js
```diff
@@ -6,13 +6,16 @@ export type CourseType = {
...
+	enroll: number,
 	gereqs?: string[],
 	instructors: string[],
 	level: number,
 	locations?: string[],
+	max: number,
 	name: string,
 	notes?: string[],
 	number: number,
+	open: boolean,
```
hawkrives: "Open" is different from "full", though… `status: open` is open (vs closed or cancelled); being "full" is different from being "closed".

A professor can let more people into a "full" class, for instance. (Honestly, I don't quite know what `status: closed` _means_, but it's different from being "full".)

Would you change this to `full` or something, please?

source/lib/course-search/update-course-storage.js
```diff
@@ -59,5 +59,15 @@ async function loadCurrentTermsFromServer(): Promise<Array<TermType>> {
...
+	const formattedTermData = formatRawData(resp)
+	storage.setTermCourseData(term.term, formattedTermData)
+}
+
+function formatRawData(rawData: Array<CourseType>): Array<CourseType> {
+	return rawData.map(course => {
+		const open = course.enroll < course.max
+		let formattedCourse = course
+		course.open = open
+		return formattedCourse
```
hawkrives:
```js
let isFull = course.enroll < course.max
return {full: isFull, ...course}
```

Otherwise, you're just essentially grabbing a pointer to an object and mutating it.

Speaking of, we should make a second `type` – RawCourseType or something – to separate what we add from what the data provides.

source/views/components/filter/active-filter-button.js
```diff
@@ -0,0 +1,56 @@
...
+		ios: 'ios-close-circle',
+		android: 'md-close-circle',
+	})
+	return (
+		<View style={[style, styles.badge]}>
+			<Text style={styles.text}>{label}</Text>
+			<TouchableWithoutFeedback
+				onPress={() => {
+					onRemove(filter)
+				}}
```
hawkrives: Collapse this down, please? `onPress={() => onRemove(filter)}`

source/views/components/filter/filter-popover.js
```diff
@@ -0,0 +1,65 @@
...
+		const {filter} = this.state
+		const {anchor, onClosePopover, visible} = this.props
+		return (
+			<Popover
+				arrowStyle={arrowStyle}
+				fromView={anchor}
+				isVisible={visible}
+				onClose={() => {
+					onClosePopover(filter)
+				}}
```
hawkrives: Same condensed arrow function request here, please.

source/views/components/filter/filter-popover.js
```diff
@@ -0,0 +1,65 @@
...
+				placement="bottom"
+			>
+				<View style={styles.popoverContainer}>
+					<FilterSection filter={filter} onChange={this.onFilterChanged} />
```
hawkrives: Can we make it so we can pass `style` to FilterSection instead of wrapping it in a view?

source/views/components/filter/filter-popover.js
```diff
@@ -0,0 +1,65 @@
+//@flow
+import * as React from 'react'
+import {StyleSheet, View} from 'react-native'
+import Popover from 'react-native-popover-view'
+import {FilterSection} from './section'
+import type {FilterType} from './types'
+import * as c from '../colors'
+
+type Props = {
+	anchor: any,
```
hawkrives: What type is `anchor`?

source/views/components/filter/filter-popover.js
```diff
@@ -0,0 +1,65 @@
...
+const styles = StyleSheet.create({
+	arrow: {
+		backgroundColor: c.iosLightBackground,
+	},
+	popoverContainer: {
+		maxWidth: 300,
+	},
+})
+
+const arrowStyle = StyleSheet.flatten(styles.arrow)
```
hawkrives: Why do you need to flatten the style?

source/views/components/filter/filter-toolbar-button.js
```diff
@@ -0,0 +1,130 @@
...
+	onClosePopover = (filter: FilterType) => {
+		this.props.onPopoverDismiss(filter)
+		this.setState(() => ({popoverVisible: false}))
+	}
+
+	render() {
+		const {filter, iconName, isActive, style, title} = this.props
+		const {popoverVisible} = this.state
+		let icon
+		if (!iconName) {
```
hawkrives: When wouldn't we have an icon here? (This question may be answered later in this review.)

source/views/components/filter/filter-toolbar-button.js
```diff
@@ -0,0 +1,130 @@
...
+		}
+
+		if (filter.type === 'list') {
+			if (!filter.spec.options.length) {
+				return null
+			}
+		}
+
+		return (
+			<View>
```
hawkrives: Use `<React.Fragment>` instead of `<View>`, if you're just using it to return multiple children. A fragment doesn't actually get rendered to anything, so it's faster to render.

source/views/menus/components/filter-menu-toolbar.js
```diff
@@ -13,40 +14,54 @@ const styles = StyleSheet.create({
...
 		flexDirection: 'row',
+		alignItems: 'center',
 	},
 })
 
 type PropsType = {
 	date: momentT,
 	title?: string,
-	onPress: () => any,
+	onPopoverDismiss: (filter: FilterType) => any,
```
hawkrives, line 24: Why are we only returning the changed filter (singular) here? Why not return all of the filters? Wouldn't that simplify usage when we use this component (being able to just replace our `state.filters` instead of update it)?

source/views/menus/components/filter-menu-toolbar.js
```diff
@@ -13,40 +14,54 @@ const styles = StyleSheet.create({
...
-				</Text>
-			</View>
-
-			<ToolbarButton
-				iconName={Platform.OS === 'ios' ? 'ios-funnel' : 'md-funnel'}
-				isActive={isFiltered}
-				title={
-					isFiltered ? `${appliedFilterCount} ${filterWord}` : 'No Filters'
-				}
+		<View>
```
hawkrives: Same `<React.Fragment>` comment here.

source/views/menus/dev-bonapp-picker.js
```diff
@@ -1,73 +0,0 @@
-// @flow
```
hawkrives: Why?

hawkrives (dismissed):
Questions have been asked and changes have been requested.

Overall, amazing work!

source/lib/course-search/update-course-storage.js
hannesmcman: I believe you meant 
```javascript
let isFull = course.enroll >= course.max
```

hannesmcman: However...I think it makes more sense that when the user toggles this filter "on", that they will now only see classes with space available. Thus, having isFull as the key for the ToggleFilter would be counterintuitive. Could we just change `open` to `spaceAvailable`?

hawkrives: That sounds fine to me. We should also think about exposing both "status: open" and "hasSpaceAvailable" in the filter, as they do represent different states.

hannesmcman: As it stands, they each have their own filter. "Enrollment" corresponds to `spaceAvailable` and "Status" corresponds to `status`. I can't think of a better name for Status, but I feel like there should be one.

source/views/components/filter/filter-popover.js
hannesmcman: I suppose...but then we'd have to wrap all three types of `FilterSection`s with another View, or pass the style prop all the way down to them, which seems like a lot. But if that's best practice, I guess we can do that. It's just more code.

source/lib/course-search/update-course-storage.js
hawkrives: I think we should leave Status as Status, for consistency with SIS if nothing else. 

I didn't realize there was already a separate filter for status, my apologies.

source/views/components/filter/filter-popover.js
hannesmcman: Weird bug in the `Popover` component...This is what we get without flattening:
screen shot 2018-08-08 at 9 57 39 pm

source/views/components/filter/filter-toolbar-button.js
hannesmcman: Yeah I think I'll just remove `iconName` from the list of props, since it will always be the down arrow.

source/views/components/filter/filter-toolbar-button.js
hannesmcman: Good to know!

source/views/menus/dev-bonapp-picker.js
hannesmcman: Wasn't being used anywhere...should we keep it?

drewvolz:
@hannesmcman amazing work! I have a question about UI. Is it possible to prevent the new horizontal list of filters and buttons from collapsing when scrolling the list? I'm finding myself wanting to always see what's been applied and scrolling back up to the top is cumbersome.

Edit: now that I think about it, I find my request nice, but not best practice. And from looking at examples like Houzz and Mail.app, it's standard to hide them when scrolled. So what you've done is fine 👍

drewvolz:
An additional bit of feedback I've received on the Department and GE dropdowns is that we should include the full titles... however we would want to implement that. That is probably out of the scope of this PR.

source/views/menus/components/filter-menu-toolbar.js
hannesmcman: We're not *returning* only the changed filter, we're *passing* it upstream through its ancestor components until it eventually gets to the component that holds the filters in its state (`FancyMenu`, in this case). The reason I did it this way is so that I could pass the same `onPopoverDismiss` prop to the `FilterToolbarButton` and the `FilterToolbar`, within the `FilterMenuToolbar` component. I guess what I'm trying to say is that the individual toolbar buttons should only be aware of the filter they control, while their parents should know about all the other filters and take care of things when one of their children (toolbar buttons) informs them that they have changed. Idk does that make sense?

hannesmcman:
@drewvolz It is possible, but it takes up quite a bit of real estate, since we can't put the search bar in the navbar...
filter-demo

~I suppose we could have the searchbar disappear on scroll though...let me see what that would look like~

On second thought, that would be pretty difficult actually, since the `<CourseResultsList/>` is below the `<AnimatedSearchBox/>`. Might be the subject of another PR..

source/views/components/filter/filter-popover.js
hawkrives: Hmm. We should make a PR to the popover repo, then.

source/views/menus/components/filter-menu-toolbar.js
hawkrives: Ah, yeah. Sure.

source/views/menus/dev-bonapp-picker.js
hawkrives: Please? :D

If I can ever figure out how to dynamically show tab bar buttons, it'd be enabled on dev builds.

hawkrives:
@drewvolz / @hannesmcman - I think the best thing for having the full names of departments and GEs would be to have a manually updated file in the course-data repo that maps from abbreviation-to-full-name. 

I have a manually updated version of this file in Gobbldygook, but I think we should move it into course-data and fetch+cache it. I can make it available on GH-pages… ah, and I may need to edit how info.json works. I'll have to think about that. 

I do think that such a change can come as a separate PR, since this one is just about moving to the toolbar; the contents of the filters I'd like to keep to minimal changes for the sake of reviewing.

source/views/components/filter/filter-popover.js
hannesmcman: Will do!

hannesmcman:
Ok I noticed another weird quirk in the react-native-popover-view repo, where the popover arrow is positioned in the middle of the button that triggers it on android. I think this is a subject of a PR to that repo though, rather than a blocker on this one.
screen shot 2018-08-09 at 9 14 32 am

hawkrives:
I'll review this again tonight

source/views/components/filter/filter-popover.js
hawkrives: Ah. Okay. I've looked at what they want… so. 

They don't actually want a "Style" here; they want an object describing the style. 

So let's just not create the arrow style in StyleSheet.create at all – just do, like, `const arrowStyle = {backgroundColor: c.iosLightBackground}`.

source/views/components/filter/filter-toolbar-button.js
```diff
@@ -0,0 +1,125 @@
...
+
+		return (
+			<React.Fragment>
+				<TouchableOpacity
+					ref={ref => (this.touchable = ref)}
+					onPress={this.openPopover}
+					style={[buttonStyles.button, activeButtonStyle, style]}
+				>
+					<Text
+						style={[
```
hawkrives: Would you mind pulling this style block out into a variable, to save on vertical lines, please?

source/views/components/filter/filter-toolbar-button.js
```diff
@@ -0,0 +1,125 @@
...
+							activeContentStyle,
+							textWithIconStyle,
+							activeTextStyle,
+							buttonStyles.text,
+						]}
+					>
+						{title}
+					</Text>
+					{icon ? (
```
hawkrives: I think we can remove the ternary here now, since the icon will always exist (on iOS and Android, at least.)

source/views/components/filter/filter-toolbar.js
```diff
@@ -0,0 +1,100 @@
...
+				/>
+			))
+		}
+		return null
+	})
+	const activeFilterButtons = flatten(allButtons)
+	const anyFiltersEnabled = filters.some(f => f.enabled)
+
+	return (
+		<View>
```
hawkrives: `<React.Fragment>`

source/views/components/filter/filter-toolbar.js
```diff
@@ -0,0 +1,100 @@
...
+			<ScrollView horizontal={true} showsHorizontalScrollIndicator={false}>
+				<Toolbar>{filterToggles}</Toolbar>
+			</ScrollView>
+			{anyFiltersEnabled && (
+				<ScrollView horizontal={true} showsHorizontalScrollIndicator={false}>
```
hawkrives: Can we put `paddingLeft` as `contentContainerStyle` on this ScrollView, instead of doing styles.first? If contentContainerStyle doesn't work, try `style`, please, but I think we want contentContainerStyle.

hawkrives: Update: Try the same thing on the filterToggles scrollview, for the same reason?

source/views/menus/components/fancy-menu.js
```diff
@@ -200,6 +199,7 @@ export class FancyMenu extends React.Component<Props, State> {
 				renderSectionHeader={this.renderSectionHeader}
 				sections={(groupedMenuData: any)}
 				style={styles.inner}
+				windowSize={5}
```
hawkrives, line 202: Do we need to change this prop on Menu? Or is this just, like, to be careful? I'd like to only override it when we actually feel like we need to…

source/views/menus/components/filter-menu-toolbar.js
```diff
@@ -13,40 +14,51 @@ const styles = StyleSheet.create({
...
+	filters,
+	onPopoverDismiss,
+}: PropsType) {
+	const mealFilter = filters.find(f => f.type === 'picker')
+	const multipleMeals =
+		mealFilter && mealFilter.type === 'picker'
+			? mealFilter.spec.options.length > 1
+			: false
+	const nonPickerFilters = filters.filter(f => f.type !== 'picker')
 	return (
```
hawkrives: Add a blank line above `return`? I tend to add a blank line before `return` if there's more than one statement before it.

i.e.

```js
let x = 5
return x
```

```js
let x = 5
let y = 6 

return x * y
```

Would you mind applying that across this PR? I wonder if there's a lint for that… (looks… wow. https://eslint.org/docs/rules/padding-line-between-statements that's not … quite what I was looking for. that's a lot of power.)

Anyway. Would you mind applying that across this PR? 😸

hawkrives (changes requested):
This is probably my last round of requests! Good work!

source/views/components/filter/filter-toolbar-button.js
```diff
@@ -0,0 +1,125 @@
...
+
+		if (filter.type === 'list') {
+			if (!filter.spec.options.length) {
+				return null
+			}
+		}
+
+		return (
+			<React.Fragment>
+				<TouchableOpacity
```
hawkrives: Only more question: why is this a TouchableOpacity instead of a Touchable? Touchable automatically switches to use "ripples" on android, which is usually nice, and is why I'm asking.

source/views/components/filter/filter-toolbar-button.js
hannesmcman: Sure, but won't we still have the same number of vertical lines if we move it into a variable?
```javascript
const buttonTextStyle = [
	activeContentStyle,
	textWithIconStyle,
	activeTextStyle,
	buttonStyles.text,
]
```

source/views/components/filter/filter-toolbar.js
hannesmcman: Adding paddingLeft to the ScrollViews makes the buttons cut off when you scroll to the right, which is not what we want. This is why I opted for setting the marginLeft on the first button. Lmk if you want a screenshot

source/views/menus/components/fancy-menu.js
hannesmcman: I did it because it prevents a noticeable lag when you turn off the "Specials Only" filter. It seemed to have a similar issue to the Course Search ListView where it was rendering a ton of rows right away...

source/views/components/filter/filter-toolbar-button.js
hannesmcman: ~no reason!~

Actually, when I tried switching it to Touchable, I got this warning:
screen shot 2018-08-10 at 12 51 53 am

The popovers no longer were positioned correctly as a result

source/views/menus/components/filter-menu-toolbar.js
hannesmcman: sure!

source/views/menus/components/fancy-menu.js
hawkrives: Kk

hawkrives: Kk

source/views/components/filter/filter-toolbar.js
hawkrives: Did you try both contentContainerStyle and style? Using padding on of those should scroll the padding off to the side… the other one won't scroll the padding.

source/views/components/filter/filter-toolbar-button.js
hawkrives: For the style, yeah, but it'll let Prettier collapse the `<Text style={}>` into one like instead of four, so we should still save on lines.

source/views/components/filter/filter-toolbar-button.js
hawkrives: Yeah, sorry. You won't be able to just switch it for a ref= on Touchable; you'll have to forward the ref through Touchable to TouchableOpacity/Highlight/NativeFeedback (when you see innerRef= or getRef=, it's forwarding the ref.)

I can give you an example if you would like.

hawkrives: Or, we can file this under "things Hawken should look at after it's merged" because I just realized that I don't know how I want you to handle typing that three-way intersection.

source/views/components/filter/filter-toolbar.js
hawkrives: I've tested contentContainerStyle with paddingLeft, and it seems to do what we want.

<details><summary>screenshot</summary>

image
</details>

<details><summary>patch</summary>

```patch
diff --git a/source/views/components/filter/filter-toolbar.js b/source/views/components/filter/filter-toolbar.js
index 83580e77..5ba239c5 100644
--- a/source/views/components/filter/filter-toolbar.js
+++ b/source/views/components/filter/filter-toolbar.js
@@ -41,18 +41,17 @@ export function FilterToolbar({filters, onPopoverDismiss}: Props) {
 		onPopoverDismiss(newFilter)
 	}
 
-	const filterToggles = filters.map((filter, index) => (
+	const filterToggles = filters.map(filter => (
 		<FilterToolbarButton
 			key={filter.spec.title}
 			filter={filter}
 			isActive={filter.enabled}
 			onPopoverDismiss={onPopoverDismiss}
-			style={index === 0 ? styles.first : null}
 			title={filter.spec.title}
 		/>
 	))
 
-	const allButtons = filters.filter(f => f.enabled).map((filter, i) => {
+	const allButtons = filters.filter(f => f.enabled).map(filter => {
 		if (filter.type === 'toggle') {
 			return (
 				<ActiveFilterButton
@@ -60,17 +59,15 @@ export function FilterToolbar({filters, onPopoverDismiss}: Props) {
 					filter={filter}
 					label={filter.spec.label}
 					onRemove={filter => updateFilter(filter)}
-					style={i === 0 ? styles.first : null}
 				/>
 			)
 		} else if (filter.type === 'list') {
-			return filter.spec.selected.map((selected, j) => (
+			return filter.spec.selected.map(selected => (
 				<ActiveFilterButton
 					key={selected.title}
 					filter={filter}
 					label={selected.label || selected.title.toString()}
 					onRemove={filter => updateFilter(filter, selected)}
-					style={i === 0 && j === 0 ? styles.first : null}
 				/>
 			))
 		}
@@ -81,11 +78,19 @@ export function FilterToolbar({filters, onPopoverDismiss}: Props) {
 
 	return (
 		<React.Fragment>
-			<ScrollView horizontal={true} showsHorizontalScrollIndicator={false}>
+			<ScrollView
+				contentContainerStyle={styles.scroller}
+				horizontal={true}
+				showsHorizontalScrollIndicator={false}
+			>
 				<Toolbar>{filterToggles}</Toolbar>
 			</ScrollView>
 			{anyFiltersEnabled && (
-				<ScrollView horizontal={true} showsHorizontalScrollIndicator={false}>
+				<ScrollView
+					contentContainerStyle={styles.scroller}
+					horizontal={true}
+					showsHorizontalScrollIndicator={false}
+				>
 					{activeFilterButtons}
 				</ScrollView>
 			)}
@@ -94,7 +99,7 @@ export function FilterToolbar({filters, onPopoverDismiss}: Props) {
 }
 
-	first: {
-		marginLeft: 10,
+	scroller: {
+		paddingLeft: 10,
 	},
 })
```
</details>

source/views/components/filter/filter-toolbar-button.js
hawkrives: See 9fdadea173e1d3f872a64004d72454533d1df45f for this.

source/views/components/filter/filter-toolbar.js
hawkrives: or see 6f9ab3356e260293b98d83e951c9460ab5561959, which is the same patch.

source/views/components/filter/filter-popover.js
```diff
@@ -0,0 +1,62 @@
...
+
+		return (
+			<Popover
+				arrowStyle={arrowStyle}
+				fromView={anchor}
+				isVisible={visible}
+				onClose={() => onClosePopover(filter)}
+				placement="bottom"
+			>
+				<View style={styles.popoverContainer}>
```
hawkrives: See 3f274975d9dae8afc0036e22ccacf45582625254 for a way to remove this View

source/views/components/filter/filter-toolbar.js
hannesmcman: Oh nice!

hawkrives (approved):

drewvolz (commented):
:+1: :shipit: great work, @hannesmcman

PR #2746, opened 2018-08-07
Keyboard dismiss props on course search list

drewvolz:
Sometimes the keyboard gets in the way. This allows for us to dismiss
it if we swipe the view.


hawkrives (dismissed):
eslint expects capital letters to come before lowercase ones

drewvolz:
Thanks, I’ll make changes in a bit. I was committing from mobile so I did not get the chance to run a lint check beforehand.

hawkrives (approved):

PR #2748, opened 2018-08-07
Allow past events in list of streams

### LINKED ISSUE #2702: Fetch a continually growing list of streams
@hawkrives this morning:

>AAO is reporting no streams, but the Olaf website has streams. Can someone check the URL that AAO uses and see if it's an AAO problem, or a St. Olaf problem?

After checking, it looks like the feed we use for our listing, [stolaf.edu/multimedia/api/collection](https://www.stolaf.edu/multimedia/api/collection) is fine, but for some reason clients aren't rendering any of the contents of that JSON slug.  Did something change?

drewvolz:
This queries the broadcast media streams for events so that we are able to see past events.

Closes #2702

### DIFF
--- a/source/views/streaming/streams/list.js
+++ b/source/views/streaming/streams/list.js
@@ -71,7 +71,6 @@ export class StreamListView extends React.PureComponent<Props, State> {
 
 	getStreams = async (date: moment = moment.tz(CENTRAL_TZ)) => {
 		try {
-			const dateFrom = date.format('YYYY-MM-DD')
 			const dateTo = date
 				.clone()
 				.add(1, 'month')
@@ -80,8 +79,6 @@ export class StreamListView extends React.PureComponent<Props, State> {
 			let params = {
 				class: 'current',
 				sort: 'ascending',
-				// eslint-disable-next-line camelcase
-				date_from: dateFrom,
 				// eslint-disable-next-line camelcase
 				date_to: dateTo,
 			}


hawkrives:
Dislike. It's inconsistent with our other event lists, and we don't control how far back it goes, and it'll be confusing, I feel, when you open the app and have to scroll down ~1 month's worth of events before you hit today. (Guessing on the month, but really I don't think we can have more than about three events before the Today section without confusing people. Myself included.)

drewvolz:
What is your alternative suggestion?

hawkrives:
Don't show old events?

hawkrives:
Or add another tab for archived events

drewvolz:
The issue was specifically made due to you thinking there were no events. I believe you asked for it to show old events, no?

hawkrives:
I thought we decided that we should just fetch more events? I have no recollection of asking for old events, but I would most likely argue with past!Hawken about why we should not show old events.

drewvolz:
I want to have some way to show past events. I don’t have a problem with them being in the same space as live and upcoming. It feels convenient to be able to watch a stream that happened recently. Are you wanting a tab separator? Perhaps this would be a great candidate to add visual filters to.

hawkrives:
We don't show live events in the tab, either, fwiw (afaik)

drewvolz:
We do show live events. There’s logic to group live events at the top.

hawkrives:
Ok. Regardless, I don't want to show archived events in the same area as upcoming events, and adding a UIToggleBar would just be confusing IMO. If we added a filter to show past events, I'm having a hard time visualizing how that would work, as well, which just makes me even more confused.

drewvolz:
It’s like any view. We have a set of data. Your intentions of visiting the streams may be different than somebody else’s. I’d like to ideally request old, new, live, and present the user with a filter set that allows for them to see their desired date range. I also believe we can filter on categories.

We can give you what you want which is “new” data up front, but I’d like to see filtering actions to let someone view the full range of streams as well as narrow down their search.

hawkrives:
Okay, sure, but that's a very different proposal from this PR, and I'd like to debate its merits in a separate space. 

I move to close this PR. Any seconds? (~~what are you doing hawken~~)

drewvolz:
I don't see a reason to change issues.

rye:
Something something infinite scrolling? I'd like to see this feature be like that. I wonder if there's a way we can do this in a better style.

1. **Near events** would be anything from today to next week. This would be ~7+ events during the school year, but would be a pretty small slug.
1. **Recent events** would be, say, 16 events prior to today.
1. **Future events** would be, say, 16 events after next week.

First, fetch near events, then fetch future events, and then fetch recent events. Render the start of the list at today. As the user scrolls around, once they hit 1–2 screen-fulls away from the end of the list, fetch the next list in the direction, and silently append it so it's ready to scroll onto.

The big thing is that we want this to be seamless. I don't want to incorporate filters unless we also want to start annotating stuff by what it is, (e.g. Chapel, etc.) at which point we'd have yet another filtered ListView in this app.  I think that is out-of-scope for solving the underlying problem here.

rye:
I think it'd be nice to replicate what the website sidebar does. (but better)

hawkrives:
I'm fine with infinite scrolling (what #2702 decided on); I'm not OK with mixing past, current, and future events in the same list view, at least not without figuring out how to start the list at "now". I'd be fine with scrolling up to see past events, like a calendar, but I don't know how to do that in RN.

I'm voting to close this PR, @drewvolz, because I think we've both agreed that we want more substantial changes than just "remove the `start_date` filter".

drewvolz:
> I'd be fine with scrolling up to see past events, like a calendar, but I don't know how to do that in RN.

You tell it a section index to scroll to

drewvolz:
If this is closed, we need to move over the salient parts of discussion into another issue, like Kris’ refined infinite scrolling idea.

rye:
How about the following sections: "This week", "Future", and "Past". Scroll to "this week" by default. WRT infinite-ish scrolling, if we can put things in a section-less section ahead of "This Week", then we don't need a section for "Past events", just "Future" and "This week."

hawkrives:
Closing, as I have moved @rye's comments to #2702 as per @drewvolz's request.

PR #2755, opened 2018-08-10
Remove old FilterView and references to it

hannesmcman:
Since we have no plans to use this anymore, might as well get rid of it!


hawkrives (approved):
😿

PR #2774, opened 2018-08-11
Downgrade google-api-client

rye:
This may resolve spurious Android build/deploy failures. We'll see.


rye:
Irrelevant in light of fastlane/fastlane#13076. Closing.

PR #2779, opened 2018-08-13
Remove xml2js

hawkrives:
We don't need it anymore*

*: since … er … um … ever? The code in lib/stalkernet code has never been used. Ever.

Update: I removed the course parsing removal; it will be in a new PR.


hawkrives:
Oh, looks like I get to remove more code! I'm going to make a separate PR for that.

rye (approved):

PR #2783, opened 2018-08-13
Course Search: Save recent filter combinations on unmount

### LINKED ISSUE #2769: Recent filter combinations are not saving or loading
I am not seeing any filter combos being loaded onto the front page of course search. I am unsure if they’re having difficulty saving, loading, or both.

hannesmcman:
This PR ensures that when a user is in browsing mode and leaves the SIS View, their current filter combination will be saved to their recent filter combinations. Resolves #2769.

### DIFF
--- a/source/views/sis/course-search/search.js
+++ b/source/views/sis/course-search/search.js
@@ -83,6 +83,12 @@ class CourseSearchView extends React.Component<Props, State> {
 		this.resetFilters()
 	}
 
+	componentWillUnmount() {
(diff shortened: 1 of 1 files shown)


hawkrives:
Should this also trigger when you hit the Cancel button? Idk

hawkrives:
Oh, it does that now… I see. Sure. I'm fine with this.

hawkrives (approved):

PR #2787, opened 2018-08-13
Wrap <FilterPopover> in <Viewport>

### LINKED ISSUE #2773: Filter popovers do not resize on rotation
8fc23ce8-36d3-43a4-b235-480bcb23d380

Can we fix this the way we’ve fixed other rotation views?

hannesmcman:
This PR is a simple solution to #2773, forcing the `<FilterPopover>` component to re-render when the screen rotates. Resolves #2773.

### DIFF
--- a/source/views/components/filter/filter-popover.js
+++ b/source/views/components/filter/filter-popover.js
@@ -39,7 +39,7 @@ export class FilterPopover extends React.PureComponent<Props, State> {
 
 		return (
 			<Viewport
-				render={viewport => (
+				render={() => (
 					<Popover
 						arrowStyle={arrowStyle}
 						fromView={anchor}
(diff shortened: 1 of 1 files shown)


hawkrives:
Have you looked into fixing this upstream, or filing an issue upstream?

hannesmcman:
@hawkrives I looked into adding the event listener to the Popover component itself unsuccessfully, but I suppose we could propose wrapping the component in a Viewport upstream?

hawkrives (approved):
Eh sure, why not. There's only going to be one open at a time, anyway.

hawkrives:
We should still probably file an issue upstream, so they know about it, but I'm good with this approach for us.

PR #2790, opened 2018-08-14
Push course search results onto the stack

hawkrives:
Yay behavioral changes 😭😆

Instead of transmogrifying the search overview into the search results, we'll just push the results view onto the stack. This better matches how searching works in YouTube, for instance, and makes logical sense to me since we don't do search-as-you-type.

We still hide the home's search box title when typing, and we show it again when you return from the results, since returning to the overview essentially resets the screen.

Tapping a search or filter will push the results view with the tapped item as the search or filter. 

We'll always show a Search icon on Android, which isn't quite platform-native, but since our search bar is separate from the titlebar the "back" button doesn't make sense IMO.

We now determine when to save a "recent" item in a nifty way: what's the best way to tell that someone found a query useful? IMO, it's that they interacted with the list, so we'll now only save the thing to the "recents" list if the person has tapped on a result row.

> IOW, to save a recent search, search for something, then tap on a row.
> To save a recent filter, browse for something, then tap on a row.

Bugfixes: 
- If the view is pushed while filters are loading, there will be a spinner to indicate that content will appear. This is impossible to trigger on master, because the filters will have loaded by the time you can search for something, but here they can be missing for up to ~1s.
- Dismissing the keyboard will now work based on your finger's position and won't dismiss by dragging on android, to better match Android conventions.
- Dismissing the keyboard on the home screen is now possible by dragging
- Course data loading now only makes network requests after the "Download" button is pressed (again; I broke it in #2737)

I haven't attached any screenshots because there really weren't any visual changes, aside from the Android back-to-search icon change.

### DIFF
--- a/source/navigation.js
+++ b/source/navigation.js
@@ -24,9 +24,11 @@ import {
 	CarletonSaylesMenuScreen,
 } from './views/menus'
 import NewsView from './views/news'
-import SISView from './views/sis'
-import {JobDetailView} from './views/sis/student-work/detail'
-import {CourseDetailView} from './views/sis/course-search/detail'
+import SISView, {
+	JobDetailView,
+	CourseDetailView,
+	CourseSearchResultsView,
+} from './views/sis'
 import {
 	BuildingHoursView,
 	BuildingHoursDetailView,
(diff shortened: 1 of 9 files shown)


drewvolz (approved):
👍 :shipit: Tested on iOS and Android. I really think the flow of pushing a view onto the stack for the search feels natural.

hannesmcman (approved):
Yeah it looks good to me! Nice work!

hawkrives:
@hannesmcman Thanks!

PR #2793, opened 2018-08-15
Switch Streams endpoint to ccc-server

drewvolz:
Uses a controlled API so that we have flexibility over the streams data.

Changes to the server are in review, but I thought I'd get this going in parallel.

Server PR: https://github.com/frog-pond/ccc-server/pull/64

### DIFF
--- a/source/views/streaming/streams/list.js
+++ b/source/views/streaming/streams/list.js
@@ -90,7 +90,9 @@ export class StreamListView extends React.PureComponent<Props, State> {
 				.map(stream => {
 					const date = moment(stream.starttime)
 					const group =
-						stream.status !== 'live' ? date.format('dddd, MMMM Do') : 'Live'
+						stream.status.toLowerCase() !== 'live'
+							? date.format('dddd, MMMM Do')
+							: 'Live'
 
 					return {
 						...stream,
--- a/source/views/streaming/streams/list.js
+++ b/source/views/streaming/streams/list.js
@@ -78,10 +78,8 @@ export class StreamListView extends React.PureComponent<Props, State> {
 
 			let params = {
 				sort: 'ascending',
-				// eslint-disable-next-line camelcase
-				date_from: dateFrom,
-				// eslint-disable-next-line camelcase
-				date_to: dateTo,
+				dateFrom,
+				dateTo,
 			}
 
 			const data = await fetchJson(API('/streams/upcoming', params))
@@ -90,7 +88,7 @@ export class StreamListView extends React.PureComponent<Props, State> {
 			const processed = data
 				.filter(stream => stream.category !== 'athletics')
 				.map(stream => {
-					const date = moment(stream.starttime, 'YYYY-MM-DD HH:mm')
+					const date = moment(stream.starttime)
 					const group =
 						stream.status !== 'live' ? date.format('dddd, MMMM Do') : 'Live'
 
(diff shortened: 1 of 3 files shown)


source/views/streaming/streams/list.js
```diff
@@ -74,31 +73,26 @@ export class StreamListView extends React.PureComponent<Props, State> {
...
-			const processed = streams
+			const processed = data
 				.filter(stream => stream.category !== 'athletics')
 				.map(stream => {
 					const date = moment(stream.starttime, 'YYYY-MM-DD HH:mm')
 					const group =
-						stream.status.toLowerCase() !== 'live'
-							? date.format('dddd, MMMM Do')
-							: 'Live'
+						stream.status !== 'live' ? date.format('dddd, MMMM Do') : 'Live'
```
hawkrives: keep the toLowerCase

source/views/streaming/streams/list.js
```diff
@@ -74,31 +73,26 @@ export class StreamListView extends React.PureComponent<Props, State> {
...
-			const data = await fetchJson(streamsAPI)
-			const streams = data.results
+			const data = await fetchJson(API('/streams/upcoming', params))
 
 			// force title-case on the stream types, to prevent not-actually-duplicate headings
-			const processed = streams
+			const processed = data
 				.filter(stream => stream.category !== 'athletics')
 				.map(stream => {
 					const date = moment(stream.starttime, 'YYYY-MM-DD HH:mm')
```
hawkrives: remove the custom parse format

source/views/streaming/streams/list.js
```diff
@@ -74,31 +73,26 @@ export class StreamListView extends React.PureComponent<Props, State> {
...
 				.clone()
-				.add(1, 'month')
+				.add(2, 'month')
 				.format('YYYY-MM-DD')
 
 			let params = {
-				class: 'current',
 				sort: 'ascending',
 				// eslint-disable-next-line camelcase
 				date_from: dateFrom,
```
hawkrives: let's make the api accept camelCase fields here so we can remove the comments

source/views/streaming/streams/row.js
```diff
@@ -30,9 +31,10 @@ function Info({item}: {item: StreamType}) {
 }
 
 function Time({item}: {item: StreamType}) {
+	const streamDate = moment(item.date)
 	const showTime = item.status !== 'archived'
 	return showTime ? (
-		<Detail>{item.date.format('h:mm A – ddd, MMM. Do, YYYY')}</Detail>
+		<Detail>{streamDate.format('h:mm A – ddd, MMM. Do, YYYY')}</Detail>
```
hawkrives: `moment(item.date).format`

hawkrives (dismissed):

source/views/streaming/streams/list.js
drewvolz: There's no reason to if the API sticks to the spec of lowercase

drewvolz: But we'll do it just to be safe.

hawkrives: You're right. We should not do this in the app. This sort of consistency check is what the server is for. 

You can remove the toLlwerCase again… sorry.

hawkrives (approved):

PR #2796, opened 2018-08-15
Travis webhook for builds

drewvolz:
Adds a webhook for travis builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,7 +78,7 @@ cache:
 notifications:
   webhooks:
     urls:
-      - $TRAVIS_WEBHOOK
+      - https://discordapp.com/api/webhooks/477651404085919747/DezH4DOZQOm2wOFHY_BjYDlOCGmS5lDsasu-P4URfqVWvzb5CLCXV9_lKiM-bu3KP_VF
     on_success: change
     on_failure: always
     on_start: change
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,3 +74,6 @@ jobs:
 
 cache:
   yarn: true
+
+notifications:
+  webhooks: $TRAVIS_WEBHOOK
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -74,3 +74,13 @@ jobs:
 
 cache:
   yarn: true
+
+notifications:
+  webhooks:
+    urls:
+      - $TRAVIS_WEBHOOK
```
hawkrives: Um. What is this.

hawkrives (dismissed):

.travis.yml
rye: I'd guess a variable called `TRAVIS_WEBHOOK`.

Probably defined in the Travis CI web configuration.

drewvolz: This is most definitely problematic. Is there a way we can make a variable from the URL without revealing it?

> Note: These webhooks are executed at the end of a build, and not by individual jobs (see builds vs jobs). This means that environment variables from the build are not available in this section.

hawkrives: Why don't we reveal it? We can destroy and recreate the webhook in Discord very easily

drewvolz: Perfect. I was concerned about privacy but I suppose there's no reason for that.

drewvolz:
~Fixed webhook to not be a url~
Fixed webhook to not be a variable

rye (approved):
If we're sure there's no way to get this webhook out into a variable, LGTM.

hawkrives:
I mean, it's still not working, since there's nothing showing up in Circle, so it could go back into a variable.

drewvolz:
The docs mentioned it not being able to see env vars at this point.

hawkrives:
Well. I'd hold off on this still; we should be able to get it to notify Discord from a PR.

drewvolz:
Close it?

hawkrives:
If you like

PR #2798, opened 2018-08-16
Use circle's workspaces to move data between jobs

hawkrives:
… instead of caching and restoring.

This takes, on average, longer to persist, but much less time to attach (from ~10s to 2s), and we don't need to run `yarn install` on each job, which saves another ~6-10s.

Workspaces only exist within a given workflow, so we don't need to worry about cache pollution, thankfully.


rye (changes requested):
Any particular reason why you are still using `restore-cache-yarn` on iOS builds instead of the same as all other jobs?

Otherwise RSLGTM. I haven't seen it in practice yet but I trust a green checkmark means good to go.

hawkrives:
I don't think I can mount a workspace across OSs… I haven't tried, though. Let's find out.

hawkrives:
Actually, yeah. I'm not going to mount between the platforms, even if it's possible, because Mac and Linux don't share native dependencies (anything compiled, ie `fsevents`).

rye (approved):

PR #2799, opened 2018-08-16
Course Search: Scroll to top of results list on refresh

### LINKED ISSUE #2345: Course search vertical scroll bar retains location
I’d expect the scroll position to reset to the top of the search results upon a new search. It currently retains its previous location. I think this can cause confusion.

hannesmcman:
Resolves #2345.

### DIFF
--- a/source/views/sis/course-search/list.js
+++ b/source/views/sis/course-search/list.js
@@ -38,7 +38,7 @@ type Props = TopLevelViewPropsType & {
 	style?: any,
 	contentContainerStyle?: any,
 	filtersLoaded: boolean,
-	setRef: (ref: any) => any,
+	getRef: (ref: any) => any,
 }
 
 function doSearch(args: {
@@ -88,7 +88,7 @@ export class CourseResultsList extends React.PureComponent<Props> {
 			contentContainerStyle,
 			style,
 			filtersLoaded,
-			setRef,
+			getRef,
 		} = this.props
 
 		// be sure to lowercase the query before calling doSearch, so that the memoization
@@ -114,7 +114,7 @@ export class CourseResultsList extends React.PureComponent<Props> {
 
 		return (
 			<SectionList
-				ref={setRef}
+				ref={getRef}
 				ItemSeparatorComponent={ListSeparator}
 				ListEmptyComponent={messageView}
 				ListHeaderComponent={header}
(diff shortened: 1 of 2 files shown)


source/views/sis/course-search/results.js
```diff
@@ -83,8 +85,19 @@ class CourseSearchResultsView extends React.Component<Props, State> {
 		}
 	}
 
+	scrollResultsToTop = () => {
+		this.resultsList.scrollToLocation({
+			animated: false,
+			itemIndex: 0,
+			sectionIndex: 0,
+			viewOffset: Platform.OS === 'ios' ? 87 : 100,
```
drewvolz, line 93: Are these rotationally dependent magic numbers?

hannesmcman: @drewvolz They shouldn’t be...It’s just the height of the filter toolbar, which is a little bigger on android, but shouldn’t change based on orientation..but I haven’t tested it so I can’t say for sure.

hawkrives: What if we just scrolled all the way to `0`?

rye (dismissed):

source/views/sis/course-search/results.js
hannesmcman: @hawkrives You mean without the viewOffset? Then it cuts off the FilterToolbar. I can post a screenshot if you want.

hawkrives: Please 😄

hannesmcman: With | Without
-- | --
with | without

@hawkrives

hannesmcman: @drewvolz I have tested it now and they are not rotationally dependent numbers, since the height of the toolbar is the same in portrait and landscape.

drewvolz (dismissed):

source/views/sis/course-search/results.js
hawkrives: Wait, so using this value makes us actually scroll to above the filterbar?

hannesmcman: @hawkrives It makes us scroll to the position where the padding above and below the filterbar is equal.

source/views/sis/course-search/list.js
```diff
@@ -38,6 +38,7 @@ type Props = TopLevelViewPropsType & {
 	style?: any,
 	contentContainerStyle?: any,
 	filtersLoaded: boolean,
+	setRef: (ref: any) => any,
```
hawkrives: Ok, just one nitpick: name this `getRef`, please.

hawkrives (dismissed):
Okay, just one nitpick. I'm fine with merging this.

drewvolz (approved):

hawkrives:
Thanks!

PR #2802, opened 2018-08-16
Centralize some pieces of the circle config

hawkrives:
- move the `docker: ...` pieces into a `&platform-js` reference
- move the iOS and Android platform configurations into `&platform-ios` and `&platform-android` references
- add a new iOS job to cache yarn and bundler separately from building the app
- also converts iOS to use a workspace – we'll see if it is arch-separated or not

Well… let's see if it works!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,6 +24,36 @@ x-config:
       paths: [./*]
     - &attach-workspace-node_modules
       at: ./node_modules
+    - &persist-workspace-bundler
+      root: ./vendor/bundle
+      paths: [./*]
+    - &attach-workspace-bundler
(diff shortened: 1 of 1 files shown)


hawkrives:
```
a abab/lib/btoa.js
tar: Failed to open 'types'
tar: Failed to open 'turf'
tar: Failed to open 'octokit'
tar: Failed to open 'mapbox'
tar: Failed to open 'hawkrives'
tar: Failed to open 'callstack'
tar: Failed to open 'babel'
a .yarn-integrity
```

Welp.

I'll open another PR sometime with the reorganization bits of this one, but without the iOS job splitting.

PR #2810, opened 2018-08-17
Add Detox for e2e testing

hawkrives:
Detox is an end-to-end testing platform thingy.

It can do a lot ([actions](https://wix.github.io/detox/docs/APIRef.ActionsOnElement.html#swipedirection-speed-percentage), for instance, like swiping, or tapping, or long-pressing, or shaking; or [performance testing](https://github.com/wix/DetoxInstruments)). There's also an [overview page](https://github.com/wix/detox/blob/master/docs/Introduction.HowDetoxWorks.md) talking about how it works, technically.

---

Things this doesn't do, but could in the future:

- run on CI
- take screenshots
- take videos(!)
- open test-specific screens (think testing the scrolling-textinputs behavior)
- run on Android (should be fairly easy)
- contribute to code coverage (??? idk)
- performance testing / benchmarking (https://github.com/wix/DetoxInstruments)

Things this does do:

- assert that the app launches and shows the homescreen on iOS
- assert that the app can open the Settings screen
- assert that the app can navigate from Home > Settings > Home

---

To run detox, run `yarn detox build`, to build your app and install it, then run `yarn detox test` to actually execute the tests. You won't need to re-run `build` until you change the native code (same as normal, interactive testing).

I think there's a way to tell it what platform to test, but I haven't dug in that far yet.

The test uses the iPhone 7 simulator; I don't know what iOS revision it wants, but mine picked iOS 11.2. If you get odd errors, please say so?

---

The other part of this PR is fixing up a few voiceover/accessibility things; namely giving proper labels to a few buttons. For instance, the "settings" button will now be read as "Open Settings, button" instead of … trying to read a Unicode private-use character. I'm not sure how it would have read that, but this has _got_ to be nicer.

I think that we'll get much better about accessibility by adding e2e tests, judging from my experience today.

---

This is not part of any milestone yet, because I kinda did this on a whim, and I don't know what the reaction will be, so I don't have a target merge window yet.

### DIFF
--- a/e2e/config.json
+++ b/e2e/config.json
@@ -1,0 +1,4 @@
+{
+    "setupTestFrameworkScriptFile": "./init.js",
+    "testEnvironment": "node"
+}
--- a/e2e/config.json
+++ b/e2e/config.json
@@ -1,4 +1,4 @@
 {
-    "setupTestFrameworkScriptFile": "./init.js",
-    "testEnvironment": "node"
+	"setupTestFrameworkScriptFile": "./init.js",
+	"testEnvironment": "node"
 }
--- a/e2e/firstTest.spec.js
+++ b/e2e/firstTest.spec.js
@@ -1,0 +1,21 @@
+/* eslint-env jest */
+/* global element, device, by */
+
+describe('Example', () => {
+	it('should have homescreen', async () => {
+		await expect(element(by.id('screen-homescreen'))).toBeVisible()
+	})
+
+	it('should show settings screen after tap', async () => {
+		await element(by.id('button-open-settings')).tap()
+		await expect(element(by.text('Sign In to St. Olaf'))).toBeVisible()
+	})
+
+	it('should show home screen after tap to exit settings screen', async () => {
+		await device.reloadReactNative()
+		await element(by.id('button-open-settings')).tap()
+		await expect(element(by.id('screen-homescreen'))).toBeNotVisible()
+		await element(by.id('header-back')).tap()
+		await expect(element(by.id('screen-homescreen'))).toBeVisible()
+	})
+})
--- a/source/app.js
+++ b/source/app.js
@@ -30,6 +30,8 @@ function getCurrentRouteName(navigationState: NavigationState): ?string {
 
 type Props = {}
 
+const navigationPersistenceKey = __DEV__ ? null : 'NavState'
+
 export default class App extends React.Component<Props> {
 	componentDidMount() {
 		OneSignal.addEventListener('received', this.onReceived)
@@ -83,7 +85,7 @@ export default class App extends React.Component<Props> {
 			<Provider store={store}>
 				<AppNavigator
 					onNavigationStateChange={this.trackScreenChanges}
-					persistenceKey="NavState"
+					persistenceKey={navigationPersistenceKey}
 				/>
 			</Provider>
 		)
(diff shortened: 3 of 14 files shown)


e2e/config.json
```diff
@@ -0,0 +1,4 @@
+{
+    "setupTestFrameworkScriptFile": "./init.js",
```
rye: Please use two-space indentation instead of four-space per the general rule of thumb with JSON in this repo.

source/app.js
```diff
@@ -30,6 +30,8 @@ function getCurrentRouteName(navigationState: NavigationState): ?string {
 
 type Props = {}
 
+const navigationPersistenceKey = __DEV__ ? null : 'NavState'
```
rye, line 33: This seems out of scope for this PR to me—this fix could get pulled out into its own PR.

rye (changes requested):
All in all, I'm okay with this. I really would just like the four-space indentation resolved. `navigationPersistenceKey` extraction seems out of scope to me, but if that was bundled into this PR I wouldn't not be happy.

rye:
Why `iPhone 7`, specifically? How much is required for Android to join along? I could easily see scaling these tests out.

Also, is there any way to strip out `testID`s from the resulting build? Or are those benign?

e2e/config.json
hawkrives: Fixed. (by using tabs, which are only not used in package.json because npm/yarn reformat it every time they run.)

hawkrives:
> Why iPhone 7, specifically? 

It was in their example, and for the initial POC pass I didn't see an advantage to running on several devices. We can add new config keys to run on other devices, and it can run them in parallel.

> How much is required for Android to join along? I could easily see scaling these tests out.

https://wix.github.io/detox/docs/Introduction.Android.html

tl;dr: pretty much add a `test:app` gradle target, add a new detox config key, and run `yarn detox test -c android.emu.debug`.

> Also, is there any way to strip out testIDs from the resulting build? Or are those benign?

Completely benign. They're pretty much just like `id` attributes in HTML, as far as the app goes. I don't even know if they get propagated to the native layers.

source/app.js
hawkrives: Done (#2813).

I'm leaving it in here until it's merged, so that the tests in this PR actually pass.

e2e/config.json
```diff
@@ -0,0 +1,4 @@
+{
+	"setupTestFrameworkScriptFile": "./init.js",
```
rye, line 2: Do we want tabs? Or two-space as is the custom with `package.json`?

hawkrives: Tabs for consistency with everything that's not package.json

hawkrives: (Plus, tab is 1 char, 2 spaces is 2 chars. 1 < 2, so 1 is better :p)

rye: Alright, I will not argue with something that I agree with. Tabs are better.

rye (dismissed):

hawkrives:
@drewvolz any opposition? I know we'll add more stuff later, but I don't know that we need any additions before we introduce this to our codebase.

e2e/firstTest.spec.js
```diff
@@ -0,0 +1,21 @@
+/* eslint-env jest */
+/* global element, device, by */
+
+describe('Example', () => {
```
drewvolz: Please rename the describe to something other than `Example`

source/app.js
drewvolz: That has been merged so this may be removed.

rye (approved):

drewvolz (approved):

PR #2813, opened 2018-08-18
Disable react-navigation session persistence on dev

hawkrives:
Extracted from #2810 

Disables React-Navigation session persistence when developing.

For posterity: if you want to enable this for a development session, you can just change `null` to some other value (but remember not to commit it!)


rye (approved):
Very reviewable. LGTM.

PR #2815, opened 2018-08-19
Test that we can reach the edit-home view and back

drewvolz:
Adds a general test to navigate from `Home > Edit Home > Home`.


source/views/components/nav-buttons/edit-home.js
```diff
@@ -18,10 +18,15 @@ export function EditHomeButton({
 }) {
 	return (
 		<Touchable
+			accessibilityComponentType="button"
+			accessibilityLabel="Open Edit Home"
```
hawkrives, line 22: This label doesn't feel right? I'd almost rather it was just kept as "Edit".

e2e/basic-smoke.spec.js
```diff
@@ -18,4 +18,12 @@ describe('Basic smoke tests', () => {
 		await element(by.id('header-back')).tap()
 		await expect(element(by.id('screen-homescreen'))).toBeVisible()
 	})
+
+	it('should show home screen after tap to exit edit-home screen', async () => {
```
hawkrives: Please put this in a new file.

e2e/basic-smoke.spec.js
```diff
@@ -18,4 +18,12 @@ describe('Basic smoke tests', () => {
...
+		await device.reloadReactNative()
+		await element(by.id('button-open-edit-home')).tap()
+		await expect(element(by.id('screen-homescreen'))).toBeNotVisible()
```
hawkrives: You don't need this one in every test; it was just me making sure that the homescreen wasn't actually visible when another screen was active.

Which should really be its own test, and not repeated in every one.

hawkrives (changes requested):
nice

e2e/basic-smoke.spec.js
drewvolz: This is part of the general homescreen smokescreen tests. Why?

drewvolz: We're testing out the two top buttons here.

hawkrives: It's part of the _homescreen_ smoke tests, not the _app_ smoke tests. This file is just to make sure that we can open the app and do a single navigation. Everything else should go into new files.

source/views/components/nav-buttons/edit-home.js
hawkrives: Okay, this makes more sense. Nevermind. (I'm now playing with voiceover.)

e2e/basic-smoke.spec.js
drewvolz: What would you like the new file named, and what else would happen in the new set of tests?

e2e/basic-smoke.spec.js
drewvolz: That's fair.

e2e/basic-smoke.spec.js
hawkrives: "e2e/views/homescreen/edit.spec.js"?

And for now, just this one; in the future, things like testing that we can drag and drop the rows (not yet possible in Detox) or ensuring that flipping a toggle switch hides the view from the homescreen. Or, in e2e/views/homescreen/home.js, making sure that you can scroll to the bottom on smaller devices.

hawkrives (dismissed):

hawkrives (approved):

PR #2817, opened 2018-08-19
More padding to bottom of credits view

drewvolz:
If the Credits view needs to scroll, the bottom is cut off.

Before | After
--|--
old-not-x | new-not-x
old-x | new-x

### DIFF
--- a/source/views/settings/credits.js
+++ b/source/views/settings/credits.js
@@ -7,10 +7,14 @@ import {Platform} from 'react-native'
 import {iOSUIKit, material} from 'react-native-typography'
 import {AppLogo} from '../components/logo'
 
-const Container = glamorous.scrollView({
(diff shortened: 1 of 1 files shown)


drewvolz:
@hawkrives I was unable to find a suitable way within Glamorous to adjust the bottom padding without an additional view. If you can think of one, I'd be happy to try it.

drewvolz:
The diff ignoring whitespace https://github.com/StoDevX/AAO-React-Native/pull/2817/files?w=1

hannesmcman (approved):

rye (approved):

hawkrives (changes requested):
I want to dig into this and find the root cause, instead of papering over it with extra padding… I will attempt to do this as soon as possible.

PR #2818, opened 2018-08-20
Update notes about environment variables

hawkrives:
Since I was just cleaning up stuff in the config files, I decided to go ahead and update the lists of environment variables that we use on our CI environments.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,8 +38,6 @@ x-config:
       name: Embed environment variables into the app
       command: |
         touch .env.js
-        echo "export const GOOGLE_CALENDAR_API_KEY = '$GCAL_KEY'" >> .env.js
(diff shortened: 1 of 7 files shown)


drewvolz:
Can we update the README to remove the bit about the calendar API key?

.circleci/config.yml
```diff
@@ -38,8 +38,6 @@ x-config:
       name: Embed environment variables into the app
       command: |
         touch .env.js
```
drewvolz, line 40: Is anything inside this file anymore?

hawkrives: No, but … something may come back in the future?

drewvolz: Alright. I think one use case we'll have for it is when the map transitions to mapbox. So we can keep it around.

drewvolz (approved):

PR #2837, opened 2018-08-20
Run [iOS] Detox in CI

hawkrives:
> I have Detox (for iOS) running in CI now… the only thing is, it adds about 6 minutes to the iOS build. I can't run them in parallel, since we only get one macOS container at a time. Do we want to enable it for a while and see how annoying the extra time is, or just, say, only enable it on nightlies for a start?
>
> I'd vote for turning it on on all builds and seeing how annoying it is.
– _Hawken_

> I’m happy running them on nightlies but no strong feelings one way or the other
– _drew_

> My argument against only nightlies is that then we have to bisect back and fix it, vs waiting an extra ~6 minutes and fixing things before master breaks.
> (Especially on greenkeeper PRs. Those be really nice to have smoke tests on.)
– _Hawken_

---

As you can see, I'm still opting to run Detox on every PR. This may need to change in the future; we can see.

### DIFF
--- a/e2e/init.js
+++ b/e2e/init.js
@@ -10,11 +10,11 @@ jasmine.getEnv().addReporter(adapter)
 
 beforeAll(async () => {
 	await detox.init(config, {launchApp: false})
-	await device.launchApp({permissions: {notifications: 'YES'}})
 })
 
 beforeEach(async () => {
 	await adapter.beforeEach()
+	await device.relaunchApp({delete: true, permissions: {notifications: 'YES'}})
 })
 
 afterAll(async () => {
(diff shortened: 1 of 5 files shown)


e2e/init.js
```diff
@@ -10,11 +10,11 @@ jasmine.getEnv().addReporter(adapter)
 
 beforeAll(async () => {
 	await detox.init(config, {launchApp: false})
-	await device.launchApp({permissions: {notifications: 'YES'}})
 })
 
 beforeEach(async () => {
 	await adapter.beforeEach()
+	await device.relaunchApp({delete: true, permissions: {notifications: 'YES'}})
```
hawkrives, line 17: This also changes Detox to relaunch the app and clear AsyncStorage between each test, since we're now testing in release mode and thus we have React-Navigation's state restoration active again.

hannesmcman (approved):

drewvolz (approved):
I still would really prefer to see this run only on `nightly` builds, but as you've said, we'll see if the increased build time becomes an annoyance.

Issue #2836, opened 2018-08-20
[stoPrint] Only list stoPrint-color printers for color jobs


hannesmcman:
## Current behavior: ##
When you select a print job destined for stoPrint-color, you are able to choose from any of the printers. If you happen to choose one that can't print in color (most of them), an error alert is triggered and you are redirected to the print jobs list. 
## Desired behavior: ##
When you select a stoPrint-color job, the only printers that should be listed are those that are able to print in color. AFAIK, the API unfortunately does not inform us of which printers can print in color. Thus, we'll probably have to hard-code the list of color printers.

PR #2839, opened 2018-08-21
Add color printers data file

hannesmcman:
Necessary for #2836.


hawkrives:
Why not just do an array of the printer names?

hannesmcman:
Idk, I figured I would just mimic the data we get from the Papercut API..but an array of names would be fine because we could match them to the data we get from the API

hawkrives (approved):

PR #2840, opened 2018-08-21
[stoPrint] how-to screen tweaks

### LINKED ISSUE #2824: [stoPrint] Include a paragraph on how to print things on the Nothing to Print screen
Something like, "We can't print anything because you haven't sent anything to the printer. You can print from a computer, or by (iOS ? "using the Print option in the Share sheet" : "however Android manages to print")".

IE, don't do any stoPrint-specific instructions, but just the minimal set for the current platform.

I think this could go under the "info" button.

drewvolz:
* Adds a `description` prop to provide more text on the stoPrint notice view
* Adds per-platform text for a better hint on what's left to do before printing
* Tweaks the stoPrint button text to read "Learn how to use stoPrint" for clarity
* Closes #2824

**Portrait**

Platform | Before | After
--|--|--|
iOS | ios-old-portrait | ios-new-portrait
Android | android-old-portrait | android-new-portrait 

**Landscape**

Platform | Before | After
--|--|--|
iOS | ios-old-landscape | image
Android | android-old-landscape | android-new-landscape

### DIFF
--- a/source/views/stoprint/components/notice.js
+++ b/source/views/stoprint/components/notice.js
@@ -83,7 +83,9 @@ export class StoPrintNoticeView extends React.PureComponent<Props, State> {
 					style={styles.notice}
 					text={text}
 				/>
-				<Text style={styles.description}>{description}</Text>
+				{description ? (
+					<Text style={styles.description}>{description}</Text>
+				) : null}
 			</ScrollView>
 		)
 	}
--- a/source/views/stoprint/components/notice.js
+++ b/source/views/stoprint/components/notice.js
@@ -1,7 +1,13 @@
 // @flow
 
 import React from 'react'
-import {Platform, ScrollView, StyleSheet, RefreshControl} from 'react-native'
+import {
+	Platform,
+	Text,
+	ScrollView,
+	StyleSheet,
+	RefreshControl,
+} from 'react-native'
 import Icon from 'react-native-vector-icons/Ionicons'
 import {NoticeView} from '../../components/notice'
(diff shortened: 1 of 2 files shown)


hannesmcman (dismissed):

source/views/stoprint/components/notice.js
```diff
@@ -76,6 +83,7 @@ export class StoPrintNoticeView extends React.PureComponent<Props, State> {
 					style={styles.notice}
 					text={text}
 				/>
+				<Text style={styles.description}>{description}</Text>
```
drewvolz: Shoot, I missed an important part of not displaying the `<Text />` element if no `description` was provided.

source/views/stoprint/print-jobs.js
```diff
@@ -103,20 +103,17 @@ class PrintJobsView extends React.PureComponent<Props> {
 				/>
 			)
 		}
-		if (this.props.loginState !== 'logged-in') {
-			return (
-				<StoPrintNoticeView
```
drewvolz: And I removed this from testing... going to add it back.

source/views/stoprint/components/notice.js
hannesmcman: ahh true

source/views/stoprint/print-jobs.js
hannesmcman: I shouldn't have been so quick to approve you! 😝

hawkrives (approved):

hannesmcman (approved):

PR #2841, opened 2018-08-21
[stoPrint]: Only list color printers for color jobs

### LINKED ISSUE #2836: [stoPrint] Only list stoPrint-color printers for color jobs

hannesmcman:
Resolves #2836.

### DIFF
--- a/source/flux/parts/stoprint.js
+++ b/source/flux/parts/stoprint.js
@@ -94,8 +94,8 @@ export function updatePrinters(): ThunkAction<UpdateAllPrintersAction> {
 			reportNetworkProblem(err)
 			return defaultData
 		})
-		const colorPrinters = allPrinters.filter(
-			printer => data.colorPrinters.includes(printer.printerName)
+		const colorPrinters = allPrinters.filter(printer =>
+			data.colorPrinters.includes(printer.printerName),
 		)
 
 		dispatch({
--- a/source/flux/parts/stoprint.js
+++ b/source/flux/parts/stoprint.js
@@ -40,6 +40,7 @@ type UpdateAllPrintersSuccessAction = {
 		allPrinters: Array<Printer>,
 		popularPrinters: Array<Printer>,
 		recentPrinters: Array<Printer>,
+		colorPrinters: Array<Printer>,
 	},
 }
 
@@ -88,11 +89,14 @@ export function updatePrinters(): ThunkAction<UpdateAllPrintersAction> {
 
 		const {recentPrinters, popularPrinters} = recentAndPopularPrinters
 
-		const colorPrintersUrl = API('/color-printers')
-		const {colorPrinters} = await fetchJson(colorPrintersUrl).catch(err => {
+		const colorPrintersUrl = API('/printing/color-printers')
+		const {data} = await fetchJson(colorPrintersUrl).catch(err => {
 			reportNetworkProblem(err)
-			return defaultData.colorPrinters
+			return defaultData
 		})
+		const colorPrinters = allPrinters.filter(
+			printer => data.colorPrinters.includes(printer.printerName)
+		)
 
 		dispatch({
 			type: UPDATE_ALL_PRINTERS_SUCCESS,
--- a/source/flux/parts/stoprint.js
+++ b/source/flux/parts/stoprint.js
@@ -5,13 +5,11 @@ import {type ReduxState} from '../index'
 import type {PrintJob, Printer} from '../../lib/stoprint'
 import {
 	fetchAllPrinters,
+	fetchColorPrinters,
 	fetchJobs,
 	fetchRecentPrinters,
 	logIn,
 } from '../../lib/stoprint'
-import {API} from '../../globals'
-import * as defaultData from '../../../docs/color-printers.json'
-import {reportNetworkProblem} from '../../lib/report-network-problem'
 
 type Dispatch<A: Action> = (action: A | Promise<A> | ThunkAction<A>) => any
 type GetState = () => ReduxState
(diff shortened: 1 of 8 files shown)


source/flux/parts/stoprint.js
```diff
@@ -9,6 +9,9 @@ import {
 	fetchRecentPrinters,
 	logIn,
 } from '../../lib/stoprint'
+import {API} from '../../globals'
+import * as defaultData from '../../../docs/color-printers.json'
```
hawkrives: I don't think we should bundle this list with the app. Unlike the other resources, it's not standalone information, and you need a network connection already to be able to see the print jobs.

source/flux/parts/stoprint.js
```diff
@@ -85,9 +89,18 @@ export function updatePrinters(): ThunkAction<UpdateAllPrintersAction> {
 
 		const {recentPrinters, popularPrinters} = recentAndPopularPrinters
 
+		const colorPrintersUrl = API('/printing/color-printers')
+		const {data} = await fetchJson(colorPrintersUrl).catch(err => {
+			reportNetworkProblem(err)
```
hawkrives: … and I'd rather not do reportNetworkProblem for fetching the file, please.

source/flux/parts/stoprint.js
```diff
@@ -85,9 +89,18 @@ export function updatePrinters(): ThunkAction<UpdateAllPrintersAction> {
 
 		const {recentPrinters, popularPrinters} = recentAndPopularPrinters
 
+		const colorPrintersUrl = API('/printing/color-printers')
+		const {data} = await fetchJson(colorPrintersUrl).catch(err => {
```
hawkrives: You should move this fetch into the Promise.all, do it can be executed simultaneously with the other two network requests, instead of sequentially afterwards.

hawkrives (dismissed):
Looking good! A few requests.

hannesmcman:
Hawken's requests prompted me to add proper error handling for all the stoPrint API requests, so that the stoPrint Error View is always shown when a critical API request fails.

source/flux/parts/stoprint.js
```diff
@@ -78,16 +80,52 @@ export function updatePrinters(): ThunkAction<UpdateAllPrintersAction> {
...
+			recentAndPopularPrintersResponse,
+			colorPrintersResponse,
+		] = await Promise.all([
 			fetchAllPrinters(username),
 			fetchRecentPrinters(username),
+			fetchColorPrinters(),
 		])
 
-		const {recentPrinters, popularPrinters} = recentAndPopularPrinters
+		if (allPrintersResponse.error) {
```
rye, line 93: This and the following two `if` statements could be condensed to

```js
if(allPrintersResponse.error || recentAndPopularPrintersResponse.error || colorPrintersResponse.error) {
  return dispatch({
    type: UPDATE_ALL_PRINTERS_FAILURE,
    payload: [either a ternary or some kind of fancy select results]
  })
}
```

Otherwise, I'd chain these together as `else if`s, because they will get executed in sequence and if one of them is selected, the chain aborts. (In fact, the function returns.)

source/lib/stoprint/api.js
```diff
@@ -40,28 +42,71 @@ export async function logIn(
 		}))
 
 	if (result.error) {
-		return 'The print server seems to be having some issues'
+		return 'The print server seems to be having some issues.'
 	}
 
 	if (!result.value.success) {
-		return 'The username and password appear to be invalid'
+		return 'The username and password appear to be invalid.'
```
rye: Maybe `Your username and password appear to be invalid.`?

source/lib/stoprint/api.js
```diff
@@ -40,28 +42,71 @@ export async function logIn(
...
+	username: string,
+): Promise<PrintJobsResponseOrErrorType> =>
 	fetchJson(`${PAPERCUT_API}/webclient/users/${username}/jobs/status`)
+		.then(response => ({
+			error: false,
+			value: response,
+		}))
+		.catch(() => ({
+			error: true,
+			value: 'Unable to fetch print jobs list from stoPrint.',
```
rye: `Unable to fetch a list of print jobs from stoPrint.`?

source/lib/stoprint/api.js
```diff
@@ -40,28 +42,71 @@ export async function logIn(
...
+	username: string,
+): Promise<AllPrintersResponseOrErrorType> =>
 	fetchJson(`${PAPERCUT_MOBILE_RELEASE_API}/all-printers?username=${username}`)
+		.then(response => ({
+			error: false,
+			value: response,
+		}))
+		.catch(() => ({
+			error: true,
+			value: 'Unable to fetch all printers from stoPrint.',
```
rye: `Unable to fetch the list of all printers from stoPrint.`?

source/lib/stoprint/api.js
```diff
@@ -40,28 +42,71 @@ export async function logIn(
...
+	ColorPrintersResponseOrErrorType,
+> =>
+	fetchJson(colorPrintersUrl)
+		.then(response => ({
+			error: false,
+			value: response,
+		}))
+		.catch(() => ({
+			error: true,
+			value: 'Unable to fetch color printers from stoPrint.',
```
rye: `Unable to fetch the list of color printers from stoPrint.`?

source/views/stoprint/printers.js
```diff
@@ -89,19 +90,26 @@ class PrinterListView extends React.PureComponent<Props> {
...
 			return (
 				<StoPrintErrorView
 					navigation={this.props.navigation}
 					refresh={this.fetchData}
+					statusMessage={this.props.error}
 				/>
 			)
 		}
+		if (this.props.loading && this.props.printers.length === 0) {
+			return <LoadingView text="Fetching Available Printers…" />
```
rye: `Fetching a list of available printers…`?

source/views/stoprint/print-jobs.js
```diff
@@ -92,17 +92,19 @@ class PrintJobsView extends React.PureComponent<Props> {
...
 			return (
 				<StoPrintErrorView
 					navigation={this.props.navigation}
 					refresh={this.fetchData}
+					statusMessage={this.props.error}
 				/>
 			)
 		}
+		if (this.props.loading && this.props.jobs.length === 0) {
+			return <LoadingView text="Fetching stoPrint Jobs…" />
```
rye: `Fetching a list of stoPrint jobs…`?

source/views/stoprint/printers.js
rye: An alternative to the `Fetching` language is `Querying available printers…`?

rye: (I like the `Querying` language a bit more.)

source/flux/parts/stoprint.js
hannesmcman: Since each `if` statement returns if triggered, isn't there no need for `else if`'s? A ternary wouldn't work since there are three conditions right?

source/lib/stoprint/api.js
drewvolz: I would prefer an “or” over an “and”

source/flux/parts/stoprint.js
rye: Well. I suppose.

I would just structure the ternary as:

```js
allPrintersResponse.error ? allPrintersResponse.value : 
  (recentAndPopularPrintersResponse.error ? recentAndPopularPrintersResponse.value :
    (colorPrintersResponse.error ? colorPrintersResponse.value : null))
```

But the more I look at that, the less I like it.

Personally I'd be tempted to write a function that, when given a list of things with `.error` and `.value` values, returns the first one with `value`.  Then use it here.

This way you DRY up the code, which is good. But it's not a major gripe of mine.

source/lib/stoprint/api.js
hawkrives: I concur with Drew.

hawkrives (changes requested):

hawkrives:
Oh I see it now

hawkrives (approved):

source/flux/parts/stoprint.js
hannesmcman: Yeah lol that's why I was afraid of the ternary route.. 😆I tried the following, but it really pissed flow off: 
```javascript
const errorResponse = [allPrintersResponse, recentAndPopularPrintersResponse, colorPrintersResponse].find(response => (
	response.error
))
if (errorResponse && errorResponse.error) {
	return dispatch({
		type: UPDATE_ALL_PRINTERS_FAILURE,
		payload: errorResponse.value
	})
}
```
It can't seem to figure out that if this if condition isn't true, then the value attribute of each response will not be a string, but the proper ResponseType...

hannesmcman: I'm tempted to leave it as is rather than entering a wrestling match with flow...is that ok @rye ?

hannesmcman: lol I guess it's too late now

hawkrives: I'm fine with the way it is. I think it's readable and doesn't have any magic. No extra functions to go look at.

PR #2844, opened 2018-08-21
Brewfile

rye:
Add a Brewfile to the repository.

This will assist with onboarding of new developers who have macOS systems.

(This _might_ (??????) be cross-compatible with LinuxBrew, which would be awesome—`homebrew-bundle` claims that it might work but is not officially supported.)

### DIFF
--- a/Brewfile
+++ b/Brewfile
@@ -1,0 +1,2 @@
+brew "node"
+brew "yarn"
--- a/Brewfile
+++ b/Brewfile
@@ -2,3 +2,8 @@ brew "node"
 brew "ruby"
 brew "watchman"
 brew "yarn"
+
+# Additional packages for Detox
+# Note: applesimutils is optional if you don't plan on doing any iOS Detox runs.
+tap "wix/brew"
+brew "applesimutils"
--- a/Brewfile
+++ b/Brewfile
@@ -1,2 +1,3 @@
 brew "node"
+brew "ruby"
 brew "yarn"
--- a/Brewfile
+++ b/Brewfile
@@ -1,4 +1,3 @@
-brew "git-lfs"
 brew "node"
 brew "ruby"
 brew "watchman"
--- a/Brewfile
+++ b/Brewfile
@@ -1,3 +1,4 @@
+brew "git-lfs"
 brew "node"
 brew "ruby"
 brew "watchman"
--- a/Brewfile
+++ b/Brewfile
@@ -1,3 +1,4 @@
 brew "node"
 brew "ruby"
+brew "watchman"
 brew "yarn"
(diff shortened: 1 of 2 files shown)


Brewfile
```diff
@@ -0,0 +1,5 @@
+brew "git-lfs"
```
hawkrives: We don't use this anywhere, and AFAICR we rewrote history to completely remove it from our repo, so we can kill it off in the Brewfile.

hawkrives (dismissed):

hawkrives:
(Also, uh… how do you use this?)

rye:
`brew bundle [install]` will:
>Install or upgrade all dependencies in a Brewfile.

- - -

Docs:

```
brew bundle command:
    Bundler for non-Ruby dependencies from Homebrew.

    bundle [install] [-v|--verbose] [--no-upgrade] [--file=path|--global]:
    Install or upgrade all dependencies in a Brewfile.

    brew bundle dump [--force] [--describe] [--file=path|--global]
    Write all installed casks/formulae/taps into a Brewfile.

    brew bundle cleanup [--force] [--zap] [--file=path|--global]
    Uninstall all dependencies not listed in a Brewfile.

    brew bundle check [--no-upgrade] [--file=path|--global] [--verbose]
    Check if all dependencies are installed in a Brewfile. Missing dependencies are listed in verbose mode. check will exit on the first category missing a dependency unless in verbose mode.

    brew bundle exec command
    Run an external command in an isolated build environment.

    brew bundle list [--all|--brews|--casks|--taps|--mas] [--file=path|--global]
    List all dependencies present in a Brewfile, optionally limiting by types.
    By default, only brew dependencies are output.

    If -v or --verbose are passed, print verbose output.

    If --no-upgrade is passed, don't run brew upgrade outdated dependencies.
    Note they may still be upgraded by brew install if needed.

    If --force is passed, uninstall dependencies or overwrite an existing Brewfile.

    If --file=path is passed, the Brewfile path is set accordingly
    Use --file=- to output to console.

    If --global is passed, set Brewfile path to $HOME/.Brewfile.

    If --describe is passed, output a description comment above each line.
    This comment will not be output if the dependency does not have a description.

    If -h or --help are passed, print this help message and exit.
```

hawkrives (dismissed):
I'd also be OK with pulling the applesimutils out into another Brewfile that you could run if you wanted to do testing, like something that could run as part of the testing/e2e init process.

hannesmcman (dismissed):
Can we add the instructions for using this to the README in this PR?

rye:
I would complain about out-of-scope but that makes sense.

hannesmcman (approved):
thx 😈

PR #2850, opened 2018-08-22
Fix course search to handle updated data source

hannesmcman:
This PR fixes what Hawken broke when he changed how the course data is formatted. :)


source/views/sis/course-search/detail/index.js
```diff
@@ -17,6 +17,8 @@ import {
 import type {TopLevelViewPropsType} from '../../../types'
 import * as c from '../../../components/colors'
 import {deptNum} from '../lib/format-dept-num'
+import * as _ from 'lodash'
```
drewvolz: Please replace with the specific imports from loadash.

```js
import groupBy from 'lodash/groupBy'
import map from 'lodash/map'
import zip from 'lodash/zip'
import zipObject from 'lodash/zipObject'
```

source/views/sis/course-search/detail/index.js
```diff
@@ -82,28 +95,38 @@ function Information({course}: {course: CourseType}) {
...
-			times: course.times,
-			locations: course.locations,
-		},
-		{groupBy: 'day'},
-	)
-	const schedule = times.map(time => {
-		const hours = time.times.map(time => {
-			const start = moment.tz(time.start, 'hmm', CENTRAL_TZ).format('h:mm A')
-			const end = moment.tz(time.end, 'hmm', CENTRAL_TZ).format('h:mm A')
+	const groupedByDay = _.groupBy(course.offerings, 'day')
```
drewvolz: Remove ` _.` from groupBy

source/views/sis/course-search/detail/index.js
```diff
@@ -82,28 +95,38 @@ function Information({course}: {course: CourseType}) {
...
+		const timesFormatted = offerings.map(offering => {
+			const start = moment
+				.tz(offering.start, 'H:mm', CENTRAL_TZ)
+				.format('h:mm A')
+			const end = moment.tz(offering.end, 'H:mm', CENTRAL_TZ).format('h:mm A')
 			return `${start} – ${end}`
 		})
+		const locations = offerings.map(offering => offering.location)
+
+		const timelocs = _.zip(timesFormatted, locations)
```
drewvolz: Remove ` _.` from zip

source/views/sis/course-search/detail/index.js
```diff
@@ -82,28 +95,38 @@ function Information({course}: {course: CourseType}) {
...
+
+		const timelocsObj = timelocs.map(timelocs =>
+			_.zipObject(['time', 'location'], timelocs),
```
drewvolz: Remove ` _.` from zipObject

source/views/sis/course-search/detail/index.js
```diff
@@ -82,28 +95,38 @@ function Information({course}: {course: CourseType}) {
...
-			locations: course.locations,
-		},
-		{groupBy: 'day'},
-	)
-	const schedule = times.map(time => {
-		const hours = time.times.map(time => {
-			const start = moment.tz(time.start, 'hmm', CENTRAL_TZ).format('h:mm A')
-			const end = moment.tz(time.end, 'hmm', CENTRAL_TZ).format('h:mm A')
+	const groupedByDay = _.groupBy(course.offerings, 'day')
+	const schedule = _.map(groupedByDay, (offerings, day) => {
```
drewvolz: Remove ` _.` from map

source/views/sis/course-search/detail/index.js
```diff
@@ -82,28 +95,38 @@ function Information({course}: {course: CourseType}) {
...
+		const locations = offerings.map(offering => offering.location)
+
+		const timelocs = _.zip(timesFormatted, locations)
+
+		const timelocsObj = timelocs.map(timelocs =>
+			_.zipObject(['time', 'location'], timelocs),
+		)
+
+		const rightDetail = timelocsObj.map(timeloc => (
+			<Text style={styles.rightDetail}>
```
drewvolz: Please add  `key` to this element.

source/views/sis/course-search/detail/index.js
```diff
@@ -82,28 +95,38 @@ function Information({course}: {course: CourseType}) {
 }
 
 function Schedule({course}: {course: CourseType}) {
-	if (!course.times) {
+	if (!course.offerings) {
 		return null
 	}
-	const times = convertTimeStringsToOfferings(
```
drewvolz, line 88: Please remove the import of `convertTimeStringsToOfferings` on line 10.

source/views/sis/course-search/detail/index.js
```diff
@@ -1,7 +1,7 @@
 // @flow
 
-import React from 'react'
-import {StyleSheet, Text, Platform} from 'react-native'
+import * as React from 'react'
+import {StyleSheet, Text, Platform, View} from 'react-native'
```
drewvolz: Please remove the `View` import as it is unused.

drewvolz (commented):
Very nice 👍

source/views/components/cells/multi-line-detail.js
```diff
@@ -20,13 +20,15 @@ export class MultiLineDetailCell extends React.PureComponent<Props> {
...
-						{leftDetail}
-					</Text>
+					{this.props.leftDetail && (
+						<Text allowFontScaling={true} style={styles.cellLeftDetail}>
+							{leftDetail}
+						</Text>
+					)}
 				</View>
-				<Text allowFontScaling={true} style={styles.cellRightDetail}>
+				<View allowFontScaling={true} style={styles.cellRightDetail}>
```
drewvolz, line 29: Changing from `Text` to `View` means we'll also need to change the style. We can't apply`letterSpacing` to a `View`.

source/views/sis/course-search/detail/index.js
```diff
@@ -82,28 +97,38 @@ function Information({course}: {course: CourseType}) {
...
 		})
+		const locations = offerings.map(offering => offering.location)
+
+		const timelocs = zip(timesFormatted, locations)
+
+		const timelocsObj = timelocs.map(timelocs =>
+			zipObject(['time', 'location'], timelocs),
+		)
+
+		const rightDetail = timelocsObj.map(timeloc => (
```
drewvolz: I don't know how I'd like to change it, but the alignment of items in the right detail looks better sometimes than others.

1 | 2
--|--
times1 | times2

drewvolz: Maybe left-aligning the times as they will be uniform, and then relying on rooms being more likely than not the same length?

hannesmcman: I think no matter whether we left-align or right-align, it’s going to look a little weird when the room numbers aren’t the same length...

hannesmcman: I honestly think left-align looks worse..
screen shot 2018-08-21 at 11 49 22 pm

drewvolz: I'll defer to what you'd prefer. I'd also ask that @rye and @hawkrives give their 2¢

rye: Most of the things will be `BLDG ###`, so the room annotation would right align itself fairly well.

Therefore I'd probably stick with right alignment.

`(TBA)` is an exception to the rule.

hawkrives: I vote right-align.

source/views/sis/course-search/detail/index.js
```diff
@@ -82,28 +97,38 @@ function Information({course}: {course: CourseType}) {
...
-			times: course.times,
-			locations: course.locations,
-		},
-		{groupBy: 'day'},
-	)
-	const schedule = times.map(time => {
-		const hours = time.times.map(time => {
-			const start = moment.tz(time.start, 'hmm', CENTRAL_TZ).format('h:mm A')
-			const end = moment.tz(time.end, 'hmm', CENTRAL_TZ).format('h:mm A')
+	const groupedByDay = groupBy(course.offerings, 'day')
```
hawkrives: Please use the function second argument instead of the key shorthand, for consistency with the rest of the code – `groupBy(course.offerings, c => c.day)`

source/views/sis/course-search/detail/index.js
```diff
@@ -82,28 +97,38 @@ function Information({course}: {course: CourseType}) {
...
+				.format('h:mm A')
+			const end = moment.tz(offering.end, 'H:mm', CENTRAL_TZ).format('h:mm A')
 			return `${start} – ${end}`
 		})
+		const locations = offerings.map(offering => offering.location)
+
+		const timelocs = zip(timesFormatted, locations)
+
+		const timelocsObj = timelocs.map(timelocs =>
+			zipObject(['time', 'location'], timelocs),
```
hawkrives: Same thing about string vs function arguments here - `[o => o.time, o => o.location]`

source/views/sis/course-search/detail/index.js
```diff
@@ -82,28 +97,38 @@ function Information({course}: {course: CourseType}) {
...
+		)
+
+		const rightDetail = timelocsObj.map(timeloc => (
+			<Text key={timeloc.time} style={styles.rightDetail}>
+				<Text style={styles.time}>{timeloc.time} </Text>
+				<Text style={styles.location}>{`(${timeloc.location})`}</Text>
```
hawkrives: Just change this to `>({timeloc.location})</Text>` – we don't need the extra string creation within JSX.

hawkrives (changes requested):
Just a couple of nits! Good work!

source/views/sis/course-search/detail/index.js
drewvolz: The right-aligns have it.

source/views/sis/course-search/detail/index.js
hannesmcman: AFAIK the first argument to the `zipObject` function is just supposed to be an array of the property names that you want for the items of the arrays that you are zipping. So in this case the string arguments would be correct?

hannesmcman: @hawkrives

hawkrives: It accepts functions, too (everything that accepts string shorthands also accepts functions, since lodash predates arrow functions and really needed the shorthand versions back then)

hannesmcman: But what is the object being passed to each function argument then? `[o => o.time, o => o.location]` doesn't work and `timelocs` is an array of arrays... @hawkrives

hawkrives: Oh! I'm sorry. Completely misread what this does. 

Okay, new suggestion: 

```
timelocs.map(([time, location]) => ({time, location}))
```

I think it does the same thing as your zipObjects call? 
https://runkit.com/hawkrives/5b7e1387bd33fc00120840c2

hannesmcman: yeah that works!

hawkrives (approved):

PR #2854, opened 2018-08-22
[stoPrint]: Fix pull-to-refresh bug and fix loading states

### LINKED ISSUE #2826: [stoPrint] Visual bug in pull-to-refresh on Print Jobs screen
When I pull-to-refresh, as soon as the refresh vibration happens, my screen jumps down about twice as far as it was right before the refresh started.

screenrecording_08-20-2018 07-48-56 min

(Apologies for the gigantic gif.)

hannesmcman:
Resolves #2826, #2855.
This PR resolves the pull-to-refresh bug by moving the loading status of the print jobs and the list of printers from Redux state to the state of the components themselves. This also allows us to distinguish between the loading state of these components when they are first mounted vs. when they have been refreshed via pull-to-refresh. This distinction allows us to present a `<LoadingView/>` only on the initial data load, while all data requests initiated by pull-to-refresh will lead to the expected behavior (with the loading icon above the SectionList or ScrollView itself). 

Bug | Before | After
-- | -- | --
#2826 | before-2 | after-2
#2855 | before | after

### DIFF
--- a/source/views/stoprint/components/notice.js
+++ b/source/views/stoprint/components/notice.js
@@ -34,9 +34,7 @@ export class StoPrintNoticeView extends React.PureComponent<Props, State> {
 	}
 
 	componentDidMount() {
-		if (this._timer) {
-			this._timer = setInterval(this.props.refresh, 5000)
-		}
+		this._timer = setInterval(this.props.refresh, 5000)
 	}
 
 	componentWillUnmount() {
@@ -69,6 +67,7 @@ export class StoPrintNoticeView extends React.PureComponent<Props, State> {
 						refreshing={this.state.refreshing}
 					/>
 				}
+				showsVerticalScrollIndicator={false}
 				style={styles.container}
 			>
 				<Icon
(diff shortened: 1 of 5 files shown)


hawkrives (commented):
Just to be absolutely sure that it works, and is waiting until the whole redux action is done, could you stick a `delay(5000)` into one of the redux actions and re-test, please? I'm trying to make sure that the `await` waits for the whole function, and not just, like, the time it takes to return the inner function.

hannesmcman:
@hawkrives Just tested it with the delay and it works (the loading is not set to false until the Redux state change is complete). However, since both of these views will always start in a loading state, I'm going to change the initial loading state of each of them to true.

source/views/stoprint/components/notice.js
```diff
@@ -34,9 +34,7 @@ export class StoPrintNoticeView extends React.PureComponent<Props, State> {
 	}
 
 	componentDidMount() {
-		if (this._timer) {
```
hawkrives, line 37: good catch 👍

hawkrives (approved):

PR #2861, opened 2018-08-23
Split out odds/ends and misc settings into separate files

drewvolz:
No visual changes. I noticed that the `odds-and-ends` file contained the `miscellany` section. This splits it into its own file.


hawkrives (commented):
I was thinking about doing this last weekend! Thanks.

hawkrives (approved):

PR #2866, opened 2018-08-24
Fix formatting of recent filter combo with department filter

hannesmcman:
This PR fixes an issue with the formatting of recent filter combinations that include a department name. I simply forgot to update the key of the old "departments" filter to "department".


hawkrives (approved):

hawkrives:
I'm not sure there's a good way to capture the list of allowed keys in flow right now, so this looks fine.

PR #2870, opened 2018-08-24
Fix course search scroll error

### LINKED ISSUE #2849: Course Search: ScrollToLocation causes exception if SectionList is empty
Within Course Search, ScrollToLocation is called when a new query is made, or when the search bar is cleared. In both of these cases, if the course results list is empty we get the following exception:

screen shot 2018-08-21 at 3 52 02 pm

This will cause the app to crash so it is critical that we get to the bottom of this

hawkrives:
Closes #2849 

[My current favorite React blog post](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#recommendation-fully-uncontrolled-component-with-a-key):

> In order to reset the value when moving to a different item […], we can use the special React attribute called `key`. When a `key` changes, React will create a new component instance rather than update the current one.

By adding `key={searchQuery}`, we completely avoid the need to scroll back to the top manually, because our "results list" gets completely unmounted and recreated.

Furthermore, because the search cache is at the module level and not cached within an instance of the component, we don't incur a speed penalty for recreating the search list.

Thoughts? I think this seems elegant.

### DIFF
--- a/source/views/sis/course-search/list.js
+++ b/source/views/sis/course-search/list.js
@@ -38,7 +38,6 @@ type Props = TopLevelViewPropsType & {
 	style?: any,
 	contentContainerStyle?: any,
 	filtersLoaded: boolean,
-	getRef: (ref: any) => any,
 }
 
 function doSearch(args: {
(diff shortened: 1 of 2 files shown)


hawkrives:
At one point during a previous approach, I had written 

```js
this._listView._wrapperListRef._listRef.scrollToOffset({offset: 0})
```

and that's when I decided that I needed to find a different way to do it lol.

hannesmcman (approved):
Nice!

PR #2882, opened 2018-08-25
[modularize] Move "colors" to a "module"

hawkrives:
The first module!

```diff
-import {sectionBgColor} from '../../../components/colors'
+import {sectionBgColor} from '@frogpond/colors'
```

Thoughts?


rye:
Why the `@frogpond` namespace?

hawkrives:
I'm trying to split things between `@app` being specific to the app and `@frogpond` being reusable between apps with no modifications; the `@frogpond` items would then be what we'd pull out into the components repository later on.

For instance, I've just broken up "storage" into a frogpond module that exposes the getItem/setItem functions, and an app module that exposes the setHomescreenOrder/etc functions, which allows analytics to continue to check and see if it's been disabled in an app-agnostic way.

Really, colors should be broken into `@frogpond/colors`, which is all of the generic colors, and then `@app/colors`, which is the stuff like `sto.lightGold`. `@app/colors` would then re-export `@frogpond/colors`, probably, so that things that need app-specific theming could just use `@app/colors`.

Assuming, that is, that we don't want to automatically allow all future apps to use the St. Olaf colors 😆 

I'll update this PR to split the colors into two parts.

hawkrives:
I now have:

- `@frogpond/colors`, housing the basic colors / gradients / etc
- `@app/lib/theme`, housing the theme variables
- `@app/lib/colors`, housing the St. Olaf colors

hawkrives:
@rye how's this look now? I'm happier with this too.

hawkrives:
Phooey

drewvolz (approved):

hawkrives:
I'm going to merge this; I'm happy to make more changes in the future.

PR #2884, opened 2018-08-26
List redux info in beta settings

drewvolz:
This is an attempt at making the app's internals a little more visible on the betas. After this PR, you'll be able to head to `Settings > Debug` to view the app's current redux state. My aim for this was for it to aid in debugging when away from a console.

~The last outstanding issue I know of is a limitation of the Cell component I've chosen. It can be fixed. Right now, settings that have quite a few characters in their values (longer than the row) do not play nicely with the key text on the left.~

~(That and merge conflicts from all of the recent reorgs)~

Critiques and feedback welcome. Big thanks to @hawkrives for helping make this possible.

<details>
<summary>Click to see the (many) screenshots</summary>

Level | Top | Detail
--|--|--
Settings | ios-settings | ios-debug
App | ios-app | 
Courses | ios-courses | course
Home | ios-home | ios-home-order
Settings | ios-debug-settings | 
Balances | ios-balances |
Buildings | ios-buildings | ios-buildings-favorites
Help | ios-help |  help-button |
stoPrint | ios-stoprint | 

</details>


source/views/settings/sections/odds-and-ends.js
```diff
@@ -14,7 +15,13 @@ type Props = TopLevelViewPropsType & {
 }
 
 class OddsAndEndsSection extends React.PureComponent<Props> {
+	onDebugButton = () => this.props.navigation.navigate('DebugView')
+
 	render() {
+		const DebugCell = () =>
+			version.includes('beta') ? (
```
hawkrives: Please pull this into source/lib/constants.js

Also, I'd prefer a check for "-beta.", and you should check for "-pre." as well, since nightlies aren't betas. 

Finally, please inline this ternary into the return statement instead of putting it before. Thanks!

source/views/settings/debug/list.js
```diff
@@ -0,0 +1,77 @@
...
+import type {NavigationState} from 'react-navigation'
+import type {TopLevelViewPropsType} from '../../types'
+
+type Props = TopLevelViewPropsType & {
+	state: any,
+}
+
+export class DebugListView extends React.PureComponent<Props> {
+	static navigationOptions = ({navigation}: NavigationState) => {
+		let titleParam = navigation.getParam('keyPath', 'Debug')
```
hawkrives: Why use getParam here and nowhere else?

source/views/settings/debug/row.js
```diff
@@ -0,0 +1,37 @@
...
+		const {data} = this.props
+
+		let rowDetail = isObject(data.value) ? null : JSON.stringify(data.value)
+		if (isArray(data.value) && data.value.length === 0) {
+			rowDetail = data.value.length
+		}
+
+		const arrowPosition = rowDetail === null ? 'center' : 'none'
+		const arrowStyle = arrowPosition !== 'none' ? 'DisclosureIndicator' : false
+		const _onPress = arrowPosition === 'none' ? null : this.props.onPress
```
hawkrives: Remove the underscore please.

source/views/settings/debug/row.js
```diff
@@ -0,0 +1,37 @@
...
+	data: any,
+	onPress: string => any,
+}
+
+export class DebugRow extends React.PureComponent<Props> {
+	render() {
+		const {data} = this.props
+
+		let rowDetail = isObject(data.value) ? null : JSON.stringify(data.value)
+		if (isArray(data.value) && data.value.length === 0) {
```
hawkrives: Use Array.isArray instead of importing from lodash please.

source/views/settings/debug/row.js
```diff
@@ -0,0 +1,37 @@
...
+type Props = TopLevelViewPropsType & {
+	data: any,
+	onPress: string => any,
+}
+
+export class DebugRow extends React.PureComponent<Props> {
+	render() {
+		const {data} = this.props
+
+		let rowDetail = isObject(data.value) ? null : JSON.stringify(data.value)
```
hawkrives: Just do typeof data.value === "object" instead of importing from lodash.

source/views/settings/debug/row.js
```diff
@@ -0,0 +1,37 @@
+// @flow
+
+import * as React from 'react'
+import {Cell} from 'react-native-tableview-simple'
+import isObject from 'lodash/isObject'
+import isArray from 'lodash/isArray'
+import type {TopLevelViewPropsType} from '../../types'
+
+type Props = TopLevelViewPropsType & {
+	data: any,
```
hawkrives: data: {key: string, value: any}

hawkrives (changes requested):
Code review time

source/views/settings/debug/list.js
drewvolz: It was a very convenient way to fallback onto the top-level name if we were not n+1 levels deep in the stack. I'm open to suggestions for here or other places.

source/views/settings/debug/row.js
drewvolz: I thought to define it that way as we're the second-level of onPress calls.

source/views/settings/debug/list.js
hawkrives: I'm asking why you aren't using getParam instead of accessing `navigation.state.params.keyPath` in the rest of the code, but you are using it here.

I think that you should use it in the rest of the code, instead of directly accessing navigation.state.params.keyPath.

source/views/settings/debug/row.js
hawkrives: I think that the local variable vs `this.props.onPress` is clear enough.

source/views/settings/debug/list.js
drewvolz: That makes sense. I'll make that happen.

source/views/settings/debug/row.js
drewvolz: Okay 👍 I'll make that change.

source/views/settings/sections/odds-and-ends.js
rye: Also, `-alpha.` and `-rc.` for completeness.

drewvolz: Do we want to see this debug functionality on all of those? `-alpha.`, `-beta.`, `-pre.`, and `-rc.`?

drewvolz: What I'm imagining is:

`constants.js`
```js
import {version} from '../../package.json'

export const IS_PRODUCTION = process.env.NODE_ENV === 'production'
export const IS_ALPHA = version.includes('-alpha.')
export const IS_BETA = version.includes('-beta.')
export const IS_PRE = version.includes('-pre.')
export const IS_RC = version.includes('-rc.')
export const IS_DEBUG = IS_ALPHA || IS_BETA || IS_PRE || IS_RC
```

`odds-and-ends.js`
```js
import {IS_DEBUG} from '../../../lib/constants'

IS_DEBUG ? (
	<PushButtonCell onPress={this.onDebugButton} title="Debug" />
) : null
```

source/views/settings/sections/odds-and-ends.js
```diff
@@ -3,18 +3,26 @@ import * as React from 'react'
...
 	onChangeFeedbackToggle: (feedbackDisabled: boolean) => any,
 	feedbackDisabled: boolean,
 }
 
 class OddsAndEndsSection extends React.PureComponent<Props> {
+	onDebugButton = () => this.props.navigation.navigate('DebugView')
+
 	render() {
+		const DebugCell = () =>
+			IS_DEBUG ? (
```
hawkrives: Please inline this ternary into the return statement.

hawkrives (changes requested):

source/views/settings/sections/odds-and-ends.js
```diff
@@ -19,24 +19,20 @@ class OddsAndEndsSection extends React.PureComponent<Props> {
...
 	render() {
-		const DebugCell = () =>
-			IS_DEBUG ? (
-				<PushButtonCell onPress={this.onDebugButton} title="Debug" />
-			) : null
 		return (
 			<React.Fragment>
 				<Section header="ODDS &amp; ENDS" sectionTintColor={sectionBgColor}>
 					<Cell cellStyle="RightDetail" detail={version} title="Version" />
-
```
hawkrives: Please leave these spaces in

hawkrives (changes requested):

source/views/settings/debug/row.js
```diff
@@ -0,0 +1,39 @@
...
+
+		if (data.value === null) {
+			rowDetail = 'null'
+		} else if (Array.isArray(data.value) && data.value.length === 0) {
+			rowDetail = 'Array(0)'
+		}
+
+		const arrowPosition = rowDetail === null ? 'center' : 'none'
+		const arrowStyle = arrowPosition !== 'none' ? 'DisclosureIndicator' : false
+		const onPress = arrowPosition === 'none' ? null : this.props.onPress
```
hawkrives:
```js
let rowDetail = '<unknown>'
let arrowPosition = 'none'

if (Array.isArray(data.value)) {
	// Array(0), Array(100), etc
	rowDetail = `Array(${data.value.length})`
	arrowPosition = 'center'
} else if (typeof data.value === 'object') {
	// [object Object], [object Symbol], etc
	rowDetail = data.value.toString()
	arrowPosition = 'center'
} else if (typeof data.value === 'string') {
	if (data.value.length > 20) {
		rowDetail = `"${data.value.substr(0, 20)}…"`
	} else {
		rowDetail = JSON.stringify(data.value)
	}
} else {
	rowDetail = JSON.stringify(data.value)
}

const arrowStyle = arrowPosition === 'none' ? false : 'DisclosureIndicator'
const onPress = arrowPosition === 'none' ? null : this.props.onPress
```

hawkrives (dismissed):

source/views/settings/debug/row.js
hawkrives: Fixing `null`

```js
let rowDetail = '<unknown>'
let arrowPosition = 'none'

if (Array.isArray(data.value)) {
	// Array(0), Array(100), etc
	rowDetail = `Array(${data.value.length})`
	arrowPosition = 'center'
} else if (typeof data.value === 'object' && data.value !== null) {
	// [object Object], [object Symbol], etc
	rowDetail = data.value.toString()
	arrowPosition = 'center'
} else if (typeof data.value === 'string') {
	if (data.value.length > 20) {
		rowDetail = `"${data.value.substr(0, 20)}…"`
	} else {
		rowDetail = JSON.stringify(data.value)
	}
} else {
	rowDetail = JSON.stringify(data.value)
}

const arrowStyle = arrowPosition === 'none' ? false : 'DisclosureIndicator'
const onPress = arrowPosition === 'none' ? null : this.props.onPress
```

drewvolz: Thanks!

hawkrives:
I hereby request that we avoid merging this until I'm done reorganizing the app, so that I don't have to propagate merge resolutions through ~30 branches.

hawkrives (dismissed):

hawkrives (dismissed):

drewvolz:
@rye @hannesmcman any objections?

hawkrives (approved):

PR #2889, opened 2018-08-26
[modularize] Move AlphabetListView into a @frogpond/listview module

hawkrives:
Moves the AlphabetListView component into a module, `@frogpond/listview`.

I envision us creating several other types of ListViews to go in here, to reduce code duplication in the rest of the codebase, but we don't currently have any.

Part of the great #1537 redo.


drewvolz (approved):

rye (commented):
Why is this not going into `@frogpond/alphabet-listview` and instead just `listview`?

hawkrives:
I was torn on that, but I opted to aim for where I want to be which is to have a `<ListView>` component that we can use and just flip on `<ListView displaySections={true} />` or `<ListView showSearchBar={true} />`, instead of having to pick between SectionList and FlatList and SearchableAlphabetListView and AlphabetListView.

We're not there yet, but that's why I put them in `@frogpond/listview`.

rye (approved):
Fair enough. RSLGTM.

I guess my prior review is mostly just questioning whether or not our primary goal is to produce publishable modules for others to use or to create modules that abstract behavior for ourselves—I think we're leaning towards the latter.

hawkrives:
I am leaning towards the latter, yes, but I think we'll wind up with a mix – maybe 25% reusable and 75% for ourselves.

PR #2925, opened 2018-08-29
Nutrition detail view for menu items

### LINKED ISSUE #1609: Create a Menu Item detail screen
Bigger title, the description, and the _nutritional information_ (and the dietary icons with labels, ofc.)

BonApp's API gives us back nutritional information:

https://github.com/StoDevX/AAO-React-Native/blob/3aaa0016509c8a195d5d28ef4080b3db4d823694/source/views/menus/types.js#L22-L28

drewvolz:
We had the data all along -- and now we're going to do something with it! The menu returns the nutritional information for quite a lot of items, and this PR provides a detail view for displaying an in-depth look into the makeup of those food items.

This adds calories, serving size, fat, saturated fat, trans fat, cholesterol, sodium, carbohydrates, fiber, sugar, and protein counts. I've also made it clear on what each dietary icon means underneath the item title. I do not count on everyone to have used the filters so this tries to make it more explicit.

Closes #1609 

<hr>

The types of screenshots
* title only
* title, icons, and description
* title, icons, description, and nutrition
* title, icons and nutrition

**iOS**

~ | ~
--|--
nothing |description-no-nutrition
icons-description-nutrition | icons-nutrition

**Android**

~ | ~
--|--
android-nothing | android-description-no-nutrition
android-icons-description-nutrition | android-icons-nutrition

### DIFF
--- a/modules/food-menu/dietary-tags.js
+++ b/modules/food-menu/dietary-tags.js
@@ -1,6 +1,7 @@
 // @flow
 import * as React from 'react'
-import {View, StyleSheet, Image} from 'react-native'
+import {View, Text, StyleSheet, Image} from 'react-native'
+import {Row} from '@frogpond/layout'
 
 import keys from 'lodash/keys'
 import pick from 'lodash/pick'
@@ -12,8 +13,16 @@ const styles = StyleSheet.create({
 		flexDirection: 'row',
 		justifyContent: 'flex-end',
 	},
-	icons: {
+	wrapper: {
+		marginVertical: 5,
+	},
+	iconsList: {
 		marginHorizontal: 3,
+		width: 15,
+		height: 15,
+	},
+	iconsDetail: {
+		marginRight: 5,
 		width: 15,
 		height: 15,
 	},
@@ -22,10 +31,12 @@ const styles = StyleSheet.create({
 export function DietaryTags({
 	corIcons,
 	dietary,
+	isDetail,
 	style,
 }: {
 	corIcons: MasterCorIconMapType,
 	dietary: ItemCorIconMapType,
+	isDetail: boolean,
 	style?: any,
 }) {
 	// filter the mapping of all icons by just the icons provided by this item
@@ -33,10 +44,27 @@ export function DietaryTags({
 		? pick(corIcons, [])
 		: pick(corIcons, keys(dietary))
 
-	// turn the remaining items into images
-	let tags = map(filtered, (dietaryIcon, key) => (
-		<Image key={key} source={{uri: dietaryIcon.image}} style={styles.icons} />
-	))
+	let tags
+
+	if (isDetail) {
+		tags = map(filtered, (dietaryIcon, key) => (
+			<Row key={key} alignItems="center" style={styles.wrapper}>
+				<Row flex={1}>
+					<Image source={{uri: dietaryIcon.image}} style={styles.iconsDetail} />
+					<Text>{dietaryIcon.label}</Text>
+				</Row>
+			</Row>
+		))
+	} else {
+		// turn the remaining items into images
+		tags = map(filtered, (dietaryIcon, key) => (
+			<Image
+				key={key}
+				source={{uri: dietaryIcon.image}}
+				style={styles.iconsList}
+			/>
+		))
+	}
 
 	return <View style={[styles.container, style]}>{tags}</View>
 }
--- a/modules/food-menu/fancy-menu.js
+++ b/modules/food-menu/fancy-menu.js
@@ -97,10 +97,7 @@ export class FancyMenu extends React.Component<Props, State> {
 	}
 
 	onPressRow(item: MenuItem, icons: MasterCorIconMapType) {
-		this.props.navigation.navigate('MenuItemDetailView', {
-			item,
-			icons,
-		})
+		this.props.navigation.navigate('MenuItemDetailView', {item, icons})
 	}
 
 	groupMenuData = (args: {
--- a/modules/food-menu/fancy-menu.js
+++ b/modules/food-menu/fancy-menu.js
@@ -11,6 +11,7 @@ import type {
 	MenuItemContainerType,
 	StationMenuType,
 } from './types'
+import {type NavigationScreenProp} from 'react-navigation'
 import size from 'lodash/size'
 import values from 'lodash/values'
 import {ListSeparator, ListSectionHeader} from '@frogpond/lists'
@@ -36,6 +37,7 @@ type FilterFunc = (filters: Array<FilterType>, item: MenuItem) => boolean
 
 type DefaultProps = {
 	applyFilters: FilterFunc,
+	navigation: NavigationScreenProp<*>,
 }
 
 type Props = ReactProps & DefaultProps
@@ -94,6 +96,13 @@ export class FancyMenu extends React.Component<Props, State> {
 		})
 	}
 
+	onPressRow(item: MenuItem, icons: MasterCorIconMapType) {
+		this.props.navigation.navigate('MenuItemDetailView', {
+			item: item,
+			icons: icons,
+		})
+	}
+
 	groupMenuData = (args: {
 		filters: Array<FilterType>,
 		stations: Array<StationMenuType>,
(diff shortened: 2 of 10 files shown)


modules/food-menu/food-item-detail.js
```diff
@@ -0,0 +1,118 @@
...
+export class MenuItemDetailView extends React.Component<Props> {
+	static navigationOptions = {
+		title: 'Nutrition',
+	}
+
+	render() {
+		const item = this.props.navigation.state.params.item
+		const icons = this.props.navigation.state.params.icons
+
+		const calculateAmount = (nutrition: NutritionDetailType) => {
```
hannesmcman: Can we pull this function out of the render() method so that it can be tested if we ever decide to write tests?

hannesmcman (dismissed):

modules/food-menu/dietary-tags.js
```diff
@@ -12,31 +13,58 @@ const styles = StyleSheet.create({
...
 		height: 15,
 	},
+	iconsDetail: {
+		marginRight: 5,
+		width: 15,
+		height: 15,
+	},
 })
 
 export function DietaryTags({
```
hawkrives: This should be turned into two components, not reused, IMO – a list vs an inline row of icons is too different to be contained within the same component.

modules/food-menu/fancy-menu.js
```diff
@@ -94,6 +96,13 @@ export class FancyMenu extends React.Component<Props, State> {
 		})
 	}
 
+	onPressRow(item: MenuItem, icons: MasterCorIconMapType) {
+		this.props.navigation.navigate('MenuItemDetailView', {
```
hawkrives: Use object shorthand here, please – `{item, icons}`. We'll get used to it if we use it.

modules/food-menu/food-item-detail.js
hawkrives: Seconded.

modules/food-menu/food-item-detail.js
```diff
@@ -0,0 +1,118 @@
...
+	navigation: {state: {params: {icons: MasterCorIconMapType, item: MenuItem}}},
+}
+
+export class MenuItemDetailView extends React.Component<Props> {
+	static navigationOptions = {
+		title: 'Nutrition',
+	}
+
+	render() {
+		const item = this.props.navigation.state.params.item
```
hawkrives: Use object destructuring here, please – `const {item, icons} = this.props.navigation.state.params`

modules/food-menu/food-item-detail.js
```diff
@@ -0,0 +1,118 @@
...
+				/>
+
+				{item.description ? (
+					<View>
+						<ListRow
+							arrowPosition="none"
+							contentContainerStyle={styles.container}
+							fullWidth={false}
+						>
+							<Row>
```
hawkrives: I don't _think_ you need these rows… isn't ListRow a column by default?

modules/food-menu/food-item-detail.js
```diff
@@ -0,0 +1,118 @@
...
+
+				<DietaryTags
+					corIcons={icons}
+					dietary={item.cor_icon}
+					isDetail={true}
+					style={styles.iconContainer}
+				/>
+
+				{item.description ? (
+					<View>
```
hawkrives: Use `<React.Fragment>` instead of `<View>` if all you need to do is wrap two elements

modules/food-menu/food-item-detail.js
```diff
@@ -0,0 +1,118 @@
...
+				Object.keys(item.nutrition_details).length > 1 ? (
+					map(item.nutrition_details, (nutrition, key: number) => {
+						return (
+							<View key={`${nutrition}-${key}`}>
+								<ListRow
+									arrowPosition="none"
+									fullWidth={false}
+									style={styles.container}
+								>
+									<Row>
```
hawkrives, line 64: Is this `<Row/>` needed?

modules/food-menu/food-item-detail.js
```diff
@@ -0,0 +1,118 @@
...
+						</ListRow>
+						<ListSeparator />
+					</View>
+				) : null}
+
+				{item.nutrition_details &&
+				Object.keys(item.nutrition_details).length > 1 ? (
+					map(item.nutrition_details, (nutrition, key: number) => {
+						return (
+							<View key={`${nutrition}-${key}`}>
```
hawkrives: `React.Fragment`

modules/food-menu/types.js
```diff
@@ -27,6 +27,7 @@ export type MenuItemType = {
 				well_being: string,
 				well_being_image: string,
 		  },
+	nutrition_details: {} | NutritionDetailContainer,
```
hawkrives, line 30: Is this still true? If we have an item with no nutritional details, will it just be a blank object, or will it be `null` or something?

hawkrives (changes requested):

modules/food-menu/food-item-detail.js
drewvolz: Removing `<Row>` from here looks like:

modules/food-menu/types.js
drewvolz: An empty `nutrition_details` element in the JSON looks like this

```json
"nutrition_details": {},
```

modules/food-menu/food-item-detail.js
drewvolz: I believe so. See the screenshot [here](https://github.com/StoDevX/AAO-React-Native/pull/2925#discussion_r214515646).

modules/food-menu/fancy-menu.js
drewvolz: Did I address this or did I modify the incorrect line?

modules/food-menu/food-item-detail.js
drewvolz: ~Good call. These are now gone.~ The one for description is now gone.

modules/food-menu/food-item-detail.js
```diff
@@ -0,0 +1,102 @@
...
+
+import * as React from 'react'
+import {Text, ScrollView, StyleSheet} from 'react-native'
+import {Column, Row} from '@frogpond/layout'
+import {ListRow, ListSeparator, Detail, Title} from '@frogpond/lists'
+import * as c from '@frogpond/colors'
+import map from 'lodash/map'
+import {DietaryTagsDetail} from './dietary-tags-detail'
+import {calculateAmount} from './lib/calculate-amount'
+import {type TopLevelViewPropsType} from '../../source/views/types'
```
hawkrives: This needs to be replaced by `import type {NavigationScreenProp} from 'react-navigation'`, as we shouldn't use anything from /source within /modules.

The `type Props` should then become `navigation: NavigationScreenProp<*> & {state: {params: {icons: MasterCorIconMapType, item: MenuItem}}}` (technically there's something we should pass instead of `*` there, but I don't recall what it is, and the alternative should work and just be a bit more verbose.)

modules/food-menu/fancy-menu.js
```diff
@@ -94,6 +96,13 @@ export class FancyMenu extends React.Component<Props, State> {
...
+			item,
+			icons,
+		})
```
hawkrives: Can you put these on one line, please? `{item, icons}` (unless Prettier is forcing the wrapping when you inline them)

modules/food-menu/food-item-row.js
```diff
@@ -16,19 +16,22 @@ type FoodItemPropsType = {|
 	style?: any,
 	badgeSpecials?: boolean,
 	spacing: {left: number},
+	onPress: any,
```
hawkrives: `onPress: () => any`, please – it needs to be a function, but we don't care what it returns, and we're not passing it anything, so it's function with no args that can return anything.

hawkrives (changes requested):
Appreciate the work!

modules/food-menu/food-item-detail.js
hawkrives: Oh right, because you're forcing these into a row. Ok.

hawkrives:
Jest is currently failing due to npm issues when installing coveralls. It will fix itself at some point.

hawkrives (approved):

hawkrives:
@drewvolz I can babysit this until npm/circle fix themselves and then merge it, if that's OK with you.

drewvolz:
That's alright by me. Thank you, @hawkrives!

Issue #2742, opened 2018-08-06
Make selected tab more apparent


drewvolz:
Either add a color highlight to it, make it darker, or something. I can hardly tell on most views which tab is active. 

The problem becomes even worse with Night Shift.

drewvolz:
I will revisit this if it truly becomes a problem.

drewvolz:
@hawkrives and I compared the blacks and grays of the icons in the tab bars for iBooks vs AAO. We still think this is an "indistinguishable-in-low-light" problem. 

App | State | Color
--|--|--
St Olaf | active | rgb(26,25,26)
iBooks | active | rgb(0,0,0)
St Olaf | inactive | rgb(128,128,128)
iBooks | inactive | rgb(144,144,144)

St. Olaf | iBooks
--|--
![photo_2018-08-14_11-15-13](https://user-images.githubusercontent.com/5240843/44107016-5ae45a74-9fb3-11e8-8fb7-4e3012fac936.jpg) | ![photo_2018-08-14_11-15-07](https://user-images.githubusercontent.com/5240843/44107018-5b05aa76-9fb3-11e8-8dad-964d6e041ab3.jpg)

hawkrives:
I'm planning to put make the inactive color configurable, and then change the colors on iOS to match iBooks.

hawkrives:
Closed by #2929, I hope.

Issue #2252, opened 2018-01-25
Change iOS tab bar tint color


hawkrives:
I personally find it difficult to read during the day, and neigh-impossible to read when Night Shift kicks on. 

I think I want to try the new St. Olaf gold / dark gold…

hawkrives:
I can't find any colors that I like that work better than `mandarin` (the current one)…

PR #2929, opened 2018-09-01
use sto.purple for the iOS tab bar

hawkrives:
See https://github.com/StoDevX/AAO-React-Native/issues/2742

So. iOS tab bar colors.

<dl><dt>Black</dt>
<dd>somehow works for iBooks; after having measured their colors again, I think they're using a "vibrant" tab bar (which we can't currently do) that tints more white when on a white background than our grey one. 

Sadly, black icons become nigh-indistinguishable from the grey icons at night / with tired eyes, _especially_ with Night Shift enabled.</dd>

<dt>Gold</dt>
<dd>I want to like gold. I can even rationalize an excuse to use it, even given the accessibility focus: something something it's an icon, and it's bigger, and maybe the text being black in the inactive state helps too. (parenthetical: I would feel better about this if our tab bar actually responded to the iOS accessibility settings, like long press to zoom the icon, but it doesn't.)

Unfortunately, it's ~9pm right now, and I already have to look twice to read the text on the icon. And night shift isn't even on yet! I think that if we use gold, I'll have to re-open https://github.com/StoDevX/AAO-React-Native/issues/2252.</dd>

<dt>Purple</dt>
<dd>Pros: It's the new St. Olaf secondary accent color, more or less, as it's used for all links on the website.

Cons: It's not gold or black, which makes it not as heavily associated with the brand yet.

I think it's readable, it's not affected by night shift, and it's easily distinguishable from the dark grey ones. Really, my only problem with it is that it's _new_ (lol).</dd>
</dl>

black | gold | purple
--- | --- | ---
screen shot 2018-08-31 at 9 09 13 pm | screen shot 2018-08-31 at 9 09 25 pm | screen shot 2018-08-31 at 9 09 05 pm

---

My vote: purple.

Thoughts? @rye, @hannesmcman, @drewvolz.

### DIFF
--- a/source/lib/theme.js
+++ b/source/lib/theme.js
@@ -70,10 +70,7 @@ export const androidTabBarForeground = firstReadable(androidTabBarBackground, [
 
 // not used in the gui; just used for calculations
 const iosTabBarBackground = '#F7F7F7'
-export const iosTabBarActiveColor = firstReadable(iosTabBarBackground, [
-	accent,
-	sto.black,
-])
+export const iosTabBarActiveColor = sto.purple
 
 export const androidStatusBarColor = tinycolor(navigationBackground)
 	.darken(20)


drewvolz (approved):

hawkrives:
@hannesmcman Now that @drewvolz has approved this, I'm going to merge – if you don't like it, we can discuss changing it again at any point until we actually release 2.6, but I'd like to get this into `master`.

PR #2932, opened 2018-09-01
Move a bunch of stuff around in Settings

hawkrives:
This doesn't pull Settings out into a module (yet), but it makes me happier when I work on stuff inside Settings.

This is a compilation of a bunch of things I've been wanting to do to Settings for a while.

- Move AppLogo into `settings/components` (since it's only used in one place)
- Make `Props` types for functional components, instead of sticking it into the arguments list – especially when both objects have to break, eurgh. So many lines of text.
- Remove PureComponent from Settings
- Rearrange the on-disk file layout – we have Settings, and it has a few other "screens" that it'll push on, so now we have `settings/screens/*` for the screens, and the overview's sections are all in `settings/overview/*.js`
- Use React.Fragment where appropriate
- Add `APP_VERSION` constant to lib/constants
- Remove `export default` (my war on default exports is almost complete)
- Consolidate Settings screen exports into `settings/index.js` and a single set of imports in `navigation/routes`


hawkrives:
This PR, like #2925, is failing due to a CircleCI/npmjs DNS issue.

source/lib/constants.js
```diff
@@ -1,4 +1,7 @@
 // @flow
 
+const {version} = require('../../package.json')
```
drewvolz, line 3: `import {version} from '../../package.json'` ?

drewvolz: This is at least what I have on the redux settings PR.

hawkrives: I'm always of a split mind when importing JSON files. They don't have exports, and it's not specified how it works. 

So I kinda like to require them instead, because that's better specified. 

But I could change it here and then do an import/require swap later one.

drewvolz: If this is your preference, we'll keep it.

rye (approved):

PR #2936, opened 2018-09-02
Update Express bus hours for 2018-09-06

rye:
- [ ] Should we indicate closure prior to the start date?
- [ ] Should we remove the Cinema 10 stop on Sunday? I get why it's listed but yeah.
- [ ] Need to test rendering to verify that `null` stop times don't break the thing.

See: https://wp.stolaf.edu/sa/transportation/localexpressbus/, and the explanation:

>The leadership team at Three Rivers has exhausted all options to ensure the community of Northfield, including the colleges, have the required transportation.  We talked to MNDot to review max hours we can have an employee operating a bus, discussed moving drivers from other towns, reviewed stats and active contracts to see if adjustments could be made to the current routes and dial-a-rides, and contacted 3rd party busing companies and temp agencies for availability.
>
>The problem is a severe driver shortage and with the FTA regulations we can’t have the few drivers we have working too many hours, so we have to make decisions based on demand for service.  The solution is to operate a modified Express route.  Hiawathaland Transit will continue to operate its blue and red routes and the AM and daily dial-a-ride services in Northfield.  We have requested and received approval from MNDot to suspend the Northfield PM dial-a-ride service and in its place operate the Express route from **6p-9p M-F, 5p-9p Sat, and 3p-6p Sun**.  This service will start as contracted on Sept 6, 2018.
>
>We will continue to review as drivers are hired and trained and will work with you to increase the serviceable transit hours.  Thank you,  TRACY HOLGUIN, Director of Transportation


rye:
Nope, `null` breaks. Tested `''` locally and it also didn't work.

Should we just put the times that the bus would reach those locations, and remove the Cinema 10 stop on Sunday?

hawkrives:
Null shouldn't break; its used on the red and blue routes.

rye:
`blue` has no holes but `red` has `false`.

hawkrives (dismissed):

rye:
Oh, we'll need to comment out the hours that don't fit in these ranges:

>Monday-Friday: 6–9pm
>Saturday: 5–9pm
>Sunday: 3–9pm

This is when the bus will actually operate.

hawkrives (approved):
Good catch

hawkrives:
I'm going to go ahead and merge this; if we find more tweaks, we can always push again.

PR #2959, opened 2018-09-05
Tweak filter buttons to be closable on entire button

drewvolz:
Makes it a bit easter to manage the list of active filters as the touch target seems too small. Instead of the close button serving as the only surface to close an active filter button, the whole button now can be used to dismiss.


drewvolz:
Much prettier diff ignoring whitespace https://github.com/StoDevX/AAO-React-Native/pull/2959/files?w=1

hawkrives (approved):

rye (approved):

PR #2943, opened 2018-09-03
Add new module: <Timer/>

hawkrives:
Usage: 

```jsx
<Timer
    interval={60000} // ms
    moment={true} // if you want a moment instance
    render={({now, refresh, loading}) => <Text>{now.calendar()}</Text>}
    timezone="America/Central" // if you want a moment-timezone instance
/>
```

The only component that didn't get significantly smaller was the Bus Map; the others all shrank by ~30sloc.

For more usage descriptions, please see the diffs.

For StoPrint, I think we can do, like

```js
renderItem = ({item}: {item: PrintJob}) => (
		<ListRow onPress={() => this.handleJobPress(item)}>
			<Title>{item.documentName}</Title>
			<Detail>
				<Timer interval={60000} moment={true} render={({now}) => <Text>{now.diff(moment(item.usageTimeFormatted, etc))}</Text>} timezone={CENTRAL_TZ} /> {' • '} {item.usageCostFormatted} {' • '}
				{item.totalPages} {item.totalPages === 1 ? 'page' : 'pages'} {'\n'}
				{item.statusFormatted}
			</Detail>
		</ListRow>
	)
```

or something.

PR #2975, opened 2018-09-06
Update timer interval to 5 seconds across the board

### LINKED ISSUE #2974: Timer component refresh is too slow
The new timer component introduced a delayed refresh that is not as quick as it used to be. Watching the transportation schedules, the bus stop updates with a delay of 25+ seconds, giving the impression the bus has not moved.

Is this intended? Does this impact other views that use this component?

drewvolz:
In #2943, a `Timer` component was introduced to make intervals friendlier. Some of the views had their intervals changed to intervals of a minute. This does not work so well when they depend upon "near instant" updates.

Rather than update views like `BuildingHours` and `BusSchedules` at an interval of a minute or every second (which is what we had prior to the `Timer` component), I propose to set the interval across the board to 5 seconds.

I do not think 20 checks a minute is out of the question, and it does not feel like overkill.

Thoughts? Concerns? Approvals?

### DIFF
--- a/source/views/building-hours/detail/index.js
+++ b/source/views/building-hours/detail/index.js
@@ -32,7 +32,7 @@ export class BuildingHoursDetailView extends React.Component<Props> {
 
 		return (
 			<Timer
-				interval={60000}
+				interval={5000}
 				moment={true}
 				render={({now}) => (
 					<BuildingDetail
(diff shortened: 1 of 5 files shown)


drewvolz:
According to @hawkrives, this papers over the issue. There is a bug somewhere else that causes a delay, so refreshing more often is not the correct fix. We'll try to fix it another way.

PR #2979, opened 2018-09-07
Move google repo to top of the list on Android

rye:
I have no idea why this might work besides priority being messed up? All I know is that I don't like it at the bottom of the list.  We'll see if `react-native android` can finally resolve `com.android.tools.lint.lint-gradle`.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -21,20 +21,20 @@ ext {
 
 allprojects {
     repositories {
-        google()
-        mavenLocal()
-        jcenter()
         maven {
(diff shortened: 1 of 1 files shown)


hawkrives:
I've got no reason not to, but I'd almost prefer it to be second, to show that we prefer local modules over all else? Idk.

rye:
Okay, that's fine. I definitely don't want `jcenter` taking precedence is all.

hawkrives:
Yeah, I completely agree on that

hawkrives (approved):

PR #2983, opened 2018-09-08
Circle: Extract Docker image selectors using anchors (and update Ruby Docker image to 2.5.1)

rye:
We only use one image for each job, and hence don't replicate across multiple different environments. Hence, this is all that we should really need.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ x-config:
     - &docker-node
       - image: 'circleci/node:8'
     - &docker-ruby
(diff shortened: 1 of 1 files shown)


hawkrives (approved):

hawkrives:
Do we need to change that file that gets written and controls the ruby version, too?

rye:
@hawkrives, nope, we already updated that one to 2.5.1, so we were out of sync. This brings us back into parity with what we were echoing to `~/.ruby-version`.

hawkrives:
👍

PR #2984, opened 2018-09-08
Circle: Extract HOMEBREW_NO_AUTO_UPDATE to iOS Job environment

rye:
Hence make it apply to all steps in this job, which seems like it is intended behavior.

I'm tempted to use `brew bundle install`, but that would install some local development stuff too.


hawkrives (approved):
Sure

PR #2987, opened 2018-09-08
Fix up the background color of Section on Android

hawkrives:
You can see the edges of the Section component … everywhere, on Android. It's a huge annoyance for me, and I'm sorry that it's taken us so long to address it.

Now that everything uses a central Tableview component, though, I can fix it in one place and get everything fixed up!

Before | After
--- | ---
before | after


drewvolz (approved):
👍 :shipit: nice!

rye (approved):
Finally! 👏

PR #2988, opened 2018-09-08
Add Push Notifications settings

### LINKED ISSUE #2811: Delay iOS Push Notifications prompt
I don't know how long we want to delay the project, but I don't think that it's a good thing to prompt immediately after installation. I think that we will have a large number of people deny access, either on purpose or accidentally, and then not be sure how to grant it again.

hawkrives:
Closes #2811 

Channels:

- Cage "Late Night" specials (manual)
- Pause "Monthly Specials" (manual)
- SGA Announcements (manual, once we tell them)
- Emergency Campus Notifications (forced on; eventually want to hook up to e2campus for automatic notifications)

The first three are controlled by OneSignal "segments," where you target "channel: active", where `channel` is something like `channel:campus/emergency`. I've created segments in the OneSignal dashboard for each channel, so we can just choose one of those segments when we send a notification.

Things to note:

- Push notifications are disabled by default on iOS. I'm not comfortable prompting on initial app launch, and we don't have any onboarding to stick it into yet.
- Push notifications are enabled by default on Android. Furthermore, Android doesn't "disable" them… it just blocks them on-device AFAICT.
- You can disable/enable any channel, except the forced ones.
- On iOS, I show a button to go to the System settings and enable/disable app notification abilities; I didn't want to add a third layer of controls (in-app global enable/disable).
- Since I took the screenshots, I've renamed the Settings row to "Push Notifications (BETA)" to indicate that they may not be reliable, either due to technical problems or due to the manual nature of the current setup.

Questions? Comments? Concerns?

<details><summary>📸 Screenshots</summary>

What | iOS | Android
--- | --- | ---
Initial Settings view | ios-disabled-settings | N/A; Android enables by default
List, push disabled | ios-disabled-list | N/A; Android doesn't "disable" them; it just … blocks them
Permissions prompt | ios-prompt | N/A; Android enables by default
List, push enabled | ios-enabled-list | android-enabled-list
Settings, pushes enabled | ios-enabled-settings | android-enabled-settings
In Settings.app | ios-settings | android-settings

</details>


hawkrives:
Danger seems to have forgotten how to delete comments, so I deleted one for it.

rye (dismissed):

drewvolz (dismissed):
Two issues still seem to be happening. 

1. For iOS, if you first decline the prompt, you have to go into the OS Settings app to enable notifications (this is expected). Once you enable them and come back, the screen still says that push notifications are off (unless you manually refresh the app). 

2. Same process, but if you accept them, and then turn them off in OS Settings, push notification settings in AAO do not refresh according to that change.

Something something enable and disable not being called correctly? I know we have the app background listener, but we're definitely not causing the right action to happen in these circumstances. Either we're not updating onesignal at the appropriate times, or some other funkiness is happening.

hawkrives:
@drewvolz I have resolved those issues. It turns out that OneSignal.getTags never calls the callback if the user isn't subscribed to push notifications, and I had tied the permissions and tags together.

I've edited the file to use callbacks instead of promisifying them, to allow parallel execution and not block one dispatch on the other.

---

However, I now can reproduce a new bug: if you accept the permissions prompt, any channels that you select are not properly persisted to OneSignal (or something; all I really know is that the channels don't come back down if you reload the app.) 

However, once you exit and re-enter the app, the channels appear, selected like you had just left them.

Since this bug only appears to happen when you _reload_ the app, not when restarting it fully, I can't figure out how anyone would encounter this in the wild. (Our only method of restarting the app is the "reset app" button.)

Thus, I don't believe that this is a blocker.

drewvolz (approved):
RSLGTM 👍. I can live with the only bug being on app reset.

hawkrives:
When iOS finishes, I'm going to merge this; we can then either beta or wait for the nightly to test pushes on real devices.

Once this is merged, there will be no more pending items in the v2.6 milestone (woot!). I think that Drew and I will start intensive testing of every screen in the app tomorrow, and we should be able to release 2.6.0 on Sunday evening/Monday! Assuming no unexpected bugs, of course.

PR #2990, opened 2018-09-08
Fix get-urls@8 crash

hawkrives:
Supercedes #2989 

A summary of my journey:
- get-urls@8 depends on normalize-url@2, which uses the [URL global](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL)
- React Native [creates an URL global](https://github.com/facebook/react-native/blob/master/Libraries/Blob/URL.js)… but only populates it with two methods
- JSDOM maintains a [whatwg-url](https://www.npmjs.com/package/whatwg-url) package, which is a spec-compliant URL global
- Because it's intended for JSDOM/Node, it happens to depend on the global `Buffer` object

What I wound up doing: 
- add `global.Buffer` so that whatwg-url can use it
- replace `global.URL` with whatwg-url
- monkey-patch the two non-standard RN methods onto whatwg-url's URL export

This fixes the issue, according to my manual testing.

I have not written a test for this simply because we don't have the infrastructure for that quite yet. I can't just use Jest, because that runs in either a JSDOM-compatible or Node-compatible environment, both of which have appropriate URL globals, and we don't yet have any "testing" screens in Detox for an e2e test, which is approximately what we'd need for this to be accurately tested.

### DIFF
--- a/package.json
+++ b/package.json
@@ -149,6 +149,7 @@
     "tinycolor2": "1.4.1",
     "titlecase": "1.1.2",
     "url": "0.11.0",
+    "whatwg-url": "7.0.0",
     "wordwrap": "1.0.0"
   },
   "devDependencies": {
--- a/source/app.js
+++ b/source/app.js
@@ -1,5 +1,9 @@
 // @flow
 
+// monkey-patches
+import './init/global--buffer'
+
+// initialization
 import './init/constants'
 import './init/fetch'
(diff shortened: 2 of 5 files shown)


hawkrives:
~~posted early. still writing.~~ _edit:_ done writing.

rye (approved):

PR #2991, opened 2018-09-08
Downgrade react-native-vector-icons to retain the outlined icons

### LINKED ISSUE #2980: Wrong icon for unfavorited places
Unsure if this is Android-only or not, but shouldn't this be an empty heart outline? Just an idle thought.

screenshot_20180908-101944

hawkrives:
Closes #2980 
Makes Hawken happy

---

`react-native-vector-icons@5` upgraded Ionicons to v4.0.

`ionicons@4` removed almost all of the outlined icons, because Apple's _starting_ to move away from them.

But… they removed too many. The iOS Share icon, for instance, should still be outlined, but they took it away 😭. 

We don't need any of the changes in rnvi@5, so I'm just downgrading to v4.6 to get back the outlined icons.

Before | After
--- | ---
screen shot 2018-09-08 at 6 56 25 pm| screen shot 2018-09-08 at 5 37 28 pm

### DIFF
--- a/modules/filter/package.json
+++ b/modules/filter/package.json
@@ -11,7 +11,7 @@
   "peerDependencies": {
     "react": "^16.4.2",
     "react-native": "^0.55.4",
-    "react-native-vector-icons": "^5.0.0",
+    "react-native-vector-icons": "^4.0.0",
     "react-native-popover-view": "^1.0.7"
   },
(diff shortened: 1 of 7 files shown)


drewvolz (approved):
RSLGTM 👍

PR #3006, opened 2018-09-09
Change navigation icons back

drewvolz:
Revert icons to follow-up pinning rn-vector-icons. This somehow slipped through.

* Revert back to an outlined version of the favorites icon
* Revert back to an outlined version of the share icon


hawkrives (approved):
Drew noticed that I missed fixing these when I rebased my branch yesterday, after taking the screenshots (I know, take the screenshots after rebasing)

PR #3011, opened 2018-09-10
Remove course times from course search rows

### LINKED ISSUE #2967: Course rows reference old course prop
Need to update to `course.offerings` (I think). 

But it's slightly more complicated because `sis-timestring-parser` uses the old prop as well. Tracking that here https://github.com/StoDevX/sis-timestring-parser/issues/33.

https://github.com/StoDevX/AAO-React-Native/blob/586d744f443597c4f0c9cc01af2e01932a7072c4/source/views/sis/course-search/row.js#L69

drewvolz:
So, this is currently broken -- `course.times` is now `course.offerings` and the dependency that parses these also looks for the old prop. That being said, even if it were fixed, I believe there are things to be gained by not listing course times on the rows.

* Creates a less-congested course list, which I've grown to like.
* Faster renderings of the rows without running the day-groupings on each row (our largest dataset by far)
* They're still available on each course in the detail
* Closes #2967

<hr>

Conversation between @hannesmcman and @hawkrives 

> **hawkrives**: At some point, we need to quit sticking stuff in the list view lol. Not saying this is the line, but just saying. I prefer notes at the bottom.

> **hannesmcman**: I agree, but I think the course schedule is a pretty important thing to have in the listview, since it may be a deal-breaker for some people in terms of whether or not they can take a class
>
> **hawkrives**: That's what the detail view is for
"oh, this looks interesting… let's get more information"
Again, not saying "no" to the schedules here
Just noting that these cells are getting pretty long(edited)
carleton is at the other end of information density from Olaf's SIS. I hope to take some inspiration from it this fall.
>
> **hannesmcman**: Agreed. Although the vast majority of classes will only have one line of text for their schedule this way...only classes with weird schedules (interim) will have long rows

### DIFF
--- a/source/views/sis/course-search/row.js
+++ b/source/views/sis/course-search/row.js
@@ -5,13 +5,7 @@ import {StyleSheet} from 'react-native'
 import type {CourseType} from '../../../lib/course-search/types'
 import {ListRow, Title, Detail} from '@frogpond/lists'
 import {deptNum} from './lib/format-dept-num'
-import moment from 'moment-timezone'
-import {convertTimeStringsToOfferings} from 'sto-sis-time-parser'
-import {formatDayAbbrev} from './lib/format-day'
-import sortBy from 'lodash/sortBy'
 import {Row} from '@frogpond/layout'
-const CENTRAL_TZ = 'America/Winnipeg'
-const fmt = 'h:mm A'
 
 type Props = {
 	course: CourseType,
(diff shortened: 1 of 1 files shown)


hawkrives (approved):
I am approving the code changes. We should still discuss this in the comments before merging.

hawkrives:
I am in favor of only showing the course times in the detail view. 

I like compact list items.

- Title: needed to find a course
- Deptnum: needed to find a course
- Professor: _only_ needed because St. Olaf distinguishes between A/B/C courses by professor sometimes (\*cough\* REL121)
- Notes: probably doesn't need to be on the list view either

I view the list view as facilitating you finding the course you're looking for, at which point you tap on the course to see all of the information. Thus, I favor minimal information in the list view, with the detail view having the rest of the content in an accessible, well organized fashion.

drewvolz:
> Notes: probably doesn't need to be on the list view either

I'm also in favor of removing notes them from the list.

hannesmcman:
I still think the schedule should be in the ListView, but I do not feel too strongly about it, so I am ok with merging this since I am outnumbered..For the sake of argument, here is a screenshot of a branch I made which adds the (proper) schedules back to the rows. Notice that the schedule is much more compact than it was when the schedule was shown in the rows previously. 
screen shot 2018-09-10 at 10 55 37 am

hannesmcman:
Didn't we add notes to the listview because it's the only way to differentiate between FYW and BTS-B courses? @hawkrives @drewvolz

drewvolz:
We have the GEs filter to separate between FYW and BTS-B courses, no? We also list them next to the department number in the form of (FYW) or (BTS-B). Am I missing something?

hannesmcman:
Sorry. I meant that all FYW classes have the same title, and are differentiated by the notes, which tells the section topic. The same is true for BTS-B's. Screenshots:

FYW | BTB-B
-- | --
screen shot 2018-09-10 at 11 13 02 am | screen shot 2018-09-10 at 11 13 30 am

drewvolz:
Ah, that does ring a bell as to why notes were included.

hannesmcman:
@hawkrives @drewvolz What do you both think about the screenshots? Totally fine if you still think we should remove the schedules altogether.

hawkrives:
I would like to remove the schedules. 

I will file an issue with an idea about notes that I just had.

drewvolz:
@hannesmcman I don't mean to steamroll your opinion here. Your condensed screenshot is nice, however, I still feel the view could use less rows.

rye (approved):

hawkrives:
I'm going to merge this.

PR #3018, opened 2018-09-11
Fix <Timer/>

hawkrives:
It was … doing not the right thing. I don't even want to try to explain what it was doing.

Now, however, it does the correct, intended thing: given `$interval`, it waits until the next even division of `$interval` into the current time (ie, the next top-of-the-minute if given 60,000), and then sets a `setInterval` to re-invoke itself every 60 seconds.

This also adds unit tests for the `msUntilIntervalRepeat` function. (As part of that, I had to update the Jest definition, which has been outdated for a while.)


drewvolz (approved):
I like your `msUntilIntervalRepeat` solution

Issue #2195, opened 2018-01-22
Username/password fields are missing the middle divider


hawkrives:
I think that if we switch out the `<View>` that wraps them for a `<React.Fragment>`, our line will come back.

Because if we do that, the interfering View won't be there anymore – fragments aren't actually anything.

PR #3021, opened 2018-09-11
Fix missing bar between username and password fields in login screen

hawkrives:
replace React.Fragment with an array, because Section is silly and iterates over `props.children` to insert the separators into the view… and a Fragment is a single child, wheras an array gets flattened before iteration.

Yay.

Before | After
--- | ---
 | image


rye (approved):

drewvolz (approved):

drewvolz:
We're closing the closed #2195 again.

PR #3024, opened 2018-09-11
Remove right navigation button's margin

drewvolz:
At one time this was needed. It no longer appears to be needed.

Before | After
--|--
broken | pause-fixed
bug| fixed


hawkrives:
@drewvolz could you include portrait screenshots as well, please?

drewvolz:
Here's portrait

Before | After
--|--
bug-portrait | fixed-portrait

hawkrives (approved):

PR #3025, opened 2018-09-11
Rename calendar back button to "Back" to fix overlapping titles

hawkrives:
before | after
--- | ---
screen shot 2018-09-10 at 11 47 21 pm | screen shot 2018-09-10 at 11 47 09 pm

I don't like hiding the back context, but sometimes you just need to.


drewvolz (approved):
Decent idea.

PR #3034, opened 2018-09-11
Patch safeareaview in react-native-popover-view

drewvolz:
I am proposing a change that I have tested on iOS 9.3 and iOS 11.4 on an iPhone 6 and an iPhone X (and Android). 

* I am expanding the `prepare` script to run `preparePopover`.
* It patches `this.safeAreaViewReady` (which I've noticed causing an issue with iOS 9.3).
* It sets it to `true`.
* This allows the popover to open.


drewvolz:
I'm now feeding `sed` line 111 which is where the variable is assigned in `componentDidMount`

drewvolz:
If we can live with the fact that this replacement impacts the rendering of long popovers on horizontal rotation, then this gets us "working" back on 9.3.

hannesmcman:
@drewvolz What do you mean by this? 

> this replacement impacts the rendering of long popovers on horizontal rotation

drewvolz:
Until we solve the underlying issue, I thought to provide a temporary solution to get our build out the door. I'm making an exchange in accuracy in order to render.

What I meant when I said it impacts rendering longer lists while in landscape: we're getting stuck within `measureContent`.

> measureContent - Waiting - requestedContentSize: {"y":0,"width":257,"x":0,"height":653} at Popover.js line 145.

```js
measureContent(requestedContentSize) {
if (requestedContentSize.width && requestedContentSize.height && !this.waitForResizeToFinish) {
  if (this.state.isAwaitingShow) {
    if ((this.props.fromView && !this.state.fromRect) || !this.getDisplayArea() || !this.safeAreaViewReady) {
      if (DEBUG) console.log("measureContent - Waiting - requestedContentSize: " + JSON.stringify(requestedContentSize));
      setTimeout(() => this.measureContent(requestedContentSize), 100);
    } 
...
```

This is picked up from my one-line `sed` change, where I set `this.safeAreaViewReady` to `true` in `componentDidMount` (comment is in the source -- not my own emphasis):

```js
componentDidMount() {
    // I found that the RN SafeAreaView doesn't actually tell you the safe area until the second layout,
    //  so we don't want to rely on it to give us an accurate display area until it's figured that out
    this.safeAreaViewReady = false;
...
```

So I thought to make `this.safeAreaViewReady` return `true` where the rendering of the popover contents happens. 

```js
setDefaultDisplayArea(evt) {
let newDisplayArea = new Rect(evt.nativeEvent.layout.x + 10, evt.nativeEvent.layout.y + 10, evt.nativeEvent.layout.width - 20, evt.nativeEvent.layout.height - 20);
if (!this.state.defaultDisplayArea || rectChanged(this.state.defaultDisplayArea, newDisplayArea)) {
  if (DEBUG) console.log("setDefaultDisplayArea - newDisplayArea: " + JSON.stringify(newDisplayArea));
  if (this.safeAreaViewReady || !isIOS()) {
    this.setState({defaultDisplayArea: newDisplayArea}, () => {
      this.calculateRect(this.props, fromRect => {
...
```
...and thus we avoid calculating something involving the `safearea` in exchange for being able to render on iOS 9.

Something about the `safearea` isn't being calculated completely in an unknown case (whatever is running infinitely and not returning. Not calculating the `safearea` provides.... less accurate results -- the popover may not render to its original correct dimensions, but _at least it renders_. It is usable if you close it and reopen it, or simply open it in portrait.

If the majority usecase is in portrait mode, we're pretty golden with this exchange in accuracy over usability. Does that make a little more sense?

hawkrives (approved):

hawkrives:
When Drew gets into full-on QA mode, which he seems to have done in this PR, I fully trust that he's tested the known-good and known bad platforms, both iOS and Android.

drewvolz:
Hoping to not have foot-in-mouth syndrome after this one... but I did test known good/bad platforms here.

hannesmcman:
I haven't tested this yet, but are you saying that this fix will mean that the Popover will not resize correctly in landscape mode? @drewvolz

drewvolz:
@hannesmcman it's not perfect. It doesn't happen every time. But it can happen. All it takes to fix that nuance is to close it and reopen it. Here are screenshots for you to compare. If you'd like, I can remove my patch, but I have not found another way to allow the popover on iOS 9. I don't know the root cause of the problem.

Correct | Incorrect
--|--
correct | incorrect

PR #3041, opened 2018-09-12
Prettier changes to badge outline

drewvolz:
This is a leftover from #3036


hawkrives (approved):
Now I need to investigate why CI didn't fail on this

PR #3047, opened 2018-09-12
Include development bonapp menu picker

drewvolz:
@hawkrives wrote the dev picker and I felt that it'd be nice to have as an exposed tab within dev. This tool lets us explore ~1980ish bonapp cafes available thru the API. 

* Check if there are any menu items to display as to handle all cafe types and menu states.
* Stop live searching as to cut down on the api calls while typing.
* Swap out the toolbar button to be an actual functioning button.

[One nuance of android](https://facebook.github.io/react-native/docs/textinput#value): it won't let you delete and type if the value is set, so opt for not setting it from state.

> The value to show for the text input. TextInput is a controlled component, which means the native value will be forced to match this value prop if provided. For most uses, this works great, but in some cases this may cause flickering - one common cause is preventing edits by keeping value the same.

Android | iOS
--|--
android-empty | ios-empty
android-100 | ios-100
android-84 | ios-84

### DIFF
--- a/source/views/menus/dev-bonapp-picker.js
+++ b/source/views/menus/dev-bonapp-picker.js
@@ -36,7 +36,7 @@ export class BonAppPickerView extends React.PureComponent<Props, State> {
 		tabBarIcon: TabBarIcon('ionic'),
 	}
 
-	_ref: any
+	_ref: ?TextInput
 
 	state = {
 		cafeId: '34',
@@ -59,12 +59,16 @@ export class BonAppPickerView extends React.PureComponent<Props, State> {
 					<TextInput
 						ref={this.setRef}
 						keyboardType="numeric"
-						onBlur={() => this.chooseCafe(this._ref._lastNativeText)}
+						onBlur={() =>
+							this._ref ? this.chooseCafe(this._ref._lastNativeText) : null
+						}
 						style={styles.default}
 						value={Platform.OS === 'ios' ? this.state.cafeId : null}
 					/>
 					<Button
-						onPress={() => this.chooseCafe(this._ref._lastNativeText)}
+						onPress={() =>
+							this._ref ? this.chooseCafe(this._ref._lastNativeText) : null
+						}
 						title="Go"
 					/>
 				</Toolbar>
--- a/source/views/menus/dev-bonapp-picker.js
+++ b/source/views/menus/dev-bonapp-picker.js
@@ -1,6 +1,7 @@
 // @flow
 import * as React from 'react'
-import {View, TextInput, StyleSheet, Platform} from 'react-native'
+import {View, TextInput, StyleSheet} from 'react-native'
+import {NoticeView} from '@frogpond/notice'
 import {TabBarIcon} from '@frogpond/navigation-tabs'
 import * as c from '@frogpond/colors'
 import {Toolbar} from '@frogpond/toolbar'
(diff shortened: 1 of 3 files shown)


source/views/menus/dev-bonapp-picker.js
```diff
@@ -35,6 +36,8 @@ export class BonAppPickerView extends React.PureComponent<Props, State> {
 		tabBarIcon: TabBarIcon('ionic'),
 	}
 
+	_ref: any
```
hawkrives: `?TextInput`, I think

source/views/menus/dev-bonapp-picker.js
```diff
@@ -47,18 +50,27 @@ export class BonAppPickerView extends React.PureComponent<Props, State> {
...
-						onBlur={this.chooseCafe}
-						onChangeText={this.chooseCafe}
+						onBlur={() =>
+							this._ref ? this.chooseCafe(this._ref._lastNativeText) : null
+						}
 						style={styles.default}
-						value={this.state.cafeId}
+						value={Platform.OS === 'ios' ? this.state.cafeId : null}
+					/>
+					<Button
```
hawkrives: Can't we just get rid of this button in favor of the "done" button on the keyboard?

drewvolz: We're using the numeric software keyboard which does not have a "done" button. I'd advise not switching away from that keyboard because we're dealing with numeric input.

Choices:
1. Remove the "go" button and go back to live-onchange submitting
    * not a fan because I don't like hammering their API onchange
2. Keep the button as-is because it's a development utility

drewvolz: I suppose a third option could be dismissing the keyboard on swipe and then firing the search. But that's more work.

hawkrives: Yeah but I like the <kbd>done</kbd> button

hawkrives: :p

drewvolz: `returnKeyType='done'` works with "number-pad" and "numeric" on v0.47.1. Looks like we found what we wanted.

source/views/menus/dev-bonapp-picker.js
drewvolz: Thanks

drewvolz:
Yes, we can switch away from the ref. I'm also going to make a change to the `TextInput` to avoid using `onBlur`.

According to [this post](https://stackoverflow.com/questions/34053315/difference-between-onendediting-and-onblur): 

>  `onBlur` is a component function, `onEndEditing` is specific for `TextInput`. Seems like `onEndEditing` is meant to pass the contents of the `TextInput`, while `onBlur` is not meant to pass any data.
>
> * [iOS source ](https://github.com/facebook/react-native/blob/40e8d8904b006dadab639d4a9fddb926961e658a/Examples/UIExplorer/js/TextInputExample.ios.js#L74;L80)
> * [Android source](
https://github.com/facebook/react-native/blob/e8198aed8d094087d5da898a325916dd44b86d5a/Examples/UIExplorer/js/TextInputExample.android.js#L59;L65)

...specifically because Android doesn't pass anything in `onBlur`.

hawkrives:
That sounds good. I'll try to review tonight.

hawkrives (approved):

PR #3061, opened 2018-09-15
Show the bonapp dev tool on beta/nightly

hawkrives:
instead of only showing it when debugging on a simulator.


drewvolz:
We also need to tweak the redux PR. We're also debating the variable names and logic for these constants.

hawkrives:
Closing in favor of #3063

PR #3067, opened 2018-09-17
Fix dev mode check in menu

drewvolz:
Do not call the constant as a function


hawkrives:
image

isDevMode is a function

drewvolz:
I'll close this and open this on the `odds-and-ends` in settings, then. Over there it's not calling it as a function.

Issue #2964, opened 2018-09-05
Changelog?


rye:
I like changelogs. Okay, that's a lie. I used to hate changelogs. I have a love/hate relationship with them. Every big project has a changelog, though. And they make the release procedure _much_ easier, because everything that needs to go in the tag body is right there.

Here's a proposal: use a variant of the format/style of: https://keepachangelog.com/en/1.0.0/.

I think that if we put this in a file in our repository, that'll make updating it a lot easier.  I thought about putting it in the Wiki, but that would require opening and maintaining the Wiki.

Here's what I think we should use as the distinction between what gets a changelog entry and what doesn't:

## Types of ChangeLog entries:

(The `monospace` text denotes a heading to be used in the body of the version section of the changelog.)

- `Added` for new features.
- `Changed` for changes in existing functionality.
- `Deprecated` for soon-to-be removed features.
- `Removed` for now removed features.
- `Fixed` for any bug fixes.
- `Security` in case of vulnerabilities.

## What gets a changelog entry

- New features¹
- Fixes of any bugs from a previous release
- Removals/deprecations
- Changes of behavior or appearance
- Security fixes

## What does not need or get a changelog entry

Except as noted in (1),
- Code style changes
- Patch or Minor version upgrades of app dependencies²
- Any upgrades of build-related dependencies

***

¹: Any new features or changes that get made and then get changed again within the same release band get combined together. For example, if one adds a new View, and then separately fixes a bug with it before the next release, one should not create a separate changelog entry denoting the change; in either case one should either update the existing changelog entry or just count the new change as part of the previous one.

²: Development version updates (e.g. `v0.56.x` to `v0.57.y`) are considered major updates.

***

How does this sound? How do people feel about changelogs, and do we think they are useful?

We can start with this with the release of v2.6.0.

hawkrives:
I like this idea.

drewvolz:
I’m a fan

drewvolz:
Anyone see the need to have this open any longer?

rye:
No, I don't really see a need.

PR #3071, opened 2018-09-17
Start a changelog

drewvolz:
Starts implementing #2964.

* Creates the initial changelog file
* Marks entries for v2.6.0 - v2.6.3

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
 *.pbxproj -text
 flow-typed/npm/* linguist-vendored
+CHANGELOG.md merge=union
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,27 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+## [2.6.3] - 2018-09-17
+### Fixed
+- Fixed an issue where the Debug view was available outside of development
+
+## [2.6.2] - 2018-09-17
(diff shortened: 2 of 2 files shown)


rye (commented):
While I like this, I think the general style we're looking at is more technical with this.

My idea currently is that each branch that gets merged into `master` should get its own 1..n ChangeLog entries. Assuming that PR's are atomic, (as they rightfully should be) this should be a no-brainer.

This means that the Release Notes (not the ChangeLog) will be a much-condensed and cherry-picked version of the ChangeLog for that version. But I think that's the easiest way to do this.

@drewvolz, would you want to take another pass at this? If not, I can work on some changes to this PR to make it more in line with the style I think we're after.

drewvolz:
@rye how about my second pass?

rye (dismissed):
Much better.

My only other possible gripe or suggestion is that we should keep things roughly in DAG order—that is, keep them relatively the same as when they entered `master`. Obviously, if two PR's are going in at roughly the same time, it may be hit or miss, but this would be a good target to hit.

This is an excellent starting point, though. Thanks for picking up the chore work.

hawkrives (dismissed):
This looks pretty good to me!

@rye, I think that as we're adding changes, we will want to investigate the two methods for avoiding append-only conflicts in this gitlab blog post – https://about.gitlab.com/2015/02/10/gitlab-reduced-merge-conflicts-by-90-percent-with-changelog-placeholders/

rye:
@hawkrives and @drewvolz, I have added `merge=union` in `.gitattributes` to match files called `CHANGELOG.md`. See extended discussion in b336bb68164769a7dc519c0feac876daa97d587c or your inboxes for the patch.

rye (approved):
My approval is much stronger now.

rye:
Since we've two (ish) approvers, is this good to go?

PR #3079, opened 2018-09-20
bump package.json version

hawkrives:
Apple won't let us upload nightlies with the same version as the currently released app.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "AllAboutOlaf",
-  "version": "2.6.3",
+  "version": "2.6.4-pre",
   "private": true,
   "scripts": {
     "android": "react-native run-android",


rye:
Why?

Apple used to let us do this just fine.

Nightlies should have -pre appended automatically by fastlane.

hawkrives:
Is that what we were doing?

All I know right now is that last night's nightly failed to upload with a message:

> [03:15:07]: [iTMSTransporter] [2018-09-20 03:15:07 PDT] <main> ERROR: ERROR ITMS-90062: "This bundle is invalid. The value for key CFBundleShortVersionString [2.6.3] in the Info.plist file must contain a higher version than that of the previously approved version [2.6.3]."

– says https://circleci.com/gh/StoDevX/AAO-React-Native/30305

I'm fine with digging into this further and not going the route of this PR.

rye:
Yes, that's what we've been doing, if I recall correctly.

drewvolz:
Do we have a bump-version script we can use here?

rye:
>Do we have a bump-version script we can use here?

We don't, and I don't think we need one. CI has been smart enough in the past to determine that it was building a nightly and act accordingly.

Judging by the error message, though,

```
This bundle is invalid. The value for key CFBundleShortVersionString [2.6.3] in the Info.plist file must contain a higher version than that of the previously approved version [2.6.3]."
```

I would interpret this as meaning we do need to change the version after we release something. This kinda makes sense: we can't keep uploading `-pre` releases with the same base version number as something that was already deployed to production.

I don't want `2.6.4-pre`, though. `2.7.0-pre` or `3.0.0-pre`.  `master` will only contain a `-pre` for the next minor version or a beta for the next patch. This was the purpose of `stable/` branches, because we would have to cherry pick patch releases onto those branches whenever doing a release from a diverged master.

If we would rather just use linear versioning, then we should stop continuing patch cycles on master.

hawkrives:
We can, I believe, upload 2.7.0 nightlies to iTC, and then upload and release 2.6.4, as long as we don't release 2.7.0 before 2.6.4. 

iTunes will not let us release 2.7.0 and then 2.6.4. 

I think that we will need to increment the version after each release. 

I don't see why we need to make branches to do patch increments.

rye:
>We can, I believe, upload 2.7.0 nightlies to iTC, and then upload and release 2.6.4, as long as we don't release 2.7.0 before 2.6.4.

That is correct.

>iTunes will not let us release 2.7.0 and then 2.6.4.

Yes. Once 2.7.0 is live, 2.6.4 goes away.

>I think that we will need to increment the version after each release.

Yes, after each release, we need to begin targeting the next minor version.

>I don't see why we need to make branches to do patch increments.

Because we have no separation of concerns between feature work and fixes. Say we have a WIP feature that is nightly/beta-only, but we need to release a fix to the currently-deployed minor version. This happens… all the time? New features that we don't want leaking into production? What I propose—and what I thought might be usable—is to just create a `stable/vX.Y` branch that targets the latest tag. All patch releases will consist of creating a branch from that tag, targeting that `stable/vX.Y` branch, and cherry-picking relevant commits from `master` onto that branch. Once the new minor version `X.Z` comes out, that `stable/vX.Y` branch gets deleted, and a new one is created for any errata.

Functionally, this means nothing different, except that we begin imposing a rule not to release patch releases on `master`. If it means that our patch release branch from `stable/vX.Y` just contains a slew of commits from master because we haven't diverged, then so be it. No worries. The idea is that patches are supposed to be direct, clean, and low-impact, and the danger in the existing system is that someone will decide to cut `v2.6.4` with unrelated history.

Long and short is that I'd rather have `master` be a next-minor branch, and then keep a `stable/vX.Y` branch for maintenance of the current minor branch. This isn't hard.

Since this bump is not what we want, though, (`2.7.0-pre` is what we're after) I'm going to close this PR and supersede it with another one shortly.

PR #3089, opened 2018-09-27
Bump version to 2.7.0-pre for 2.7-cycle development on master

rye:
Supersedes #3079; this is more what we want.


hawkrives (approved):
I am fine with this.

Separately, we should document the need for this as part of the release process for stable builds.

PR #3091, opened 2018-09-27
CircleCI: Sleep for up to 600 seconds at start of nightly

### LINKED ISSUE #3027: Sleep randomly for up to an hour before beginning nightly
This is a common practice, especially when doing things e.g. certbot renewals, to cause the build to begin at a different time each day. Not only is this good practice, but it's also better for us, because it may allow us to avoid intermittent failures on remote servers due to large spikes of load on the hour.

rye:
Closes #3027.

This PR changes CircleCI nightlies to sleep for up to 600 seconds at the start of the nightly, by forcing `yarn` and `bundle` to wait for a sleep job to finish.

This shouldn't incur any complaints from CircleCI, because sleeping is a very low-impact kind of thing to do. Yes, it does build up some build minutes, but we are an open source project so it doesn't matter.

This is done primarily to avoid any nightly failures due to load spikes on the services we depend on.

Another choice we could make is to start the nightly at a non-start-of-the-hour time, or (even better) start our build at a random offset of up to 10 minutes from that non-start-of-the-hour time.

This additional change would just involve applying a diff like the following:

```diff
diff --git a/.circleci/config.yml b/.circleci/config.yml
index e1bb7333..a1a4529d 100644
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,7 +92,7 @@ workflows:
   nightly:
     triggers:
       - schedule:
-          cron: 0 10 * * *
+          cron: 17 10 * * *
           filters: {branches: {only: [master]}}
     jobs:
       - sleep-random
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -109,7 +109,7 @@ workflows:
 
 jobs:
   sleep-random:
-    docker: [{image: alpine:latest}]
+    docker: [{image: 'alpine:latest'}]
     steps:
       - run:
           name: Sleep for up to 10 minutes
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -95,9 +95,8 @@ workflows:
           cron: 0 10 * * *
           filters: {branches: {only: [master]}}
     jobs:
-      - sleep-random
-      - cache-yarn-linux: {requires: [sleep-random]}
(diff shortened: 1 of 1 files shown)


hawkrives (approved):
I'm approving the version that changes from 10:00UTC to 10:17UTC and does not add `sleep`s.

PR #3108, opened 2018-10-05
Privacy Policy updates

### LINKED ISSUE #2647: Update Privacy Policy
Our Privacy Policy seems to have just been… taken from the Oleville Privacy Policy, and the things that that Policy describes are not isomorphic to what we do.

I'd like to update the Privacy Policy, especially to address GDPR concerns, but also to prune down the language and make it actually useful.  I'd like users to be able to read about their privacy and actually understand where their data is.  In general, we shouldn't use the Privacy Policy as a shield to hide behind and not be transparent; this should be a living document that gets changed as new data wrangling protocols take effect.

I'll open a PR sometime soon with some proposed changes, which will obviously be open to suggestions and debate.

rye:
This change is not entirely finished, but I have gotten started, and this is where I'm at right now.

Closes #2647.


data/privacy.md
```diff
@@ -16,55 +79,71 @@ We may use the information we collect from you when you register, make a purchas
...
+Our website is scanned on a regular basis for security holes and known vulnerabilities in order to make your visit to our site as safe as possible.
+We use regular Malware Scanning.
+Your personal information is contained behind secured networks and is only accessible by a limited number of persons who have special access rights to such systems, and are required to keep the information confidential.
+In addition, all sensitive/credit information you supply is encrypted via Secure Socket Layer (SSL) technology.
+We implement a variety of security measures when a user places an order enters, submits, or accesses their information to maintain the safety of your personal information.
+All transactions are processed through a gateway provider and are not stored or processed on our servers.
 
 ### Do we use 'cookies'?
```
hawkrives: we can probably just remove this section?

hawkrives:
I just took a look at Drew's link to Panic's policy, and I like it as well.

hawkrives (dismissed):
I like it!

Do we want to mention the "Submit WiFi problem" data collection as well?

data/privacy.md
```diff
@@ -1,97 +1,73 @@
...
+  In its current form, the "suggest an update" feature requires you to
+  send an email, and this email is sent to our developer mailing list.
+
+Most importantly: **the developers and maintainers of All About Olaf
+_will not_, under any circumstances, grant access to any personally
+identifiable information that we may have collected to a third party,
+unless as required by law**.  We treat your data as confidential and
+have the utmost respect for your privacy, and we think that sharing
+data with other companies for marketing purposes or compensation is a
+blatant violation of trust; we will never share these data.
```
drewvolz: > we will never share these data

Should this be “this data”? Data always confuses me as a singular or plural.

data/privacy.md
```diff
@@ -1,97 +1,73 @@
...
+    data is intentionally vague so that we can also not use this to
+    identify you.
+  - _What you requested and how long it took_, which is widely
+    considered to be fully anonymous.
+
+- We also use a few tools to track our most used features and to
+  gather crash reports.  These tools are:
+  - _Google Analytics_, which can be disabled via the "Share
+    Analytics" toggle in the Settings view.
+  - _BugSnag_, which can also be disabled via the "Share Analytics"
```
drewvolz: I believe we still collect BugSnag reports from disabled analytics toggles... but I’d have to check the code.

data/privacy.md
```diff
@@ -1,97 +1,73 @@
...
+  credentials_.  These credentials become inaccessible when you delete
+  the app.  We only transmit these credentials to on-campus servers,
+  and the security of those servers is not controlled by us.
+
+- When you fetch any data besides your personal information, your
+  requests may reach servers that we control.  Whenever your app makes
+  such a request, the following information is logged:
+  - _Your public IP address_, for diagnostic purposes.  We do not use
+    this to identify you or for tracking purposes, and this is logged
+    for diagnostic purposes.
```
drewvolz: Nit: “for diagnostic purposes” is repeated twice. Not the worst thing.

data/privacy.md
```diff
@@ -1,97 +1,73 @@
...
+  - _The version of All About Olaf you are using and information about
+    your phone_, which is sent as part of your device's request.  This
+    data is intentionally vague so that we can also not use this to
+    identify you.
+  - _What you requested and how long it took_, which is widely
+    considered to be fully anonymous.
+
+- We also use a few tools to track our most used features and to
+  gather crash reports.  These tools are:
+  - _Google Analytics_, which can be disabled via the "Share
```
drewvolz, line 49: Do we want to mention the detailed device information that’s being collected thru GA and BugSnag?

drewvolz (commented):
Thanks again. This is something I’ve kept coming back to because having it accurately reflect our practices is important to me. I have a few comments we can discuss.

data/privacy.md
drewvolz: @hawkrives raises a good point about the WiFi issue reporting. That’d be an important section to have.

data/privacy.md
hawkrives: If we leave it plural, can we make it "data points"?

data/privacy.md
drewvolz: Oh, and something to think on. Do we want to say anything about OneSignal and notifications?

hawkrives (approved):
:shipit:

drewvolz (approved):
👍 :shipit:

data/privacy.md
drewvolz: 1. Is it worth adding links to these tools?
2. Should we link to `ccc-server` so anyone can see its source?

rye:
Woop woop! I'll remove the WIP header on this PR to indicate that it's somewhat ready to be shipped.

data/privacy.md
rye: I addressed most of your suggestions in this thread, but:

1. I don't think we really need to—this would just be distracting to me.
1. I think the link to `StoDevX/AAO-React-Native` is sufficient; anyone who wants to dig deeper will probably realize rather quickly how this works.  Since it's referred to as "our server" I don't think it'll really need much more detail.

Feel free to mark this conversation as unresolved if you think this is a blocker.

drewvolz:
Anybody want changes to anything? Looks good to me. Ship it?

rye:
Ship for now? (@hawkrives)

hawkrives:
LGTM.

hawkrives (approved):

PR #3136, opened 2018-10-20
fix "yarn run prettier"

hawkrives:
It looks like Yarn changed their binary/script priority in favor of binaries, which is a reversal of the status quo. IOW, `yarn prettier` will now run `./node_modules/.bin/prettier` instead of the package.json script "prettier".

Thus, I have changed the package.json script to "pretty".

Regarding the other change... For a number of months now, I've been using "yarn p" as an alias for "yarn pretty-quick", and I really like it, so I'd like to change that here as well. (The old names pre-date `pretty-quick`.)


rye (dismissed):

rye (dismissed):

hawkrives:
Tested via 

```
env task=JS-general DANGER_DISABLE_TRANSPILATION=true ./node_modules/.bin/danger pr https://github.com/StoDevX/AAO-React-Native/pull/3136
```

rye (approved):

drewvolz (approved):

rye:
Merging!

PR #3138, opened 2018-10-20
Use _our_ build tools versions when compiling dependencies

rye:
This explicitly forces each subproject to use our version of the build tools, and suppresses any warnings about out-of-date build tools versions.  It does so by redefining the compileSdkVersion and buildToolsVersion for each subproject after they have been evaluated.

For what it's worth, this does compile and run properly locally.

This _could_ go into `android/app/build.gradle` and use the defined constants from there? Something like that?

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -38,3 +38,14 @@ allprojects {
         }
     }
 }
+
+subprojects {
+    afterEvaluate { project ->
+        if (project.hasProperty("android")) {
+            android {
+                compileSdkVersion 28
+                buildToolsVersion "27.0.3"
+            }
+        }
+    }
+}


hawkrives:
> This could go into `android/app/build.gradle` and use the defined constants from there? Something like that?

I like that!

hawkrives (approved):
I like it! Especially with this:

> This could go into android/app/build.gradle and use the defined constants from there? Something like that?

(otherwise I'm afraid we'll forget to change them and get confused)

rye:
While I'd love to just up and copy my added lines over to `android/app/build.gradle`, we can't do that because then they wouldn't apply to our subprojects. (`android/build.gradle` handles our app plus our deps)

rye:
@hawkrives, would you be fine with me putting my self-suggestion into a tracking issue and merging this as-is? This lets us use this on our other PRs and helps me keep "pretty safe" config changes away from breaking config changes…

PR #3146, opened 2018-10-21
Use the US/Central timezone, and centralize logic for it

rye:
This could have been two PR's, but it's not.

First, I changed every time zone to US/Central. Then I made a module and moved everything into that module.

Jest broke because init is not called beforehand.  So I needed to include a fallback, which I did.


hawkrives (approved):
LGTM

hawkrives:
(I've wanted to do this for literally years now, so thanks :-))

drewvolz:
Shoutout to https://github.com/StoDevX/AAO-React-Native/issues/807 even if we don’t touch it going forward

rye:
@drewvolz we can explore that in #next-PR, yes?

rye:
Merging.

PR #3147, opened 2018-10-24
Remove the check command

drewvolz:
I feel that these can be run individually and are also run via CI individually.


hawkrives (approved):

rye (approved):
These would best be a script in a `script` directory rather than part of our node package. LGTM.

PR #3161, opened 2018-11-01
Add `yarn d` deduplication script

rye:
This can be used to reconcile accidental duplication that yarn did not clean up for itself.

This PR also adds a `yarn-dedupe` step and Danger configuration to handle duplication.

We _could_ actually just do `yarn-deduplicate --list yarn.lock` which will produce a list of things that could get changed. However, that script does not return `1`, which would make it a bit harder to change things.  If we want CI to be more neutral about duplication, however, that would be an approach—we could just have Danger comment on the PR.

#### Why did I add something to `.flowconfig`?

https://github.com/atlassian/yarn-deduplicate/issues/14
https://github.com/atlassian/yarn-deduplicate/pull/15

That's why.


rye:
I'll upgrade `yarn-deduplicate` and remove the patch for flow.

hawkrives (approved):

PR #3162, opened 2018-11-01
Upgrade react-native-vector-icons to version 6.0.2

rye:
This skips forward a bit, but keeps us in sync with some of our dependencies and allows for further deduplication.

Yes, we've already done this before with #2710 and #2991, but I think we should be working on tracking forward.  If that means this needs to turn into an epic to also upgrade any lacking parts of our design, then so be it.

* * *

#### v6.0:
- Upgrade to gradle build tools 3.1.4/compile -> implementation
- **Update minimum supported iOS version to 9**

#### v5.0:

- Use fontReference instead of fontFamily for web.
- Fix Octicons padding/centering. (If we have manually compensated, we may need to fix that.)


hawkrives:
We reverted because the filled-in icons were ugly for the toolbar icons. 

We've already bitten the bullet and removed the outlined icons in the tab bars.

rye:
@hawkrives so does what you say mean that we can go ahead with this, or does it mean that we can't because we have designed around it thus far?

rye:
This needs a partial `react-native link` done on it to update `android/app/src/main/assets/fonts/Ionicons.ttf`, as some of the icons have changed…

rye:
For what it's worth, I have tested this now and the icons look great on Android after I updated the .ttf file.

hawkrives:
@rye The issue was with how to represent active/inactive icons in the toolbar. 

Namely Building Hours' "favorite" button. 

Currently, we use outline for inactive and solid for active. 

---

Originally, we had an issue with the tab bar icons as well, but we've designed around that by using accent colors / grays / stuff, so it's really just the toolbar icons.

rye:
@hawkrives I fixed those icons and got them working on v6. `-outline` was replaced by `-empty`, and that's all I had to account for. See the commits.

modules/navigation-tabs/tabbar-icon.js
```diff
@@ -17,6 +17,14 @@ type Props = {
 	focused: boolean,
 }
 
-export const TabBarIcon = (icon: string) => ({tintColor}: Props) => (
-	<Icon name={`ios-${icon}`} style={[styles.icon, {color: tintColor}]} />
-)
+export const TabBarIcon = (icon: string) => ({tintColor}: Props) => {
+	if (icon === 'ionic') {
+		return (
+			<Icon name={`logo-${icon}`} style={[styles.icon, {color: tintColor}]} />
```
hawkrives: The "ionic" icon is used in the dev tool for changing the BonApp menu. It should be changed there, not in TabBarIcon.

hawkrives (dismissed):

modules/navigation-tabs/tabbar-icon.js
hawkrives: I see your other reply now. I shall think.

hawkrives: I propose that we change from "ionic" to "medal", since that's just a random icon that I pulled out for dev builds a long time ago.

rye: No.

See what I said about this:

image



The current version of `TabBarIcon` prefixes `icon` with `ios-` no matter what its contents were, so if I revert the changes in this file, I would get `ios-logo-ionic`. The only other way I thought of doing this was changing all usages of `TabBarIcon` to include their prefix (i.e. changing `TabBarIcon('blah')` to `TabBarIcon('ios-blah')`) and changing `TabBarIcon` so that it does not prefix its input. Would you prefer that?

hawkrives: I want to continue being able to call TabBarIcon with no prefixes. I also want to keep TabBarIcon simple; I have no attachment to the "ionic" icon. 

Thus, I would like to remove the use of the ionic icon from Menus, allowing us to continue using the simple code.

hawkrives (approved):

drewvolz (approved):

rye:
Thanks to the both of you for helping me get this landed!

hawkrives:
Filed an issue for icon tweak followups in #3461

PR #3166, opened 2018-11-06
Remove login credentials from redux

hawkrives:
> yeesh
> – @hawkrives, today

hokay. This PR's purpose is to remove credentials from Redux, but I also finished removing Balances caching because apparently that was still around (although it should have never been used, thanks to the `forceFromServer` flag.)

---

I enabled the ESlint [`eqeqeq: smart`](https://eslint.org/docs/rules/eqeqeq) rule, which allows `x == null` comparisons to avoid having to do `x === null || x === undefined`, but won't allow `x == 1` or `x == "str"`.

I also added `varsIgnorePattern: ^_` to [the `no-unused-vars` rule](https://eslint.org/docs/rules/no-unused-vars), to match what we allow in function arguments.

---

I got to remove the 144 lines of Balances caching from `lib/cache.js`.

---

Instead of storing username/password in Redux, I now load it on-demand in the Settings view's Login section, and in StoPrint's PrintJob view.

---

I also removed the `isConnected` section from the App part of Redux, because it's considered better practice to try and then handle the network failure instead of checking and then hitting the network. (Otherwise, the network may vanish between the check and the request, and also checking `isConnected` is somewhat unreliable in the first place – what really is a network connection? what if the thing you want to access is just down, but the rest of the internet is working? what if the device is on wifi but has no internet? etc.)

See also ["Boy, Have I Been Misusing SCNetworkReachability"](http://blog.jaredsinclair.com/post/179083065595/til-boy-have-i-been-misusing), [`isMounted` is an Antipattern](https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html), and the similar reasoning in [the Node docs for the `fs` module](https://nodejs.org/dist/latest-v11.x/docs/api/fs.html#fs_fs_exists_path_callback):

> Using `fs.exists()` to check for the existence of a file before calling `fs.open()`, `fs.readFile()`, or `fs.writeFile()` is not recommended. Doing so introduces a race condition, since other processes may change the file's state between the two calls. Instead, user code should open/read/write the file directly and handle the error raised if the file does not exist.

---

`lib/login.js` now returns an enum of values, instead of a boolean, to allow more detailed reporting – no credentials, bad credentials, no network, or an actual login success.

---

There were also the couple of misc. code cleanups that I couldn't resist while I was wandering through.

---

Now that `isConnected` is gone from Redux's app reducer, I removed it entirely; it was only actually used for the `isConnected` state.

---

I removed the concept of "validating" the Olaf credentials; instead, we just try the login, and then we can handle failures at the time of us trying to use the credentials. As before, having a separate concept of "valid" is subject to race conditions; for instance, someone may change their password while the app is in the background. 

> edit: this is now under re-consideration, due to Hawken's confusion about being logged in in app settings after a reload

> edit 2: this is still gone; my change was to trigger a login when the Settings view mounts, so that we can display the "log out" button if your credentials are valid

---

The changes in `login-credentials.js` should also fix the bug that I've never filed, about how if you resume the app on the Settings page, your credentials never populate the form fields.

---

~~If you are logged out or your credentials are invalid, I now only show the "open Settings" button, instead of showing the empty boxes and then the "open Settings" button.~~

> edit: this has been reverted, because the screen just looked sad

---

… that should be everything.

Now I shall try to open and run the app. Expect more commits later.


hawkrives:
@rye, @hannesmcman, @drewvolz This is now ready for review. I'd like to get this into tomorrow's nightly.

hannesmcman (dismissed):
Just tested this. Found a few bugs. 

I also am not quite sold on the notion of completely removing login info from redux state. For example, we don't want the login action to fire every time we open the settings view, right? On this branch, navigating to settings triggers the login action, which shows the "Logging in..." text until the action is completed, rather than presenting the "Sign out" button right away. 

Another related bug that I noticed was that if you open the app without ever navigating to settings, the balances view thinks that you're not logged in, presenting the "Log in to St. Olaf button".

hawkrives:
@hannesmcman I actually do want to fire the login action every time you open settings…

My rationale is that there's really no way for use to accurately store the login state. No matter when we check it, the credentials could have become invalid by the next time we try to use them – a password change, for example – so we ought to handle the failure case everywhere anyway.

Balances saying "you're not logged in"… grumble. I thought I fixed that before I opened this. I'll take another look.

hawkrives:
@hannesmcman you're probably right.

I'm talking about "handling login failures everywhere" and then I'm only calling the "log in" method (which is the only one with proper error handling) in one place.

Let me think about this some more.

hawkrives:
I'm realizing that we have several APIs that all handle auth errors differently, and I'm wondering if we can unify them. (StoPrint, OleCard, and I suppose the general "validate login credentials" call.)

Still thinking about that.

hannesmcman:
Maybe we keep login credentials in redux state until a request (either stoPrint, balances, or whatever) fails, at which point we make you log back in?

hawkrives:
I honestly really dislike keeping the credentials themselves in redux – I want them to be out of the keychain for as little time as possible… it'd be too easy to, say, dump the entire state of their redux store for, let's say a new debugging/support feature, and forget that we store passwords in it.

I think I've almost got something, too. I'm defining more of a consistent interface between the three things that do login (verify, stoprint, and olecard)

rye (approved):
I'm just gonna give this a thumbs up for now.

hawkrives:
I'm heading to bed; I'll work on this more in the morning.

drewvolz (commented):
Builds and runs on iOS and Android simulators. Otherwise, I can't test the APIs.

rye:
I'm building and running this on an emulator—results incoming.

rye:
As far as normal usage, it seems to work fine. The Settings view has a couple of observable phases when you first open it (it first presents the "Log in" with the credentials filled, then it logs in…)

In principle, I dislike the idea of storing anything privileged via redux. It's a gigantic security issue. Login status and a timestamp for predictive thought would be as far as I would go.

rye:
Can we make a decision about whether or not we want to merge this?

Bumping because I still think this is nightly-ready.

hawkrives:
I think we can put it on nightly; I do want to make a few more changes before we ship, but they can come after the nightly. 

Things to do:

- Push the "login info" into a new view nested inside of Settings, with login status rows for "OleCard" and "StoPrint", so we can validate the login against each service (they all use the same credentials, but then we can re-use each service's logIn function to validate the login.)
- Only check the login info when you open that view, instead of when you open Settings
- Trigger OleCard fetch (which implies a login, so that we get a new session) when you open OleCard
- Trigger StoPrint jobs fetch (which implies a login, so we get a session) when you open StoPrint
- Make sure that all StoPrint methods handle 403 errors and automatically trigger login-and-retry behavior
- Make sure OleCard and StoPrint both retry with exponential fallback 3 times, then somehow point you to the new Login view in Settings (we can push just that view from both locations, too)

rye:
@hawkrives and @hannesmcman, then I propose we merge, try it out in the nightly, and open issues with feedback—as well as the suggestions made here.

I will dismiss @hannesmcman's review to make the merge button green. But also because I think we came to a consensus that this seeming regression was actually an important step towards more dependable behavior.

hawkrives:
(Hannes did notice that if you open the app without going to Settings, SIS won't know that you're logged in, because that state doesn't get updated properly right now. But I'm working on fixing that.)

PR #3178, opened 2018-11-14
danger: Add a rule to check whether or not the CHANGELOG was updated

rye:


rye:
Maybe I should update my regex to `/\.js$/`?

rye:
Good job, Danger! A correct assessment of this PR. I shall make a change to the CHANGELOG.

In all seriousness, the "Enforce Changelogs" example on https://danger.systems/js/ is a much simpler approach. It would also potentially enable a "trivial" label to be added to PRs that don't matter.

However, I think checking _which_ files were changed is also a good thing, since I don't think we really need to note e.g. data or dep changes in the changelog.

rye:
In order to preserve linear history, we should merge #3177 before this one.

hawkrives (dismissed):

hawkrives (dismissed):

hawkrives (approved):
donk

rye:
Mergey mergey.

PR #3181, opened 2018-11-15
Fix Failed Login Bug

### LINKED ISSUE #2847: Balances should hide "login needed" text until the login fails
It's really odd … a lot of the time, when I open the app and go into SIS, it fetches my updated balances but also says I need to log in. 

The auto-login-at-app-launch seems to be failing.

hannesmcman:
All this PR does is set the default number of login attempts to 2. I have tested this and it works, since the login request only fails on the first request made after booting the app up. This is definitely a hot fix, and may not be needed in the future if the login logic changes. Resolves #2847.

### DIFF
--- a/source/lib/financials/balances.js
+++ b/source/lib/financials/balances.js
@@ -1,7 +1,5 @@
 // @flow
-import {loadLoginCredentials} from '../login'
-import buildFormData from '../formdata'
-import {OLECARD_AUTH_URL, OLECARD_DATA_ENDPOINT} from './urls'
+import {OLECARD_DATA_ENDPOINT} from './urls'
 import type {BalancesShapeType, OleCardBalancesType} from './types'
 
 type BalancesOrErrorType =
--- a/source/lib/financials/balances.js
+++ b/source/lib/financials/balances.js
@@ -9,24 +9,7 @@ type BalancesOrErrorType =
 	| {error: false, value: BalancesShapeType}
 
 export async function getBalances(): Promise<BalancesOrErrorType> {
-	const {username, password} = await loadLoginCredentials()
-
-	if (!username || !password) {
-		return {error: true, value: new Error('Not logged in')}
-	}
-
-	const form = buildFormData({username, password})
-
 	try {
-		let loginResponse = await fetch(OLECARD_AUTH_URL, {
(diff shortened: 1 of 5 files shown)


rye (dismissed):
Do we want to try exponential back off? This is a good first starting point though.

hawkrives (dismissed):
Could you change the calling location instead of the default argument, please?

hannesmcman:
Ok so I was annoyed by a lot of little quirks of how/when we were performing logins, so I changed some more stuff. Below is a summary of the changes I made:

1. Only log in before balances update if the redux loginState is not 'logged-in'
   - This one I am not entirely sure about, as I removed the login call to the OLECARD_AUTH_ENDPOINT from within the `getBalances()`, which might need to be undone. I was trying to unify logins to the redux tracked `loginWithCredentials` function. @hawkrives LMK if this change needs to be reverted. Idk what the timeout for the OLECARD_BALANCES endpoint is..
2. Do not store credentials in keychain if login failed because of bad credentials
3. Initiate redux `loginWithCredentials()` function when refreshing balances data if not logged in. 
4. Other minor tweaks :)

hawkrives:
Yeah, we need to always log in when we get balances.

hawkrives:
Ideally, I'd like to remove the redux loginWithCredentials entirely. I didn't get around to it last night, but I'm going to work on that this weekend.

hannesmcman:
@hawkrives Why do you want to completely remove loginState from redux again? I understand removing credentials from redux state, but I think that keeping loginState in redux allows us to centralize error handling with logging in, rather than having to handle login errors within every individual component that needs to login..

hannesmcman:
I agree that something needs to change, perhaps loginWithCredentials itself, but I'm not entirely convinced we should remove loginState from redux entirely.

hawkrives:
Because loginState is too ephemeral for redux IMO. It's more a property of the request than a persistent variable.

If you successfully log in in Settings, there's no guarantee that you'll be able to fetch balances – your password might have changed, or your account might have been disabled, or… so you have to handle all of the cases everywhere you make an authenticated request, too.

I just don't see a benefit to tracking `is logged in` separately.

PR #3182, opened 2018-11-17
Add "open webpage" row to student work detail

### LINKED ISSUE #3169: Add “open webpage” row to student work detail
Rather than make someone hit the share button, it might be helpful to provide the original webpage as part of the detail listing

drewvolz:
Closes #3169

* Introduces a function to create a full job URL from the id
* Shows a pushbutton on iOS to open a webpage
* Shows a card with text on Android to open a webpage

iOS | Android
--|--
screen shot 2018-11-17 at 2 23 28 pm | android-jobs


hawkrives:
I don't know that I like the wording "Open original posting webpage". 

I don't have anything better though. 

I also don't think that the Android one feels like a thing you can interact with, but I'll defer to Kris on that I think.

drewvolz:
> I don't know that I like the wording "Open original posting webpage". I don't have anything better though.

I feel similarly, and can't think of an alternative.

> I also don't think that the Android one feels like a thing you can interact with, but I'll defer to Kris on that I think.

That's our current convention with android links... is text inside a card. Would you prefer the url to be visible instead?

rye (commented):
I may do a more thorough code review, but towards Hawken's suggestion I think the wording should be "Open posting".

On Android, I would honestly prefer to do away with the card and use a FAB. I also think the usage of the `open` ionicon would help a lot. Maybe one or the other, or both.

rye:
I think a FAB is an excellent style element that lots of Android users anticipate, but we don't use.

hawkrives:
Okay, but isn't a FAB intended for the "primary" action in a view?

Although I'm not sure what other actions you would take from the view…

rye:
@hawkrives, you're exactly right.

From [the Material spec](https://material.io/design/components/buttons-floating-action-button.html):

>A FAB promotes an important, constructive action such as:
>- Create
>- Favorite
>- Share
>- Start a process

I think of opening the webpage as "starting a process".  Of course, for strange job listings like the lovely one that we have here, people should instead be going to the google form, but I digress. Ideally, the Student Work postings page would adequately render on a phone so that people can get to the application through that page.

Since this view is basically just a representation of the data on a web page, I think having a FAB with the `open` icon would not be considered a minor action in this view.

Like this, but round…:

FAB with icon-open, but rectangular

I think anything short of a button for this is kinda strange on Android.

Also, can someone remind me why links don't look like links?

drewvolz:
> Also, can someone remind me why links don't look like links?

We have that ability on iOS thanks to TextInput but we don’t have a natural link parser for android

source/views/sis/student-work/detail.android.js
```diff
@@ -165,6 +165,20 @@ function Timeline({job}: {job: JobType}) {
...
 
+function OpenWebpage({job}: {job: JobType}) {
+	return job.id ? (
+		<Card header="Webpage" style={styles.card}>
+			<Text
+				key={job.id}
+				onPress={() => openUrl(createJobFullUrl(job))}
+				style={styles.cardBody}
+			>
+				Open original posting webpage
```
hawkrives:
```suggestion
				Open Posting
```

source/views/sis/student-work/detail.ios.js
```diff
@@ -151,6 +158,17 @@ function Timeline({job}: {job: JobType}) {
 	) : null
 }
 
+function OpenWebpage({job}: {job: JobType}) {
+	return job.id ? (
+		<Section header="">
+			<PushButtonCell
+				onPress={() => openUrl(createJobFullUrl(job))}
+				title="Open original posting webpage"
```
hawkrives:
```suggestion
				title="Open Posting"
```

hawkrives (changes requested):
I think we're OK with not doing a FAB in this PR; that will be a much more significant change.

Let's file an issue and do the FAB in another PR, as much as I want it.

source/views/sis/student-work/detail.ios.js
rye: Downcase second word?

hawkrives: On iOS, at least, these are usually in title case

rye:
@hawkrives I am okay with doing FABs elsewhere. I will admit that they are out of scope since we never have done them and I think that'll require an added dependency.

hawkrives (approved):

rye (approved):

PR #3184, opened 2018-11-18
[RN Upgrade] iOS: Remove ReactCommon from HEADER_SEARCH_PATHS in buildSettings

rye:
This was done at the suggestion of a couple of random Internet people.

It seems to properly build… but the question is whether or not this is the only such required change.  And if this is indeed a good idea.


hawkrives (approved):
Let's go ahead and merge this; I'm going to operate with the philosophy of "this is ~C; if it needed those headers, it would have yelled."

Kris, I know you're doing more work on this, but let's merge this one in and I'll start poking at Detox?

rye:
@hawkrives, sounds good. I'll merge this now.  I'll have another PR that prunes off the maximum number of HEADER_SEARCH_PATHS possible shortly.

PR #3185, opened 2018-11-18
[RN Upgrade] iOS: Remove most HEADER_SEARCH_PATHS

rye:
This PR is an interesting change.  Removes most of our HEADER_SEARCH_PATHS, without having much of a noticeable impact on the build outcome.  Detox is failing, but that's just because it's cranky at the moment.

The only reason that this causes problems is if native dependencies are included by our source, but most of them compile entirely internally and we don't shell out to their native code yet. Soon, maybe. But not for now.

Had to keep OneSignal's headers in scope so that `ios/OneSignalNotificationServiceExtension/NotificationService.m` was still happy.


hawkrives (approved):
So the only reason I'm hesitant about this is future debugging things - I'm afraid I'm going to bang my head against a wall for hours in the future when something winds up needing this. 

But I'm game to merge it.

rye:
Yeah, @hawkrives, I hear ya.

There was a layer of indirection today when I had to go and google the initial error message to figure out that we could do this.

*shrug*

I don't think this will really need be touched again, though. `react-native link` should probably take care of this? No clue.

Anyhow. Merging. 🚀

PR #3188, opened 2018-11-18
circle: Suppress Detox build stdout

### LINKED ISSUE #3186: Detox: Suppress build output
Since we build the app in the `Run Fastlane` step, can't we ignore the build output (at least to stdout) when running `yarn detox build --configuration ios.sim.release`? It's basically the same output, but unprettified.

rye:
It's very noisy. Let's turn it down a notch.

Closes #3186.


hawkrives (commented):
Oh, I forgot that there was a separate step. 

Um. 

Given that Circle collapses this row after completion, why do we even need to hide this?

rye:
Given that this output is redundant, why do we need to keep it?

Also, it lags out my browser. It's _very_ noisy.

hawkrives:
Have you tested that it properly prints errors when it errors?

hawkrives:
I'm still not convinced that this won't rob us of debugging output that we might need in the future.

I would prefer to not hide this.

rye:
We should see if we can pipe this to `xcpretty`, then.  I don't like `xcodebuild`'s unnecessary verbosity.

rye:
Interestingly enough, Detox themselves use `xcpretty`. 

https://github.com/wix/Detox/blob/5ef4ae1fa2731b94e95faf0a65920bf9298af0c8/scripts/unit.ios.sh#L3-L4

I have pushed a change that will give that a try.

hawkrives (dismissed):

hawkrives:
I'm nervous about piping test output through xcpretty - it's Jest, IIRC, not Xcode testing, so xcpretty could break it at any time…?

rye:
@hawkrives You will notice that (1) the linked Detox unit `test` step pipes to `xcpretty`, and (2) build [35185](https://circleci.com/gh/StoDevX/AAO-React-Native/35185) includes piping non-`xcodebuild` output to `xcpretty`, and does not fail.

rye:
As such, I'm pretty sure xcpretty is just doing pattern matching. If it doesn't find any patterns, they don't get formatted.

hawkrives:
Yes, but I'm asking if we can rely on that not breaking in the future.

hawkrives:
It definitely hides some of the output lines

hawkrives:
I'd need to at the _least_ see a test failure piped through Detox to make sure it doesn't hide the actual fails, because that'd be bad.

.circleci/config.yml
```diff
@@ -338,8 +338,8 @@ jobs:
           command: bundle exec fastlane ios ci-run | tee ./logs/build
           environment:
             GIT_COMMIT_DESC: $(git log --format=oneline -n 1 $CIRCLE_SHA1)
-      - run: yarn detox build --configuration ios.sim.release
-      - run: yarn detox test --configuration ios.sim.release --cleanup
+      - run: yarn detox build --configuration ios.sim.release | xcpretty
+      - run: yarn detox test --configuration ios.sim.release --cleanup | xcpretty
```
hawkrives: Let's just not pipe test output through `xcpretty`

```suggestion
      - run: yarn detox test --configuration ios.sim.release --cleanup
```

rye:
I will make a branch off of this branch, invert at least one of the assertions in our tests, and watch that fail. We'll see if it actually slurps the output. (I'm 99% sure it'll still work.)

If you actually read [the business logic in the xcpretty source](https://github.com/supermarin/xcpretty/blob/4b90228e80dd74dbdb209553668178023447ff85/bin/xcpretty#L83-L86), xcpretty is simply reading from STDIN and writing to STDOUT, and the [printer](https://github.com/supermarin/xcpretty/blob/4b90228e80dd74dbdb209553668178023447ff85/lib/xcpretty/printer.rb#L18-L24) is simply forwarding (in our case to the [simple](https://github.com/supermarin/xcpretty/blob/4b90228e80dd74dbdb209553668178023447ff85/lib/xcpretty/formatters/simple.rb) formatter) the input; if Jest manages to output something that triggers an EMPTY handler, then that's really special and we can handle that.

hawkrives (dismissed):

.circleci/config.yml
rye: Why, though? I still don't understand how xcpretty, which is only slurping stuff that looks like `xcodebuild` output, would possibly slurp jest output.  Expect the Proof of Concept that it actually does report failures shortly, but I see no reason to make an exception and keep the noise in the test output, especially if that's wholly unnecessary.

hawkrives:
@rye okay, but what does piping our Jest test output to xcpretty actually get us? even hypothetically.

I see only potential downsides, and no reason to pipe it in.

.circleci/config.yml
hawkrives: I see no reason to even pipe it through. By that argument, why not pipe literally every command through xcpretty? 

I just don't see any reason to pipe this through.

rye:
> @rye okay, but what does piping our Jest test output to xcpretty actually get us? even hypothetically.

The output of detox is not only Jest.  There's also xcodebuild output, is there not?

- - -

Okay, there's not.

image

becomes

image

- - -

Fine, I'll remove it.  But not because it might be eating output—only because there isn't `xcodebuild` output. Also, the colors go missing when it goes through `xcpretty`.

hawkrives:
> Fine, I'll remove it. But not because it might be eating output—only because there isn't xcodebuild output. Also, the colors go missing when it goes through xcpretty.

I mean, that was kinda my argument – there's no point to sending the test output through xcpretty because there's nothing but `jest` output.

And, I will note that it _is_ eating output, currently – there's no more `yarn` ending line, for instance.

hawkrives (approved):

rye:
FWIW, here is a [deliberate failure](https://circleci.com/gh/StoDevX/AAO-React-Native/35197) which demonstrates that failure context is not messed with.

But whatever.

PR #3189, opened 2018-11-24
Update express bus schedule

rye:
Per https://threeriverscap.org/transportation/hiawathaland-transit/northfield-service-area:

Screenshot of Three Rivers' Hours table

I don't think it's just from 6–9.


hawkrives:
… I was going to ask that someone wander over to Buntrock at 5 and check, but I think I'm the only one around and I don't feel like making the trek :-)

hawkrives (approved):

PR #3209, opened 2018-11-27
Automatically skip native builds if not needed

### LINKED ISSUE #2582: Update "android" and "ios" to skip native builds on data/ changes
If the only changes are in data/, we should skip running the "native" builds (`bundle exec fastlane`) to speed up PR times. 

We'll still have queue the VMs, but that's a non-issue on Circle. 

~Should we extend that to the jest and flow jobs too?~

hawkrives:
Closes #2582 

```
[22:00:03]: -------------------------------------------------
[22:00:03]: --- Step: git diff --name-only 'master' 'foo' ---
[22:00:03]: -------------------------------------------------
[22:00:03]: $ git diff --name-only 'master' 'foo'
[22:00:03]: ▸ .circleci/config.yml
[22:00:03]: should_build? some native file changed, so yes
```

(this PR should trigger a native build; submitting early to test.)

---

With this PR, `fastlane` will automatically skip native iOS/Android builds when only JS code has changed, as it can't really affect the native results.

So we check if any of these files have changed:

```ruby
  source_globs = [
    '.circleci/**',
    'e2e/**',
    'fastlane/**',
    'scripts/**',
    'Gemfile.lock',
  ]

  case lane_context[:PLATFORM_NAME]
  when :android
    platform_globs = [
      'android/**',
      '{modules,source}/**/*.java',
    ]
  when :ios
    platform_globs = [
      'ios/**',
      '{modules,source}/**/*.{h,m,mm}',
    ]
  end
```

_or_, if any package.json dependencies/devDependencies that include the word "react" or "jsc" have changed.

### DIFF
--- a/fastlane/lib/util.rb
+++ b/fastlane/lib/util.rb
@@ -110,7 +110,7 @@ def should_build?
   end
 
   # 1. need to get files changed in the current build since master
-  changed_files = sh("git diff --name-only '#{source_branch}' '#{current_branch}'").lines
+  changed_files = sh("git log --name-only --format='' '#{source_branch}'..'#{current_branch}'").lines.sort.uniq
 
   # 2. check for "packages we care about"
   if changed_files.include? 'package.json' and npm_native_package_changed?
--- a/fastlane/lib/util.rb
+++ b/fastlane/lib/util.rb
@@ -51,6 +51,17 @@ def deps_diff(old_hash, new_hash, path)
   (new_hash.dig(*path).to_a - old_hash.dig(*path).to_a).to_h
 end
 
+def npm_native_package_changed?
+  old_package = JSON.parse(sh "git show '#{source_branch}:package.json'")
+  new_package = JSON.parse(sh "git show '#{current_branch}:package.json'")
+
+  changed_dependencies = deps_diff(old_package, new_package, ['dependencies']).keys
+  changed_devdependencies = deps_diff(old_package, new_package, ['devDependencies']).keys
+
+  dep_name_regex = /react|jsc/
+  (changed_dependencies + changed_devdependencies).any? { |dep| dep =~ dep_name_regex }
+end
+
 # checks if the native build needs to be run
 def should_build?
   # Essentially, we want to avoid native builds if the only thing that changed was JS code.
@@ -71,18 +82,9 @@ def should_build?
 
   changed_files = sh("git diff --name-only '#{source_branch}' '#{current_branch}'").lines
 
-  if changed_files.include? 'package.json'
-    old_package = JSON.parse(sh "git show '#{source_branch}:package.json'")
-    new_package = JSON.parse(sh "git show '#{current_branch}:package.json'")
-
-    changed_dependencies = deps_diff(old_package, new_package, ['dependencies']).keys
-    changed_devdependencies = deps_diff(old_package, new_package, ['devDependencies']).keys
-
-    dep_name_regex = /react|jsc/
-    if (changed_dependencies + changed_devdependencies).any? { |dep| dep =~ dep_name_regex }
-      UI.message('should_build? some react|jsc dep changed, so yes')
-      return true
-    end
+  if changed_files.include? 'package.json' and npm_native_package_changed?
+    UI.message('should_build? some react|jsc dep changed, so yes')
+    return true
   end
 
   source_globs = [
--- a/fastlane/lib/util.rb
+++ b/fastlane/lib/util.rb
@@ -62,6 +62,39 @@ def npm_native_package_changed?
   (changed_dependencies + changed_devdependencies).any? { |dep| dep =~ dep_name_regex }
 end
 
+def native_build_globs
+  base_globs = [
+    '.circleci/**',
+    'e2e/**',
+    'fastlane/**',
+    'scripts/**',
+    'Gemfile.lock',
+  ]
+
+  case lane_context[:PLATFORM_NAME]
+  when :android
+    base_globs + [
+      'android/**',
+      '{modules,source}/**/*.{java,gradle}',
+    ]
+  when :ios
+    base_globs + [
+      'ios/**',
+      '{modules,source}/**/*.{h,m,mm}',
+    ]
+  else
+    base_globs
+  end
+end
+
+def native_file_changed?(changed_files)
+  native_build_globs.any? do |glob|
+    changed_files.any? do |path|
+      File.fnmatch?(glob, path)
+    end
+  end
+end
+
 # checks if the native build needs to be run
 def should_build?
   # Essentially, we want to avoid native builds if the only thing that changed was JS code.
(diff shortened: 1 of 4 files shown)


fastlane/lib/util.rb
```diff
@@ -45,3 +45,76 @@ def travis?
...
+
+    dep_name_regex = /react|jsc/
+    if (changed_dependencies + changed_devdependencies).any? { |dep| dep =~ dep_name_regex }
+      UI.message('should_build? some react|jsc dep changed, so yes')
+      return true
+    end
+  end
+
+  source_globs = [
+    ".circleci/**",
```
rye:
```suggestion
    '.circleci/**',
```

fastlane/lib/util.rb
```diff
@@ -45,3 +45,76 @@ def travis?
...
+    "e2e/**",
```
rye:
```suggestion
    'e2e/**',
```

fastlane/lib/util.rb
```diff
@@ -45,3 +45,76 @@ def travis?
...
+    "fastlane/**",
```
rye:
```suggestion
    'fastlane/**',
```

fastlane/lib/util.rb
```diff
@@ -45,3 +45,76 @@ def travis?
...
+    "scripts/**",
```
rye:
```suggestion
    'scripts/**',
```

fastlane/lib/util.rb
```diff
@@ -45,3 +45,76 @@ def travis?
...
+    "Gemfile.lock",
```
rye:
```suggestion
    'Gemfile.lock',
```

fastlane/lib/util.rb
```diff
@@ -45,3 +45,76 @@ def travis?
...
+  ]
+
+  case lane_context[:PLATFORM_NAME]
+  when :android
+    platform_globs = [
+      "android/**",
```
rye:
```suggestion
      'android/**',
```

fastlane/lib/util.rb
```diff
@@ -45,3 +45,76 @@ def travis?
...
+      "{modules,source}/**/*.java",
```
rye:
```suggestion
      '{modules,source}/**/*.java',
```

fastlane/lib/util.rb
```diff
@@ -45,3 +45,76 @@ def travis?
...
+    ]
+  when :ios
+    platform_globs = [
+      "ios/**",
```
rye:
```suggestion
      'ios/**',
```

fastlane/lib/util.rb
```diff
@@ -45,3 +45,76 @@ def travis?
...
+      "{modules,source}/**/*.{h,m,mm}",
```
rye:
```suggestion
      '{modules,source}/**/*.{h,m,mm}',
```

fastlane/lib/util.rb
```diff
@@ -45,3 +45,76 @@ def travis?
...
+    ]
+  end
+
+  native_file_changed = (source_globs + platform_globs).any? {
```
rye: For multi-line blocks, you should use `do...end` instead of `{}`.  (They are identical in this case.)

Also, I'd put `|glob|` either on its own line or after `do`.

e.g.
```ruby
(source_globs + platform_globs).any? do |glob|
  [...]
end
```

fastlane/lib/util.rb
```diff
@@ -45,3 +45,76 @@ def travis?
...
   ENV['CIRCLECI'] == 'true'
 end
+
+# diff two hashes at a path and return the changed items
+def deps_diff(old_hash, new_hash, path)
+  (new_hash.dig(*path).to_a - old_hash.dig(*path).to_a).to_h
+end
+
+# checks if the native build needs to be run
+def should_build?
```
rye, line 99: RuboCop wants this to get split up a bit; it's a bit long, and the ABC and Cyclomatic complexity are a bit high. But this is a very good start.

rye (changes requested):
General feedback, mostly. I'd love to have the RuboCop-y things (especially single quotes over double quotes) get taken care of.

hawkrives:
Something broke… it's not doing the Android build.

fastlane/lib/util.rb
hawkrives: I pulled the big block in the middle out into a separate function; hopefully that helps

rye: Those metrics are still high, but more acceptable this time around.

rye (dismissed):
One potential issue I still see is that `git diff` will also include forwards changes that we don't currently have in our branch.

For instance, if `master` has additional changes to native code, that'll trigger native builds in a JS-only branch, because the diff expands to include stuff that we're missing.

We should use `git log` instead of `git diff`.

I think that the following pipeline satisfies what we need:

```
git log --name-only --format='' 'master'..'<branch>' | sort | uniq
```

- The `..` is required here
- `sort` and `uniq` can be replaced by pipeline magic in Ruby.

This would also give us the opportunity to check, say, how _many_ times a file was changed.

rye:
We could also just use `git log [...] 'master'..HEAD` if we wanted, lol

fastlane/lib/util.rb
```diff
@@ -45,3 +45,86 @@ def travis?
...
+end
+
+def npm_native_package_changed?
+  old_package = JSON.parse(sh "git show '#{source_branch}:package.json'")
+  new_package = JSON.parse(sh "git show '#{current_branch}:package.json'")
+
+  changed_dependencies = deps_diff(old_package, new_package, ['dependencies']).keys
+  changed_devdependencies = deps_diff(old_package, new_package, ['devDependencies']).keys
+
+  dep_name_regex = /react|jsc/
```
rye: Can we pull this out into a constant and freeze it?

À la `DEP_NAME_REGEXP = /react|jsc/.freeze` outside this method, and change your comparison to `dep =~ DEP_NAME_REGEXP`.

fastlane/lib/util.rb
```diff
@@ -45,3 +45,86 @@ def travis?
...
+    UI.message("should_build? branch == #{source_branch}, so yes")
+    return true
+  end
+
+  # 1. need to get files changed in the current build since master
+  changed_files = sh("git log --name-only --format='' '#{source_branch}'..'#{current_branch}'").lines.sort.uniq
+
+  # 2. check for "packages we care about"
+  if changed_files.include? 'package.json' and npm_native_package_changed?
+    UI.message('should_build? some react|jsc dep changed, so yes')
```
rye: I ask about extracting `DEP_NAME_REGEXP` so that we can include that here:

Once you have it extracted, this should work:

```suggestion
    UI.message("should_build? some dependency matching #{DEP_NAME_REGEXP.inspect} changed, so yes")
```

rye (commented):
One more suggestion and then I'm ready to ship.

fastlane/lib/util.rb
hawkrives: yeah just wasn't sure how to include it in the string before

rye (approved):
Looks great. :shipit: 

Thanks for taking this on, @hawkrives! 🎉

Side note: At some point I want to start testing this with full mocking just so we don't have any guessing games about logic or stuff like that.  But this was a good expedition.

PR #3215, opened 2018-11-27
fastlane#auto_beta: Explicitly write 0 build-status when skipping

### LINKED ISSUE #3214: Native build skipping confuses Danger
Danger claims that "Android" or "iOS" fails whenever the builds get skipped by the nascent build skipper in #3209.

This is because we aren't calling any of our `build` actions, which means we don't trigger any code that writes into the file.

Should be as easy as just manually writing `0` when we skip.

rye:
Closes #3214.


hawkrives:
I should really go contribute to Danger and make it write warnings back as a comment…

hawkrives (approved):

Issue #3250, opened 2018-11-28
Centralize logic for build skipping into reusable scripts


rye:
I can think of a few possible use cases here, but the main thing would be the creation of a few scripts called something like `script/ci-should-continue`, taking one of three arguments: `native`, `ios`, or `android` and exiting with `0` if the build should not be skipped or `1` if the build should be skipped.

This way, we run things as follows:
```
script/ci-should-continue ios && <ios commands> || echo "Skipping iOS build because we said so."
```

These should be Ruby-based? Ideally, we'll centralize the logic into a central library file which we can require from `fastlane` to dedupe the logic.

rye:
We should only call this on "expensive" operations, such as `fastlane`, `detox`, and the like.  Obviously, it should run as fast as possible.

rye:
We could build this as a Ruby Gem, if we wanted to! ~(Under `frogpond` namespace?)~ (RubyGems doesn't have namespaces)

This would make it _very_ easy to test and distribute.

hawkrives:
is this done‽‽?!

rye:
Yea

PR #3252, opened 2018-11-28
Skip native builds - part 2

hawkrives:
followup to #3209 

part of #3250? maybe?

now with actual build skipping!

I pulled the should-build script out into a separate Ruby file - it could probably stand some poking, @rye.

- [x] ~I should go make it print '0' to build-status, because Danger's still going to want that~ done

image


hawkrives:
@rye meh - went another route :p

hawkrives:
(thanks though)

hawkrives:
So! Reviews? @rye, @hannesmcman, @drewvolz?

This seems to work!

scripts/should-build.rb
```diff
@@ -0,0 +1,108 @@
...
+    puts('should_build? some native file changed, so yes')
+    return true
+  end
+
+  # 4) if none matched, return false
+  puts('should_build? no')
+  false
+end
+
+def main
```
rye: This method is kinda pointless.  Consider just putting the body outside of the function (at the top level?)

scripts/should-build.rb
```diff
@@ -0,0 +1,108 @@
+require 'json'
+
+PLATFORM = ENV.fetch('task', 'unknown')
```
rye: I prefer the following syntax:
```suggestion
PLATFORM = ENV['task'] || 'unknown'
```

Also, freeze it?

```suggestion
PLATFORM = (ENV['task'] || 'unknown').freeze
```

scripts/should-build.rb
```diff
@@ -0,0 +1,108 @@
...
+def git_branch
+  sh('git rev-parse --abbrev-ref HEAD')
+end
+
+# diff two hashes at a path and return the changed items
+def deps_diff(old_hash, new_hash, path)
+  (new_hash.dig(*path).to_a - old_hash.dig(*path).to_a).to_h
+end
+
+NPM_DEP_NAME_REGEXP = /react|jsc/.freeze
```
rye, line 21: Can we get some whitespace between this and the method that follows?

scripts/should-build.rb
```diff
@@ -0,0 +1,108 @@
...
+  end
+
+  # 4) if none matched, return false
+  puts('should_build? no')
+  false
+end
+
+def main
+  if should_build?
+    File.open(File.join(File.dirname(__FILE__), '..', 'logs', 'build-status'), 'w') { |file| file.write('0') }
```
rye: I would imagine RuboCop wants this line truncated.

Maybe use `do...end` syntax instead?

rye (changes requested):
Just a couple of stylistic nits

scripts/should-build.rb
hawkrives: Rubocop said something about `Do not freeze immutable objects, as freezing them has no effect.` on the regexp one… I can still freeze this one too

rye: Oh, no need to do that.  We should not freeze the regexp literal then.

hawkrives: Oh, that went away after upgrading rubocop.

rye: Oh, then we should freeze it this one too :P

scripts/should-build.rb
hawkrives: are you sure we can't just inline this? I really want to inline this again… so concise…

rye: Yes, I'm sure.  We print it out elsewhere.

hawkrives: although then we can't inline it into the other string… bah.

rye (approved):
Looks good to me!

hawkrives:
(cancelled the last two commits to speed things along)

hawkrives:
https://status.circleci.com/incidents/vvdpb19g0n21 – something broke inside Circle, so the macOS build is stuck for a while.

PR #3254, opened 2018-11-28
Clean up Dangerfile

### LINKED ISSUE #3244: Hide "ios slow build times" report if empty
Currently, it posts even if it's empty. It should not post if there's no content.

hawkrives:
closes #3244 

- hide the 'ios report' if it's empty
- update all blocks to report as "markdown" instead of "warn/fail" – other things will update the build status, so it's OK; plus, this way they'll comment inline instead of in Checks
- remove .pbxproj warnings – they were disabled anyway because they didn't work right
- clean up unused functions and dependencies

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -53,7 +53,7 @@ async function main() {
 // task=ANDROID
 //
 
-async function runAndroid() {
+function runAndroid() {
 	const logFile = readLogFile('./logs/build').split('\n')
 	const buildStatus = readLogFile('./logs/build-status')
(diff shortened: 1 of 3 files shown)


rye:
Uh rip

hawkrives:
@rye I can no longer reproduce the thing doing the commenting when it shouldn't…

rye (approved):
I approve, then.

If we start getting weird behavior we'll have to revisit, but this is an overall improvement.

Issue #3242, opened 2018-11-28
[github] Investigate an "auto-merge" github app


hawkrives:
https://github.com/SvanBoxel/auto-merge

It has an option to auto-merge approved and green PRs with the `auto-merge` label -

```yaml
# .github/auto-merge-settings.yml
merge_method: 'merge'
strategy: 'label'
```

I think it'd be really nice 😸 

Thoughts?

hawkrives:
(cc @rye @drewvolz)

rye:
I think it'd be… nice…

As long as it only happens on PR's with a given label.

rye:
And maybe can we customize which label to check? E.g. `merge-me-when-green`

hawkrives:
I didn't see anything about customizing the label, but it is open source, so we might be able to contribute that.

hawkrives:
Closing this as the thing isn't public, so we'd have to figure out how to run our own instance.

PR #3255, opened 2018-11-28
"git merge-base"-based native build skipping

rye:
This PR changes the logic behind native build skipping to involve using `git merge-base` to find the latest common point in history to use as the beginning of a range log, rather than comparing strings.

This PR also removes the explicit halt if the current branch is called `master`, instead using a general purpose algorithm that should work in all cases, assuming the history is unshallowed enough to manage finding the beginning of a branch. (If a clone is shallow to depth `n`, a branch can be no longer than `n - 1`.)

This plumbing command is used internally for finding the best ancestor to start a merge at, so it makes a _lot_ of sense for us to use that when trying to figure out where we should start our comparison.


hawkrives:
> This PR also removes the explicit halt if the current branch is called master, instead using a general purpose algorithm that should work in all cases, assuming the history is unshallowed enough to manage finding the beginning of a branch. (If a clone is shallow to depth n, a branch can be no longer than n - 1.)

Let's say `n = 50` – what happens if a merged branch has more than 50 commits? I think we should just do the native build then, but I want to make sure that's what happens.

.circleci/config.yml
```diff
@@ -343,7 +343,7 @@ jobs:
       - run: touch logs/build-status
       - run: 
           name: Skip build if possible
-          command: ruby scripts/should-build.rb || circleci step halt
+          command: ruby scripts/should_build.rb || circleci step halt
```
hawkrives: it annoys me when cli tools use underscores. i get that rubocop complains the other way, but it annoys me.

hawkrives (approved):

.circleci/config.yml
rye: Yes, but this is not just a cli tool, it's also a library file.

If we were instead to call it `scripts/should-build` with no file extension (which we _can_ do if it's `+x` and includes the `#!` line it needs) then RuboCop would not complain. I want to do that, if that sounds good to you.

hawkrives: I would prefer that. 

Actually, if you wanted (probably in another PR), we could split this into a library file and an executable script – `should_build.rb` and `should-build`, and then `should-build` can require and invoke the function from `should_build.rb`.

rye: That's a better way to do it anyhow, because then we can mock `sh` and such functions and test our actual behavior!

I'll rename the script as the last commit in this PR, then merge when green.

hawkrives: (speaking of merge-when-green… #3242)

rye:
>Let's say `n = 50` — what happens if a merged branch has more than 50 commits? I think we should just do the native build then, but I want to make sure that's what happens.

I "replicated" this by checking out an orphaned branch and trying to compare it with master with `git merge-base`.  Turns out that `git merge-base` just exits with code `1` when the histories are not connected at any point.

I _think_ that a shallow clone might just look like two orphaned histories in CI?

- - -

That said, we'll actually never hit this condition.

The `checkout` script,

image

includes nothing about `--depth`; Circle clones everything and also forces an update to make sure it's in sync.

hawkrives (approved):

hawkrives (approved):

PR #3261, opened 2018-11-29
Explicitly build nightlies

rye:
Since `ios-nightly` and `android-nightly` are built off of the `ios` and `android` builds, I don't think we actually have any safeguards to ensure that we do actually build nightlies besides setting environment variables.

We should explicitly state that we want nightlies to be built.

This commit does that; if `ENV['IS_NIGHTLY']` is anything besides `nil` or false-y values likke that, it will report that the build should continue as planned.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Fixed
 - Fixed an issue where Fastlane was reporting build failures despite having skipped the build (#3215)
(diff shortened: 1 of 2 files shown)


hawkrives:
So uh

Should we also check that there has been a commit in the last 26 hours?

We could finally skip nightlies when nothing has changed

rye:
Yes, we can do that too! I'll include it in this PR for completeness.

hawkrives:
I think 26 hours is a good time limit, in case it got delayed for some reason or whatever. Not being too cautious, but just a little bit cautious.

hawkrives (approved):

PR #3272, opened 2018-12-07
String#unique_lines: Chomp after uniquing

rye:
This is done to make it possible to actually do sane filename inclusion checks.

Previously, `String#unique_lines` was returning filenames with `\n` appended… making it impossible to do a simple set inclusion check.

This should also make all of our PRs that aren't building because they're "not significant" more happy to build.


hawkrives:
well… I guess if this fails, we get to bisect some! whee. let's hope it passes.

hawkrives (approved):

PR #3278, opened 2018-12-09
[circle2.1] add "executors" to centralize executor configuration

hawkrives:
Instead of using YAML references, which get a bit messy after a while, and can't really handle multiple top-level keys, we can use CircleCI 2.1's new `executors` top-level structure to define custom "executors" for ourselves.

A simple one:

```yaml
executors:
  node:
    docker:
      - image: 'circleci/node:10'
```

A more complex one:

```yaml
executors:
  ios:
    macos:
      xcode: '9.4.1'
    shell: /bin/bash --login -o pipefail
    environment:
      LC_ALL: en_US.UTF-8
      LANG: en_US.UTF-8
```

---

This is really a question of "will Circle do anything with this config". I'm just idly poking at this while I wait for local native android builds.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,13 +1,33 @@
 ---
-version: 2
+version: 2.1
+
+executors:
+  node:
+    docker:
+      - image: 'circleci/node:10'
+
+  ruby:
+    docker:
+      - image: 'circleci/ruby:2.5.3'
+
+  android:
+    docker:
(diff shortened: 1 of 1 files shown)


rye (changes requested):
It doesn't seem to be building.

hawkrives:
Very true.

hawkrives:
Closing due to lack of interest on my part in investigation.

Issue #3164, opened 2018-11-06
Release v2.6.4


hawkrives:
It's about time.

I want to do two things: #3108 and an as-yet unfiled issue to disable state restoration (because I have managed to, once again, completely break my app by suspending on a calendar detail view, and while we can fix that bug, IMO we should disable state restoration until we've audited the rest of the app for this sort of problem.)

drewvolz:
- [x] Privacy policy
- [x] State restoration 

...is the state restoration actually _state_ or _navigation_?

hawkrives:
Navigation state restoration. 

It's done.

hawkrives:
I need to fix login before 2.6.4 I think (only because I told someone it would be fixed in the "next release")

drewvolz:
Oof. We do need to fix that.

rye:
We need to fix all of the crashes that were introduced by RN 0.57 as well.

hawkrives:
Do those have issues? We should make sure that all of the crashes have associated issues.

rye:
I'm specifically concerned about #3260 and #3243 (which isn't a crash so much as login being broken)

We should also have something filed for popovers crashing the app.

rye:
Given that we've stepped forward with RN I think it's also v2.7 rather than v2.6.4.

hawkrives:
(skipping 2.6.4 and going straight to 2.7)

PR #3279, opened 2018-12-09
Fix popover crash on interaction

drewvolz:
Part of #3164. Fixes the touchable ref from not passing from `filter toolbar button` to the `popover`.

**Known issue:** the toggles, switches, checkboxes, scrolling menus, etc do not apply. This purely attempts to fix the crashing app. Getting them to work again is the next step.

Much thanks to @hawkrives for the clarity on why this was broken.

> createRef and ref={cb} both work similarly
> They start null, and then get populated after stuff is mounted
> ~stuff
> ref={cb} needs to be on the _view_ that we want to use as the anchor
> And then we pass anchor={this.refName} to the popover
> 
> Similarly with createRef
> pass ref={this.refName} to the actual anchor, and then anchor={this.refName} as well
> Without actually _attaching_ the ref anywhere, it's always going to be null

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Made sure that nightly builds will get built if something changed (#3261)
 - Fixed a syntax error in build skipping script (#3262)
 - Forced build skipping script to actually query Git history for nightly skip check (#3275)
+- Fixed the touchable ref from not passing from the filter toolbar button to the popover (#3279)
 
 ## [2.6.3] - 2018-09-17
 ### Fixed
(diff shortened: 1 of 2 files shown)


drewvolz:
Also, flow types will need tweaking. Note from hawken:
>I wrote those types. They expect one thing. But that doesn't mean it's the only allowed thing.

hawkrives (dismissed):
Flow's going to complain until you switch back to createRef

rye (approved):
Very nice. This looks like exactly the CL we needed after our investigation a few days ago.

drewvolz:
Switched back. One flow error left.

hawkrives (approved):

rye (approved):

PR #3288, opened 2018-12-13
disable nightly checks

hawkrives:
Circle's not building nightlies currently; we think its due to Circle
not having enough git log information in its clones.

### DIFF
--- a/scripts/should-skip-build
+++ b/scripts/should-skip-build
@@ -87,7 +87,9 @@ def anything_changed_yesterday?
 end
 
 def should_build_nightly?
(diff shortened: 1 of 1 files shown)


drewvolz (approved):

rye:
Why not try something like `git fetch --all --unshallow` instead?

hawkrives:
Can we do this tonight and try that tomorrow? I'd like a guaranteed nightly tonight :smile:

I'd be willing to spearhead that update tomorrow.

rye:
Fine by me, then.

PR #3289, opened 2018-12-13
[WIP] Re-enable nightly change detection

hawkrives:
Reverts and fixes #3288 

For merge after a successful nightly.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -268,6 +268,7 @@ jobs:
             git reset --hard origin/master
             git checkout "$CIRCLE_BRANCH"
             git reset --hard "origin/$CIRCLE_BRANCH"
+            git fetch --all --unshallow
       - run: mkdir -p logs/
       - run: touch logs/build-status
       - run:
@@ -339,6 +340,7 @@ jobs:
             git reset --hard origin/master
             git checkout "$CIRCLE_BRANCH"
             git reset --hard "origin/$CIRCLE_BRANCH"
+            git fetch --all --unshallow
       - run: mkdir -p logs/
       - run: touch logs/build-status
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -268,7 +268,6 @@ jobs:
             git reset --hard origin/master
             git checkout "$CIRCLE_BRANCH"
             git reset --hard "origin/$CIRCLE_BRANCH"
-            git fetch --all --unshallow
       - run: mkdir -p logs/
       - run: touch logs/build-status
       - run:
@@ -340,7 +339,6 @@ jobs:
             git reset --hard origin/master
             git checkout "$CIRCLE_BRANCH"
             git reset --hard "origin/$CIRCLE_BRANCH"
-            git fetch --all --unshallow
       - run: mkdir -p logs/
       - run: touch logs/build-status
       - run:
--- a/scripts/should-skip-build
+++ b/scripts/should-skip-build
@@ -81,9 +81,7 @@ def git_log_between(source, target)
 end
 
 def anything_changed_yesterday?
-  hours_ago = 26
-  seconds_ago = hours_ago * 60 * 60
-  yesterdays_changes = changes_after(seconds_ago.to_s).unique_lines
+  yesterdays_changes = changes_after('26 hours ago').unique_lines
 
   yesterdays_changes.count > 0
 end
--- a/scripts/should-skip-build
+++ b/scripts/should-skip-build
@@ -87,9 +87,7 @@ def anything_changed_yesterday?
 end
 
 def should_build_nightly?
-  # disabled due to failures to build nightlies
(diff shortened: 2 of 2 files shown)


hawkrives:
circle says:

```
fatal: --unshallow on a complete repository does not make sense
```

/cc @rye

rye:
That error makes sense.

Is it not possible that our `git log` is doing something messy?

I'd prefer to avoid using `--since="26 hours ago"` since it's ambiguous.  Why don't we try doing something like `$(date -Iseconds --date="26 hours ago")` and shoving that into the `--since` flag?

hawkrives:
I mean, I'm fine with that, but I think that supporting natural language parsing is a well-established part of `git log`.

Anyway. I'll update this PR to do that, instead, and we can try it tonight.

rye:
That's probably not gonna work. We should instead use `Time` (core) and/or `DateTime` (stdlib)

hawkrives:
oh, that's true. it does want an absolute date.

rye:
Oh and we could shovel `2>&1` and if `^warning` matches we could throw and diagnose that way

hawkrives:
i like that too

hawkrives:
but um. the `sh` function should be dumping the output of the command to stdout?

hawkrives:
also I just realized that this block

```
git checkout master
git reset --hard origin/master
git checkout "$CIRCLE_BRANCH"
git reset --hard "origin/$CIRCLE_BRANCH"
```

will result in un-reproducible builds. Can we make it set the tip of CIRCLE_BRANCH to the given commit sha?

hawkrives:
or something?

rye:
> but um. the sh function should be dumping the output of the command to stdout?

Yes, but `sh` doesn't capture stdout, I think.

> also I just realized that this block
> 
> ```
> git checkout master
> git reset --hard origin/master
> git checkout "$CIRCLE_BRANCH"
> git reset --hard "origin/$CIRCLE_BRANCH"
> ```
> will result in un-reproducible builds. Can we make it set the tip of CIRCLE_BRANCH to the given commit sha?

Sure, but that's out of the scope of this PR.

We could drop the `git reset --hard "origin/$CIRCLE_BRANCH"` and instead `git reset --hard $CIRCLE_SHA1`. I _think_ that'll rewind properly.

We should _definitely_ be keeping `master` in sync with origin, that's the whole point of doing that block in the first place, and besides doing some weird reflog-fu which would likely break whenever Circle updates their checkout script, I'm pretty sure this is all we have.

That's why I wanted a custom checkout script in the first place. Then we have control.

hawkrives:
```ruby
output = `#{cmd}`.chomp
puts output
```

Shouldn't that capture stdout?

hawkrives:
And I agree with the out of scope thing. I'll file an issue for that.

rye:
Towards this actual issue, I'll note the following:

It's entirely possible that Git is actually doing the right thing and saying `git log --after='26 hours ago'` is empty. Note how if you change `26 hours` with `2 seconds`, you'll likely get an exit 0 and no output. This means we might actually just have issues with timestamps and relative time. I think that's a much more likely thing to troubleshoot.

A great idea would be to `git show --format="%aI" --quiet HEAD` which would grab the commit date in ISO8601 form which can definitely be parsed.  We then pass that as our `--after=`, which would guarantee that we have our commits be correct.

Let's… merge this PR, so we get our "broken" code back. I can play around with this some more tonight.

- - -

Yes, @hawkrives, your code example does capture stdout.

But I'm concerned that git is writing to stderr, and it's just getting eaten somewhere or we're not seeing it.

Note that

```ruby
`>&2 echo "I'm an error"`.chomp
# => ""
```

That is to say, stderr completely bypasses the capture. So if git does as I hope—what it's supposed to—and prints its errors on stderr, it doesn't get caught by our script, and it's entirely possible that that gets lost.

Compare that with:

```ruby
`{ >&2 echo "I'm an error"; } 2>&1`.chomp
# => "I'm an error"
```

(Braces added to force the output redirection to take place on both ends)

So that's why we'd need redirection, to force stderr back into stdout for our consumption, if we want it.

Some defense of Ruby's logic here:

- The double-` style is likely intended for hacky settings where you don't care about stderr, or if stderr is emitted you probably don't want it mixed in with your stdout.
- There are better alternatives which keep both streams separate (e.g. `Open3.popen3`) and let you consume them as pipes rather than getting one giant string. They also let you inspect the exit status, which the double-` doesn't.

hawkrives:
@rye I completely agree with ruby's default of only capturing stdout; I got confused from your comment,

> image

and failed to think straight.

---

Regarding merging this PR… okay, if we do a beta tomorrow. I want to make sure the login fixes actually work 🤞

rye:
Sounds good to me. I may have a fix for the skipping logic before then.

rye (approved):
This final CL looks good. (Essentially just `git revert` lol)

PR #3298, opened 2018-12-13
Delete .env.sample.js

hawkrives:
We no longer use this file.


drewvolz (approved):
RSLGTM 👍 although we should consider this impacting Carls if/when we do any work on moving the code bases back to parity

hawkrives:
Doesn't. All of the API stuff has moved to ccc-server atm.

hawkrives:
But good thought!

PR #3308, opened 2018-12-16
Fix olecard login

hawkrives:
Closes #3181

---

`fetch()` doesn't send cookies by default (good).

`fetch()` in RN@<0.57 apparently sent cookies along with a redirect (hmm).

`olecard/login.cfm` redirects you to `olecard/index.cfm`, which checks to make sure you're logged in… except that because the cookies weren't sent, it can't find your session, and it kicks you back to the login page.

so we'll just pass along the credentials.

---

and that one-line change is all it took.

I went ahead and refactored some of the other bits after fixing the core issue; now we just check the status code to detect failed login, instead of string matching on the page looking for the phrase "Password". (We couldn't do this before because the server didn't send a 403. Now it does.)

I'm using `rawFetch` here because our stock fetch() wrapper auto-rejects the request if it returns a non-2xx statuscode.

### DIFF
--- a/source/lib/login.js
+++ b/source/lib/login.js
@@ -47,6 +47,7 @@ export async function performLogin({
 		loginResult = await fetch(OLECARD_AUTH_URL, {
 			method: 'POST',
 			body: form,
+			credentials: 'include',
 		})
 	} catch (err) {
 		let wasNetworkFailure = err.message === 'Network request failed'
--- a/source/lib/login.js
+++ b/source/lib/login.js
@@ -30,6 +30,8 @@ export type LoginResultEnum =
 	| 'network'
 	| 'bad-credentials'
 	| 'no-credentials'
+	| 'server-error'
+	| 'other'
 
 type Args = {attempts?: number}
 
@@ -41,14 +43,28 @@ export async function performLogin({
 		return 'no-credentials'
 	}
 
-	const form = buildFormData({username, password})
-	let loginResult = null
+	let form = buildFormData({username, password})
+
 	try {
-		loginResult = await fetch(OLECARD_AUTH_URL, {
+		let {status: statusCode} = await rawFetch(OLECARD_AUTH_URL, {
 			method: 'POST',
 			body: form,
 			credentials: 'include',
 		})
+
+		if (statusCode >= 400 && statusCode < 500) {
+			return 'bad-credentials'
+		}
+
+		if (statusCode >= 500 && statusCode < 600) {
+			return 'server-error'
+		}
+
+		if (statusCode < 200 || statusCode >= 300) {
+			return 'other'
+		}
+
+		return 'success'
 	} catch (err) {
 		let wasNetworkFailure = err.message === 'Network request failed'
 		if (wasNetworkFailure && attempts > 0) {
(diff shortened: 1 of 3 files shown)


source/redux/parts/login.js
```diff
@@ -39,6 +39,13 @@ const showNetworkFailureMessage = () =>
 		[{text: 'OK'}],
 	)
 
+const showServerErrorMessage = () =>
+	Alert.alert(
+		'Server Failure',
+		'The OleCard Server is having issues right now. Please try again later.',
```
rye: This language is problematic.

Something more along the lines of:

```
Unable to get a good response from the OleCard server; please try again later.
```

seems like it'd be less likely to cause a student to complain to IT, for example. Puts the blame on us rather than IT, even though (after scrubbing forward in the commits) it seems we're only showing this when we get a 5xx.

source/lib/login.js
```diff
@@ -41,14 +43,28 @@ export async function performLogin({
...
+
 	try {
-		loginResult = await fetch(OLECARD_AUTH_URL, {
+		let {status: statusCode} = await rawFetch(OLECARD_AUTH_URL, {
 			method: 'POST',
 			body: form,
 			credentials: 'include',
 		})
+
+		if (statusCode >= 400 && statusCode < 500) {
```
rye, line 55: These three ifs are screaming "DRY me up" but I'm not sure of an easy way to do that in JS. Ideas?

rye (changes requested):
On the whole, looks good. Two nits.

source/redux/parts/login.js
hawkrives: I think I'd interpret yours in the same way as mine, ie as though it's an IT problem. 

The clearest thing would be to just say something like "We're having issues talking to OleCard. Please try again later"?

(But yes, it should only trigger when something is seriously wrong inside of OleCard; it has no code to explicitly handle and return a 500 at the moment, for instance.)

rye: @hawkrives, yes, that's the idea. It's probably better to just do `We're having issues talking to OleCard. Please try again later.` Even if the issue is on their end, we're accepting responsibility that it might be because we're triggering an exception due to an incorrect request format.

source/lib/login.js
hawkrives: be explicit 😛 

sadly JS doesn't have pattern matching (yet?) so there's not much more you can do here.

source/redux/parts/login.js
hawkrives:
```suggestion
		"We're having issues talking to the OleCard server. Please try again later.",
```

source/lib/login.js
rye: Oh let's definitely do

```js
switch(true) {
case statusCode >= 400 && statusCode < 500:
  // ...
}
```

That's an excellent idea.

/s

source/redux/parts/login.js
rye: Looks good.

source/lib/login.js
hawkrives: oh dear

rye (approved):

drewvolz (approved):
RSLGTM 👍 :shipit:

PR #3309, opened 2018-12-16
should-skip-build: Start abstracting away ambiguous nightly skipping wording, and potentially fix nightlies

rye:
This way, we know exactly what time `git log` is going off of, which will make any other necessary troubleshooting much easier.

I began by adding a `Commit` class, which can be used as a holder for some business logic.  I implemented dereferencing of SHAs, so this class can accept anything (commit, ref, tag, etc.) that can go to `git rev-parse`.

Then I added a `#time` helper which is then used by `anything_changed_yesterday?`, in addition to some idiomatic `Numeric` helpers shamelessly borrowed from Rails.

`anything_changed_yesterday?` now queries the `HEAD` commit's commit date, subtracts 26 hours' worth of seconds, and converts back to ISO8601 which then gets fed to `git log --after=`, hopefully disambiguating the history.

### DIFF
--- a/scripts/should-skip-build
+++ b/scripts/should-skip-build
@@ -98,6 +98,10 @@ end
 class Commit
 	attr_reader :hash
 
+	def self.HEAD
+		self.new('HEAD')
+	end
+
 	def initialize(ref)
 		@hash = dereference(ref)
 	end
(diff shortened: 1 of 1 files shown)


hawkrives (approved):
Sure

drewvolz (approved):

PR #3310, opened 2018-12-16
Make it easy to cache requests, and do it

hawkrives:
This PR introduces a new module, `@frogpond/fetch`, which has one export: 

```js
async function fetch(input: string | URL | Request, init: ?RequestOptions): ResponsePromise<any>;
```

It wraps the `fetch` we all know and (sometimes) love, but automatically caches cachable requests.

---

I was inspired by [`ky`](https://github.com/sindresorhus/ky), which is a `fetch` wrapper by sindresorhus. I didn't implement everything that `ky` does, but I did do one thing that it doesn't – automatic caching of cachable requests.

You see, there's a library called [`http-cache-semantics`](https://github.com/kornelski/http-cache-semantics), which implements the, well, HTTP caching semantics around the JS stdlib `Request` and `Response` classes.

> The following is all part of the `fetch()` standard.

- `fetch(url, {cache: 'default'})`
	1. The cache is checked for a matching URL
	2. If it doesn't exist, it's requested and cached
	3. If it's "fresh" (ie, if the `Cache-control: max-age=N` header from the original server response hasn't been passed yet), it's returned from the cache
	4. If it's "stale", the `Etag` header from the original response is attached to the request, which is sent to the server for re-validation
	5. If the server responds with a `304 Not Modified`, the original response's headers are updated, and it's returned from the cache
	6. If the server responds with `200 OK`, the new request is cached and returned
- `fetch(url, {cache: 'no-store'})`: bypasses the cache entirely – use this for logins
- `fetch(url, {cache: 'no-cache'})`: always checks for freshness before returning
- `fetch(url, {cache: 'only'})`: returns an error if the request isn't in the cache already
- `fetch(url, {cache: 'reload'})`: always do the fetch, then cache and return it
- `fetch(url, {cache: 'force-cache'})`: if there's a match in the cache, return it; otherwise, fetch, cache, and return.

Now, I don't trust myself to implement all of that correctly – much less to properly handle the `Etag`, `Cache-control`, `Expires`, and assorted other headers! (Okay, I probably could, but I'd really really want to write tests.)

What `http-cache-semantics` does, then, is to boil that list down into this:

<details><summary>a copy of the <code>cachedFetch</code> function</summary>

```js
let key = `urlcache:${url}`
let {response: oldResponse, policy: oldPolicy} = await getItem(key)

// if nothing has ever been cached, go fetch it
if (!oldPolicy) {
    let response = await fetch(request)
    let policy = new CachePolicy(request, response)

    if (policy.storable()) {
      await cacheItem({key: request.url, response, policy})
    }

    return response
}

// if we can re-use the cached data, return it
if (oldPolicy.satisfiesWithoutRevalidation(request)) {
    oldResponse.headers = oldPolicy.responseHeaders()
    return oldResponse
}

// otherwise, we're serving requests from the cache

// Change the request to ask the origin server if the cached response can be used
request.headers = oldPolicy.revalidationHeaders(request)

// Send request to the origin server. The server may respond with status 304
let newResponse = await fetch(request)

// Create updated policy and combined response from the old and new data
let {policy, modified} = oldPolicy.revalidatedPolicy(request, newResponse)

let response = modified ? newResponse : oldResponse

// Update the cache with the newer/fresher response
await cacheItem({key: request.url, policy, response})

// And proceed returning cached response as usual
response.headers = policy.responseHeaders()
return response
```

</details>

<p></p>

> Additionally, by re-using the `{cache}` parameter to do things like tell it "no, don't ever stick this in a cache", we will also automatically skip the OS' network cache!

---- 

This new `fetch` wrapper has three other tricks, two of which are inspired by `ky`:

1. [ky] You can pass `fetch(url, {searchParams: {query: 'a-name'}})`, and it will automatically stick the passed `searchParams` on the URL as the query string! Note that it replaces the query string, as that's the standard behavior for similarly named APIs.
2. [ky] By default, it will behave like our current `fetch` wrapper, and will throw an error when it gets a request outside of the 2xx range. If you pass `fetch(url, {throwHttpErrors: false})`, it won't throw errors.
3. [not-ky] You can pass `fetch(url, {delay: 500})` to ensure that the request won't complete in less than N milliseconds. This is to facilitate the slow-reload pattern in our ListViews.

---- 

One last thing – you'll notice the `ResponsePromise` return type, above? That's because (again, thank you `ky`) I was inspired to add two properties to the normal Promise that you get from `fetch`. 

The reason we had `fetchJson` was so that we didn't have to write `fetch(url).then(r => r.json())`.

`ky` showed me a better way: `fetch(url).json()`. If you monkey-patch the instance of the response promise to add `.json()` and `.text()` methods, you're not monkeying with any globals, and you can just call `fetch().json()` to get the JSON content!

---- 

I have migrated all of the `fetch()` calls in the app to use either the new `@frogpond/fetch` module or the standard `global.fetch()`, depending on the use.

---- 

Current issues: 

- [x] … I somehow introduced a `require()` cycle between RN's `Networking/fetch.js` and RN's `vendor/whatwg-fetch` modules. Still … investigating … what happened there.
- [x] Once that's dealt with, I need to make sure that this code actually works.
- [~] I'm going to move the bundled-data insertion to a new `init/data.js` file, I think. Which will centralize it a bit more than I might like, but I'd rather have it centralized than spread out in this case, I think. Anyway, it'll essentially just prime the cache with "stale" data from the app bundle, if nothing is cached at app launch time. **edit**: maybe done? I haven't _tested_ this, but there sure are a lot of cached files in my localstorage now.


source/lib/cache.js
```diff
@@ -1,97 +1,147 @@
...
-	return {isCached: true, isExpired, value: stored.value}
+async function setItem(
+	key: string,
+	value: any,
+	{ttl, source}: {ttl?: [number, string], source?: string},
+) {
+	console.log('setItem', {key, value, ttl, source})
+	await AsyncStorage.multiSet([
+		[`fp:${key}:data`, JSON.stringify(value)],
+		[`fp:${key}:ts`, new Date().toUTCString()],
```
rye: Any particular reason that `toUTCString` is used here? I prefer ISO strings, personally.

source/lib/cache.js
```diff
@@ -1,97 +1,147 @@
...
+		if (shouldDelay && elapsed < 500) {
+			// 0.5s delay for ListViews – if we let them go at full speed, it feels broken
+			await delay(500 - elapsed)
+		}
+
+		return getItem(key)
 	} catch (error) {
-		if (cachedValue.isCached && cachedValue.value) {
-			return cachedValue.value
+		// fall back to the local data
```
rye: This comment doesn't seem to match the code that follows it, is this meant for the "catch"?

source/lib/cache.js
```diff
@@ -1,97 +1,147 @@
...
+		delay: shouldDelay,
+	} = args
+
+	let start = Date.now()
 
-	if (process.env.NODE_ENV === 'development') {
-		return fetchHelpToolsBundled()
+	console.log('fetchAndCacheItem', {key, url, cbForBundledData, ttl})
+
+	// if (process.env.NODE_ENV === 'development' && cbForBundledData) {
```
rye: Dead code now?

modules/ccc-calendar/index.js
```diff
@@ -79,35 +80,25 @@ export class CccCalendarView extends React.Component<Props, State> {
...
-			data = await fetchJson(url)
-		} catch (err) {
-			reportNetworkProblem(err)
-			this.setState({error: err.message})
-			console.warn(err)
-		}
+		let {value} = await fetchAndCacheItem({
+			key: `calendar:${url}`,
+			url: url,
+			ttl: [1, 'minute'],
```
rye: Any particular reason why this TTL is `1 minute`?

source/redux/parts/help.js
```diff
@@ -2,7 +2,19 @@
...
-import {fetchHelpTools} from '../../lib/cache'
+import {fetchAndCacheItem, type CacheResult} from '../../lib/cache'
+import {API} from '@frogpond/api'
+
+function fetchHelpTools(): CacheResult<?Array<ToolOptions>> {
+	return fetchAndCacheItem({
+		key: 'help:tools',
+		url: API('/tools/help'),
+		afterFetch: parsed => parsed.data,
+		ttl: [1, 'hour'],
```
rye: Any particular reason for the 1 hour timeout?

rye (commented):
Here's some general feedback, some of the comments of which I am going to mark as resolved because I realized what was happening later on.

hawkrives:
@rye yeah, there's some dead code, and debugging stuff left in there – I've needed it often enough while I was working on these five that I was going to leave it in until all of the views were finished.

re:toUTCString – because … oh. that does _not_ do what I wanted. not going to use that… **edit**: now switched to toISOString.

source/lib/cache.js
hawkrives: Only dead until we're ready to ship the feature.

I like to have this in so that we run off of local data when developing… saves several minutes of head-scratching about "why aren't my changes showing up???" whenever you're doing that.

source/lib/cache.js
hawkrives: Switched to toISOString. I didn't realize that toISOString used UTC, and I also didn't realize that toUTCString returned … not an ISO string.

anyway. it's gone now.

source/lib/cache.js
rye: Yep, that's all good.

source/lib/cache.js
hawkrives: Removed.

hawkrives:
After sleeping on this, I think I can just remove the `afterFetch` callback entirely. By doing so, most uses of this API can be as simple as `fetchCached(url, {json: true})`. (I also want to redo the fetch() wrapper to be `fetch(url, {json?, status?})`, but that'll be a separate PR.)

So to switch something from fetch() to fetchCached(), it'll be just two changes: 

```
let body = fetch(API('/spaces/hours'), {json: true, status: true})
let {value: body, error} = fetchCached(API('/spaces/hours'), {json: true, status: true})
```

(status:true causes it to throw an error / return the error message on non-200 responses)

Words

hawkrives:
I updated the OP with documentation for the new direction of this PR.

rye (approved):
RSLGTM

drewvolz (approved):
RSLGTM

hawkrives:
Yolo nightly test

Issue #3318, opened 2018-12-17
OleCard login is still iffy


hawkrives:
- [ ] on fresh app launch, it doesn't trigger the actual "log in" process until you open settings (I remember this now)
- [ ] the first attempt at logging in always seems to trigger "invalid username or password"
- [ ] @rye doesn't get any balances even after logging in (android)

see also https://github.com/StoDevX/AAO-React-Native/issues/2847

rye:
Spitballing: could https://github.com/StoDevX/AAO-React-Native/blob/c43e5a16f87202961a788a2f74789a359d8740d7/source/lib/financials/balances.js#L21-L24
be the issue? This is a fetch which doesn't have `credentials: 'include'` on it.

rye:
That was the only such instance.

To the other two issues, I'd also point to views not refreshing on open too—things like hours presenting stale information until the next tick of the update timer, for example.

hawkrives:
I get that feeling, but I think those are two separate underlying issues – one being how FlatList doesn't sometimes like to update unless you set things up just right, and the other (balances) being … well, something else, because it's just raw Views all the way down.

hawkrives:
words.

rye:
Have we made any further progress on this one?

hawkrives:
I have not.

hannesmcman:
Can we make this a top priority? It's probably the most annoying and frequent bug in the app..

rye:
I'm tempted to blame a few of the other network related issues on the same underlying issue.

Yes, this is top priority, but I'm not working on it atm.

hawkrives:
It is a top priority for me as well, but I likely won't have the motivation to work on it for at least a month (Blame degree auditing and phone work - ie work-work is interesting enough that I don't have a lot of motivation when I get home.)

Just as an aside - the reason I haven't put a ton of work into this yet is because it only affects the dev versions, so worst case you could switch back to the release channel. 

I myself have a workaround that I do, where I open Settings twice and then trigger Balances; that manages to log in properly.

hawkrives:
okay looks like I'm going to be working on this tonight. we'll see what happens.

rye:
I think we can call this one fixed by #3575, can't we. Closing for now.

hawkrives:
Both iOS and Android log in for me, now, and I haven't seen any login errors since merging #3503 and #3575, so I think so!

PR #3319, opened 2018-12-17
getBalances: Include credentials through CORS

rye:
We "trust" this endpoint to redirect to the right place, so we need this because CORS policy is not correct, it seems.

Maybe a fix for #3318?


hawkrives (approved):
Yeah probably. 

I think there are still some other bugs lurking (like my "failure on first attempt" one) but this is also good.

PR #3320, opened 2018-12-18
Return stale-but-cached data if offline

hawkrives:
> A followup to #3310 

Turns out that I had never included explicit "hey, if you're offline, just return the cached data, even if it's stale" instructions to the `cachedFetch` function.

If we had any data that was ever marked as fresh, #3310 would have let you use that offline, I believe, but I don't have any of that so 🤷‍♀️.

I recommend looking through this PR one commit at a time; the final commit reorganizes the file somewhat to reduce the `cachedFetch` function to 24loc.

I also tweaked it to always return bundled data when in development.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Added Renovate as our new automated dependency management tool, with a nice configuration (#3193)
 - Add "open webpage" row to student work detail
 - Added some logic to skip native builds if nothing that might affect them has changed (#3209)
-- All network requests are now cached according to the server's caching headers (#3310)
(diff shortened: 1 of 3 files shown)


hawkrives:
and, can I just say "wow" – the build was finished before I finished writing the PR!

rye (approved):
All looks good to me.

hawkrives:
telegram-cloud-file-1-806028907-131513-328283485387648148
telegram-cloud-file-1-806027187-133429--8857728037636458102

state | image 
--- | ---
before internet | telegram-cloud-file-1-806623257-32556--3470422520429554501
after internet | image

I have now tested this; it seems to work. 😁

drewvolz (approved):

drewvolz:
Changelog entry?

rye (approved):

PR #3322, opened 2018-12-18
Fix reconcile git histories commands

drewvolz:
image


rye:
But… this'll only work on tags.

The whole point of that step was to keep `master` and `CIRCLE_BRANCH` separate; Circle keeps resetting `master` to the tip of `CIRCLE_BRANCH` which is naughty.

drewvolz:
This isn't what we want.

PR #3329, opened 2018-12-19
Ignore specific proptype warning for react markdown

### LINKED ISSUE #3328: Ignore specific PropTypes warning in YellowBox
there's way to have YellowBox ignore specific warnings by way of string matching.

we should ignore the following warning: 

```
Warning: Failed prop type: Invalid prop `containerTagName` of type `object` supplied to `ReactMarkdown`, expected `function`
```

derived from #3260

drewvolz:
Closes #3328

### DIFF
--- a/source/root.js
+++ b/source/root.js
@@ -3,6 +3,6 @@
 import {AppRegistry, YellowBox} from 'react-native'
 import App from './app'
 
-YellowBox.ignoreWarnings([])
+YellowBox.ignoreWarnings(['Invalid prop `containerTagName`'])
 
 AppRegistry.registerComponent('AllAboutOlaf', () => App)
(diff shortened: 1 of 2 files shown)


source/root.js
```diff
@@ -3,6 +3,6 @@
 import {AppRegistry, YellowBox} from 'react-native'
 import App from './app'
 
-YellowBox.ignoreWarnings([])
+YellowBox.ignoreWarnings(['Invalid prop `containerTagName`'])
```
hawkrives:
```suggestion
YellowBox.ignoreWarnings([
  'Failed prop type: Invalid prop `containerTagName` of type `object` supplied to `ReactMarkdown`, expected `function`',
])
```

I want to be as specific as possible with this error silencing.

I think it does some substring matching, which is why `Invalid prop` matched `Failed prop type: Invalid prop`, but I'd rather be specific.

hawkrives (dismissed):
(also you can do a single-to-multiline suggestion now, but I don't think you do a multi-to-single yet)

hawkrives (approved):

PR #3337, opened 2018-12-22
Use the scrollEnabled prop and nix the prepare call

### LINKED ISSUE #2538: Find a better solution to fixing the scrolling textinputs
Reminder to check on react-native often to see if we can revert #2535.

drewvolz:
Closes #2538

We no longer need to run a `sed` to tweak React Native's `RCTMultilineTextInputView`  value for `scrollEnabled`. 

I've tested this by:

1. Removing `node_modules`
2. Removing the `prepare` script
3. Reinstalling packages with `yarn`
4. Rebuilding the project
5. Uncommenting the `scrollEnabled` prop
6. Reloading
7. Going to different views that use our `<TextInput />`
8. Verifying that scrolling is enabled/disabled per the setting

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Told Circle to use Xcode 10.1 to build and test the iOS app
 - Updated Fastlane support url to point to the project's issues tracker (#3314)
 - Ignore specific proptype warning for react markdown (#3329)
+- Only set `scrollEnabled` if `multiline` is true (#3337)
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


hawkrives (dismissed):
https://github.com/StoDevX/AAO-React-Native/blob/e399b3ec674196fb04ae757fdab25c07188f7830/modules/tableview/cells/textfield.js

needs the prop change too. 

Once you do it there, please test opening settings; that's the one that caused crashes. 

Which… may be because it unconditionally sets it. Try only setting it when multiline is also true in there.

drewvolz:
Huh, I wasn't bugged to add a CHANGELOG entry for that.

hawkrives:
I think because you also edited package.json, and it might short-circuit then

hawkrives (approved):

PR #3344, opened 2018-12-23
Fix broken filters

### LINKED ISSUE #3281: Filters are not applying
To do:
- [x] Toggles, lists, pickers not applying their new values to the UI or data
- [x] Calculation issue that prevents filters from rendering -- the view also freezes

<hr>

* A blocker on our release PR #3164 
* Partly addressed by fixing the crash on interaction in #3279

hannesmcman:
Resolves #3281.

The `getDerivedStateFromProps` in both the popover component and the fancy menu were overwriting the changes made to the state of each component when the filters were supposed to update, which was causing the unexpected behavior. I tested this and it resolves the bug.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Resolved some circular `require` statements in our code (#3280)
 - Resolved issue with OleCard login just never working (#3308)
 - Made build tooling always build tagged commits (#3323)
+- Fixed bug where filters were not applying correctly in menu and course search views (#3344)
 
 ### Removed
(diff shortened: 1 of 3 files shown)


rye (dismissed):
RSLGTM for nightly.

rye (dismissed):
Needs CHANGELOG update.

drewvolz (approved):
👍 :shipit:

hawkrives:
I think this breaks the case when filters change outside of the component, like when we clear them from the filter toolbar.

hannesmcman:
Ahh right, since the FilterPopover never unmounts...ok I will change this so that the filter stored in state updates to the filter passed through props ONLY if the props changed.. PR incoming.

PR #3347, opened 2018-12-24
Select, rather than deselect first filter option of OR filter

### LINKED ISSUE #3346: Select, rather than deselect, OR filter option when "Show All" is active
After showing the new filter toolbar feature to a few friends that use the app, I received the following suggestion, which will make the selection of filters more intuitive.

Setting:

When an OR filter is disabled, all of its options are selected. 

Current behavior:

When the user selects an option, that option is deselected, while all other options remain selected. 

Desired behavior:

When the user selects an option, that option should now be the only option selected.

hannesmcman:
Resolves #3346.

### DIFF
--- a/modules/filter/section-list.js
+++ b/modules/filter/section-list.js
@@ -22,7 +22,11 @@ export function ListSection({filter, onChange}: PropsType) {
 	function buttonPushed(tappedValue: ListItemSpecType) {
 		let result
 
-		if (selected.some(val => isEqual(val, tappedValue))) {
+		if (mode === 'OR' && selected.length === options.length) {
+			// if all options of an OR filter are selected and a user selects
+			// an option, make that the only sleected option
+			result = [tappedValue]
+		} else if (selected.some(val => isEqual(val, tappedValue))) {
 			// if the user has tapped an item, and it's already in the list of
 			// things they've tapped, we want to _remove_ it from that list.
 			result = reject(selected, val => isEqual(val, tappedValue))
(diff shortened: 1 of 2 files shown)


modules/filter/section-list.js
```diff
@@ -22,7 +22,11 @@ export function ListSection({filter, onChange}: PropsType) {
 	function buttonPushed(tappedValue: ListItemSpecType) {
 		let result
 
-		if (selected.some(val => isEqual(val, tappedValue))) {
+		if (mode === 'OR' && selected.length === options.length) {
+			// if all options of an OR filter are selected and a user selects
+			// an option, make that the only sleected option
```
drewvolz: ✂️ 
```suggestion
			// an option, make that the only selected option
```

drewvolz (approved):
I'm okay with this behavior change if others think it's intuitive 👍 .

I'd like to hear what @rye and @hawkrives have to say about this, though.

rye (approved):
I'm willing to give it a try in the nightly. Not sure how I feel about the semantics of this, but yeah.

hawkrives:
I don't like this.

hawkrives:
I also can't review on mobile atm.

hawkrives:
When I tap an option, I want to exclude it from the list; I don't want to only see it. 

That's why it works the way it does. 

If other people see if differently, we should think about that. But I don't like this. Sorry.

hannesmcman:
When this was first pointed out to me, I definitely agreed with you @hawkrives. But the more I thought about it, I think that our intimate knowledge of the app has clouded our judgement on what is most intuitive here..In most cases with the OR filters, I find myself first pressing the "Show All" button to unselect all of the filters (unintuitive if you're not used to it), and then pressing individual options to get the desired combination. I think many first time users are confused by the need to select "Show All" in order to unselect all before selecting individual options. I understand your desire to have the option to deselect individual options easily, but I think it is much more common to first deselect all and then pick a combination of options...

hawkrives:
Sure. Let's merge this and I'll live with it for a while. I may get used to it.

PR #3350, opened 2018-12-24
Fix Filter Popover Bug

hannesmcman:
In response to @hawkrives comment on #3344, this PR makes sure that the filter popover component is unmounted when it is closed, and remounts when a filter button is pressed. This way, we don't have to mess around with `getDerivedStateFromProps` at all.


hawkrives:
Oh. That works too. 

Good work!

(Approves. Mobile app can't submit reviews currently.)

rye (approved):

drewvolz (approved):
👍 :shipit:

PR #3352, opened 2018-12-24
Only rebuild filters when no filters are applied, not on PTR (Menus)

### LINKED ISSUE #3351: PTR clears filters
Problem first stated in #3348 by @drewvolz:

>Pull to Refresh when filters are applied -- the button does not change colors nor are the filters listed >on the popup, nor are they applied to the list after PTR-ing. It takes tapping the button again for the >filters to be applied.

Desired behavior: On PTR, filters should not be cleared. I think it is safe to assume that the filter categories will not change after a PTR, so we should be fine just preventing the rebuilding of fresh filters in `getDerivedStateFromProps`.

hannesmcman:
Resolves #3351.


drewvolz (approved):
👍 :shipit:

drewvolz:
(approval pending changelog entry)

PR #3353, opened 2018-12-25
Disable sticky headers for menus and courses

### LINKED ISSUE #3348: List items are misbehaving
I think something is mutating? When I scroll the menu list, it sort of... jitters. After that strange UI animation, it inserts a whole bunch of duplicate items in the list each time it loads. It continues to do so when I continue to scroll. Could you help debug this, please, @hannesmcman?

drewvolz:
Closes #3348

If we can compromise on the headers not being sticky for `FancyMenu` and `CourseSearch`, this appears to hide the react native issue with headers not remembering their positions when a list is scrolling.

I do not think we need to apply this to any of the other `SectionList` components.

Thoughts?

### DIFF
--- a/modules/food-menu/fancy-menu.js
+++ b/modules/food-menu/fancy-menu.js
@@ -207,6 +207,8 @@ export class FancyMenu extends React.Component<Props, State> {
 				renderItem={this.renderItem}
 				renderSectionHeader={this.renderSectionHeader}
 				sections={(groupedMenuData: any)}
+				// remove this after react native fixes the sticky header positioning
+				stickySectionHeadersEnabled={false}
 				style={styles.inner}
 				windowSize={5}
 			/>
--- a/source/views/sis/course-search/list.js
+++ b/source/views/sis/course-search/list.js
@@ -121,6 +121,8 @@ export class CourseResultsList extends React.PureComponent<Props> {
 				renderItem={this.renderItem}
 				renderSectionHeader={this.renderSectionHeader}
 				sections={(results: any)}
+				// remove this after react native fixes the sticky header positioning
+				stickySectionHeadersEnabled={false}
 				style={style}
 				windowSize={10}
 			/>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Ignore specific proptype warning for react markdown (#3329)
 - Only set `scrollEnabled` if `multiline` is true (#3337)
 - Select, rather than deselect, first selected filter option of OR filters (#3347)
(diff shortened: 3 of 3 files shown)


hawkrives (dismissed):
Before we merge this, I want someone to tell me when this was introduced. Because this wasn't an issue on Saturday's build, so somehow something we've done in the last three days broke this.

And I want to know what, instead of just papering over it.

There haven't been any giant branches merged in that timeframe, so you should just be able to bisect everything.

hawkrives:
For instance, I don't seem to be able to reproduce this problem on build 43564, from two days ago, although I may just not be doing it right.

hawkrives:
Ah, I missed that the bug only occurs when there are no filters applied, so we only just exposed it.

modules/food-menu/fancy-menu.js
```diff
@@ -207,6 +207,8 @@ export class FancyMenu extends React.Component<Props, State> {
 				renderItem={this.renderItem}
 				renderSectionHeader={this.renderSectionHeader}
 				sections={(groupedMenuData: any)}
+				// remove this after react native fixes the sticky header positioning
+				stickySectionHeadersEnabled={false}
```
hawkrives, line 211: Please include the relevant RN issue numbers in these comments

hawkrives (changes requested):

hawkrives:
@drewvolz Can you remind me why Calendar and Contacts and Streams can keep using the sticky headers? What makes them different?

drewvolz:
> Ah, I missed that the bug only occurs when there are no filters applied, so we only just exposed it. 

Exactly. The broken filters have hidden this problem from us. It’s only just now that it’s visible.

> Can you remind me why Calendar and Contacts and Streams can keep using the sticky headers? What makes them different?

I haven’t been able to reproduce this issue on them. I was only trying to tweak the noticeably impacted lists. It seems to me that our “big networked lists” (that can load a ton of rows that change via the filters) are the impacted ones.

If you’d prefer for it to be applied to all SectionLists, that’s fine, too.

hawkrives:
No, I'd just like to understand why it only seems to be the filtered lists. Streams and Hours can get pretty large… (I understand that you don't have the answers either.)

modules/food-menu/fancy-menu.js
rye: I second this—can you include `see facebook/react-native#20956, facebook/react-native#21468, facebook/react-native#22444, facebook/react-native#20127` here?

source/views/sis/course-search/list.js
```diff
@@ -121,6 +121,8 @@ export class CourseResultsList extends React.PureComponent<Props> {
 				renderItem={this.renderItem}
 				renderSectionHeader={this.renderSectionHeader}
 				sections={(results: any)}
+				// remove this after react native fixes the sticky header positioning
```
rye, line 124: Same here.

PR #2205, opened 2018-01-23
Version 2.5.0, β4

rye:
View [the comparison from v2.4.1](https://github.com/StoDevX/AAO-React-Native/compare/v2.4.2...v2.5.0-beta.3).
View [the comparison from v2.5.0-beta.1](https://github.com/StoDevX/AAO-React-Native/compare/v2.5.0-beta.1...v2.5.0-beta.3).
View [the comparison from v2.5.0-beta.2](https://github.com/StoDevX/AAO-React-Native/compare/v2.5.0-beta.2...v2.5.0-beta.3).
View [the comparison from v2.5.0-beta.3](https://github.com/StoDevX/AAO-React-Native/compare/v2.5.0-beta.3...v2.5.0-beta.4).

## Checklist
- [x] [Release PR]() successfully builds against `master` ([follow](https://circleci.com/workflow-run/21508433-e369-4ab2-8c7c-282a593d077c))
- [x] Signed-off/approved
- [x] GitHub release notes drafted ([wiki](https://github.com/StoDevX/AAO-React-Native/wiki/Release-Notes#v250))
- [x] [Release PR]() merged to `master`.
- [x] GitHub release created
- [x] Tag build passes ([follow](https://circleci.com/workflow-run/5d3909d7-7101-4e8d-87e9-b45779bbaaa8))

PR #3357, opened 2018-12-26
Apply upstream patch from #2205

drewvolz:
> Fix regression of VirtualizedList jumpy header

I'm in favor of applying this to the repo instead of #3353, so this would close #3353.

This takes the unverified fix from upstream to patch the VirtualizedList's leadSpace calculation. It appears to work. I like this better than disabling sticky headers altogether in menus and course search.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Added some logic to skip native builds if nothing that might affect them has changed (#3209)
 - All network requests are now cached according to the server's caching headers, even offline (#3310, #3320)
 - Added "Safety Concerns" tile that links to St. Olaf's official form for documenting safety concerns (#3345)
+- Added a prepare statement to apply an upstream fix to the VirtualizedList sticky header calculation (#3357)
 
 ### Changed
 - Adjusted and deduplicated logic in API scaffolding
--- a/contrib/0000-rn-22025-fix-jumpy-lists.patch
+++ b/contrib/0000-rn-22025-fix-jumpy-lists.patch
@@ -6,8 +6,8 @@
                const stickyBlock = this._getFrameMetricsApprox(ii);
 -              const leadSpace = stickyBlock.offset - initBlock.offset;
 +              const leadSpace =
-+                stickyBlockOffset -
-+                initBlockOffset -
++                stickyBlock.offset -
++                initBlock.offset -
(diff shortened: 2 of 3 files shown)


rye:
@drewvolz, would this be better formulated as a patch application?

drewvolz:
I don't have a preference. I like the `prepare` calls because they're really visual (meaning I see them all the time because we touch the package.json a lot) and a good reminder to keep searching for when the actual fix is released. I could be convinced to have this be an actual patch, though. Would you prefer that?

rye:
An actual patch would be applied only where we expect it—that's my main reason for thinking of that. I don't have a strong preference one way or the other… but this would allow us to commit the patch in our source tree. (still including something in `prepare` that could apply multiple patches) For the future, this would mean that we would just need to delete a `.patch` file once something goes upstream.

rye:
Here's an example patch file:

```diff
--- ./node_modules/react-native/Libraries/Lists/VirtualizedList.js
+++ ./node_modules/react-native/Libraries/Lists/VirtualizedList.js
@@ -805,7 +805,10 @@ class VirtualizedList extends React.Pure
             if (stickyIndicesFromProps.has(ii + stickyOffset)) {
               const initBlock = this._getFrameMetricsApprox(lastInitialIndex);
               const stickyBlock = this._getFrameMetricsApprox(ii);
-              const leadSpace = stickyBlock.offset - initBlock.offset;
+              const leadSpace =
+                stickyBlockOffset -
+                initBlockOffset -
+                (this.props.initialScrollIndex ? 0 : initBlock.length);
               cells.push(
                 <View key="$sticky_lead" style={{[spacerKey]: leadSpace}} />,
               );
```

(You can apply it with `patch -p0 -i <file>`.)

@drewvolz, if you want, I can go and adjust this PR to contain that patch in a `contrib` directory.

drewvolz:
@rye that would be great

hawkrives:
Lgtm

drewvolz:
[patch-package](https://www.npmjs.com/package/patch-package) looks neat and worth looking into

drewvolz:
But probably too much overhead for the task of applying a patch file

rye:
I've got a commit ready, I'll push it this afternoon

rye:
Hmm. I never flowed this. I'll poke it at 5pm.

rye (approved):

hawkrives (approved):
Concept approved; merge on green

drewvolz:
Thanks @rye for converting this into a patch. This is very clean 👍

PR #3361, opened 2018-12-27
Prompt user when they open the safety home button

drewvolz:
It's a little jarring to leave the app without being warned. This provides an alert to inform the user.

This is behavior that happens currently with `openUrlInBrowser` which we're calling when we open the top-level "Safety Concern" home tile.

> **Leaving All About Olaf**
> A web page will be opened in a browser outside of All About Olaf. Are you sure you want to open it?
> [Cancel]  [Open]

iOS | Android
--|--
ios-open-url | android-open-url

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - All network requests are now cached according to the server's caching headers, even offline (#3310, #3320)
 - Added "Safety Concerns" tile that links to St. Olaf's official form for documenting safety concerns (#3345)
 - Added a prepare statement to apply an upstream fix to the VirtualizedList sticky header calculation (#3357)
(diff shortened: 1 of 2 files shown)


rye (approved):
Looks great! Thanks for taking screenshots so I don't have to fire up a sim.

hawkrives:
I'd rather not ask "Are you sure?" - that sounds to me like we're subtly recommending that they not open the link after all. 

I like the rest of the wording.

rye:
This is a good catch on @hawkrives' part—maybe just say "Press Open to continue" or something?

drewvolz:
I can happily nix the second part of the message and avoid questioning altogether. Would that work?

hawkrives:
@drewvolz Yeah, that sounds good.

rye (approved):
Updated approval.

hawkrives (approved):

Issue #3260, opened 2018-11-29
<Markdown/> component crashes on nested lists


drewvolz:
<img src="https://user-images.githubusercontent.com/5240843/49233534-114cc080-f3ab-11e8-81e4-77794137a7a8.jpg" width=350 />

<details>

<summary>[full error] Invariant Violation: Nesting of <View> within <Text> is not currently supported</summary>

```
This error is located at:
    in glamorous(glamorous(glamorous.View)) (created by ReactMarkdown)
    in RCTText (at Text.js:151)
    in TouchableText (at Text.js:274)
    in SelectableText (created by glamorous(glamorous(SelectableText)))
    in glamorous(glamorous(SelectableText)) (at list.js:25)
    in RCTView (at View.js:45)
    in glamorous.View (at list.js:23)
    in ListItem (created by glamorous(ListItem))
    in glamorous(ListItem) (created by ReactMarkdown)
    in RCTView (at View.js:45)
    in glamorous(glamorous(glamorous.View)) (created by ReactMarkdown)
    in RCTView (at View.js:45)
    in ReactMarkdown (at markdown.js:51)
    in Markdown (at privacy.js:18)
    in PrivacyView (at SceneView.js:9)
    in SceneView (at StackViewLayout.js:483)
    in RCTView (at View.js:45)
    in AnimatedComponent (at screens.js:51)
    in Screen (at StackViewCard.js:42)
    in Card (at createPointerEventsContainer.js:26)
    in Container (at StackViewLayout.js:507)
    in RCTView (at View.js:45)
    in ScreenContainer (at StackViewLayout.js:401)
    in RCTView (at View.js:45)
    in StackViewLayout (at withOrientation.js:30)
    in withOrientation (at StackView.js:49)
    in RCTView (at View.js:45)
    in Transitioner (at StackView.js:19)
    in StackView (at createNavigator.js:57)
    in Navigator (at createKeyboardAwareNavigator.js:11)
    in KeyboardAwareNavigator (at createNavigationContainer.js:376)
    in NavigationContainer (at app.js:65)
    in ThemeProvider (at app.js:64)
    in Provider (at app.js:63)
    in App (at renderApplication.js:34)
    in RCTView (at View.js:45)
    in RCTView (at View.js:45)
    in AppContainer (at renderApplication.js:33)

<unknown>
    index.bundle?platform=ios&dev=true&minify=false:10566:20
updateContextConsumer
[...]

PR #3362, opened 2018-12-27
Paper over Privacy Policy crash by removing nested lists

hawkrives:
Part of https://github.com/StoDevX/AAO-React-Native/issues/3260, which I will re-title to <q>Nested Lists in Markdown components crash the app</q> once this is merged.

I removed the list items under the "What data do you collect?" heading, to allow the nested lists to remain listed.

I also re-wrapped all lines to 80col, which I thought was what we were targeting?

### DIFF
--- a/data/privacy.md
+++ b/data/privacy.md
@@ -32,12 +32,12 @@ reach servers that we control. Whenever your app makes such a request, the
 following information is logged:
 
 1. _Your public IP address_. We do not use this to identify you or for tracking
-purposes, and this is logged only for diagnostic purposes.
+  purposes, and this is logged only for diagnostic purposes.
 2. _The version of All About Olaf you are using and information about your
-phone_, which is sent as part of your device's request. This data is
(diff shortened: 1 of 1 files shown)


rye (commented):
Why does this PR change dot double-space (`.  `) to dot space (`. `)?

My argument for keeping it the way it was before `.  `:
- `.  ` is the standard for monospace systems.
- `.  ` gets rendered the same as `. ` on GH, but `.  ` has different breaking meaning from `. `
- Changing it in this PR introduces debate whereas keeping it the way it was does not.

hawkrives:
Because I hate editing files with two spaces after a period, and I hate how `man` inserts two spaces.

I didn't realize that you had introduced them originally, or I would have fought it then. 

I will remove that change from this PR.

hawkrives:
The double-space removal has been removed.

rye (approved):

drewvolz (approved):

PR #3363, opened 2018-12-27
s|US/Central|America/Chicago|g

### LINKED ISSUE #3198: Migrate to America/Chicago
According to the [(unofficial) table of IANA time zones](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones), `US/Central` is deprecated.

In fact, if you download an unofficial copy of the [time zone database](https://timezonedb.com/download), you will not find a mention to `US/Central` at all.

If you prefer "official" sources, IANA's [tzdb download](https://www.iana.org/time-zones) (current version: 2018g) has no mention of "US/Central" except in the `backward` file:
```console
$ rg 'US/Central' *
backward
118:Link	America/Chicago		US/Central
```
This `backward` file includes the following header:
```
# tzdb links for backward compatibility

# This file is in the public domain, so clarified as of
# 2009-05-17 by Arthur David Olson.

# This file provides links between current names for timezones
# and their old names.  Many names changed in late 1993.

# Link	TARGET			LINK-NAME
[...]
Link	America/Chicago		US/Central
[...]
```

I am led to believe what I thought was the case before, which is that `US/Central` is an old way of referring to a time zone.  `America/Chicago` is the best convention for us to follow, and while unlikely it is possible that `US/Central` could be removed later.

I move that we consolidate all of our time zone usage on `America/Chicago`, since that is what St. Olaf uses as its standard.

hawkrives:
Closes #3198.

There should be no functional difference here, as documented in #3198.

I also tweaked Jest to automatically set the timezone when doing tests, which I figure is OK since it's kinda equivalent with the app; the app sets this as a global thing during init, so the tests can too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Changed
 - Adjusted and deduplicated logic in API scaffolding
-- Switched time zone definitions from America/Winnipeg to US/Central
+- Switched time zone definitions from America/Winnipeg to US/Central, and then to America/Chicago
 - Pulled out time zone constants from all over the place into a central constant
 - Removed list footer credit for student activities dictionary (we're the source now)
 - Updated the privacy policy to more closely match what we do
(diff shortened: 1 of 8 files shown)


rye (approved):
Looks good.

PR #3370, opened 2018-12-28
Clean up and expand .gitignore file

rye:
- Adds `/logs` to the `.gitignore`.
- Removes `#`-preceded lines after comments

Anyone have recommendations on stuff I can remove in addition to what I've done here? Worth a pass through just to see if anything here actually hits the ignores.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,14 +33,8 @@ npm-debug.log
 yarn-error.log
 .eslintcache
 
-# BUCK
-buck-out/
-\.buckd/
-android/app/libs
+# Android Signing Keys
 *.keystore
-
-# KEYS
-.env.js
 
 # Tests and code coverage
 coverage
--- a/.gitignore
+++ b/.gitignore
@@ -54,7 +54,7 @@ scripts/travisci_deploy_key
 fastlane/play-private-key.json
 
 # generated files
-docs/
+/docs/
 
 # android keystores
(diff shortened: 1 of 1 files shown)


hawkrives:
You could remove

```diff
-# BUCK
-buck-out/
-\.buckd/
-android/app/libs
+# Android Signing Keys
*.keystore

-# KEYS
-.env.js
```

hawkrives (approved):

.gitignore
```diff
@@ -23,29 +21,21 @@ DerivedData
...
 node_modules/
 npm-debug.log
 yarn-error.log
 .eslintcache
 
-# BUCK
-buck-out/
-\.buckd/
-android/app/libs
+# Android Signing Keys
```
hawkrives: looking at this again, this section could move down and be joined up with the `# android keystores` section

.gitignore
```diff
@@ -64,7 +54,7 @@ scripts/travisci_deploy_key
 fastlane/play-private-key.json
```
hawkrives: and this line could also move down to `# android keystores`, and we could probably remove the `travisci_deploy_key` line, too.

.gitignore
rye: Actually, the `*.keystore` rule is a duplicate.

PR #3372, opened 2018-12-28
circle: Add --jit flag to Fastlane runs

rye:
Just for fun, and to see if this affects performance at all.  I doubt it will, since most of Fastlane's work involves waiting around for everything else to run.  But still.

I wonder about adding it to the `scripts/should-skip-build` thing… I've experienced a bit of slowdown though because it's so short.


rye:
We'll need to wait until Ruby 2.6.0 lands in the `circleci` Docker images, I guess.

rye:
I'm closing this for now, but I'll protect the branch from deletion until my PR to CircleCI's images repository is merged. (Business is starting back up soon so the wait will be over shortly.)

PR #3374, opened 2018-12-29
Show special BonApp menu messages for closed cafs

### LINKED ISSUE #3373: Show BonApp special message for closed cafeterias
Right now, if the menu retrieved is empty, we always return the vague message "No items found." Instead, we should see if BonApp has provided a special message, which is often more descriptive, and if so, show that instead, along with the date.

Current Behavior | Desired Behavior
-- | --
screen shot 2018-12-29 at 5 02 21 pm | screen shot 2018-12-29 at 5 23 12 pm

hannesmcman:
Resolves #3373.

### DIFF
--- a/modules/food-menu/fancy-menu.js
+++ b/modules/food-menu/fancy-menu.js
@@ -187,12 +187,13 @@ export class FancyMenu extends React.Component<Props, State> {
 		let messageView = <NoticeView style={styles.message} text={message} />
 
 		// If the requested menu has no food items, that location is closed
-		const isClosed = Object.keys(foodItems).length === 0
+		const isOpen = Object.keys(foodItems).length !== 0
+
 		let header = (
 			<FilterToolbar
 				date={now}
 				filters={filters}
-				isClosed={isClosed}
+				isOpen={isOpen}
 				onPopoverDismiss={this.updateFilter}
 				title={mealName}
 			/>
--- a/modules/food-menu/fancy-menu.js
+++ b/modules/food-menu/fancy-menu.js
@@ -186,8 +186,11 @@ export class FancyMenu extends React.Component<Props, State> {
 
 		let messageView = <NoticeView style={styles.message} text={message} />
 
+		// If the requested menu has no food items, that location is closed
+		const isClosed = Object.keys(foodItems).length === 0
 		let header = (
 			<FilterToolbar
+				isClosed={isClosed}
 				date={now}
 				filters={filters}
 				onPopoverDismiss={this.updateFilter}
--- a/modules/food-menu/fancy-menu.js
+++ b/modules/food-menu/fancy-menu.js
@@ -190,9 +190,9 @@ export class FancyMenu extends React.Component<Props, State> {
 		const isClosed = Object.keys(foodItems).length === 0
 		let header = (
 			<FilterToolbar
-				isClosed={isClosed}
 				date={now}
 				filters={filters}
+				isClosed={isClosed}
 				onPopoverDismiss={this.updateFilter}
 				title={mealName}
 			/>
--- a/modules/food-menu/filter-menu-toolbar.js
+++ b/modules/food-menu/filter-menu-toolbar.js
@@ -23,6 +23,7 @@ const styles = StyleSheet.create({
 
 type PropsType = {
 	date: momentT,
+	isClosed: boolean,
 	title?: string,
 	onPopoverDismiss: (filter: FilterType) => any,
 	filters: FilterType[],
@@ -30,6 +31,7 @@ type PropsType = {
 
 export function FilterMenuToolbar({
 	date,
+	isClosed,
 	title,
 	filters,
 	onPopoverDismiss,
@@ -57,10 +59,12 @@ export function FilterMenuToolbar({
 					/>
 				) : null}
 			</Toolbar>
-			<FilterToolbar
-				filters={nonPickerFilters}
-				onPopoverDismiss={onPopoverDismiss}
-			/>
+			{!isClosed && (
+				<FilterToolbar
+					filters={nonPickerFilters}
+					onPopoverDismiss={onPopoverDismiss}
+				/>
+			)}
 		</React.Fragment>
 	)
 }
--- a/modules/food-menu/filter-menu-toolbar.js
+++ b/modules/food-menu/filter-menu-toolbar.js
@@ -23,7 +23,7 @@ const styles = StyleSheet.create({
 
 type PropsType = {
 	date: momentT,
-	isClosed: boolean,
+	isOpen: boolean,
 	title?: string,
 	onPopoverDismiss: (filter: FilterType) => any,
 	filters: FilterType[],
@@ -31,7 +31,7 @@ type PropsType = {
 
 export function FilterMenuToolbar({
 	date,
-	isClosed,
+	isOpen,
 	title,
 	filters,
 	onPopoverDismiss,
@@ -59,7 +59,7 @@ export function FilterMenuToolbar({
 					/>
 				) : null}
 			</Toolbar>
-			{!isClosed && (
+			{isOpen && (
 				<FilterToolbar
 					filters={nonPickerFilters}
 					onPopoverDismiss={onPopoverDismiss}
--- a/source/views/menus/menu-bonapp.js
+++ b/source/views/menus/menu-bonapp.js
@@ -217,9 +217,9 @@ export class BonAppHostedMenu extends React.PureComponent<Props, State> {
 		const dayparts = cafeMenu.days[0].cafe.dayparts
 
 		// either use the meals as provided by bonapp, or make our own
-		const mealInfoItems = dayparts[0].length ? dayparts[0] : DEFAULT_MENU
+		const mealInfoItems = dayparts.length !== 0 && dayparts[0].length ? dayparts[0] : DEFAULT_MENU
 
-		const ignoreMenus = dayparts[0].length ? ignoreProvidedMenus : true
+		const ignoreMenus = dayparts.length !== 0 && dayparts[0].length ? ignoreProvidedMenus : true
 		return mealInfoItems.map(mealInfo =>
 			this.prepareSingleMenu(mealInfo, foodItems, ignoreMenus),
 		)
--- a/source/views/menus/menu-bonapp.js
+++ b/source/views/menus/menu-bonapp.js
@@ -222,6 +222,7 @@ export class BonAppHostedMenu extends React.PureComponent<Props, State> {
 
 		const ignoreMenus =
 			dayparts.length !== 0 && dayparts[0].length ? ignoreProvidedMenus : true
+
 		return mealInfoItems.map(mealInfo =>
 			this.prepareSingleMenu(mealInfo, foodItems, ignoreMenus),
 		)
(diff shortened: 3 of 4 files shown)


drewvolz (changes requested):
This reverts some logic that still needs to be taken into account.

source/views/menus/menu-bonapp.js
```diff
@@ -261,11 +261,6 @@ export class BonAppHostedMenu extends React.PureComponent<Props, State> {
 			return <NoticeView text={msg} />
 		}
 
-		if (Object.keys(this.state.cafeMenu.items).length === 0) {
```
drewvolz, line 264: Right, so I don't think we can outright remove this. I added it originally back in #3047 because we have to handle all ranges of cafes provided by bonapp. Sometimes there are no dayparts and no items, yet the cafeteria is open.

I'm testing this with cafe id 2 and the error that this block of code originally helped catch is back without it. So while it may cause another issue where it's at now, this still needs to be taken into account _somewhere_.

hannesmcman: Ahh I see..My latest commit should fix this issue. I tested it with cafe id 2 and it works as expected.

hawkrives (commented):
I haven't fully looked through this, but I just want to make sure that this doesn't hide the toolbar when you're looking at a closed menu (as opposed to a closed cafe).

hannesmcman:
If there are no food items, only the top portion of the toolbar will show, the part with the date in it. 

screen shot 2018-12-29 at 8 33 20 pm

hawkrives:
That would prevent you from changing menus, then, if you wanted to check Lunch and Breakfast was empty?

hannesmcman:
Sorry I should be more clear.. the only part of the filter toolbar that will be hidden is the part where the filters themselves can be changed. If the location offers multiple meals that toggle will still be available, which is something that is not currently possible before this PR.

https://github.com/StoDevX/AAO-React-Native/blob/31fbc658ca5cd4673139f614046e8afd5ea5e6ea/modules/food-menu/filter-menu-toolbar.js#L46-L69

source/views/menus/menu-bonapp.js
drewvolz: Okay. I'm noticing that we are unable to parse cafe ids 0 and 3000. The cafes returns `{"cafe":[]}`. We should add a check for no cafe data...

hannesmcman: Good point! The latest commit should resolve this..

drewvolz: I don’t know how many edge cases this spans to

hannesmcman: What kinds of edge cases are you envisioning?

hannesmcman:
Jest test failures seem to be unrelated to the contents of this PR...

hawkrives:
Re:jest; agreed. Can you reproduce locally?

hawkrives:
Re:menu switching, ok cool. Sgtm.

hannesmcman:
Re:jest, Yes I can reproduce locally...looking into it now, but could use some help as I did not write these tests or the code that they are testing..

rye:
Judging by the Jest output it looks like this is related to #3363, which was merged a bit ago. Does Jest run fine on `master`?

rye:
Oh I bet I know what it is, we had build issues last year around new year too, didn't we.

hannesmcman:
Ah yes I remember that...not sure exactly why these tests are failing though

rye:
Okay given I've spent ~1h looking at the tests and trying to make them green again to no avail, I think that as long as no `building-hours` or `bus-schedule`-related failures show up in your Jest output, they're safe to ignore. But obviously a build will not continue to native if that is not the case… Hmm.

rye:
Once #3376 gets merged, I think the Jest failure should go away.

modules/food-menu/fancy-menu.js
```diff
@@ -186,10 +186,13 @@ export class FancyMenu extends React.Component<Props, State> {
 
 		let messageView = <NoticeView style={styles.message} text={message} />
 
+		// If the requested menu has no food items, that location is closed
+		const isClosed = Object.keys(foodItems).length === 0
 		let header = (
```
rye, line 192: Please add a newline separating the previous line and this line.

source/views/menus/menu-bonapp.js
```diff
@@ -261,13 +262,14 @@ export class BonAppHostedMenu extends React.PureComponent<Props, State> {
...
-			const msg = 'No items were found.'
-			return <NoticeView text={msg} />
-		}
-
 		const {ignoreProvidedMenus = false} = this.props
 		const {now, cafeMenu, cafeInfo} = this.state
+		// The API returns an empty array for the cafeInfo.cafe value if there is no
```
rye, line 269: New line before this line please.

source/views/menus/menu-bonapp.js
```diff
@@ -248,11 +250,10 @@ export class BonAppHostedMenu extends React.PureComponent<Props, State> {
 			return <NoticeView buttonText="Again!" onPress={this.retry} text={msg} />
 		}
 
+		const cafe =
+			typeof this.props.cafe === 'string' ? this.props.cafe : this.props.cafe.id
```
rye, line 255: Type checking is a smell but it wasn't added here so don't mind this comment.

source/views/menus/menu-bonapp.js
```diff
@@ -217,9 +217,11 @@ export class BonAppHostedMenu extends React.PureComponent<Props, State> {
 		const dayparts = cafeMenu.days[0].cafe.dayparts
 
 		// either use the meals as provided by bonapp, or make our own
-		const mealInfoItems = dayparts[0].length ? dayparts[0] : DEFAULT_MENU
+		const mealInfoItems =
+			dayparts.length !== 0 && dayparts[0].length ? dayparts[0] : DEFAULT_MENU
 
-		const ignoreMenus = dayparts[0].length ? ignoreProvidedMenus : true
+		const ignoreMenus =
+			dayparts.length !== 0 && dayparts[0].length ? ignoreProvidedMenus : true
```
rye, line 224: What happens if `dayparts.length === 0`? Then this becomes `null` or `undefined`, I guess. Is that handled?

source/views/menus/menu-bonapp.js
```diff
@@ -217,9 +217,11 @@ export class BonAppHostedMenu extends React.PureComponent<Props, State> {
 		const dayparts = cafeMenu.days[0].cafe.dayparts
 
 		// either use the meals as provided by bonapp, or make our own
-		const mealInfoItems = dayparts[0].length ? dayparts[0] : DEFAULT_MENU
+		const mealInfoItems =
+			dayparts.length !== 0 && dayparts[0].length ? dayparts[0] : DEFAULT_MENU
```
rye, line 221: What happens if `dayparts.length === 0`? Then this becomes `null` or `undefined`, I guess. Is that handled?

modules/food-menu/filter-menu-toolbar.js
```diff
@@ -57,10 +59,12 @@ export function FilterMenuToolbar({
 					/>
 				) : null}
 			</Toolbar>
-			<FilterToolbar
-				filters={nonPickerFilters}
-				onPopoverDismiss={onPopoverDismiss}
-			/>
+			{!isClosed && (
```
rye: Could this be changed to `isOpen` instead? `!isClosed` is a double negative.

rye (commented):
Some nits.

source/views/menus/menu-bonapp.js
hannesmcman: If `dayparts.length === 0`, the ternary "short-circuits" and doesn't bother checking the second condition and the condition is evaluated as `false`. Right? @rye

rye: Well for one, I don't think this logic covers the case where `dayparts[0]` doesn't have a `.length`?

In any case, here's what I get when `dayparts.length === 0`:

```console
> let ary = []
undefined
> ary.length !== 0 && ary[0].length ? ary[0] : 'default'
'default'
```

***

Ohh… precedence here means the condition on the ternary is `dayparts.length !== 0 && dayparts[0].length`, not `dayparts[0].length`. (My bad.)

You're right, the `&&` stops evaluating after the first condition becomes `false`.

My new question is: can you wrap the condition in parens? (Will prettier change it back?) It's more readable that way.

rye:
@hannesmcman, I'll fire up a sim and give this a look-see now. Thanks for resolving those nits.

source/views/menus/menu-bonapp.js
drewvolz: I guess this actually covers it 👍

source/views/menus/menu-bonapp.js
hawkrives: Prettier will remove the parens.

hannesmcman: If `dayparts[0]` doesn't have a `.length`, that part of the condition will return `undefined`, which will make the whole condition evaluate as `undefined` and so the ternary will return the default schedule, as desired.

drewvolz (approved):

rye (approved):
Looks to be operating as intended to me, but please wait for additional feedback from @hawkrives and @drewvolz. BonApp's legacy API is serving us seemingly faulty information over breaks. (I'm specifically looking at Stav, which is marked as "open" on this API:)

image

As to the functionality of this, it definitely seems to be doing the right thing. The Legacy API is just inconsistent with their website, which was what confused me.

hawkrives:
Dammit. @rye, can you check to see if there's a new API powering their website, please, or if it's just server-side HTML?

source/views/menus/menu-bonapp.js
```diff
@@ -261,14 +263,16 @@ export class BonAppHostedMenu extends React.PureComponent<Props, State> {
...
-			const msg = 'No items were found.'
-			return <NoticeView text={msg} />
-		}
-
 		const {ignoreProvidedMenus = false} = this.props
 		const {now, cafeMenu, cafeInfo} = this.state
 
+		// The API returns an empty array for the cafeInfo.cafe value if there is no
+		// matching cafe with the inputted id number, otherwise it returns an non-array object
+		if (cafeInfo.cafe instanceof Array) {
```
hawkrives: Array.isArray(stuff), please

hawkrives (dismissed):
lgtm, pending comment

source/views/menus/menu-bonapp.js
hannesmcman: sure

rye:
@hawkrives, their website appears to be driven by its own HTML. I believe this because when I opened it, I didn't see any calls to anything like `legacy.cafebonappetit.com`, and everything was HTML or otherwise. No JSON.

drewvolz:
something something the site is wordpress

something something there are network requests but maybe server-side because `Bamco.menu_items` is already populated?
```js
<script type="text/javascript">
(function() {
	Bamco = (typeof Bamco !== "undefined") ? Bamco : {};
	Bamco.api_url = {
		items: 'https://legacy.cafebonappetit.com/api/2/items?format=jsonp',
		cafes: 'https://legacy.cafebonappetit.com/api/2/cafes?format=jsonp',
		menus: 'https://legacy.cafebonappetit.com/api/2/menus?format=jsonp'
	};
	Bamco.current_cafe = {
		name: 'Stav Hall',
		id: 243				};
	Bamco.view_tier = 1;
	Bamco.menu_items = [redacted];
                        Bamco.dietary_filters = [];
    			})();
</script>
```

rye:
@drewvolz the uh

https://legacy.cafebonappetit.com/api/2/cafes?cafe=261

That's what I was talking about tho.

So they're doing their own override thing? That'd explain it—if they don't fetch unless they're missing stuff… But the site still says Stav's closed whereas the API says it's open.

Still, it's BonApp's fault, I think. 😈 But maybe we should think about doing more than just an API request, e.g. more rigorous closure checks.

drewvolz:
On another note, from my code snippet, why does their script say 243 is stav’s id?

PR #3376, opened 2018-12-30
Fix Jest around the New Year

### LINKED ISSUE #3375: Jest is broken around New Year?
Q.v. #3374.

Looking at our nightly from today, [Jest seems to be broken](https://circleci.com/workflow-run/c2d6bf2d-c722-46ee-a208-db5e2525aeb9) (read: it's not producing the correct results).

Initially, I thought I would blame #3363, but:
- #3363 was merged 3 days ago, and we've had a successful nightly since then.
- I remember something about us having issues around the New Year last year or the year before that.
- The only tests that have magically gone red are the `building-hours` ones.

I'm pretty sure this is because of the New Year coming up—I think it's messing with our date parsing, or with those tests specifically because they're too general? Or maybe `isScheduleOpenAtMoment` is a bit wonky.

I will note that the app is running fine, and is producing expected results at this moment.

I'll poke around this. Others who actually wrote the tests are welcome to join.

rye:
Closes #3375.

My notes on `parseHours`:

>Previously, `m`'s dayOfYear was applied to both `open` and `close`, but if `m` was in the next year, then its `dayOfYear` would be low, and `open` and `close` would be freely wrapped around to the beginning of the year.
>
>Now, `parseHours` checks if `open` or `close` are six months prior to `m`, and if so, adds `1` year.  `m` is used only to provide context to "10am" and such kinds of strings, so I think this is a safe bet, and could probably be done less aggressively.

CC @hawkrives for another set of eyes.  I think a similar fix will be needed for bus schedules.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Fixed bug where filters were not applying correctly in menu and course search views (#3344, #3350)
 - Fixed the text color of the safety concerns button (#3349)
 - Fixed bug where applied filters would be cleared on a pull-to-refresh of the BonApp menus (#3352)
+- Fixed some strange behavior with hours and bus schedules around the new year (#3376)
 
 ### Removed
 - Removed the `prepare` script patching `ScrollEnabled` inside `RCTMultilineTextInputView` (#3337)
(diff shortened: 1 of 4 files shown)


rye:
Force-pushed to clean up the commit logs. (Squashed prettification into its prior commit, reworded to add `Tested-by`, and completely removed the messing around with time I was doing in CI.)

Now ready-for-review.

hawkrives (dismissed):
Could you duplicate your comment from parseHours into the Bus' parseTime function, please? Other than that I'm m/l fine with this.

rye:
Sure, I will go ahead and do that for consistency.

hawkrives (approved):

PR #3377, opened 2019-01-01
Render markdown with line breaks appropriately

### LINKED ISSUE #3369: Fix rendering of fill-columns on privacy policy
The privacy policy is rendering with sudden newlines. I think this is due to the fill-columns.

<details>
<summary>📸Screenshot</summary>

a1cc3716-7cb8-455d-acc0-1b5f6cd5b243

</details>

drewvolz:
Closes #3369.

This PR converts softbreaks to a single space.

<details>

~ | ~
--|--
Before | image
After | image

</details>

### DIFF
--- a/modules/markdown/markdown.js
+++ b/modules/markdown/markdown.js
@@ -15,7 +15,7 @@ import {Link} from './link'
 import {Image} from './image'
 import {List, ListItem} from './list'
 
-const Softbreak = () => <Text>{'\n'}</Text>
+const Softbreak = () => ' '
 const Hardbreak = () => <Text>&lt;br&gt;</Text>
 const HorizontalRule = glamorous.view({
 	width: '100%',
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Fixed the text color of the safety concerns button (#3349)
 - Fixed bug where applied filters would be cleared on a pull-to-refresh of the BonApp menus (#3352)
 - Fixed some strange behavior with hours and bus schedules around the new year (#3376, #3378)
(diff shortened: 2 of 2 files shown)


drewvolz:
Looks like it's time to update the jest tests/snapshots.

modules/markdown/markdown.js
```diff
@@ -15,7 +15,7 @@ import {Link} from './link'
 import {Image} from './image'
 import {List, ListItem} from './list'
 
-const Softbreak = () => <Text>{'\n'}</Text>
+const Softbreak = () => ' '
 const Hardbreak = () => <Text>&lt;br&gt;</Text>
```
hawkrives, line 19: I suppose we could go ahead and render hard breaks (lines ending with two spaces) as newlines now.

hawkrives: If you wanted.

hawkrives (approved):

hawkrives:
The rest failures appear to be due to the new year; let's rebase this once #3378 is merged?

rye:
Rebase away.

rye (approved):

modules/markdown/markdown.js
drewvolz: Let's not worry about that for now? This is in the readme for our version:

> Inline HTML is broken

If we upgrade, okay. But since we're pinned to this version of the dep I vote we worry about it later.

hawkrives: Inline HTML has nothing to do with hard linebreaks, fwiw. 

(Embedding literal HTML in the markdown source is broken; _I_ chose to print an escaped br tag here, when I was seeing what things did, and then I never encountered a hard linebreak to cause me to change it.)

But OK.

drewvolz: You're right. Thanks. I'll follow up with an additional change.

PR #3378, opened 2019-01-01
Fix Jest after the New Year too

rye:
Apparently the logic from #3376 cuts both ways.

No snapshot updates needed? That's fun.


rye:
We will possibly need an additional follow-up for any additional Snapshot work, but since the failing snapshots all seemed to declare their own moment I think this is fine.

rye:
Also, I think the followup is going to be to refactor this into a range inclusion check, but since that would require adding an additional dependency (dev) this should do the trick.

hawkrives (approved):

PR #3382, opened 2019-01-02
DatePicker: Remove getDerivedStateFromProps definition

### LINKED ISSUE #3381: Building Hours Editor does not keep changes to hours
[Screencap goes here]

1. Open Hours
1. Click a building
1. Tap "Suggest an Edit"
1. Make a change to one of the hours
1. Tap "Ok"
1. Notice that the time did not change, even if you tap back and tap back in

rye:
See #3381. Actually, fixes #3381.

All of a sudden, state persists! Amazing.


hawkrives (commented):
I need to do an audit of everywhere that we use this component - with this change, if we change the passed-in date outside of DatePicker, it won't get the new date. 

(Side note: we should also maybe tweak the API so that we don't have an initialDate prop somehow, but that's not required in this PR.)

hawkrives (approved):
This is only used in one location, so I'll approve+merge this. I still want to go through and rework the API, but that'll come later.

PR #3383, opened 2019-01-02
DictionaryEditorView: Remove getDerivedStateFromProps

### LINKED ISSUE #2577: Dictionary definition problem-reporter is editable but reverts all changes
This is for Android. I blame TextInput being multiline and already having text in the field.

rye:
Yeah. It works better now, but it may not handle outside-in prop changes.

Fixes #2577.


hawkrives (approved):
This one I think I'm okay with unconditionally, since there shouldn't be anything passing in new data.

PR #3416, opened 2019-01-14
ssb: Add detox to the list of important deps

rye:
We don't really want any changes to the Detox version to _not_ trigger a native build. This commit just adjusts the regexp.

Inspired by #3414.


hawkrives (approved):
Lol

Very good

PR #3436, opened 2019-01-27
apply two small circle config tweaks

hawkrives:
I can't find a reference to `cmd` in the docs, but they do talk about `command`. If the build fails, we'll know not to do the change.

The other is just clarification; numbers will get stringified anyway during env insertion, so let's be consistent.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -256,7 +256,7 @@ jobs:
       - run: *run-danger
 
   android: &android
-    # cmd: key is passed to solve a weird issue (see GitHub PR's 2170 and 2173).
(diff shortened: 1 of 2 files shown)


drewvolz (approved):

.circleci/config.yml
```diff
@@ -259,7 +259,7 @@ jobs:
     # cmd: key is passed to solve a weird issue (see GitHub PR's 2170 and 2173).
```
rye: Comment string needs an update

rye (changes requested):

rye (changes requested):
Needs changelog, then good to go.

rye (approved):

PR #3457, opened 2019-02-03
[WIP] Add safearea to the homescreen buttons

drewvolz:
This idea comes from @hawkrives. Adding safearea view to the buttons _does_ get us home screen buttons that respect the safe area, but they do have weird calculation side effects.

photo_2019-02-03 14 50 02

photo_2019-02-03 14 49 57


hawkrives:
@drewvolz I think this has been fixed now? The buttons are respecting SafeAreaView for me, at least.

drewvolz:
Looks like it.

PR #3459, opened 2019-02-04
TabBarIcon: Add platformPrefixIcon helper and use it

rye:
Instead of prefixing all input with `ios-`, (which necessitated #3458 to fix some breakage) use some reusable logic on both platforms.

The `platformPrefixIcon` helper has the following behavior:

| `${name}` is available¹ | `ios-${name}` is available¹ | `md-${name}` is available¹ | Result of `platformPrefixIcon(name)` |
|---|---|---|---|
| 🚫| any | any | `name` |
| ✅ | 🚫 | 🚫 | `name` |
| ✅ | ✅ | 🚫 | `name` |
| ✅ | 🚫 | ✅ | `name` |
| ✅ | ✅ | ✅ | `ios-${name}` or `md-${name}` depending on the value of `Platform.OS` |

* * *

Here are some screenshots of the newly-available `md-`-prefixed tab bar icons. I have no complaints as an Android user, and I think this nets us an improvement.

| screenshot_1549239084 | screenshot_1549238146 |
|---|---|
| screenshot_1549238160 | screenshot_1549238169 |

***
¹: Availability of a given icon is computed by `Icon.hasIcon`, which was introduced in `react-native-vector-icons` version `4.6.0` and works by querying the available glyphTable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Added a prepare statement to apply an upstream fix to the VirtualizedList sticky header calculation (#3357)
 - Added a prompt before the user leaves the app from open-browser-url (#3361)
 - Added more descriptive messages provided by Bon Appetit for closed cafeterias (#3374)
+- Added `@frogpond/icon` module with a helper for platform-prefixing modules (#3459)
 
 ### Changed
 - Adjusted and deduplicated logic in API scaffolding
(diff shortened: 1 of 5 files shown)


hawkrives (approved):
This looks good to me. 

Would you be interested in expanding the use of this helper into nav-buttons/open-settings and radio/controller? (A hardcoded ios- prefix is also in DisclosureArrow and the iOS EditHome, but those aren't an issue because Android will never run those code paths.)

rye:
@hawkrives, I had the same thought. I propose moving this helper out into the `navigation-buttons` module, since we use this mostly with navigation buttons. Then, I can adjust all variants of the `ios-` prefixing to use this instead. What do you think?

drewvolz (approved):
👍 :shipit:

rye:
ToT builds and works fine for me.

Merging this.

PR #3462, opened 2019-02-04
Fix color banding in balances (and clean up other spots too)

hawkrives:
This PR gets rid of that annoying gray banding on the Android balances screen.

I also went through the app and removed all other hard-coded references to `iosBackgroundColor`, except the ones in `.ios.js` files.

before | after
--- | ---
 |

### DIFF
--- a/modules/filter/filter-popover.js
+++ b/modules/filter/filter-popover.js
@@ -54,5 +54,5 @@ const popoverContainer = {
 }
 
 const arrowStyle = {
-	backgroundColor: c.iosLightBackground,
+	backgroundColor: c.sectionBgColor,
 }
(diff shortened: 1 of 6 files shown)


modules/filter/filter-popover.js
```diff
@@ -51,8 +50,5 @@ export class FilterPopover extends React.PureComponent<Props, State> {
 const popoverContainer = {
 	minWidth: 200,
 	maxWidth: 300,
-}
-
-const arrowStyle = {
-	backgroundColor: c.iosLightBackground,
```
hawkrives, line 57: turns out that RNPopoverView will take the background color from the popover, if provided, so I opted to provide it. this should also address the rubber-banding with white BG on iOS.

drewvolz (approved):
👍 :shipit:

rye (approved):

PR #3500, opened 2019-02-22
Require exact props for CellTextField

hawkrives:
By default, until Flow 0.92, which we don't use yet, Flow treats objects defined with `{ blah }` as somewhat-malleable objects – for instance, when we use a `{}` as the Props to a component, we can pass along any prop keys we want, and Flow won't complain.

The "exact object" syntax, `{| blah |}`, which is the _default_ in 0.92, I think, restricts the allowable keys down to just the ones listed.

By restricting the CellTextField props, I found that we were passing an unknown prop `hideLabel` to CellTextField, and removed it.

### DIFF
--- a/modules/tableview/cells/textfield.js
+++ b/modules/tableview/cells/textfield.js
@@ -32,7 +32,7 @@ const styles = StyleSheet.create({
 	},
 })
 
-type Props = {
+type Props = {|
 	label?: string,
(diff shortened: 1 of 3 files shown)


hawkrives:
image

see, I even missed one while preparing this PR

drewvolz (approved):

PR #3503, opened 2019-02-22
Fix OleCard login (maybe, probably… at least on the simulator)

hawkrives:
Part of #3318 

I attacked this from two angles:

---

I needed to address the way that LoginCredentials (fails at) rendering with async credential loading, and I also wanted to fix up a bit of the jitter that it has when it's busy running the initial can-you-log-in check with saved credentials. 

I did this with a few extra state variables, to track the progress through the loading sequence. Fairly straightforward.

---

Next, I needed to address (at the very least) the way in which the Balances screen always says "not logged in".

I did this by triggering the login mechanism if you aren't marked as logged in. Also fairly straightforward.

---

I'd like to merge this and test it in the nightly; I kinda doubt that this will fix all of our OleCard issues, but it should hopefully make it a little better. I'm just having a hard time replicating the issues on the simulator.


rye (approved):
Had a lot of questions as I read through it but they answered themselves. Looks good 👍

drewvolz (approved):
👍 :shipit: excellent job

rye:
(needs prettification and then green stamp from me)

PR #3509, opened 2019-02-24
Check if disclaimer is agreed to before fetching balances

### LINKED ISSUE #3506: Balances logs-in when user has not agreed to terms
Fallout from #3503. 

> Next, I needed to address (at the very least) the way in which the Balances screen always says "not logged in".
> 
> I did this by triggering the login mechanism if you aren't marked as logged in. Also fairly straightforward.

If you have not agreed to our notice that BonApp is the source of truth, you’ll still be able to use this screen anyways as it’s now going through the motions of logging you in and refreshing.

I think we need to tie-in the redux value of that alert before auto-logging-in and refreshing.

drewvolz:
Resolves #3506

It was possible to not agree to the disclaimer about BonApp and have a successful data fetch, subverting the check. This actually checks if the disclaimer has been agreed to before running `fetchData` which calls `getBalances`.

I'm testing this locally with invalid credentials (as I do not have a login) –– so let me know if I've missed something!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Fixed rendering of markdown hardbreaks (#3379)
 - Fixed the dictionary editor and made it handle user input again (#3383, #3387)
 - Fixed DatePicker by removing an unnecessary call to getDerivedStateFromProps (#3382)
+- Fixed the balances view from refreshing if the user has not agreed to the alert (#3509)
 
 ### Removed
 - Removed the `prepare` script patching `ScrollEnabled` inside `RCTMultilineTextInputView` (#3337)
--- a/source/views/sis/balances.js
+++ b/source/views/sis/balances.js
@@ -69,11 +69,11 @@ class BalancesView extends React.PureComponent<Props, State> {
 	}
 
 	componentDidMount() {
-		// calling "refresh" here, to make clear to the user
-		// that the data is being updated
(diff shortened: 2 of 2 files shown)


hawkrives:
How about

```js
		if (this.props.alertSeen) {
			// calling "refresh" here, to make clear to the user
			// that the data is being updated
			this.refresh()
		} else {
			Alert.alert('', LONG_DISCLAIMER, [
				{text: 'I Disagree', onPress: this.goBack, style: 'cancel'},
				{text: 'Okay', onPress: this.props.hasSeenAcknowledgement},
			])
		}
```

?

drewvolz:
In order to _not_ call the alert bit twice (it's currently happening inside `cDM`), should we move the alert out of `cDM`?

hawkrives:
componentDidMount is only called once…?

drewvolz:
`componentDidMount` is _also_ calling `this.refresh()`

hawkrives:
Ah, I forgot some context.

hawkrives:
```js
	componentDidMount() {
		if (this.props.alertSeen) {
			// calling "refresh" here, to make clear to the user
			// that the data is being updated
			this.refresh()
		} else {
			Alert.alert('', LONG_DISCLAIMER, [
				{text: 'I Disagree', onPress: this.goBack, style: 'cancel'},
				{text: 'Okay', onPress: this.props.hasSeenAcknowledgement},
			])
		}
	}
```

is my proposal.

hawkrives (approved):
changelog?

PR #3512, opened 2019-02-24
🗿Circle 2.1

rye:
# What?

This PR updates our CircleCI configuration to Circle 2.1.

To-Do list:
- [x] Change out "commands" for actual commands, and use those in the job configurations.
- [x] DRY any commands up so that we don't have to look at multiple copies of the same thing.
- [x] Look at possibly extracting out into orbs so this CI config can be less "special" and more reusable.
- [x] DRY up workflows.

# Why?

The old `<<`- and `*`-based DRY config strategy is (i) hard to understand, and (ii) no longer the best possible option.  Circle added Commands, Executors, and such, so that we can re-use bits of our configuration.  Plus, 2.1 is greater than 2.0, so we're cooler if we change the numbers to be bigger. 😎


.circleci/config.yml
```diff
@@ -1,50 +1,79 @@
...
+  danger:
+    description: 'Run Danger'
+    steps:
+      - run:
+          command: if ! [[ $CIRCLE_PR_NUMBER ]]; then yarn run danger --id $task; fi
+          when: always
+          environment:
+            DANGER_DISABLE_TRANSPILATION: 'true'
+
+  embed-env-vars:
```
hawkrives: this one could just vanish, tbh

.circleci/config.yml
```diff
@@ -1,50 +1,79 @@
...
+    description: 'Embed environment variables into the app'
+    steps:
+      - run: touch .env.js
+
+  restore-cache-bundler:
+    description: 'Restore the Bundler cache'
+    steps:
+      - restore_cache:
+          key: 'v2-ruby-dependencies-{{ arch }}-{{ checksum "Gemfile.lock" }}'
```
hawkrives: These giant strings should use YAML references, I think, so we don't wind up (say) changing `save` to v4 and forgetting `restore` at v3. We can just annotate this first string like `key: &bundler-cache-key 'v2-…'`, and then use it in the `save` command as `*bundler-cache-key`.

IMO anyway.

.circleci/config.yml
rye: @hawkrives are you saying we do not need this anymore?

.circleci/config.yml
rye: That's out of scope for this PR, but a good idea.

.circleci/config.yml
drewvolz: I agree with @hawkrives. We're no longer relying on the `.env.js` file since api keys have been moved to `ccc-server`.

rye: Removed.

rye:
I think this is now ready-for-review in a more official sense. Obviously more work can be done to get us farther down the modularization path, but this MWE makes me happier.

.circleci/config.yml
```diff
@@ -340,7 +345,6 @@ jobs:
       task: JS-bundle-android
     steps:
       - checkout
-      - attach_workspace: *attach-workspace-node_modules
```
hawkrives, line 320: Pretty sure we still want this one?

rye: Oh yeah

hawkrives: (insert kool-aid guy)

hawkrives (approved):
Sure

drewvolz (approved):
👍

PR #3517, opened 2019-02-26
Remove Google Analytics

rye:
Analytics is not helpful in our use case, and presents a possible privacy concern for users.  We have decided to at least remove analytics for the time being, especially considering that Google Analytics for mobile apps will be sunset in October of this year.

We do not want to move towards "Firebase" or other Google-things, instead opting to add analytics back selectively where we think information may be useful.

As we move towards other crash-tracking systems such as Sentry, we can also investigate whether it is sufficient to send copious breadcrumbs.  Otherwise, we will implement our own analytics system, so we can tune our retention and privacy policies accordingly.

This PR will end up doing a large set of things, but here is what it does right now.

- Remove `react-native-google-analytics-bridge` from our dependency manifests and build configurations.

- Remove redux wiring for "Feedback" (which is Analytics, I believe.)

- Remove google-analytics from our modules/analytics thing, and remove all references to its exports.

Flow, jest, and prettier seem happy locally, so I am pushing for initial review.

Left to do:

- [ ] Test on Android
- [ ] Test on iOS (can one of @hawkrives, @drewvolz do this?)
- [x] Update the privacy policy
- [x] CHANGELOG


data/privacy.md
```diff
@@ -39,13 +39,8 @@ following information is logged:
...
   fully anonymous.
 
-We also use a few tools to track our most used features and push notification
-receipts, as well as collect crash reports.  These tools are:
-
-- _Google Analytics_, which can be disabled via the "Share Analytics" toggle in
-  the Settings view.  Google Analytics records slightly more in-depth information
-  about your device, which we use for ensuring that our app looks good on all
-  screen sizes.
+We also use a couple of tools to track our most used features and push
```
hawkrives: Tracking most used features is part of GA, so we can strike it

modules/open-url/open-url.js
```diff
@@ -60,8 +58,7 @@ export function openUrl(url: string) {
 	}
 }
 
-export function trackedOpenUrl({url, id}: {url: string, id?: string}) {
-	trackUrl(id || url)
+export function trackedOpenUrl({url, _id}: {url: string, _id?: string}) {
```
hawkrives: Leave the right-side alone; waking won't complain about it. Revert the _-prefix on the left side.

modules/open-url/open-url.js
```diff
@@ -74,8 +71,7 @@ export function canOpenUrl(url: string) {
 	return true
 }
 
-export function openUrlInBrowser({url, id}: {url: string, id?: string}) {
-	trackUrl(id || url)
+export function openUrlInBrowser({url, _id}: {url: string, _id?: string}) {
```
hawkrives: Same comments about renaming

hawkrives (dismissed):
Overall, LGTM

modules/open-url/open-url.js
rye: >waking

(what do you mean by this?)

If I revert this change, I get an error, I'm pretty sure. `id` is unused which triggers an eslint warning which gets coerced to an error.

hawkrives: … I meant eslint. 

If you do `{url}: {url: string, id?: string}`, ESLint won't complain.

hawkrives (approved):
If you're up for it, I'd be interest to know the size difference of `node_modules` and the Android app apk before/after :-)

drewvolz (approved):
💯

PR #3528, opened 2019-03-04
Sectionlist tweaks on menu, contacts, and transportation

drewvolz:
Two things are happening in this PR:

First, the "Important Contacts" and "Transportation Other Modes" section lists do not have filters or search, so I ask that we remove<code><sup>1</sup></code> the `data` prop from them –– which I think should be `extraData`.

We can handle this in two ways:
* a) `data` could be passed-in as [`extraData`](https://facebook.github.io/react-native/docs/sectionlist#extradata)... or
* b) it could be removed because we're not searching or filtering on either of these views

Next, the "Fancy Menu" component is a filterable list, so we should rename<code><sup>1</sup></code> `data` to `extraData` because the data is passed-in within the `sections` prop, and we need to tell it to re-render because it is a pure component.  

<hr>

<code><sup>1</sup></code> `section.data` is a thing, but `data` is not a top-level prop of `SectionList` as per [the docs](https://facebook.github.io/react-native/docs/sectionlist.html#props).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Addressed color banding in SIS/Balances on Android (#3462)
 - [wip] Fix OleCard login stuff (#3503)
 - Adjusted how we present the BonApp ultimatum on first visiting the Balances tab (#3515)
+- Changed the `data` prop on fancy-menu to be `extraData` (#3528)
 
 ### Fixed
 - Fixed an issue where Fastlane was reporting build failures despite having skipped the build (#3215)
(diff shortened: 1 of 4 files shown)


hawkrives (approved):
Just to be clear – the reason Menus needs `extraData` isn't due to _filtering_, because filtering changes the data that we pass in to the `sections` prop.

It's needed because `renderItem` depends on data beyond the individual row item:

https://github.com/StoDevX/AAO-React-Native/blob/bc76c9894f3b186949688c056e5efb370261660d/modules/food-menu/fancy-menu.js#L148-L152

… which, yes, happens to be the filters, but that tangential.

---

TL;DR this looks OK to me.

rye:
@drewvolz thanks for the ship!

Looking at this on my phone, it may just be due to wireless reception where I am now, but this seems to load the menus _much_ faster than it was before.  Is there any possible explanation for this change in performance to be given by this PR?

hawkrives:
No; even if I've completely misunderstood extraData, it doesn't speed up rendering; it just tells React that it actually needs to re-render when data changes. 

Especially the loading spinner won't have been affected.

PR #3555, opened 2019-03-16
🔍 Searchable food menus

### LINKED ISSUE #2388: Search courses by description?
Followup from #2341 and #2383 

@hawkrives:

> I feel like searching description is going to generate false positives and be really slow.
>
> I'd like some benchmarks on search speed with and without description searching.
>
> Nothing fancy; just, like, log `performance.now` before and after the filter process happens.

### LINKED ISSUE #3523: Searchable menus
From a user:

> It would be nice to be able to search within a menu (e.g. The Cage) to find items by name!

(This might fit under an existing issue; "Related Issues" didn't find anything.)

Thoughts?

drewvolz:
Closes #3523.

This PR makes the menu titles and descriptions on iOS and Android searchable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Added more descriptive messages provided by Bon Appetit for closed cafeterias (#3374)
 - Added `@frogpond/icon` module with a helper for platform-prefixing modules (#3459)
 - Added consideration of `optionalDependencies` to our build scripts (#3489)
+- Added search to the food menus (#3555)
 
 ### Changed
 - Adjusted and deduplicated logic in API scaffolding
--- a/modules/food-menu/fancy-menu.js
+++ b/modules/food-menu/fancy-menu.js
@@ -28,7 +28,7 @@ import uniq from 'lodash/uniq'
 import map from 'lodash/map'
 import fromPairs from 'lodash/fromPairs'
 import deburr from 'lodash/deburr'
-import {SearchBar} from '@frogpond/searchbar'
+import {AnimatedSearchBar} from '@frogpond/searchbar'
 import {white} from '@frogpond/colors'
 
 type ReactProps = {
@@ -54,7 +54,9 @@ type Props = ReactProps & DefaultProps
 type State = {
 	filters: Array<FilterType>,
 	cachedFoodItems: ?MenuItemContainerType,
-	query: string,
+	typedQuery: string,
+	searchQueery: string,
+	isSearchbarActive: boolean,
 }
 
 const styles = StyleSheet.create({
@@ -77,7 +79,9 @@ export class FancyMenu extends React.Component<Props, State> {
 	state = {
 		filters: [],
 		cachedFoodItems: null,
-		query: '',
+		typedQuery: '',
+		searchQuery: '',
+		isSearchbarActive: false,
 	}
 
 	static getDerivedStateFromProps(props: Props, prevState: State) {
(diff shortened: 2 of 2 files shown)


hawkrives:
I will request to block this on a redesign of the menus screen. 

In my opinion, having Navigation > MenuInfo > FilterableItems > ActiveFilters > SearchBar (in whatever order) is too much for one screen. 

I do want to be able to search, but I am against just adding a search bar here.

drewvolz:
> In my opinion, having Navigation > MenuInfo > FilterableItems > ActiveFilters > SearchBar (in whatever order) is too much for one screen.

This is not the case while searching.

hawkrives:
But it is the case when you're scrolling, before you search. If you ever search; not every visit to the menu will involve a search.

drewvolz:
> But it is the case when you're scrolling, before you search.

Search and filters are not sticky.

drewvolz:
This is not the case while scrolling regardless of searching, on both platforms.

drewvolz:
It’s also not the case while searching because filters are hidden while searching.

hawkrives (changes requested):
I will request to block this on a redesign of the menus screen.

In my opinion, having Navigation > MenuInfo > FilterableItems > ActiveFilters > SearchBar (in whatever order) is too much for one screen.

I do want to be able to search, but I am against just adding a search bar here.

(just moving to an official review-comment comment)

drewvolz:
What do you suggest @hawkrives

drewvolz:
My first thought is to add a `headerRight` option to the menu screens, so that lands us at

```js
MenusView.navigationOptions = {
	title: 'Menus',
	headerRight: <SearchButton />,
}
```

where `SearchButton` looks _something_ like this

```js
type Props = {
	onSearch: () => mixed,
	searching: boolean,
}

export function SearchButton(props: Props) {
    const icon = `${Platform.OS === 'ios' ? 'ios' : 'md'}-search`
	return (
		<Touchable
			highlight={false}
			onPress={props.onSearch}
			style={styles.button}
		>
			<Icon name={icon} style={styles.icon} />
		</Touchable>
	)
}
```

but that leads to a tricky situation where we don't have the menu view's right button in communication with the view it's acting on. I'm thinking of then storing whether we've tapped that a button inside redux so that I can handle view changes to show/hide one of search, active filters, filterable items, or menu info as appropriate. 

The goal would be to show search IFF the nav button for search is tapped and hide the rest and reverse when the "cancel" button is tapped on the search bar.

Do we go the redux route? I'd like to get an MVP search into the menu as it was a feature directly requested by one of our users, so advice and suggestions from you would be appreciated.

hawkrives:
I do not have a suggestion. I just know that I'm opposed to _ever_ having 5 rows of stuff before the content appears, whether or not it persists when you scroll.

hawkrives:
(Also, I don't know if this changes the calculus, but the person who requested this is a friend and a coworker. He's understanding of taking time to do things right.)

drewvolz:
> I do not have a suggestion. I just know that I'm opposed to ever having 5 rows of stuff before the content appears, whether or not it persists when you scroll.

It's 4, is that okay? While not everyone uses the filters, that would make it 3 if you're browsing the menu. I'll mention this is the same amount of things that appear on course searching with filters activated.

drewvolz:
@hawkrives I don't think we're going to have a change to our overall navigation anytime soon. Thoughts on if this should be closed since it won't be accepted?

drewvolz:
Closing this out until it can resolve the concerns about navigation items. Perhaps we can take the approach GH takes with their issue/PR list in the future.

PR #3557, opened 2019-03-16
Upgrade to React-Native v0.59

hawkrives:
Closes #3433 (by superseding it.)

I have tested that both apps build and launch; I have not gone through them thoroughly.

A checklist:

- [x] Run `react-native upgrade`
- [x] Update flow to v0.92
    - [x] update package.json for flow-bin
    - [x] Install flow-typed locally and run `flow-typed update`
    - [x] run scripts in scripts/ to trim and find duplicate files, and remove the duplicate files as needed
- [x] Ensure Android app launches, due to `You need to use a Theme.AppCompat theme (or descendant) with this activity.` change

I'd like to do these in follow-up PRs:

- [x] Enable "inline requires" – https://facebook.github.io/react-native/docs/0.58/performance#adding-a-packager-config-file (see #3573)
- [x] Switch to external repo for async-storage (see #3565)
    - [x] Add dep on `@react-native-community/async-storage`, and run `react-native link`
    - [x] Switch `import {AsyncStorage} from 'react-native'` to `import AsyncStorage from '@react-native-community/async-storage'`
- [x] Switch to external repo for netinfo (see #3566)
    - [x] Add dep on `@react-native-community/netinfo`, and run `react-native link`
    - [x] Switch `import { NetInfo } from "react-native"` to `import NetInfo from "@react-native-community/netinfo"`

(Switching to the external repos can wait because they've been deprecated in RN0.59, but haven't been removed yet.)


drewvolz (approved):
**Wow** nice work with this one. I can't see anything wrong from the multiple passes I've made over this. I've yet to build it on my own machine, though. :shipit:

android/gradle.properties
```diff
@@ -10,8 +10,7 @@
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
 # Default value: -Xmx1024m -XX:MaxPermSize=256m
-# org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
-org.gradle.jvmargs=-Xmx512m -XX:+HeapDumpOnOutOfMemoryError
+org.gradle.jvmargs=-Xmx1024m -XX:MaxPermSize=256m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
```
rye, line 13: Do we need to keep these arguments in here, or are they actually "defaults"? (I.e. would removing them mean the same behavior?)

CHANGELOG.md
```diff
@@ -46,6 +46,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Updated to CircleCI 2.1 configuration syntax (#3512)
 - Adjusted how we present the BonApp ultimatum on first visiting the Balances tab (#3515)
 - Changed the `data` prop on fancy-menu to be `extraData` (#3528)
+- Upgraded to RN 0.59 (#3557)
```
rye, line 49: Does this need to be abbreviated?

android/app/src/debug/res/xml/react_native_config.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<network-security-config>
```
rye, line 2: Do we really need this file in here? Or can this be ignored?

(10.0.2.2 means something entirely different on STO vs elsewhere)

rye (approved):
Seems fine to me. Couple of nits but nothing that should block merging for nightly.

android/gradle.properties
hawkrives: Well, the "default" line would be doubled RAM and wouldn't dump on OOM (I recall adding the dump-on-OOM because that was actually an issue where it would OOM and not tell me.)

So I don't know? But I don't think I want the defaults.

android/app/src/debug/res/xml/react_native_config.xml
hawkrives: That is the default file from RN upstream. I don't know the answer to your question.

CHANGELOG.md
hawkrives: It's the developer changelog, for us; I don't particularly care, but I will forget and abbreviate it next time as well.

android/app/src/debug/res/xml/react_native_config.xml
hawkrives: I think the file is needed, and the localhost is needed, because I think I remember that Android 8 or 9 did or will be doing forced-HTTPS suggestions like iOS does? (If that's already a thing, then IDK.)

PR #3558, opened 2019-03-16
📇 Native Directory view

drewvolz:
Implementing a native directory client for iOS and Android using the directory API.

📸 Screenshots

~ | Android | iOS
--- | --- | ---
Initial | android initial | ios initial
Results | android results | ios results
Detail | android detail | ios detail

<hr>

Known issues
- [x] flowtypes
- [x] error when searching about splitToArray being called on null
- [x] android searchbar eating the query while typing
- [x] android searchbar covering the first result
- [x] no idea what on-campus results will look like
- [x] no results not differentiated
- [x] unknown if error state presents itself
- [x] office view changes
  - [x] finish implementing the office section changes on iOS
  - [x] implement the office section changes to android
- [x] android searchbar is very broken on Orgs, Dictionary, and ClassLab Search, but works well here
- [x] department chips need to wrap when it gets really long
- [x] android searchbar is broken (fixed in https://github.com/StoDevX/AAO-React-Native/pull/3715)
- [x] office hours tweaks
- [ ] telling fetch() to not cache search results isn't working
- [ ] need to differentiate student vs fac/staff props


source/views/directory/types.js
```diff
@@ -0,0 +1,36 @@
...
+	email: string,
+	homePhone: null,
+	onLeave: boolean,
+	officePhone: string,
+	homeAddress: {
+		zip: string,
+		city: string,
+		country: string,
+		state: string,
+		street: Array<string>,
```
drewvolz: not sure why this is an array on the backend?

hawkrives: Because there are up to two street addresses and I didn't want to put two props in

hawkrives (commented):
Just remember that the API will be changing in as-yet-undetermined ways on either Monday or Tuesday

hawkrives:
however! this looks really nice!

source/views/directory/list.js
```diff
@@ -0,0 +1,244 @@
...
+					...item,
+					$sortableName: sortableName,
+					$groupableName: head(startCase(sortableName)),
+				}
+			})
+
+			const sorted = sortBy(withSortableNames, '$sortableName')
+			const grouped = groupBy(sorted, '$groupableName')
+			this.setState(() => {
+				results: grouped
```
hawkrives: This is not returning an object; it's returning `undefined`, after evaluating a block with a `results:` label. You will need to wrap the {…} in parentheses in order for it to be parsed as an object.

drewvolz:
Help wanted on the known issues listed in the PR description 😅

hawkrives:
(I just rebased onto master)

rye (commented):
👀 All the eslint issues are just related to variables being created but never used. Remove them or prefix with `_`?

Then the `flow` stuff… I'm of no use on that.

hawkrives:
I've updated the screenshots and TODO list

drewvolz:
Comment from @hawkrives 

> One thing to note for the native directory - I expect that the directory API will start returning results for departments in the main list - ie, searching for History will return all the people in the history department, as well as a card representing the History department itself - name, AAA location, main contact number, etc.
>
> We may want to visually distinguish them in the results list, and they may require a different API format. We shall see.
>
> Also office hours are getting tweaked ~soon

drewvolz:
@hawkrives so where are we at again? 

edit: updating the list up top

modules/tableview/cells/multi-line-left-detail.js
```diff
@@ -23,7 +25,13 @@ export class MultiLineLeftDetailCell extends React.PureComponent<LeftDetailProps
 				</Text>
 			</View>
 		)
-		return <Cell cellContentView={cellContent} />
+		return (
+			<Cell
+				accessory={accessory}
+				cellContentView={cellContent}
+				onPress={onPress ? onPress : null}
```
hawkrives: If you've defined `onPress` to be potentially null, why the runtime check here?

source/views/directory/detail-android.js
```diff
@@ -93,11 +95,11 @@ export function DirectoryDetailView(props: Props) {
 					))}
 				</View>
 
-				{officeHours ? (
+				{officeHours && (officeHours.prefix || officeHours.content) ? (
```
hawkrives: You should be able to just check if `officeHours` is null; the API will return `null` if the person doesn't have anything set.

hawkrives (dismissed):

source/views/directory/detail-android.js
drewvolz: It’s more that I’m finding myself wanting to know if these individual props have content or not.

modules/tableview/cells/multi-line-left-detail.js
drewvolz: To prevent the cell from being pressable

drewvolz: Ah — I see your point now. I was thinking about a different file.

drewvolz:
Screenshots have been updated. 

Edit: But they're already out of date

drewvolz:
@rye @hawkrives I think we're at a place where we can merge this branch and make follow-ups afterwards.

source/views/directory/lib.js
```diff
@@ -0,0 +1,61 @@
...
+export const homeAddressText = (homeAddress: Address) => {
+	const {street, city, state, zip, country} = homeAddress
+	return `${street.join('')}\n${city} ${state} ${zip} ${country}`.trim()
+}
+
+export const departmentOrMajorTitle = (
+	departments: Array<Department>,
+	classYear: string,
+) => {
+	if (classYear !== null) {
```
rye: The Rustacean in me is wary of a `null` comparison, especially one that's imbued with meaning as this one seems to be. Is there another way to do this check?

Also, what if `classYear` is non-`null` for fac/staff? Like if staff want to indicate that they were '69, we should support that.

(I think it's better to have @hawkrives figure out how to do this server-side first, personally.)

drewvolz:
@hawkrives what are the longstanding plans to the directory API? I am unsure if/when the updated API results for departments will land so that this client must visually distinguishing between people vs departments.

Can we get this merged and reevaluate the API once that actually happens?

drewvolz:
@hawkrives 😄

hawkrives:
Okay, we should merge!

drewvolz:
Next step here is to rebase this changeset on top of head.

rye:
@drewvolz I can rebase if you want me to!

drewvolz:
@rye thanks for offering. I think I've got it, but we'll have to triple-check. This branch was pretty far behind.

drewvolz:
Let's rebase and see where this is at!

drewvolz:
Thoughts on merging this after a rebase?

rye:
Has it been rebased? I think I would be in favor of this one.

hawkrives:
Sure thing

drewvolz:
Current status: I'm still cleaning this up. I'll comment and relabel afterwards.

source/views/types.ts
```diff
@@ -1,11 +1,11 @@
...
-	navigation: NavType
+export type TopLevelViewPropsTypeWithParams<P = {}> = {
+	navigation: {
+		navigate: (string, ?Object) => mixed,
+		push: (string, ?Object) => mixed,
+		goBack: () => mixed,
+		getParam: (string, any) => any,
+		state: {params: P},
+	},
 }
```
drewvolz: I'm not sure how to make prettier happy with this file yet.

source/views/directory/detail.android.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+	const fabActions = [
+		...campusLocations.map((loc: CampusLocation) => ({
+			icon: 'phone',
+			label:
+				campusLocations.length === 1
+					? 'Call'
+					: `Call ${loc.buildingabbr} ${loc.room}`,
+			onPress: () => callPhone(loc.phone),
+		})),
```
rye: Could this `map` body get extracted to a function?

source/views/directory/detail.android.tsx
```diff
@@ -0,0 +1,194 @@
...
+		...(email
+			? [
+					{
+						icon: 'email',
+						label: 'Email',
+						onPress: () => sendEmail({to: [email], subject: '', body: ''}),
+					},
+			  ]
+			: []),
+	]
```
rye: In a similar vein to the previous idea on the `map` body — could this be a function as well?  Even if it is just an if-statement.

source/views/directory/detail.android.tsx
```diff
@@ -0,0 +1,194 @@
...
+		displayTitle,
+		photo,
+		officeHours,
+		profileUrl,
+		email,
+		departments,
+		username,
+	} = props.navigation.state.params.contact
+
+	const [fabOpen, setFabOpen] = React.useState(false)
```
rye: To make it more clear what `fabOpen` means, could you prefix it with `is`? (`isFabOpen`, and `setIsFabOpen`)

source/views/directory/detail.android.tsx
```diff
@@ -0,0 +1,194 @@
...
+						onPress: () => sendEmail({to: [email], subject: '', body: ''}),
+					},
+			  ]
+			: []),
+	]
+
+	const showFab = fabActions.length >= 1
+
+	return (
+		<>
```
rye, line 52: Was there any particular reason to not use `React.Fragment`?  (More just curious, but I had to look up this syntax.)

source/views/directory/types.ts
```diff
@@ -0,0 +1,70 @@
...
+	onLeave: ?OnLeave
+	photo: string
+	profileUrl: ?string
+	suffixName: ?string
+	thumbnail: string
+	title: ?string
+	username: ?string
+	// added from ccc-server
+	description: string
+}
```
rye: Typescript complains about these `?<type>` keys.

If these are optional (i.e. are optional if I'm constructing one of these), these could be written as e.g. `email?: string`.

If these are nullable, that should be explicit via `| null`, e.g. `email: string | null`.

source/views/types.ts
```diff
@@ -1,11 +1,14 @@
...
 
-export type TopLevelViewPropsType = {
-	navigation: NavType
+export type TopLevelViewPropsTypeWithParams<P = AnyObject> = {
+	navigation: {
+		navigate: (routeName: string, params?: AnyObject) => mixed
+		push: (routeName: string, params?: AnyObject) => mixed
+		goBack: () => mixed
+		getParam: (paramName: string, defaultValue: any) => any
+		state: {params: P}
```
rye: TypeScript is complaining about these things too.  Is it possible to turn `mixed` into a union type at all?

source/views/directory/detail.ios.tsx
```diff
@@ -0,0 +1,157 @@
...
+								title={email}
+							/>
+						) : null}
+
+						{officeHours && (
+							<MultiLineLeftDetailCell
+								accessory={officeHours.href ? 'DisclosureIndicator' : undefined}
+								detail={officeHours.title}
+								onPress={
+									officeHours.href ? () => openUrl(officeHours.href) : null
```
rye: TypeScript is complaining about this line and a couple of the other `onPress` lines.

```
No overload matches this call.
  Overload 1 of 2, '(props: LeftDetailProps | Readonly<LeftDetailProps>): MultiLineLeftDetailCell', gave the following error.
    Type '(() => Promise<any>) | null' is not assignable to type '(() => any) | undefined'.
      Type 'null' is not assignable to type '(() => any) | undefined'.
  Overload 2 of 2, '(props: LeftDetailProps, context: any): MultiLineLeftDetailCell', gave the following error.
    Type '(() => Promise<any>) | null' is not assignable to type '(() => any) | undefined'.ts(2769)
```

Do we need `null`? If so, the type definition needs to include that case. Or could the false case just be `undefined`?

(A side note: I'm not against giving anonymous functions a name and promoting them to functions — that helps make things traceable.)

rye (dismissed):
Couple of ideas so far.  Looking good!

source/views/directory/detail.android.tsx
drewvolz: No particular reason. I like the shorthand but if it is hard to read, I can switch it?

source/views/directory/detail.ios.tsx
drewvolz: Good call. I'll opt to return `undefined` in the false case to match the updated type.

And I feel similarly about extracting inlined functions! I suppose I don't have a set rule other than thinking about how much work they do, how many lines they are, or if they can be shared.

drewvolz: > and a couple of the other onPress lines.

As for other `onPress` return types like `<Cell />` (react-native-tableview-simple):

Return type is `() => void | false` but we tend to do other things like call `openUrl()` ... `() => Promise<any>`.

Other than forking the project and updating the type or patching locally, are there other options?

source/views/directory/types.ts
drewvolz: Oh yes, great suggestion. I forgot to modify these types!

source/views/directory/detail.ios.tsx
drewvolz: What do we think of this solution? The type that ships for `CellInterface['onPress']` looks like an oversight as we know we are able to return, pass, and do things other than the provided signature `(() => any) | undefined)`.

I am suggesting a replacement `onPress` type because the one that ships with the library is too restrictive. I would prefer if we overwrite it and use our own less restrictive `onPress`.

```ts
type CellInterfaceModifiedType = Omit<CellInterface, 'onPress'> & {
	onPress?: () => void   // (or some variant of `void | undefined`)
}
```



Does this sound okay to you, @rye?

drewvolz:
- Brought this branch up to date with the latest changes
- Refactored to be in-line with the current navigation + route hooks
- Added additional typescript type to work with tableview-simple's types
- Added additional typescript type to work with FABGroupAction types
- Replaced the searchbar with the native one.

drewvolz:
Anything more to review here, @hawkrives and @rye? Thinking this is in pretty good shape!

source/navigation/types.tsx
```diff
@@ -17,9 +17,14 @@ import type {
 	MasterCorIconMapType,
 	MenuItemType as MenuItem,
 } from '../views/menus/types'
+<<<<<<< HEAD
+import {DirectoryItem} from '../views/directory/types'
+||||||| fe7026171
```
hawkrives: Merge conflict

modules/searchbar/searchbar-android.tsx
```diff
@@ -0,0 +1,26 @@
+import * as React from 'react'
```
hawkrives: Didn't we get rid of this file?

source/views/directory/types.ts
```diff
@@ -0,0 +1,70 @@
...
+	lastName: string
+	officeHours: OfficeHours | null
+	onLeave: OnLeave | null
+	photo: string
+	profileUrl: string | null
+	suffixName: string | null
+	thumbnail: string
+	title: string | null
+	username: string | null
+	// added from ccc-server
```
hawkrives, line 59: We're not proxying through ccc, are we?

hawkrives (changes requested):

source/navigation/types.tsx
drewvolz: Dang it, thanks.

modules/searchbar/searchbar-android.tsx
drewvolz: Somehow my merge really botched this. Git said it was removed when I committed this...

source/views/directory/types.ts
drewvolz: I had proposed a changeset to ccc to proxy before we realized we'd end up losing results that way. The proxied endpoint never was merged, but I can take a look to see if I forgot something on the client side. Good catch.

drewvolz: We're good, looks like the comment remained but the logic is all on the client.

```tsx
//  https://github.com/StoDevX/AAO-React-Native/blob/directory/source/views/directory/helpers.ts#L80-L83

const shortRoomDetail = (loc: CampusLocation): string => {
	const shortRoom = `${loc.buildingabbr} ${loc.room}`.trim()
	return `${shortRoom ? `${shortRoom} • ` : ''}${loc.phone}`
}

// https://github.com/StoDevX/AAO-React-Native/blob/directory/source/views/directory/helpers.ts#L107-L110

const locations = campusLocations.map((loc: CampusLocation) => ({
	...loc,
	shortLocation: shortRoomDetail(loc),
}))
```

hawkrives:
Do we need android/app/src/main/assets/fonts/MaterialIcons.ttf? Git says it's added by this PR

drewvolz:
Looks like that's commit [846826e](https://github.com/StoDevX/AAO-React-Native/commit/846826e2db50cf6952353dec52e1329354b4d03a)

> copy MaterialIcons in for react-native-paper
> 
> hawkrives authored and drewvolz committed on Oct 6, 2021

... so I suppose we should check out whether RN paper uses this font?

hawkrives (approved):

drewvolz:
@rye any thoughts on these changes?

drewvolz:
@hawkrives addressing MaterialIcons.ttf:
- 👍 We can and should remove from this pull request, I see an intermittent duplicate assets issue if I do not remove it 
- ❓ I'm not seeing icons from react-native-paper appear in the directory detail view for Android. I now see failed prop type warnings for icon names such as "room", "more-vert", "sentiment-satisfied"

hawkrives:
Oh, they probably do need material icons, don't they. Hmm. 

Don't we already link a copy of material icons? How do we render the Back button on android without them?

drewvolz:
✅ @hawkrives material icons building and linking issues are now fixed for Android!

hawkrives:
Is it too greedy for me to ask you to rebase such that the extra .ttf is never added to the repo?

Either way, I'm finally ready to approve this!

drewvolz:
Alright, as requested I removed both font commits:
- the commit which brought a duplicate copy of MaterialIcons.ttf into the repo
- the commit which removed that copy.

hawkrives (approved):

PR #3560, opened 2019-03-17
Dev API explorer

drewvolz:
This comes out of an irritation of making and changing my apple Shortcuts.app recipes to test the data when it makes sense to do this in the app itself.

<hr>

Developer settings
* Separate developer section in the settings groups
* Adds bon app dev picker to developer settings 

API Explorer
* Adds an API explorer powered by the proxy server in developer settings
  * note: the segmented view (JSON vs list is only iOS at this time)
* Adds the explorer to the developer settings menu

<details>

~ | ~
--|--
image | image
image | image
image | image
</details>

<hr>

Todo
- [x] android textview isn't rendering at the top
  - (tried to address this with textAlignVertical via the [docs](https://facebook.github.io/react-native/docs/textinput#multiline))
- [x] android textview isn't scrolling


source/views/menus/dev-bonapp-picker.js
```diff
@@ -73,3 +73,7 @@ export class BonAppPickerView extends React.PureComponent<Props, State> {
 		)
 	}
 }
+BonAppPickerView.navigationOptions = {
+	title: 'BonApp',
+	tabBarIcon: TabBarIcon('cog'),
```
hawkrives: There's no tab bar here anymore

source/views/settings/screens/api-test.js
```diff
@@ -0,0 +1,93 @@
...
+		results: '',
+		error: '',
+	}
+
+	fetchData = async (path: string) => {
+		try {
+			let responseData: Array<any> = await fetch(API(path), {
+				cache: 'no-store',
+			}).json()
+			let stringifiedResults = JSON.stringify(responseData)
```
hawkrives: Why not just call .text() instead of json+stringing?

drewvolz:
Some results aren't displaying, yet they are fetching okay. I'm trying to find out why that is.

hawkrives (approved):
A nice follow-up would be to expose common routes into the UI - so I don't have to go remember the path to fetch a calendar by ID value. But nice!

It would also be nice to be able to toggle a parsed-JSON view on, which would reuse the redux debugger.

source/views/menus/dev-bonapp-picker.js
drewvolz: So... this is still being used in the menus. I'm needing to have it set here so that I can have a title when it's launched from developer settings, and I need to set the tabBarIcon here because we're now setting the title here 🤷‍♂️

source/views/settings/screens/api-test.js
drewvolz: That's an excellent suggestion!

source/views/menus/dev-bonapp-picker.js
hawkrives: Just take it out of the Menus section :p

source/views/settings/screens/api-test.js
hawkrives: Also on the previous bit - it won't always be an array. Don't cast it to Array.

source/views/menus/dev-bonapp-picker.js
drewvolz: Alrighty!

drewvolz:
> Some results aren't displaying, yet they are fetching okay. I'm trying to find out why that is.

Out of our locally-bundled `init` data, I can fetch these
`/contacts`
`/faqs`
`/food/named/menu/the-pause`
`/transit/modes`
`/webcams`

but not these:
`/spaces/hours`
`/dictionary`

and not our remote sources like `calendar` or `news` etc

Although unable to display, their data is returned and fetched appropriately. Hmm.

drewvolz:
> It would also be nice to be able to toggle a parsed-JSON view on, which would reuse the redux debugger.

@hawkrives that's an awesome idea

source/views/settings/screens/api-test.js
```diff
@@ -61,12 +64,10 @@ export class APITestView extends React.PureComponent<Props, State> {
 		try {
 			let responseData: Array<any> = await fetch(API(path), {
```
hawkrives: Remove array here

hawkrives (changes requested):

drewvolz:
Update: data not appearing due to scrollview
Update: fixed!

source/views/menus/dev-bonapp-picker.js
```diff
@@ -73,3 +71,6 @@ export class BonAppPickerView extends React.PureComponent<Props, State> {
 		)
 	}
 }
+BonAppPickerView.navigationOptions = {
```
hawkrives: This block isn't needed at all, since it's a class property as well.

source/views/settings/screens/api-test.js
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	state = {
+		results: '',
+		error: '',
+	}
+
+	fetchData = async (path: string) => {
+		try {
+			let responseData: string = await fetch(API(`/${path}`), {
```
hawkrives: Could you check and only prepend the slash if it's not given?

source/views/settings/screens/api-test.js
```diff
@@ -0,0 +1,104 @@
...
+}
+
+export class APITestView extends React.PureComponent<Props, State> {
+	static navigationOptions = {
+		title: 'API Tester',
+	}
+
+	state = {
+		results: '',
+		error: '',
```
hawkrives: Why empty strings instead of null?

hawkrives (changes requested):

source/views/menus/dev-bonapp-picker.js
```diff
@@ -30,6 +30,7 @@ type State = {
 
 export class BonAppPickerView extends React.PureComponent<Props, State> {
 	static navigationOptions = {
+		title: 'BonApp',
 		tabBarLabel: 'BonApp',
```
hawkrives: Don't need tabBarLabel anymore either

source/views/settings/screens/api-test.js
```diff
@@ -0,0 +1,159 @@
...
+
+const Segment = glamorous(SegmentedControlIOS)({})
+
+type Props = {
+	navigation: NavigationScreenProp<*>,
+}
+
+type State = {
+	results: string | null,
+	error: string | null,
```
hawkrives: ?string

source/views/settings/screens/api-test.js
```diff
@@ -0,0 +1,159 @@
...
+
+	state = {
+		results: null,
+		error: null,
+		selectedIndex: 0,
+	}
+
+	fetchData = async (path: string) => {
+		try {
+			let correctedPath = path.charAt(0) === '/' ? path : `/${path}`
```
hawkrives: .startsWith('/')

source/views/settings/screens/api-test.js
```diff
@@ -0,0 +1,159 @@
...
+	View,
+	SegmentedControlIOS,
+} from 'react-native'
+import {Toolbar} from '@frogpond/toolbar'
+import {fetch} from '@frogpond/fetch'
+import {API} from '@frogpond/api'
+import glamorous from 'glamorous-native'
+import {iOSUIKit, material} from 'react-native-typography'
+import type {NavigationScreenProp} from 'react-navigation'
+import {DebugListView} from '../../../views/settings/screens/debug'
```
hawkrives: This file should probably be moved around

hawkrives: Actually the paths just needs to be updated. That's technically correct, but it's much more work than needs to be done here.

source/views/settings/screens/debug/index.js
```diff
@@ -1,3 +1,4 @@
 // @flow
 
 export {ConnectedDebugListView as DebugView} from './list'
+export {DebugListView} from './list'
```
hawkrives: Merge this into the previous line

hawkrives (changes requested):

drewvolz:
@hawkrives I will note that the list-view json explorer isn't working quite correctly at the moment. I don't think my drop-in replacement of data is working as intended, and I think I've mucked up navigation state because the navbar glitches out because... I'm unsure.

I'll also note that I think I've done everything to get android's TextInput to behave the same as iOS' in that it should be scrollable and start the loading position at the top, but even with all of the appropriate props applied from the docs, it fails to scroll or start at the top. I'm tempted to rip out the TextView for android. 

~(this is also not on RN 0.59, so maybe it'd help to merge that in)~ (doesn't actually fix it)

source/views/settings/screens/debug/list.js
```diff
@@ -28,9 +29,13 @@ export class DebugListView extends React.PureComponent<Props> {
 	}
 
 	onPressRow = (row: any) => {
+		let testing = this.props.testing
 		let keyPath = this.props.navigation.getParam('keyPath', [])
 		keyPath = [...keyPath, row.key]
-		this.props.navigation.push('DebugView', {keyPath: keyPath})
+		this.props.navigation.push('DebugView', {
```
hawkrives:
```suggestion
		this.props.navigation.push('DebugView', {keyPath, testing})
```

source/views/settings/screens/debug/list.js
```diff
@@ -44,14 +49,23 @@ export class DebugListView extends React.PureComponent<Props> {
...
+
+		let keyed = toPairs(parsedData).map(([key, value]) => {
 			return {key, value}
 		})
 
 		return (
 			<FlatList
 				ItemSeparatorComponent={ListSeparator}
-				ListEmptyComponent={<NoticeView text="Nothing found in redux." />}
+				ListEmptyComponent={<NoticeView text={nothingFoundMessage} />}
```
hawkrives:
```suggestion
				ListEmptyComponent={<NoticeView text="Nothing found." />}
```

maybe? (I don't think we need a distinction between the two modes here.)

source/views/settings/screens/api-test.js
```diff
@@ -0,0 +1,159 @@
...
+
+				{Platform.OS === 'ios' ? (
+					<Segment
+						onChange={event => {
+							this.setState({
+								selectedIndex: event.nativeEvent.selectedSegmentIndex,
+							})
+						}}
+						selectedIndex={selectedIndex}
+						values={['JSON', 'List']}
```
hawkrives:
```suggestion
						values={['Text', 'Parsed']}
```

I think?

source/views/settings/screens/api-test.js
```diff
@@ -0,0 +1,159 @@
...
+				multiline={true}
+				scrollEnabled={true}
+				style={styles.data}
+				// use multiline with textAlignVertical="top" for the same behavior in both platforms
+				textAlignVertical="top"
+				value={results}
+			/>
+		) : null
+
+		let showIndex = () => {
```
hawkrives: This needn't be a function.

source/views/settings/screens/api-test.js
```diff
@@ -0,0 +1,159 @@
...
+				multiline={true}
+				scrollEnabled={true}
+				style={styles.error}
+				// use multiline with textAlignVertical="top" for the same behavior in both platforms
+				textAlignVertical="top"
+				value={error}
+			/>
+		) : null
+
+		let jsonData = results ? (
```
hawkrives: Inline these into the fragment in `showIndex`?

source/views/settings/screens/api-test.js
```diff
@@ -0,0 +1,159 @@
...
+				value={results}
+			/>
+		) : null
+
+		let showIndex = () => {
+			return selectedIndex === 0 ? (
+				<>
+					{jsonData}
+					{jsonError}
+				</>
```
hawkrives:
```suggestion
			error ? (
				<Output
					editable={false}
					// this aligns the text to the top on iOS, and centers it on Android
					multiline={true}
					scrollEnabled={true}
					style={styles.error}
					// use multiline with textAlignVertical="top" for the same behavior in both platforms
					textAlignVertical="top"
					value={error}
				/>
			) : selectedIndex === 0 ? (
				<Output
					editable={false}
					// this aligns the text to the top on iOS, and centers it on Android
					multiline={true}
					scrollEnabled={true}
					style={styles.data}
					// use multiline with textAlignVertical="top" for the same behavior in both platforms
					textAlignVertical="top"
					value={results}
				/>
			) : (
				<DebugListView
					navigation={this.props.navigation}
					state={results}
					testing={true}
				/>
			)
```

instead of the inside of this function? Prevents errors from hiding when you're looking at a Debug view, and reduces nesting? And just remove the two `json*` variables above.

source/views/settings/screens/api-test.js
hawkrives: (and it needs a better name)

source/views/settings/screens/debug/list.js
```diff
@@ -13,6 +13,7 @@ import type {NavigationState} from 'react-navigation'
 import type {TopLevelViewPropsType} from '../../../types'
 
 type Props = TopLevelViewPropsType & {
+	testing?: boolean,
```
hawkrives: I'm going to have to ask for a better name for this prop than `testing`…

hawkrives (dismissed):
still enjoying this idea!

source/views/settings/screens/api-test.js
```diff
@@ -0,0 +1,149 @@
...
+						placeholder="path/to/resource"
+						returnKeyType="done"
+						style={styles.default}
+					/>
+				</Toolbar>
+
+				{Platform.OS === 'ios' ? (
+					<Segment
+						onChange={event => {
+							this.setState({
```
hawkrives: Use the functional setState form please

hawkrives (changes requested):
I think I'm good after this change

hawkrives (approved):

drewvolz:
I still need to fixup Android’s non-scrolling and non-positioning textinput

source/views/settings/screens/api-test/api-test-android.js
```diff
@@ -0,0 +1,96 @@
...
+
+		let APIResponse = error ? (
+			<Output>
+				<Text>{error}</Text>
+			</Output>
+		) : (
+			<Output>
+				<Text>{results}</Text>
+			</Output>
+		)
```
hawkrives:
```js
			let APIResponse = (
				<Output>
					<Text>{error ? error : results}</Text>
				</Output>
			)
```

hawkrives (dismissed):
plz

looks good otherwise

source/views/settings/screens/api-test/api-test-android.js
drewvolz: Oh, this is a decent suggestion!

source/views/settings/screens/api-test/api-test-android.js
```diff
@@ -65,16 +65,6 @@ export class AndroidAPITestView extends React.PureComponent<Props, State> {
 	render() {
 		let {error, results} = this.state
 
-		let APIResponse = error ? (
```
hawkrives: as another note… capitalized variables should be reserved for … er, component creation functions? Constructors and etc. 

APIResponse, I would expect to need `<APIResponse/>`, whereas this would be `{APIResponse}`.

I prefer and try to keep the code to `{apiResponse}` for this sort of "data" variable.

Just a side note.

hawkrives (approved):
let's merge it

source/views/settings/screens/api-test/api-test-android.js
drewvolz: noted, thanks!

PR #3564, opened 2019-03-17
Fix course search

drewvolz:
Return the right item in courses sorting/grouping.


hawkrives (approved):
the downside of avoiding variable shadowing strikes again

PR #3569, opened 2019-03-17
Fix signed Gradle builds

hawkrives:
Somehow, configuring Gradle's signing options separately from the main `android{}` block results in an error about "can not find .hasProperty on project"

and like

wtf gradle

anyway

this avoids that by doing the configuration inside the `android{}` block

---

I've tested this locally; it works for both debug and release builds, and on both signed and unsigned builds.


hawkrives:
actually this is broken. please hold.

hawkrives:
now it should be fixed.

drewvolz (approved):
:shipit: it's green!

PR #3573, opened 2019-03-17
Enable inline requires in metro bundler

drewvolz:
As per the 0.59.0 [blogpost](http://facebook.github.io/react-native/blog/2019/03/12/releasing-react-native-059#faster-app-launches-with-inline-requires), we're going to try enabling inline requires for performance enhancements.

> **Faster app launches with inline requires:** 
> 
> We want to help people have performant React Native apps by default and are working to bring Facebook's optimizations to the community. Applications load resources as needed rather than slowing down launch. This feature is called "inline requires", as it lets Metro identify components to be lazy loaded. Apps with a deep and varied component architecture will see the most improvement.
>
> We need the community to let us know how it works before we turn it on by default. When you upgrade to 0.59, there will be a new `metro.config.js` file; flip the options to true and give us your feedback! Read more about inline requires in the performance docs to benchmark your app.


hawkrives:
Can you run the `yarn *-bundle` commands both before and after this commit, and let us know the results? There should be a file-size difference, if nothing else.

If there are no changes, we'll need to try and follow the longer instructions.

drewvolz:
Platform | Before `62183cf` | After `62183cf`
--|--|--
Android | 8.0M (bundle) and 14M (map)  | 8.2M (bundle) and 13M (map)
iOS | 3.2M (bundle) and  8.7M (map) | 3.2M (bundle) and 8.6M (map)

hawkrives:
I vote we merge this tomorrow morning; I'd like a nightly with the OleCard fixes :-)

rye (approved):
Now that's a diff I can review!

RSLGTM.

rye (approved):
Approving once more. @hawkrives' last comment indicates a second approval, so once green I will merge.

hawkrives (approved):

PR #3574, opened 2019-03-18
BugSnag ↦ Sentry 

rye:
This PR deletes the integrations with BugSnag and adds the barebones integrations with Sentry. We now no longer have a `notify` function, as I anticipated using Sentry's provided methods over that.  We can always write some helpers if we want.

Still yet to figure out:
- [x] Source map + dSYM uploading (from CI)
  - [x] Need to include `auth-token` in `sentry.properties`
- [ ] Git integration


android/sentry.properties
```diff
@@ -0,0 +1,3 @@
+defaults.url=https://sentry.frogpond.tech/
+defaults.org=frog-pond-labs
+defaults.project=all-about-olaf
```
hawkrives, line 3: One thought - could we include a commented out version of the API key line in both of these files?

ios/AllAboutOlaf.xcodeproj/project.pbxproj
```diff
@@ -5,7 +5,6 @@
 	};
 	objectVersion = 46;
 	objects = {
-
```
hawkrives: Keep this line [xcode]

ios/AllAboutOlaf.xcodeproj/project.pbxproj
```diff
@@ -1100,7 +1094,7 @@
 		83CBB9F71A601CBA00E9B192 /* Project object */ = {
 			isa = PBXProject;
 			attributes = {
-				LastSwiftUpdateCheck = 0810;
+				LastSwiftUpdateCheck = 810;
```
hawkrives: Keep the 0 [xcode]

ios/AllAboutOlaf.xcodeproj/project.pbxproj
```diff
@@ -1111,7 +1105,7 @@
 					};
 					13B07F861A680F5B00A75B9A = {
 						DevelopmentTeam = TMK6S7TPX2;
-						LastSwiftMigration = 0810;
+						LastSwiftMigration = 810;
```
hawkrives: Keep the 0 [xcode]

ios/AllAboutOlaf/AppDelegate.m
```diff
@@ -14,6 +14,11 @@
 #import <React/RCTBridge.h>
 #import <React/RCTBundleURLProvider.h>
 #import <React/RCTRootView.h>
+#if __has_include(<React/RNSentry.h>)
+#import <React/RNSentry.h> // This is used for versions of react >= 0.40
+#else
+#import "RNSentry.h" // This is used for versions of react < 0.40
+#endif
```
hawkrives: Remove the else branch

ios/AllAboutOlaf.xcodeproj/project.pbxproj
```diff
@@ -1730,7 +1713,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "export NODE_BINARY=node\n../node_modules/react-native/scripts/react-native-xcode.sh";
+			shellScript = "export SENTRY_PROPERTIES=sentry.properties\nexport NODE_BINARY=node\n../node_modules/@sentry/cli/bin/sentry-cli react-native xcode ../node_modules/react-native/scripts/react-native-xcode.sh";
```
hawkrives: Something tells me that we need another \n in here, or we'll wind up running "sentry-cli React-native" which I don't think is a command

source/init/sentry.js
```diff
@@ -0,0 +1,14 @@
...
+import {Sentry, SentryLog} from 'react-native-sentry'
+import {IS_PRODUCTION} from '@frogpond/constants'
+
+export function bootSentry() {
+	let config = Sentry.config(SENTRY_DSN, {
+		logLevel: SentryLog.Debug,
+		handlePromiseRejection: true,
+	})
+
+	if (!IS_PRODUCTION) config.install()
```
hawkrives: Make this if multi-line please

modules/add-to-device-calendar/lib.js
```diff
@@ -1,7 +1,6 @@
 // @flow
 
 import type {EventType} from '@frogpond/event-type'
-import {notify} from '@frogpond/analytics'
```
hawkrives, line 4: Thinking into the app logic… I see a lot of notify/breadcrumb removals. Do we want to keep those? Does sentry automatically do them?

hawkrives (commented):
Some thoughts

source/init/sentry.js
rye: Prettier did this.

ios/AllAboutOlaf.xcodeproj/project.pbxproj
rye: This was autogenerated, and I don't think it should be a problem.

source/init/sentry.js
hawkrives: Put braces around it and prettier will leave it multi-line

android/sentry.properties
rye: Yeah, we could.

ios/AllAboutOlaf.xcodeproj/project.pbxproj
hawkrives: I'm game - if CI builds, it's not a problem here

ios/AllAboutOlaf/AppDelegate.m
hawkrives: And by proxy, leave only the content of the if - there's no need for the conditional here at all, since we're running 0.59

modules/add-to-device-calendar/lib.js
rye: Sentry "should" handle crashes and the like, I believe. For sure, it includes the necessary beef to handle unhandled promise rejections, but beyond that I am not sure if it will e.g. catch uncaught errors.

rye: Addendum: `notify` was mostly used to explicitly send out an event or error message, and we will need to manually use the Sentry equivalent of that.

ios/AllAboutOlaf/AppDelegate.m
rye: Okay, yeah.

rye:
Some notes for not-tired Kris:

- Source maps can be uploaded via `npx @sentry/cli releases files <VERSION> [...]`… `<VERSION>` may or may not be `tech.frogpond.all-about-olaf-<version>` or just what we have coming from `current_bundle_version`, I think.
- Can use `SENTRY_PROPERTIES` and `SENTRY_AUTH_TOKEN` environment variables to get around dumping `auth.token` into a file which is bad security.
- `current_bundle_version` would probably be the right thing to use, I think, once `propagate` has been called and stuff

rye:
@hawkrives and @drewvolz, what are your thoughts on version numbering?

I'm considering just piping the output of `git describe --tags HEAD`, which looks like `v2.7.0-beta.2-669-gbbe0e05ce`.  On tags, it's just the tag value, (e.g. `v2.7.0-beta.2`) and otherwise the format also includes the number of commits ahead of the tag (e.g. `669`) and the SHA sum, with `g` as a prefix. I like this format because it's concise and standardized.

rye:
I have skipped dSYM uploading for now because it was disabled on iOS already.

rye:
And I think the Git integration is "out of scope" despite being a feature we will likely use. As such, I would like to get through this so we can test it out on the nightly. (Sourcemap uploading only occurs on auto_beta builds)

hawkrives:
Sure. We can certainly keep turning on integrations after we do the switch.

fastlane/lib/sourcemaps.rb
```diff
@@ -44,18 +44,21 @@ def generate_sourcemap
 	                                      print_command: true)
 end
 
-# Upload source map to Bugsnag
+# Upload sourcemap to bugsnag
 def upload_sourcemap_to_bugsnag
```
hawkrives: Rename this method?

source/views/help/wifi.js
```diff
@@ -50,8 +50,7 @@ export class ToolView extends React.Component<Props, State> {
 
 		this.setState(() => ({status: 'collecting', error: ''}))
 		const [position, device] = await Promise.all([
-			getPosition().catch(error => {
-				notify(error)
+			getPosition().catch(() => {
```
hawkrives: Put the notify call back in?

hawkrives (dismissed):
A few tweaks have been requested by me

hawkrives (approved):

drewvolz (approved):
👏 good job, @rye! :shipit:

rye:
Thanks y'all! I'm gonna merge this.

PR #3595, opened 2019-03-23
👨‍🔬️ Remote js-bundle downloading and refreshing

### LINKED ISSUE #3593: Internal alternative codepush
Internal branch build testing: What this would look like is a listview on each platform with rows for branches that were built and have active jsbundles that we can download and launch on the device. These would get uploaded from CI/CD as artifacts on a per-branch basis signified by a pr label.

- Fetch list of branches and their jsbundles
- Display list in developer/debug listview
- Select a row to download the bundle to the app’s fs
- Point the native code to the new “active” bundle location
- Trigger a restart of the app

Could possibly use [react-native-dynamic-bundle](https://github.com/mauritsd/react-native-dynamic-bundle) 

...and download the file with `react-native-fs`

<hr>

Current progress is at [jsbundle-updater](https://github.com/StoDevX/AAO-React-Native/tree/jsbundle-updater)

drewvolz:
This PR aims to grant us the ability to seamlessly switch between remote JS bundles built and hosted on CircleCI as artifacts.

The flow:
1. Requests the open GitHub pull requests
2. Tap on a pull request row
3. A network request is dispatched to request the _latest_ artifacts for a branch on CircleCI
4. If the data returned looks okay, we'll go ahead and make a second request to download the bundle with `react-native-fs` and save it to disk
    * There is some error checking going on! If we get a response (errors are empty arrays here), technically you're allowed to download it right now... which isn't great if it contains native code. @hawkrives suggested keying off of GitHub labels to prevent this? We should discuss how we want to handle this.
5. Supposedly at this point, we can kill the app (crash? refresh?) and now restart into the newly downloaded bundle using `react-native-dynamic-bundle`

Resolves #3593 

<details>

~ | ~
--|--
image | image
image | image
image | 

</details>

<hr>

Notes:
- [x] Many `any` types until I work up the courage to type out GitHub and CircleCI types
- [ ] Don't know if the `restart()` call is necessary
- [ ] Don't know what happens when you switch _back_ (no other PR/branch has this)
- [ ] Android isn't working correctly yet (issue with `onReactContextInitialized`)
- [ ] Bundled assets/images disappear. Do we mind?

I really have only seen it work so far after switching the physical `AppDelegate.m` and `AppDelegate.h` code back to loading the other bridge... so this could be not working from what I can tell 🤷‍♀️


source/views/settings/screens/bundle-picker/downloader.js
```diff
@@ -0,0 +1,139 @@
...
+		const {circleResponse, loading, error} = this.state
+
+		return (
+			<ScrollView style={styles.container}>
+				<Text selectable={true} style={styles.title}>
+					{item.title}
+				</Text>
+
+				{loading ? (
+					<Button disabled={true} title="Checking for artifacts…" />
```
drewvolz, line 118: I used too many `Button` elements but I think we can simplify this and come up with another way to say what `disabled`, `title`, and `onPress` should be.

source/views/settings/screens/bundle-picker/list.js
```diff
@@ -0,0 +1,105 @@
...
+
+type Props = TopLevelViewPropsType & {
+	onRefresh: () => any,
+	refreshing: boolean,
+}
+
+type State = {
+	loading: boolean,
+	refreshing: boolean,
+	results: any,
```
drewvolz: I hope to clean up all of the `any` types once I write out all of the types for Circle and GitHub

source/views/settings/screens/overview/odds-and-ends.js
```diff
@@ -32,7 +36,13 @@ export class OddsAndEndsSection extends React.Component<Props> {
 				<ConnectedNotificationsCell onPress={this.onNotificationsButton} />
 
 				{isDevMode() && (
-					<PushButtonCell onPress={this.onDebugButton} title="Debug" />
+					<>
+						<PushButtonCell onPress={this.onDebugButton} title="Debug" />
+						<PushButtonCell
```
drewvolz, line 41: Once the other "Debug Section" for settings is merged in, this will move into that file.

drewvolz:
TL;DR: This needs TLC and help.

The core logic centers around fetching open GitHub pull requests and CircleCI artifacts. If we see an open GitHub PR, we look to fetch the output of those builds from CircleCI so that we may run them locally on-device.

First, ideally, the heavy lifting of this core logic (checking and parsing of the APIs) would happen via ccc-server. As of now, this logic happens on the client app so that needs to be changed.

Next, I'm unsure if the current logic is functioning properly. I am pretty sure it is broken on Android (it errors out), and is most likely broken on iOS.

This will most likely sit in limbo for the time being.

source/views/settings/screens/bundle-picker/downloader.js
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	state = {
+		circleResponse: null,
+		loading: true,
+		error: false,
+	}
+
+	componentDidMount() {
+		this.fetchCirlceData()
```
rye: Typo,
```suggestion
		this.fetchCircleData()
```

source/views/settings/screens/bundle-picker/downloader.js
```diff
@@ -0,0 +1,140 @@
...
+				this.setState(() => ({loading: false, error: true}))
+			})
+	}
+
+	onDownloadPushed = () => {
+		this.downloadBundle()
+		this.replaceBundle()
+	}
+
+	fetchCirlceData = async () => {
```
rye:
```suggestion
	fetchCircleData = async () => {
```

rye (commented):
@drewvolz, I think this PR will require a bit more backend work before we can get this ready for acceptance.

Do I have it right that this PR works for fetching, downloading, and inserting bundles? If not—can you just get it to that point? I want to be able to give it a URL (with the assumption that it is a valid bundle) and have it load the bundle.

The goal with CCC is to add a painless abstraction such that AAO and CARLS can all talk to the same endpoint, instead of making their requests directly to the various destinations, which shrinks the bundle down quite significantly and makes updating that logic less of a nightmare. CCC can thus be the source of bundle download lookups and such; we could add an endpoint to serve up a list of bundle URLs, or we could proxy the bundle downloads themselves.

GitHub Actions has an "artifacts" thing somewhat similar to Circle's, which we might be able to eventually use for our storage.

drewvolz:
@rye I agree with you. This will require quite a bit more work. It’s not in a good state and needs a refactoring.

PR #3598, opened 2019-03-25
Update rn-vector-icons and associated changes

hawkrives:
Closes #3539

- rn-vector-icons is now flow-typed, and partially prevents us from using an invalid icon name
- I say "partially" because I can't figure out how to type the optional `md-`/`ios-` prefixes… so anything that goes through that helper becomes `any`

also I gave up and forced the solid "share" icon on iOS for now, so it's not broken anymore, but we still need to figure out something for that before 2.8


drewvolz (approved):
Thanks @hawkrives!

PR #3599, opened 2019-03-25
Remove *:release package.json scripts

### LINKED ISSUE #3072: Build script for android:release is broken
`yarn run android:release` runs the `react-native` equivalent for building a release android app:

https://github.com/StoDevX/AAO-React-Native/blob/479add9f1ab9cd1caf79a848afaf9d6ae09c59c3/package.json#L7

but the console returns:

```console
FAILURE: Build failed with an exception.

* What went wrong:
Task 'installRelease' not found in root project 'AllAboutOlaf'. Some candidates are: 'uninstallRelease'.
```

hawkrives:
We don't use them, and the Android one in particular has been broken for a while.

Closes #3072

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,6 @@
   "private": true,
   "scripts": {
     "android": "react-native run-android",
-    "android:release": "react-native run-android --variant=release",
     "bundle-data": "node scripts/bundle-data.js data/ docs/",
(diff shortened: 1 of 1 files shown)


drewvolz (changes requested):
I’d prefer to keep the iOS one, sorry. The issue was about the android one and the iOS one works. I use it to verify production vs dev things.

hawkrives:
How do you use it?

drewvolz:
I've used it to verify production build sizes, version numbers, push notification changes per environment, etc. It's ultimately a good shortcut in my book. If you feel strongly about nixing it, that's okay. I suppose I can try to remember it or add it to my aliases.

drewvolz (approved):

PR #3609, opened 2019-03-28
(Maybe) fix sentry sourcemaps

rye:
There's a comment in the fastlane configs about the commands being run
from within the fastlane dir. (as the cwd)

This commit prefixes the `SENTRY_PROPERTIES` environment variable with
`..` so that it... points to the right thing?

Just guessing.


hawkrives (approved):
Let's try it!

drewvolz (approved):

PR #3658, opened 2019-04-13
circle: Use Ruby 2.5.5 everywhere

rye:
This might actually work both from a caching standpoint and also allow macOS to function properly?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ executors:
     docker: [{image: 'circleci/node:10'}]
 
   ruby:
-    docker: [{image: 'circleci/ruby:2.6.1'}]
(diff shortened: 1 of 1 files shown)


drewvolz (approved):
🤞

rye:
Just adding to the background: new minor versions of Ruby use a different gem directory; we were referencing incorrect Ruby versions, so `chruby` was failing and falling back to macOS' stupidly out-of-date default.

I changed the Ruby Docker image version to make sure that Android and iOS use the same version. Note well that the cache cannot be shared between the two, because of the Linux/Darwin disparity.

Awaiting second approval, this PR should be good to go.

hawkrives (approved):

PR #3663, opened 2019-04-15
Fastlane: Use Netrc gem

### LINKED ISSUE #3662: authorize_ci_for_keys inappropriately writes to ~/.netrc
Troubleshot not being able to log into GitHub only to find that… in the default mode of operation, it just appends to `~/.netrc` with very little regard.

Consider using the `netrc` gem, i.e.:

```ruby
require 'netrc'

n = Netrc.read
unless n["github.com"]
  # fill it in
end
n.save
```

rye:
Rather than (naïvely and indiscriminately appending to `~/.netrc`, this commit introduces the `netrc` gem, which we can use for reading from and writing to `netrc` files just the same.

Resolves #3662.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,7 @@
 source 'https://rubygems.org'
 
 gem "fastlane"
+gem "netrc"
 gem 'json'
 gem 'xcodeproj'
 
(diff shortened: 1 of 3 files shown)


hawkrives (commented):
Would you mind leaving the comment?

hawkrives (approved):

PR #3665, opened 2019-04-15
fastlane/README: Remove dangling ref to ios:refresh_dsyms

rye:
We might bring it back under a different or similar name, but it's now out of date. (Subsequent fastlane invocations will keep removing these lines.


drewvolz:
Didn’t @hawkrives say this file is generated?

rye:
It is. Nobody has run `fastlane` locally, so this change has not been observed by us.

hawkrives:
Yes, the readme there is generated, but we have to run it locally in order to update it and commit the changes.

hawkrives (approved):

PR #3730, opened 2019-05-13
Fix stoprint release

### LINKED ISSUE #3485: stoPrint Jobs are not releasing
I noticed this today when I was trying to use the stoPrint feature. Jobs are not releasing properly, and trigger the following error:

hannesmcman:
Resolves #3485.

Turns out that we were trying to extract the username from the navigation params, rather than loading them using `loadLoginCredentials`. I guess this never got updated when we moved away from storing login credentials in redux itself. I tested this though and it resolves the issue!


rye (approved):

drewvolz (approved):

hannesmcman:
Oops, forgot to update the changelog!

PR #3732, opened 2019-05-13
Add loginViaCredentials to Stoprint view

### LINKED ISSUE #3731: StoPrint view shows "Not Logged In" message if opened before settings view
The StoPrint view never calls `performLogin`, and so the credentials loaded from storage are never verified against the Olecard server, which they probably should be. Thus, if you open the stoprint view before accessing Settings or Balances (the other views that trigger a login), the "Not Logged In" message appears.

hannesmcman:
Resolves #3731.

This way, the credentials will be validated by the Olecard API before loading the Stoprint jobs, which is a good thing. Furthermore, the "Not Logged In" message will no longer appear, because the credentials will be validated and the loginState in Redux will be updated!

### DIFF
--- a/source/views/stoprint/print-jobs.js
+++ b/source/views/stoprint/print-jobs.js
@@ -96,7 +96,7 @@ class PrintJobsView extends React.PureComponent<Props, State> {
 
 	fetchData = async () => {
 		await this.logIn()
-		this.props.updatePrintJobs()
+		await this.props.updatePrintJobs()
 	}
 
 	keyExtractor = (item: PrintJob) => item.id.toString()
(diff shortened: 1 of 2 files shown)


source/views/stoprint/print-jobs.js
```diff
@@ -80,7 +82,22 @@ class PrintJobsView extends React.PureComponent<Props, State> {
...
+
+		let {username = '', password = ''} = await loadLoginCredentials()
+		if (username && password) {
+			await this.props.logInViaCredentials(username, password)
+		}
+	}
+
+	fetchData = async () => {
+		await this.logIn()
+		this.props.updatePrintJobs()
```
hawkrives: You should `await` this line as well

hawkrives (dismissed):

rye (approved):

hawkrives (approved):

drewvolz (approved):

PR #3741, opened 2019-05-17
Move the Bookstore into the Post Office section

hawkrives:
I don't think the "convenience store" really exists as a distinct entity any longer.


rye (changes requested):
I'm fine with merging the Convenience Store and the Bookstore together, but I think they are independent of the Post Office + Print Center.

hawkrives:
@rye I don't like one-item sections, which is why I merged them.

I would appreciate suggestions towards how best to combine or reorganize them.

drewvolz:
I think the suggestion is how not to do something; I think we would all prefer to not combine the Post Office and Bookstore, as per the out-of-band discussion.

hawkrives:
> I think we would all prefer to not combine the Post Office and Bookstore, as per the out-of-band discussion.

No, that is incorrect - I wish to combine the two sections, so as to avoid a 1-item section.

rye:
N.B.: We have a `Museum` section with one item in it already.

Is it possible to have a title-less section?

I would be fine with merging the now-three places under `Mail, Packages, Printing, and Books` or something like that.

PR #3746, opened 2019-05-18
🆗 Show alert after dev sentry button is pressed

drewvolz:
A confirmation because feedback is nice.

### DIFF
--- a/source/views/settings/screens/overview/developer.js
+++ b/source/views/settings/screens/overview/developer.js
@@ -1,6 +1,7 @@
 // @flow
 import {Sentry, SentrySeverity} from 'react-native-sentry'
 import * as React from 'react'
+import {Alert} from 'react-native'
 import {Section, PushButtonCell} from '@frogpond/tableview'
 import {type NavigationScreenProp} from 'react-navigation'
 
@@ -14,9 +15,11 @@ export class DeveloperSection extends React.Component<Props> {
 		Sentry.captureMessage('A Sentry Message', {
 			level: SentrySeverity.Info,
 		})
+		Alert.alert('Sentry message sent.')
 	}
 	sendSentryException = () => {
 		Sentry.captureException(new Error('Debug Exception'))
+		Alert.alert('Sentry exception sent.')
 	}
 
 	render() {
--- a/source/views/settings/screens/overview/developer.js
+++ b/source/views/settings/screens/overview/developer.js
@@ -19,7 +19,7 @@ export class DeveloperSection extends React.Component<Props> {
 	}
 	sendSentryException = () => {
 		Sentry.captureException(new Error('Debug Exception'))
-		Alert.alert('Sentry exception sent.')
+		Alert.alert('Sentry exception sent')
 	}
 
 	render() {
--- a/source/views/settings/screens/overview/developer.js
+++ b/source/views/settings/screens/overview/developer.js
@@ -4,6 +4,7 @@ import * as React from 'react'
 import {Alert} from 'react-native'
 import {Section, PushButtonCell} from '@frogpond/tableview'
 import {type NavigationScreenProp} from 'react-navigation'
+import {isDevMode} from '@frogpond/constants'
 
 type Props = {navigation: NavigationScreenProp<*>}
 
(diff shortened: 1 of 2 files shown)


source/views/settings/screens/overview/developer.js
```diff
@@ -14,9 +15,11 @@ export class DeveloperSection extends React.Component<Props> {
 		Sentry.captureMessage('A Sentry Message', {
 			level: SentrySeverity.Info,
 		})
+		Alert.alert('Sentry message sent.')
```
hawkrives:
```suggestion
		Alert.alert('Sentry message sent')
```

source/views/settings/screens/overview/developer.js
```diff
@@ -14,9 +15,11 @@ export class DeveloperSection extends React.Component<Props> {
...
 	}
 	sendSentryException = () => {
 		Sentry.captureException(new Error('Debug Exception'))
+		Alert.alert('Sentry exception sent.')
```
hawkrives:
```suggestion
		Alert.alert('Sentry exception sent')
```

hawkrives (dismissed):
iOS, at least, doesn't do full sentences in alert titles

rye:
I like this idea.

Another idea: perhaps also add a hook for `isDevMode` only:

```js
// set a callback after an event was successfully sent
// its only guaranteed that this event contains `event_id` & `level`
Sentry.setEventSentSuccessfully((event) => {
  Alert.alert(`Sent an event to Sentry: ${event.event_id}`)
});
```

source/views/settings/screens/overview/developer.js
drewvolz: > HIG: If the title is a complete sentence, use sentence-style capitalization and appropriate ending punctuation. If the title is a sentence fragment, use title-style capitalization and don’t add ending punctuation.

...alright. I'll accept it.

source/views/settings/screens/overview/developer.js
drewvolz: Same as message above.

drewvolz:
@rye I actually looked for that bit of docs -- thank you for finding it!

rye (approved):
Looks alright.

drewvolz:
I am unable to get this hook to work. If anyone can advise, I would love to add it.

rye:
What is "this hook"? Do you mean `Sentry.setEventSentSuccessfully`?

drewvolz:
Correct, for `Sentry.setEventSentSuccessfully` and maybe `Sentry.setShouldSendCallback`

rye:
Bear in mind that `Sentry` will not be installed (read: activated) in local development environments unless you remove the `if (IS_PRODUCTION)`

rye:
Actually, my suggestion is illogical.

Sentry only boots when `IS_PRODUCTION`, but we only want an alert for local testing, I think, which is `!IS_PRODUCTION`… and those have no overlap.

rye (approved):
Green light from me.

hawkrives:
But if Sentry is not active, those buttons won't do anything at all, will they? They won't actually send the event?

rye:
@hawkrives correct. Pressing the button will fire the callback, but nothing will happen internally.

hawkrives:
I would argue for using the Sentry callbacks here, so that we know when the event has actually been sent. 

I'm fine with using an Alert if we're !IS_PRODUCTION, to announce that no alert will be sent, but I'd like the success notification in betas and nightlies.

rye:
@hawkrives, if `!IS_PRODUCTION`, no `Sentry` callback will be set either, though, because `bootSentry()` will not install the `config`.

hawkrives (dismissed):
Callback in production, but alert that no event was sent in development please.

drewvolz:
@hawkrives that doesn't make sense to me.

hawkrives:
Two cases:

1. **In development**: No data will ever be sent, so we should tell the user that nothing will appear in Sentry's dashboard.

2. **In production**: Data will be sent, so it makes sense to alert the user _after_ the data has been sent, instead of onPress, so that they know when they can expect to see it.

drewvolz:
> so it makes sense to alert the user _after_ the data has been sent, instead of onPress, so that they know when they can expect to see it.

but we don't show these features outside of development?

hawkrives:
We show them on betas and nightlies

drewvolz:
I think this last commit covers that

hawkrives:
Yes, it should; GitHub did not show me that commit when I was responding to your previous comment.

hawkrives (approved):

PR #3789, opened 2019-06-12
(Try to) Fix Sentry sourcemap association

rye:
First commit:

>Previously we were just calling `current_build_number` which immediately passed with the results from `build_number`.  Instead, just call `build_number`, which doesn't need `args[:track]`.

Second commit:

>We were previously just sending the value of current_bundle_version as both the `RELEASE_NAME` and `DISTRIBUTION_NAME` parameters for the Sentry CLI, however, this was incorrect.  It seems that Sentry instead associates with the provided `DISTRIBUTION_NAME`, (i.e. iOS'
`CFBundleVersion` or Android's `versionCode`)
>
>So we were uploading with `2.8.0-pre` in both `RELEASE_NAME` and
`DISTRIBUTION_NAME` instead of `2.8.0-pre` for `RELEASE_NAME` and the
distribution version code for `DISTRIBUTION_NAME`
>
>Now, we have a helper that can query against the gradle stuff and get the appropriate values out, and we use that in the sourcemap upload step.

(The commit messages summarize the PR but are included here for reference.)

### DIFF
--- a/fastlane/lib/versions.rb
+++ b/fastlane/lib/versions.rb
@@ -1,19 +1,3 @@
-# Gets the version, be it from Travis, Testflight, or Google Play
-def current_build_number(**args)
-	return build_number if build_number
-
-	begin
-		case lane_context[:PLATFORM_NAME]
-		when :android
-			(google_play_track_version_codes(track: args[:track]) + 1).to_s
-		when :ios
-			(latest_testflight_build_number + 1).to_s
-		end
-	rescue StandardError
-		'1'
-	end
-end
-
 # get the current build number from the environment
 def build_number
 	travis = ENV['TRAVIS_BUILD_NUMBER']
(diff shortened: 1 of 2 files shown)


fastlane/lib/versions.rb
```diff
@@ -1,27 +1,3 @@
-# Gets the version, be it from Travis, Testflight, or Google Play
-def current_build_number(**args)
-	return build_number if build_number
-
-	begin
```
hawkrives, line 5: For posterity's sake – we haven't used this chunk of code (the `begin`/`end` block) since … oh gosh, at least a while _before_ we left Travis.

fastlane/lib/versions.rb
```diff
@@ -1,27 +1,3 @@
...
-			(latest_testflight_build_number + 1).to_s
-		end
-	rescue StandardError
-		'1'
-	end
-end
-
-# get the current build number from the environment
-def build_number
-	travis = ENV['TRAVIS_BUILD_NUMBER']
```
hawkrives, line 19: I OK'd removal of the Travis check here, since we're firmly on Circle atm.

hawkrives (approved):

PR #3875, opened 2019-07-20
Fix android tabbar crashing

drewvolz:
Fixes a crash on all Android tabbed views. They were expecting to receive a hex color string. This uses first readable color (string) instead of that returned instance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Building signed Android APKs was broken after RN 0.59; now it is fixed (#3569)
 - Fixed an issue where StoPrint jobs failed to release properly (#3730)
 - Fixed StoPrint login issue (#3732)
+- Fixed a crash that prevented all Android tab views from loading
 
 ### Removed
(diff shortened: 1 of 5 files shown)


drewvolz (commented):
⚠️ This isn't correct yet, working on it...

drewvolz:
✅ This is correct!

rye (approved):
Sounds good to me!

hawkrives (changes requested):
`toString()` is documented to return differently formatted values depending on the input – can we switch to `toHexString()` (or another format) so we can be sure what we're getting out?

I think RN can handle any of the string formats, but I'd like consistency :-)

drewvolz:
Okay. I opted to be consistent with the existing codebase which uses `toString()` for the tinycolor calls, but I'm happy to change it.

hawkrives:
After talking with Drew, who's testing this PR, we've opted to switch to `toRgbString()` everywhere, as we have some semi-transparent colors which can be handled by toRgbString but not toHexString.

drewvolz:
So yep, we're changing to that for everything.

After looking closer, my message about being consistent doesn't make sense. There are existing parts of the codebase that do use `toRgbString()` which is what we want!

hawkrives (approved):

Issue #3646, opened 2019-04-10
App crashes on boot (2019-04-10 nightly)


rye:
Furthermore, it does not send a report to Sentry.

rye:
Attempted to reproduce on a Pixel 2 API 28 build, didn't get anywhere.

rye:
This has been confirmed on both platforms, and occurs when running in Release mode only. Kris suspects something changed that we ignored. Bisect in progress!

rye:
Bisect logs:

```
git bisect start
# bad: [1e96cf673e023dd24056b3880f7035707f11bdff] Merge pull request #3658 from StoDevX/xcode-10.2-ruby-updates
git bisect bad 1e96cf673e023dd24056b3880f7035707f11bdff
# good: [23605301bd98a59d38b2940f5286cd569c627512] Merge pull request #3602 from StoDevX/renovate/detox-12.x
git bisect good 23605301bd98a59d38b2940f5286cd569c627512
# good: [1abe9c4ab3782198e3a29a0c104828f6342aa9d4] Update dependency danger to v7.1.0
git bisect good 1abe9c4ab3782198e3a29a0c104828f6342aa9d4
# bad: [3109d592d1bf89f6b2c5201b170bd2f28a61d497] Merge pull request #3640 from StoDevX/renovate/react-native-device-info-1.x
git bisect bad 3109d592d1bf89f6b2c5201b170bd2f28a61d497
# bad: [503ad2bd745bf6279c6255cb7b82feb8e2caeb8c] Merge pull request #3573 from StoDevX/inline-requires
git bisect bad 503ad2bd745bf6279c6255cb7b82feb8e2caeb8c
# good: [3622f1d5f85b6204d3d2d356e4e1ffd005a98268] Merge pull request #3634 from StoDevX/renovate/react-native-calendar-events-1.x
git bisect good 3622f1d5f85b6204d3d2d356e4e1ffd005a98268
# bad: [db3f1298da3472b2d2b2b31781fdb9f3b2d45031] Merge branch 'master' into inline-requires
git bisect bad db3f1298da3472b2d2b2b31781fdb9f3b2d45031
# bad: [db3f1298da3472b2d2b2b31781fdb9f3b2d45031] Merge branch 'master' into inline-requires
git bisect bad db3f1298da3472b2d2b2b31781fdb9f3b2d45031
# bad: [62183cf21a045d3aa43814462fb267f0a9c67498] enable inline requires in metro bundler
[...]

Issue #3002, opened 2018-09-09
Experiment with "ram bundles"


hawkrives:
I've been seeing things about this for a few months now, and there's a bit of a description of them in this issue: https://github.com/react-native-community/discussions-and-proposals/pull/25

I'm just filing this so I can close the browser tab, really. I'd like to play with them once 0.57 is out and we've upgraded.

elicwhite:
RAMBundles aren't a new feature in 0.57 and you should be able to use them without upgrading (I'm not sure what version you are on.) You can find more information about RAMBundles and Inline Requires from the performance page in the RN docs: https://facebook.github.io/react-native/docs/0.56/performance#unbundling-inline-requires

hawkrives:
@TheSavior thanks! We're on 0.55 currently, which called them "unbundles" IIRC (or just didn't have them; I've been meaning to poke at this for a while now.)

I really appreciate the comment 😄 and I really like working with React Native, as well.

hawkrives:
closing this in favor of keeping the app build simple(r). 

Current docs are at https://archive.reactnative.dev/docs/next/ram-bundles-inline-requires

PR #3916, opened 2019-08-05
Trying inline-requires again

drewvolz:
#### Context
It's time to poke at enabling inline requires again, but it might take more setup. The initial blog post made it seem as if all that was needed  was the flip of a switch but we've since learned that we'll crash if we don't do _something_ else in the setup instructions. What that is, I'm unsure.

#### History of this feature
#3002, #3573, #3646, #3666

Related [performance instructions](https://facebook.github.io/react-native/docs/performance#ram-bundles-inline-requires) and [blog post](http://facebook.github.io/react-native/blog/2019/03/12/releasing-react-native-059).

---
⚠️ I'm [pretty confident](https://github.com/stodevx/aao-react-native/issues/3646 ) that this will cause crashing if merged as-is. 
🙇‍♂️ I'd like to figure out why that happens on release mode.


drewvolz:
The code ownership feature of our repo solicited reviews on this PR, and while it doesn’t hurt that it’s automatic, I’m honestly not looking for a review right now.

drewvolz:
Closing again, for now.

Issue #3921, opened 2019-08-08
Fix regression in HTML parser


rye:
Introduced by #3912, likely, we're getting exceptions like:

```
TypeError: TypeError: TypeError: undefined is not an object (evaluating 'u.default.parseDOM')
```

Nightly 2.8.0-pre+65526 on Android works fine, +65786 does not. If it's not #3912, it's something else in that range.

hawkrives:
Thanks all

rye:
Resolution confirmed in Android 2019-08-09 nightly.

PR #3922, opened 2019-08-08
html-lib: Update usage of htmlparser2

rye:
Inspired by [this line][a] and [this line][b], I decided to try simply adding a 2.  Flow, Jest, and the rest of the tools seem to agree that this looks like a good change.

Have not verified that this actually fixes the broken views, but for now I will note that this seems to be the resolution for #3921.

[a]: https://github.com/fb55/htmlparser2/commit/759b1220c03e55a895f971deab9f1e94c30a7f61#diff-04c6e90faac2675aa89e2176d2eec7d8L19
[b]: https://github.com/fb55/htmlparser2/commit/759b1220c03e55a895f971deab9f1e94c30a7f61#diff-04c6e90faac2675aa89e2176d2eec7d8R21

### DIFF
--- a/modules/html-lib/index.js
+++ b/modules/html-lib/index.js
@@ -1,5 +1,5 @@
 // @flow
-import htmlparser from 'htmlparser2'
+import htmlparser2 from 'htmlparser2'
 import cssSelect from 'css-select'
 export {cssSelect}
 
@@ -8,7 +8,7 @@ import {AllHtmlEntities} from 'html-entities'
 export const entities = new AllHtmlEntities()
 
 export function parseHtml(string: string): Object {
-	return htmlparser.parseDOM(string, {
+	return htmlparser2.parseDOM(string, {
 		withDomLvl1: true,
 		normalizeWhitespace: false,
 		xmlMode: false,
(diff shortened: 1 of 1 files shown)


hawkrives:
This should have absolutely no bearing on the functionality; in JS, `import blah from 'module-name'` binds the "default export" of the module to whatever name you give it.

IE, doing `import itsRainingCatsAndDogs from 'htmlparser2'` would do the same thing here.

If you use the `import {identifier} from 'module-name'` form, you are importing a "named export", instead of the "default export", and that _does_ matter, but that's not what this is doing.

rye:
Okay, well this is what the documentation changed to.

~A better idea is incoming.~ I have no clue what the problem is and will not be investigating further.

drewvolz:
@rye @hawkrives I opted to import `parseDOM` directly and use that. We aren't crashing with this usage. Looks like the TypeScript rewrite on their end uses named exports.

hawkrives (approved):

rye (commented):
My approval is latent in this comment. I say merge to see how nightly goes. (Thanks, @drewvolz!)

PR #3929, opened 2019-08-13
Remove coveralls from CircleCI config

drewvolz:
> at least it's not a 503 today

We're hitting a 405 on our PRs, so let's say goodbye for now.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -216,16 +216,6 @@ jobs:
       - danger
       - store_test_results:
           path: ./test-results
-      - run: yarn global add coveralls
-      - run:
-          name: coveralls
-          command: |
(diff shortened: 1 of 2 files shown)


drewvolz:
I checked to see the usage with ripgrep after this commit.

```
$ rg -i coveralls --iglob=[!./node_modules] ./*

./CHANGELOG.md
89:- Removed coveralls coverage reporting
```

rye:
Would it be worth replacing with codecov?

drewvolz:
I think it'd be nice to have a replacement. Would it make sense to have that be a follow-up to this PR instead of as a part of this?

rye:
I think that'd be suited to go into this PR

hawkrives (approved):

PR #3932, opened 2019-08-14
🏗 Adopt GitHub Actions and switch to CodeCov for coverage!

rye:
# Background

This repository finally has access to the GitHub Actions-based CI/CD offering directly by GitHub, so we thought it was time to migrate! A lot of the commit history here is me "battling" GitHub Actions to get everything functional, but I did try to keep each commit relatively small and reviewable, so it might be easier to go through each commit and review individual items.

# Motivation

CI/CD with GitHub is free for OSS projects, of which we are one. This means it's a great idea for us to switch so that we get multiple parallel macOS executors. Plus, porting CI configurations is one of our specialties at this point, so another exercise was long due.

# Architecture

I have placed things into a total of five builds. Instead of inter-dependent (staged) builds, all will run at once. This is because our prior strategy was to test the stuff that was most likely to break first. However, with unlimited resources, it isn't necessary. We can always add `needs` configurations in the future if we need to have structure/prevent swamping GitHub.

# Subtasks

- [x] Remove `analyze-and-build` workflow from CircleCI configuration
- [x] Port `analyze-and-build` workflow from Circle into GitHub Actions
- [x] Disable Coveralls
- [x] Set up CodeCov

(We also had some stuff here for the new nightlies, but those aren't blockers on this PR)

# Questions to be answered

- [x] Should we get rid of Danger in favor of (hopefully) future support for the annotations API? Or does it need to be added back? @hawkrives, might I defer work on that to you?

### DIFF
--- a/.github/workflows/analyze-and-build.yml
+++ b/.github/workflows/analyze-and-build.yml
@@ -10,4 +10,5 @@ jobs:
     - uses: actions/checkout@master
     - run: yarn install
     - run: yarn run bundle-data
+    - run: yarn run pretty
     - run: yarn run flow check
--- a/.github/workflows/analyze-and-build.yml
+++ b/.github/workflows/analyze-and-build.yml
@@ -8,4 +8,5 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@master
+    - run: yarn install
     - run: yarn run flow
--- a/.github/workflows/analyze-and-build.yml
+++ b/.github/workflows/analyze-and-build.yml
@@ -22,28 +22,6 @@ jobs:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
         run: bash <(curl -s https://codecov.io/bash)
 
-  bundle-android:
-    name: Test Android Bundle
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/setup-node@master
-        with: {node-version: ^12.0}
-      - uses: actions/checkout@master
-      - run: yarn install --frozen-lockfile
-      - run: yarn run bundle-data
-      - run: yarn run --silent bundle:android
-
-  bundle-ios:
-    name: Test iOS Bundle
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/setup-node@master
-        with: {node-version: ^12.0}
-      - uses: actions/checkout@master
-      - run: yarn install --frozen-lockfile
-      - run: yarn run bundle-data
-      - run: yarn run --silent bundle:ios
-
   android:
     name: Build for Android
     runs-on: ubuntu-latest
--- a/.github/workflows/analyze-and-build.yml
+++ b/.github/workflows/analyze-and-build.yml
@@ -7,4 +7,5 @@ jobs:
     name: Flow
     runs-on: ubuntu-latest
     steps:
+    - uses: actions/checkout@master
     - run: yarn run flow
--- a/.github/workflows/analyze-and-build.yml
+++ b/.github/workflows/analyze-and-build.yml
@@ -45,6 +45,7 @@ jobs:
       - run: yarn run --silent bundle:ios
 
   android:
+    name: Build for Android
     runs-on: ubuntu-latest
     steps:
       - uses: actions/setup-node@master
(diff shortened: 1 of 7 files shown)


.github/workflows/analyze-and-build.yml
```diff
@@ -0,0 +1,57 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-node@master
+        with: { version: ^12.0 }
+      - uses: actions/setup-ruby@master
+        with: {ruby-version: ^2.6}
+      - uses: actions/checkout@master
+      - run: yarn install --frozen-lockfile
+      - run: gem install bundler && bundle check || bundle install --frozen --path vendor/bundle
+      - run: android/gradlew androidDependencies --console=plain
```
hawkrives: This might need to `cd android &&` instead

rye: Done.

rye:
I'm currently diagnosing a hang in Fastlane as it runs the step:

> Generating 'OneSignalNotificationServiceExtension.appex.dSYM'

fastlane/lib/commands.rb
```diff
@@ -30,7 +30,7 @@ def authorize_ci_for_keys
 	netrc = Netrc.read
 	unless netrc["github.com"]
 		UI.message "An entry for github.com was not found in ~/.netrc; setting..."
-		netrc["github.com"] = token
+		netrc["github.com"] = token, ''
```
hawkrives, line 33: We have no idea how this was working on Circle - the netrc gem hasn’t updated in 4 years, and is documented to expect a tuple as input here…

fastlane/lib/versions.rb
```diff
@@ -25,15 +25,9 @@ def current_bundle_code
...
-
-	# android's build number goes way up because we need to exceed the old build
-	# numbers generated for the x86 build.
-	ci_build_num = build
-	build = ((2 * 1_048_576) + build.to_i).to_s if lane_context[:PLATFORM_NAME] == :android
-
-	{ci: ci_build_num, build: build}
+def build_number
+	# Should last until ~2080 for Android.
+	DateTime.now.to_time.to_i - DateTime.parse("2014-01-01").to_time.to_i
```
hawkrives, line 30: We decided to simplify the build number logic, and just use Unix epoch time. Downside: we can only submit one build a second. Upside: no more strange calculations.

hawkrives (approved):

.github/workflows/analyze-and-build.yml
```diff
@@ -0,0 +1,93 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-node@master
+        with: {node-version: ^12.0}
+      - uses: actions/checkout@master
+      - run: yarn install --frozen-lockfile
+      - run: yarn run bundle-data
+      - run: yarn run --silent bundle:ios
+
+  android:
```
rye, line 25: Needs `name` value, I think, to be consistent.

.github/workflows/analyze-and-build.yml
```diff
@@ -0,0 +1,93 @@
...
+          MATCH_PASSWORD: ${{ secrets.MATCH_PASSWORD }}
+          MAPBOX_KEY: ${{ secrets.MAPBOX_KEY }}
+          ONESIGNAL_KEY: ${{ secrets.ONESIGNAL_KEY }}
+          GIT_COMMIT_DESC: $(git log --format=oneline -n 1 $GITHUB_SHA)
+          FASTLANE_SKIP_UPDATE_CHECK: '1'
+          FASTLANE_DISABLE_ANIMATION: '1'
+          SENTRY_PROPERTIES: '../android/sentry.properties'
+          GITHUB_KEYS_REPOSITORY_TOKEN: ${{ secrets.GITHUB_KEYS_REPOSITORY_TOKEN }}
+
+  ios:
```
rye, line 50: Needs `name` value.

rye (commented):
Couple of things I forgot. Mental notes!

hawkrives:
@rye, re:danger, I’m OK with removing the invocations from this PR. I would like to keep the dangerfile around for a couple of months, though, for easy reference, because I’d like to either port some of its things to actions or to the annotations API. 

But that can definitely happen later.

rye:
@hawkrives okay, great—I didn't _add_ the Danger invocations anyhow. That makes life easier. 😄

hawkrives (approved):

hawkrives (approved):

drewvolz (approved):

PR #3949, opened 2019-08-23
💅 Update the Old Main Iconography

rye:
This PR:

- Removes the old `roundedIcon` variants from our AndroidManifest, per the documentation:

  >You must only use the `android:roundIcon` attribute if you require a different icon asset for circular masks, if for example the branding of your logo relies on a circular shape.
- Updates the iconography to a more modern design standard. This is primarily in accordance with the new Android guidelines, but iOS comes along for the ride, while the theme of "silhouetted Old Main" remains, instead gaining a more "airy" feel. The lack of borders allows us to offload the tuning work to the respective publishers.
- Instead of hand-tracing the entire shape and adapting, I worked from both the old logo and a reference image of OMH, tracing the contours of _one_ of the sides, then creating a symbol and duplicating it with a transform. This means that the lower part of the OMH contour is symmetrical about the vertical axis; the flag is a separate object.
- Since both of our distributors want "square" icons "without rounded corners," this ends up bringing us _into compliance_! Good.

Old | New
---|---
Old Icon | New Icon


rye:
N.B.: iOS builds are still failing while certificate problems are resolved.

hawkrives (dismissed):
something … happened to this icon? https://raw.githubusercontent.com/StoDevX/AAO-React-Native/ff4e7c48bea64b229632ebaa1d14c3e695e2cfff/ios/AllAboutOlaf/Images.xcassets/AppIcon.appiconset/Icon-29%401x.png

I see this: image

I don't … think the line should be there.

(It's on the Icon-29@3x, Icon-57@1x, and Icon-83.5@2x files, as well)

hawkrives:
To clarify: The black bar at the top is my web browser.

The white line down the middle is what I don't think should be there.

rye:
That's right, there's a seam between the two halves. I have not the slightest clue why, especially since they do start and end on the same pixel value, but I'll just extend the flagpole to the ground, I guess.

hawkrives (commented):
superb!

hawkrives (approved):
superb!

drewvolz (approved):
:shipit:

PR #3959, opened 2019-08-29
Update contributor documentation

rye:
I think it's better if this is more of an actual list of things to do rather than just notes, which seemed to be the old style. Kinda WIP, but can get merged whenever.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,10 +1,9 @@
 # Contributing
 
-Would you like to contribute? Great! Have a look at
-[React Native's "getting started" guide][rn-gs] to help you get your
-computer set up, and make a pull request! (If you feel like you
-should have write access to the repo, please make an issue and we can
-discuss it.)
+Would you like to contribute?
+Great!
+Have a look at [React Native's "getting started" guide][rn-gs] to help you get your computer set up.
+Then, fork the repository, and make a pull request!
 
 If you are looking to get started with the project, issues tagged with
 [`good first issue`][gfi] are a great place to start. These issues are
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -76,10 +76,28 @@ As always, please keep the [Code of Conduct][cc] in mind.
 
 ## Maintainers
 
-This stuff is more of the maintenance material.
+This stuff is more of the documentation for maintainers.
 
 ### Deploying New Versions
 
-The bare minimum: `npm version <version>`, e.g. `npm version 2.7.0`.
-It is probably a good idea to run a beta before doing a full release.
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -1,15 +1,12 @@
...
+Would you like to contribute?
+Great!
+Have a look at [React Native's "getting started" guide][rn-gs] to help you get your computer set up.
+Then, fork the repository, and make a pull request!
 
-If you are looking to get started with the project, issues tagged with
-[`good first issue`][gfi] are a great place to start. These issues are
-ones that the team has identified as simple, small changes that won't
-get too deep, and will be easy to approve for merging.
+If you are looking to get started with the project, issues tagged with [`good first issue`][gfi] are a great place to start.
```
drewvolz, line 8: Looks like we run out of those types of issues from time to time (like now!). Should we also note that the rest of our issues need help, too?

CONTRIBUTING.md
```diff
@@ -22,81 +19,51 @@ get too deep, and will be easy to approve for merging.
...
-- `npm run prettier`: Prettier enforces a common style on the JS code,
-  without us needing to edit anything
+- `npm run prettier`: Prettier enforces a common style on the JS code, without us needing to edit anything
 
-We have a special command that runs all four of those for you: `npm
-run check`.  So, before you commit and push, you may wish to run `npm
-run check`.  If you don't, TravisCI will run them for you and our
-StoDevX Bot will comment in the PR to let you know exactly what
-happened.
+We have a special command that runs all four of those for you: `npm run check`.
```
drewvolz: This needs to be removed since we removed the `check` command from package.json.

CONTRIBUTING.md
```diff
@@ -109,18 +76,28 @@ As always, please keep the [Code of Conduct][cc] in mind.
...
+(`-beta.W` where `W` is a monotonically increasing number.)
+Technically this can be whatever you want.
+
+The bare minimum: `npm version <version>`, e.g. `npm version 2.7.0`.
+It is probably a good idea to run a beta before doing a full release, to ensure that all kinks in the uploading pipeline are worked out.
+This also lets you distribute beta builds to beta testers.
+
+#### Release Commit Style
+
+@rye prefers this style _a lot_ to the `npm version` style.
```
drewvolz: 😆

CONTRIBUTING.md
rye: Yeah, I'll add that.

drewvolz (approved):

rye:
I'll ship this for now and continue playing around with it as I have time.

PR #3977, opened 2019-09-07
Remove the homescreen editor

hawkrives:
This is my proposal.

- It's not a good native experience on either platform; Android has these strange buttons, and iOS has a _really hecking bad_ scrolling list.

Okay, that's really my only reason.

@drewvolz is investigating upgrading the sortable-list package that underlays this; if that works, we can upgrade instead of removing it.


drewvolz (approved):
I upgraded and it fixed our current scrolling issue, but other issues are still present, so let's remove it altogether.

rye (approved):
Fine. In the long run we're going to redesign the Home view anyhow, so this will (probably) not be needed.

PR #3980, opened 2019-09-07
Disable in-app radio player by default

### LINKED ISSUE #2747: Open Radio streams in an in-app browser, instead of attempting to play it inline
Radio streams are tricky.

The current state is: 
- sometimes iOS will start on the first tap; sometimes it won't.
- Android always opens the stream in a web browser.

I propose that we change iOS to also open in a web browser.

hawkrives:
Closes #2747 

I've left in all the code, because I would like to eventually get this working in the app.

For now, though, this puts the Android behavior of just opening the player websites on iOS as well.


drewvolz (approved):

hawkrives:
@rye To be clear, this PR is like three lines; all of the (non-working) code remains, unlike my other PRs from today, in order to allow us to continue experimenting with it in development.

hawkrives:
@dependabot merge

PR #3983, opened 2019-09-08
Convert uses of const to let

drewvolz:
Requested by @hawkrives.


hawkrives (approved):
wow, we have a lot of code here.

thanks for making this change drew

PR #3989, opened 2019-09-09
Fix prettier check to actually fail if needed

hawkrives:
I discovered today that our prettification check has been happily passing everything since we moved it off of CircleCI.

This fixes that issue.


hawkrives:
bddddbf failed the check, as expected: https://github.com/StoDevX/AAO-React-Native/pull/3989/checks?check_run_id=216295395

drewvolz (approved):

PR #4009, opened 2019-09-14
Fix radio crashes

### LINKED ISSUE #4005: Tapping on radio tabs is crashing at the moment
We suspected the `parser` (#3996) and/or `buffer` (#3999) upgrades.

I've gone thru the steps of reverting both of these changes locally on iOS but the crash remains.

drewvolz:
Closes #4005 

Fixes the rn-theme import in the radio views


rye (approved):
Sure? Sure.

PR #4060, opened 2019-09-29
fastlane: Use yarn instead of npx

rye:
`npx` _might_ happily move us up into versions of react-native we have not explicitly upgraded to, which can break stuff pre-emptively. This PR removes the last uses of `npx` in the codebase.


hawkrives (approved):
It _shouldn’t_, but it seems to be. This will definitely use our versions, though.

drewvolz (approved):

PR #4061, opened 2019-09-29
Resign first responder on searchbar blur

### LINKED ISSUE #1141: [iOS] Fix keyboard showing up over Orgs detail
The keyboard consistently appears over the detail view of a student org with these steps:

1. Open student orgs list
2. Search for org with website in detail (Agnes A Cappella, GLOW, etc)
3. Tap on result and go to detail
4. Tap link to open browser
5. Go back, notice the keyboard on the detail view

Something something resign first responder.

drewvolz:
Resolves #1141 

This handles preventing the keyboard from reappearing on the detail view as well as fully closes the keyboard when returning to the view that instantiated it.

The keyboard was never fully resigning first responder. We observed the keyboard disappear when navigating to the Student Orgs detail view, but it would reappear when navigating to a view beyond such as a webview.

This aims to use the searchbar's `onBlur` event to resign the searchbar's textinput as first responder since we're not inputting text after navigating away from the searchbar.

### DIFF
--- a/modules/searchbar/searchbar-ios.js
+++ b/modules/searchbar/searchbar-ios.js
@@ -34,6 +34,10 @@ export class SearchBar extends React.Component<Props> {
 		this._ref = ref
 	}
 
+	onHandleBlur = () => {
+		this._ref && this._ref.unFocus()
+	}
+
 	render() {
 		return (
 			<NativeSearchBar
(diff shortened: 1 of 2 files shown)


modules/searchbar/searchbar-ios.js
```diff
@@ -34,12 +34,17 @@ export class SearchBar extends React.Component<Props> {
 		this._ref = ref
 	}
 
+	onHandleBlur = () => {
+		this._ref && this._ref.unFocus()
```
drewvolz, line 38: Calling `unFocus` here instead of `unBlur`

> 
> * `blur()` - uses the iOS endEditing:true method on the underlying UISearchBar view.
> * `unFocus()` - calls resignFirstResponder on the UITextField used by the UISearchBar.

hawkrives (approved):

rye (approved):

PR #4067, opened 2019-09-30
fastlane: Add support from beta-ing from GitHub Actions

rye:
In two parts:

- Add code for testing the `GITHUB_REF` environment variable, which is present on GitHub Actions builds and will be set to `refs/tags/<tag>` on tag builds.

- Add a fallback that shells out to `git show` to retrieve commit messages in case we want to use that instead.

The workflow should be otherwise identical… so no changes to the GitHub Actions config are required. (Yay!)

### DIFF
--- a/fastlane/lib/util.rb
+++ b/fastlane/lib/util.rb
@@ -27,7 +27,7 @@ end
 def tagged?
 	travis = ENV['TRAVIS_TAG'] && !ENV['TRAVIS_TAG'].empty?
 	circle = ENV['CIRCLE_TAG'] && !ENV['CIRCLE_TAG'].empty?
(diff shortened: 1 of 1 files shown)


drewvolz (approved):

fastlane/lib/util.rb
```diff
@@ -27,15 +27,21 @@ def travis_cron?
 def tagged?
 	travis = ENV['TRAVIS_TAG'] && !ENV['TRAVIS_TAG'].empty?
 	circle = ENV['CIRCLE_TAG'] && !ENV['CIRCLE_TAG'].empty?
-	travis || circle
+	github = ENV['GITHUB_REF'] && ENV['GITHUB_REF'] =~ /^refs\/tags\/(.+)$/
```
hawkrives: Maybe just `/^refs\/tags\//`?

hawkrives (approved):

PR #4093, opened 2019-10-05
Convert FAQ view to use hooks

drewvolz:
* Remove FaqView Class –– replace with functional component
* Replace state with hooks –– `useState` and `useEffect`


source/views/faqs/index.js
```diff
@@ -1,5 +1,5 @@
 // @flow
-import * as React from 'react'
+import React, {useState, useEffect} from 'react'
```
hawkrives: We typically do `* as React`, because there are certain things that Flow only puts on the React object if you do, and that prevents us from spending ten minutes diving into the Flow internals.

source/views/faqs/index.js
```diff
@@ -15,79 +15,62 @@ const styles = StyleSheet.create({
...
-	}
-
-	fetchData = async (reload?: boolean) => {
+	let fetchData = async (reload?: boolean) => {
 		let {text}: {text: string} = await fetch(API('/faqs'), {
 			forReload: reload ? 500 : 0,
 		})
 			.json()
-			.catch(() => ({text: 'There was a problem loading the FAQs'}))
+			.catch(() => setText('There was a problem loading the FAQs'))
```
hawkrives: This should not have changed

source/views/faqs/index.js
```diff
@@ -15,79 +15,62 @@ const styles = StyleSheet.create({
...
 
-		this.setState(() => ({text}))
+		setText(text)
```
hawkrives: If you make this return text, instead, this function could be pulled out of the component entirely ;-)

hawkrives (changes requested):

source/views/faqs/index.js
```diff
@@ -15,79 +15,62 @@ const styles = StyleSheet.create({
...
-	loading: boolean,
-	refreshing: boolean,
-}
-
-export class FaqView extends React.PureComponent<Props, State> {
-	static navigationOptions = {
-		title: 'FAQs',
-	}
+	useEffect(() => {
+		fetchData().then(() => setLoading(false))
```
hawkrives: This can use the wrong setLoading. You need to put [setLoading] as a dependency of the useEffect hook.

hawkrives (changes requested):

source/views/faqs/index.js
drewvolz: I don't understand the advantage of pulling this out, unless you're hinting at centralizing all of the refresh functions across the codebase?

source/views/faqs/index.js
drewvolz: Could you explain a little more?

source/views/faqs/index.js
drewvolz: Understood.

source/views/faqs/index.js
```diff
@@ -15,79 +14,60 @@ const styles = StyleSheet.create({
...
-			forReload: reload ? 500 : 0,
+		await fetchData(true).then(text => {
+			setText(text)
 		})
-			.json()
-			.catch(() => ({text: 'There was a problem loading the FAQs'}))
 
-		this.setState(() => ({text}))
+		setRefreshing(false)
 	}
```
hawkrives, line 48: can we put the 

```js
if (loading) {
	return <LoadingView />		
}
```

back here, instead of returning the result of the ternary, please?

hawkrives (dismissed):

source/views/faqs/index.js
hawkrives: We chatted over the phone.

drewvolz: I can only recall bits about promise cancellations. Did we chat about more than that?

hawkrives (approved):

rye (approved):

PR #4097, opened 2019-10-06
Migrate Nightly Builds to GitHub Actions

### LINKED ISSUE #4096: Migrate Nightlies to GitHub Actions
As the last part of the GitHub Actions migration, we need to migrate nightlies to GitHub Actions. This will involve

- Adding a "schedule" to the `on` block.
- Re-tooling Fastlane configs so that they understand schedule builds.
- Removing the CircleCI configuration and disabling Circle's integration with our repository.

rye:
Closes #4096.

This is a surprisingly trivial PR, and basically just required adding support for scheduled builds to the fastlane check logic. Since all steps are run, fastlane will (read: _should_) pick up that a build is a nightly and go through the motions it usually would.

It's very nice, of course, that we don't have to use an environment variable here.


hawkrives (approved):

drewvolz (approved):
👍 :shipit:

PR #4098, opened 2019-10-06
Use fetch api's delay in some of the refresh functions

drewvolz:
In some cases, the fetch API param `delay` was named `forReload`. This PR renames that prop to swizzle where the delay logic happens. 

Renamed delay prop and removed from refresh altogether 
* Bonapp Menu
* News

Added fetch delay logic but removed from refresh altogether
* Student work
* Transportation (other modes)

### DIFF
--- a/source/views/transportation/other-modes/list.js
+++ b/source/views/transportation/other-modes/list.js
@@ -1,6 +1,5 @@
 // @flow
 import * as React from 'react'
-import delay from 'delay'
 import {OtherModesRow} from './row'
 import {TabBarIcon} from '@frogpond/navigation-tabs'
 import * as defaultData from '../../../../docs/transportation.json'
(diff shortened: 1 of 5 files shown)


source/views/transportation/other-modes/list.js
```diff
@@ -57,23 +56,17 @@ export class OtherModesView extends React.PureComponent<Props, State> {
...
 		this.setState(() => ({refreshing: false}))
 	}
 
-	fetchData = async () => {
+	fetchData = async (reload?: boolean) => {
 		let {data: modes}: {data: Array<OtherModeType>} = await fetch(
 			transportationUrl,
+			{
+				delay: reload ? 500 : 0,
+			},
```
hawkrives:
```suggestion
			{delay: reload ? 500 : 0},
```

instead of the three-line version, please? :)

hawkrives (changes requested):

hawkrives (approved):

PR #4157, opened 2019-11-02
🖨 stoPrint tweaks

### LINKED ISSUE #3986: Failed login in stoPrint spams invalid prompt ad infinitum
I'm hitting infinite loading / spamming of the "invalid login" prompt.

1. Open Settings, enter bogus credentials, press login
2. Open Home > stoPrint

_Originally posted by @drewvolz in https://github.com/StoDevX/AAO-React-Native/pull/3985#issuecomment-529255724_

drewvolz:
Potentially fixes #3986 

The diff is a little misleading. I recommend expanding the collapsed lines to gain full context.

Two things:
1. I'm noticing an error view that reports that login was incorrect (alert) but the stoprint error view reports that the print server was having issues. I've flipped these two conditionals. Not sure if this works in the _other_ case now. Can someone with credentials check, please?

2. The `StoPrintErrorView` and `StoPrintNoticeView` each are wrapped by the `Timer` component with intervals. I've opted to remove the timer/interval from the case where login fails. This is where that infinite alert view is happening, I think. Someone with credentials will have to test this, too, please.

I'm not 100% confident on these changes so please help test these. Thanks!


source/views/stoprint/components/error.js
```diff
@@ -49,6 +49,32 @@ export class StoPrintErrorView extends React.PureComponent<Props> {
...
 		)
+
+		const errorWithoutTimer = (
+			<ScrollView
+				contentContainerStyle={styles.content}
+				showsVerticalScrollIndicator={false}
+				style={styles.container}
+			>
+				<Icon color={sto.black} name={iconName} size={100} />
+				<NoticeView
```
hawkrives, line 60: Isn't this the view that's provided for an invalid username/password? 

If so, we shouldn't send them to the Connection Issue page.

hawkrives (changes requested):

source/views/stoprint/components/error.js
drewvolz: The connection issue page is being used as the error page for invalid username/password once we've confirmed credentials aren't blank. I agree that we should render something else.

drewvolz:
Unsure what the right approach here should be. Closing this PR as we have the tracking issue.

source/views/stoprint/components/error.js
drewvolz: If we route to a different view, is this approach valid?

hawkrives:
@drewvolz should we close this MR? Or is it still valuable?

drewvolz:
We can close if we've resolved infinite error alerts for incorrect credentials.

hawkrives:
Enough of the implementation has changed that I feel confident in saying that we will no longer have infinite alerts.

drewvolz:
Alright! Closing this.

PR #4159, opened 2019-11-03
Upgrade to React-Native v0.61

hawkrives:
- [x] iOS: builds
- [x] Android: builds
- [x] Flow: passes

This was … really painless.

Changes:

- You need Xcode 11 to build from the command line (Xcode 10 can only build from the GUI)
- Android fonts are now copied at build time, instead of being part of our repo
- To fix type errors with React-Redux, I've replaced all `connect()` calls with the new `useDispatch` / `useSelector` hooks. Informal testing says that everything still works.
- To skip Cocoapods, you can `export FP_PODS=no` (for you, @rye)
- I added Cocoapods to our Gemfile, because it's a ruby dep
- We now use cocopods for iOS builds, because they're required
- We've switched from `react-native-sentry` to `@sentry/react-native`, because Sentry said to
- I removed `@frogpond/dimensions` because it just needed the "hasNotch" check from rn-device-info
- Added `useNativeDriver` where prompted
- assorted other changes for associated upgrades

Please test and lmk if things look broken! Otherwise we'll merge and test on nightly. It works at least enough for that.


hawkrives:
hmm

```
❌  ld: library not found for -lBVLinearGradient
❌  clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

.github/workflows/analyze-and-build.yml
```diff
@@ -62,8 +62,8 @@ jobs:
       - uses: actions/setup-node@master
         with: {node-version: ^12.0}
       - uses: actions/checkout@master
+      - run: gem install bundler:1.17.2 && bundle check || bundle install --frozen --path vendor/bundle
```
rye, line 65: Why was this necessary? CocoaPods' version requirement on bundler of `~> 1.3`?

ios/Podfile
```diff
@@ -0,0 +1,44 @@
...
+  pod 'Folly', :podspec => '../node_modules/react-native/third-party-podspecs/Folly.podspec'
+
+  target 'AllAboutOlafUITests' do
+    inherit! :search_paths
+    # Pods for testing
+  end
+
+  use_native_modules!
+end
+
```
rye: Our usual code style is tabs instead of spaces, if I remember right.

Also, please remove the trailing newline.

scripts/pods.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+set -e -o pipefail
+
+USE_PODS=${FP_PODS:-yes}
+
+if [[ $USE_PODS = 'yes' ]]; then
+	if [[ $(uname) = 'Darwin' ]]; then
+		bundle install --path .bundle
```
rye: Is this required? I don't use `.bundle` in my Ruby projects, but if it's necessary in this case then I guess it's necessary.

modules/timer/index.js
```diff
@@ -80,6 +80,10 @@ export class Timer extends React.Component<Props, State> {
 		this.setState(() => ({loading: false}))
 	}
 
+	refresh = () => {
+		this._refresh()
+	}
+
```
rye, line 86: This seems somewhat kludgey to me. Is there a particular reason why we need a shim?

rye (approved):
I gave this a pretty rapid once-over and it looks good enough to merge, just a couple of nitpicks and questions that don't really affect whether or not it can get shipped in nightly. (I'm perfectly happy to dogfood as long as it builds and makes it through detox on iOS.)

.github/workflows/analyze-and-build.yml
hawkrives: If I don’t specify a version, then whatever `gem` wound up installing didn’t work, and the error message suggested this.

modules/timer/index.js
hawkrives: RN’s flow types require that functions passed to RefreshControl’s refresh prop be of the type `() => void`, and all Async functions return Promise<>, which causes type errors at each call site. 

My other alternative here is to typecast this.refresh as a variable in the render method here, but that requires that we cast through `any`, so I almost prefer this way. 

Idk.

scripts/pods.sh
hawkrives: Where does bundler install packages by default? I don’t actually know; this is what my shell autocompletes when I do “bundle install” 😅

If I shouldn’t need this, then I’m happy to take it out.

modules/timer/index.js
hawkrives:
```js
refresh = () => {
	let start = Date.now()
	this.setState(() => ({loading: true}))

	this.updateTime()

	let tick = Promise.resolve(null)

	let elapsed = Date.now() - start
	if (elapsed < 500) {
		tick = tick.then(() => delay(500 - elapsed))
	}

	tick.then(() => {
		this.setState(() => ({loading: false}))
	})
}
```

That's my other option for this method lol

drewvolz (approved):
@rye's comments make sense. Thank you @hawkrives for putting in the legwork! 🥳

scripts/pods.sh
rye: You shouldn't need this; the default installation location is fine.

You can use `gem environment` to figure out where it's installing gems. `~/.gem` would be a place that it'd install.

hawkrives:
I’m going to remove the Cocoapods Update step from ./scripts/pods.sh; it takes > 30s on my laptop even after only five minutes have passed. 

I think I’ll see if “pod install” fails, and advise running it manually if the exit code is non-0.

.github/workflows/analyze-and-build.yml
rye: Ah okay, that clears things up.

modules/timer/index.js
rye: Alright, just curious.

PR #4165, opened 2019-11-04
ci: Pin build executors to their current versions

rye:
In case something changes, we'd rather not be surprised, I suppose.

I think I'm fine with it as long as it's not specifically for macOS, and as long as we keep updating things quickly, which hasn't been an issue.

### DIFF
--- a/.github/workflows/analyze-and-build.yml
+++ b/.github/workflows/analyze-and-build.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   js:
     name: JavaScript Checks
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


hawkrives (approved):
Thanks!

drewvolz:
What will be our reminder to update this?

hawkrives:
The build failing, probably.

Or, through a more optimistic lens, we'll remember to do it when Apple or Ubuntu release a new version.

I presume that Github will also announce fairly loudly when they update their macOS version availability, as well.

drewvolz (approved):

PR #4183, opened 2019-11-07
macos 10.15

hawkrives:
Hi GitHub. 

GitHub is requiring at we switch to “macos-latest” as they don’t want to support multiple MacOS environments. 

BD0FB7E6-E6F7-40B3-84CD-D2016F4DEA3E

### DIFF
--- a/.github/workflows/analyze-and-build.yml
+++ b/.github/workflows/analyze-and-build.yml
@@ -57,7 +57,7 @@ jobs:
 
   ios:
     name: Build for iOS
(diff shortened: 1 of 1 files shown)


drewvolz (approved):
👍 :shipit:

rye (approved):
N.B.: The [docs](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/virtual-environments-for-github-hosted-runners) now have `macos-latest` as the slug we should use. (This PR is correct.)

PR #4184, opened 2019-11-07
Upgrade react-navigation to v4.0.10

drewvolz:
I noticed that `react-navigation` had fallen behind on versions. This upgrades us to v4.0.10 which introduces the following changes:

* `createStackNavigator()` moved into `react-navigation-stack`
* `createBottomTabNavigator()` moved into `react-navigation-tabs`
* the library now depends upon `react-native-reanimated`


hawkrives:
There should have also been a checksum change in Podfile.lock, I think?

Also I want to dig through that lockfile to find out how we now have two deps on rn-tabs…

hawkrives:
So `react-navigation-material-bottom-tabs@1.1.1` also has updates, to v2.1.5. It doesn't look like they changed too much; [the 2.0.0 changes say](https://github.com/react-navigation/material-bottom-tabs/releases/tag/v2.0.0)  "change activeTintColor -> activeColor" so that should be pretty straightforward.

drewvolz:
Thanks for finding that

rye:
@drewvolz any chance this could get a rebase onto the typescript changes? Or, should we close and upgrade to v5 at the same time? Here's [an upgrade guide](https://reactnavigation.org/docs/upgrading-from-4.x/) for the v4-to-v5 migration.

drewvolz:
Yeah, I'll have a look at this soon!

drewvolz:
Okay, this has been rebased onto the tip of `master`. If we want to upgrade as well, I'm thinking that could be done in a separate PR. What do you think?

rye (approved):
Looks good! We can follow up with the latest version later.

PR #4187, opened 2019-11-07
License viewer

drewvolz:
Integrates the [license-checker](https://github.com/davglass/license-checker) package in order to parse _all_ of our deps and their licenses. 

📸 Screenshots

~ | ~
--|--
Simulator Screen Shot - iPhone 11 - 2019-11-17 at 12 53 01 | Simulator Screen Shot - iPhone 11 - 2019-11-06 at 23 44 40

## Todo
- [ ] Flow types


source/views/settings/screens/licenses/index.js
```diff
@@ -0,0 +1,82 @@
...
+	static navigationOptions = {
+		title: 'Open Source Licenses',
+	}
+
+	extractNameFromGithubUrl = (url: string) => {
+		if (!url) {
+			return null
+		}
+
+		const reg = /((https?:\/\/)?(www\.)?github\.com\/)?(@|#!\/)?([A-Za-z0-9_]{1,15})(\/([-a-z]{1,20}))?/iu
```
drewvolz: Would like to find a cleaner way to pull out the parts of the url

source/views/settings/screens/licenses/index.js
```diff
@@ -0,0 +1,82 @@
...
+		const components = reg.exec(url)
+
+		if (components && components.length > 5) {
+			return components[5]
+		}
+		return null
```
drewvolz, line 30: I'd love to find a better way to do all of this

source/views/settings/screens/licenses/index.js
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	sortDataByKey = (data: Array<LicenseType>, key: string) => {
+		data.sort((a, b) => {
+			return a[key] > b[key] ? 1 : b[key] > a[key] ? -1 : 0
+		})
+		return data
```
drewvolz: This allows for us to sort by a key. I've chosen to group by repository, but it needs to be tweaked to be case-insensitive.

source/views/settings/screens/licenses/index.js
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	render() {
+		const {navigation} = this.props
+
+		let data = licenseData.data.filter(item => Object.keys(item).length)
+		let nonEmptyData = data[0]
```
drewvolz: Our custom js bundling added an extra layer of unwrapping... can we do this cleanly?

source/views/settings/screens/licenses/item.js
```diff
@@ -0,0 +1,67 @@
...
+					this.props.navigation.navigate('LicenseDetailView', licenseText)
+				}
+			>
+				<Row alignItems="center">
+					{image.length ? (
+						<Image
+							accessibilityIgnoresInvertColors={true}
+							source={{uri: image}}
+							style={styles.image}
+						/>
```
drewvolz: Unfortunately, github doesn't redirect on username's avatar's like it does for repos, so it's possible to render a blank image here.

hawkrives:
I’d still like to not show any images in the list, but maybe that’s just me.

source/views/settings/screens/licenses/index.js
hawkrives: Check lodash’s sortBy method, and probably just call .toLowerCase() on both keys.

hawkrives:
So… how do I look at a nested dependency’s license? Are they all in one huge giant list?

drewvolz:
It’s all one big list at the moment.

drewvolz:
I'm happy to remove the images @hawkrives. I'm thinking they're more effort than they're worth.

hawkrives:
@drewvolz That's kinda where I'm at. They look nice, but there's a bunch of edge cases, not to mention memory use and rate limiting and bandwidth and… 

I'm pretty firmly of the opinion that we should just not show images there.

source/views/settings/screens/licenses/index.js
```diff
@@ -0,0 +1,69 @@
...
+	static navigationOptions = {
+		title: 'Open Source Licenses',
+	}
+
+	extractNameFromGithubUrl = (url: string) => {
+		if (!url) {
+			return null
+		}
+
+		const reg = /((https?:\/\/)?(www\.)?github\.com\/)?(@|#!\/)?([A-Za-z0-9_]{1,15})(\/([-a-z]{1,20}))?/iu
```
rye, line 24: According to the form validation parameters on [GitHub's signup page](https://github.com/join), the parameters for a username are

- A username may only contain alphanumeric characters or hyphens.
- A username cannot have multiple consecutive hyphens.
- A username cannot begin or end with a hyphen.
- Usernames have a maximum length of 39 characters.

There are similar constraints somewhere for the repository name.

All of this is to say that this regex doesn't match the following types of GitHub urls:

- URLs with a username component containing hyphens. (e.g. https://github.com/frog-pond/ccc-server)
- URLs with a repository component containing numbers. (e.g. https://github.com/rye/rust-netaddr2)

source/views/settings/screens/licenses/index.js
```diff
@@ -0,0 +1,69 @@
...
+		const components = reg.exec(url)
+
+		if (components && components.length > 5) {
+			return components[5]
+		}
+		return null
```
rye, line 30: I think ES2018's [named capture groups](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Groups_and_Ranges) might be great solution here. The prior groups are apparently needed for optional matching, but if we were to just use something like `(?<username>[a-z0-9]*)`, then we could just yank out `components.groups.username`.

source/views/settings/screens/licenses/index.js
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	render() {
+		const {navigation} = this.props
+
+		let data = licenseData.data.filter(item => Object.keys(item).length)
+		let nonEmptyData = data[1] // TODO: don't do this
```
rye, line 37: What do you want to do instead?

rye (changes requested):
Some small nits to pick. Looks good from the screenshots, but a couple of logical changes to make the code more robusts.

Might want to mark `data/licenses/licenses.json` as autogenerated (via [`linguist-generated`](https://help.github.com/en/github/administering-a-repository/customizing-how-changed-files-appear-on-github)), and add checks to CI to keep that information up-to-date.  (Or, perhaps, remove the file entirely and include it in the data build?)

rye:
Also, I was able to successfully run this in an Android emulator just now, and did not encounter any crashes.

source/views/settings/screens/licenses/index.js
drewvolz: If this could avoid indexing that would be ideal

source/views/settings/screens/licenses/index.js
drewvolz: I'm open to exploring implementing it with named capture groups!

source/views/settings/screens/licenses/index.js
drewvolz: Thank you for listing the form validation parameters. I think a few things could use tweaking seeing your feedback.

1. We can add capturing for hyphens in the username
1. We can add capturing for numbers in the repo name
1. We can completely ignore max length of username

What are your thoughts on something like this, then?

```diff
-([A-Za-z0-9_]{1,15})(\/([-a-z]{1,20}))
+([A-Za-z0-9_-]{1,})(\/([-a-z0-9]{1,}))
```

* Here's a [repl.it](https://repl.it/@drewvolz/TriangularDistinctQuadrants) to demonstrate the changes to the username capturing
* Here's a [regex101](https://regex101.com/r/AKwSjV/1) to demo all captured parts

drewvolz:
Thinking it could be beneficial to reimplementing this with LicensePlist and be done with it.

PR #4292, opened 2019-12-18
ios: Unshallow Git history after checking out

rye:
We want to have the full history back to the most recent tag.

This commit causes us to unshallow and prune up the Git history before moving on in the build.

Should clear up the spurious nightly failure we've had since actions/checkout v2 came out.


drewvolz (approved):

rye:
I wonder if we should just do this everywhere in case there is anything using the output of `git describe`?

hawkrives (approved):
Thanks!

PR #4351, opened 2020-01-18
Fix time tests

hawkrives:
The core of this change is to pin the time tests to a specific date; I chose `2019-12-18T18:39:45-0600`, because it is a recent date when all of the tests passed.

I then set the moments created by our moment helper functions to the year/month/day of the "base moment" [^1].

I followed up by removing the ±6-month logic from #3375; it seems to not be needed anymore.

Thoughts?

---

[^1]: I'm setting y/m/d instead of just dayOfYear because I need to set year as well. I could probably do just year().dayOfYear(), but I feel that this requires less thinking when you see it again.


.github/workflows/analyze-and-build.yml
```diff
@@ -35,7 +35,7 @@ jobs:
         with: {ruby-version: ^2.6}
       - uses: actions/checkout@master
       - run: yarn install --frozen-lockfile
-      - run: gem install bundler && bundle check || bundle install --frozen --path vendor/bundle
+      - run: gem install bundler:1.17.2 && bundle check || bundle install --frozen --path vendor/bundle
```
rye, line 38: This certainly makes sense to me. Until Fastlane supports v2 or whatever the problem is preventing us from upgrading, we should pin.

source/views/building-hours/lib/__tests__/moment.helper.js
```diff
@@ -3,10 +3,28 @@ import moment from 'moment-timezone'
...
+export const dayMoment = (time: string, format: ?string = 'ddd h:mma') => {
+	let parsed = moment.tz(time, format, false, CENTRAL_TZ)
+
+	let dow = parsed.day()
+
+	parsed
+		.year(baseTime.year())
+		.month(baseTime.month())
+		.date(baseTime.date())
+		.day(dow)
```
rye: Nit: change `dow` -> `dayOfWeek` for legibility?

rye (approved):
This all looks very good to me, with just one little nitpick change…

Thanks!

source/views/building-hours/lib/__tests__/moment.helper.js
hawkrives: Done.

rye (approved):

drewvolz:
Thanks @hawkrives!

PR #4423, opened 2020-02-15
Rename CMIE to Taylor Center

drewvolz:
We received a suggestion to rename the CMIE to the Taylor Center

### DIFF
--- a/data/building-hours/10-taylor-center.yaml
+++ b/data/building-hours/10-taylor-center.yaml
@@ -1,5 +1,4 @@
 name: Taylor Center
-subtitle: Center for Equity and Inclusion
 category: Offices
 
 schedule:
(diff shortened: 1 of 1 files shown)


hawkrives (commented):
Should we rename the subtitle to Taylor Center for … 

?

drewvolz:
I wanted to ask that here. It seemed redundant but I'm not sure.

rye:
I'd say the subtitle should be the full name, "Taylor Center for Equity and Inclusion"?

It's a bit redundant and it might look weird but it's probably the best possible choice.

hawkrives:
Yeah. I'd treat it like Piper Center - it's never abbreviated to "Center for Vocation and Career".

drewvolz:
If it’s not an acronym and we’re not providing additional context for something, what do you all think of simply removing the subtitle here, like we do elsewhere?

hawkrives:
I would be fine with not having a subtitle here, since it's not an acronym anymore

hawkrives (approved):

hawkrives:
idk, Piper Center is just called Piper Center… don't know if we need the full name on Taylor

PR #4656, opened 2020-04-27
add dependabot config file

hawkrives:
By default, Dependabot will update Ruby and JS dependencies "live".

That is, as soon as the update is published to the package repository. However, we have been innundated with updates to certain packages, `*cough*`aws-* via fastlane`*cough*`, which we don't even directly depend on.

So, this commit sets Ruby updates to update weekly, and JS updates to update daily.

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,0 +1,9 @@
+version: 1
+update_configs:
+  - package_manager: "javascript"
+    update_schedule: "daily"
+    directory: "/"
+
+  - package_manager: "ruby:bundler"
+    update_schedule: "weekly"
+    directory: "/"


rye (approved):
Looks good as-is.

Possible follow-up: update "direct" (declared via `Gemfile` and `package.json`) dependencies live, with all others updating on a weekly/daily interval?

hawkrives:
@rye I wasn't sure how to do that. Dependabot's configs are a bit limited. https://dependabot.com/docs/config-file/

rye:
So I took a stab at expanding the Ruby config to only apply live updates for direct dependencies:

```yaml
version: 1
update_configs:
  - package_manager: "javascript"
    update_schedule: "daily"
    directory: "/"

  - package_manager: "ruby:bundler"
    update_schedule: "live"
    directory: "/"
    allowed_updates:
      # Only allow "all" updates of "direct" dependencies
      - match:
          dependency_type: "direct"
          update_type: "all"

  - package_manager: "ruby:bundler"
    update_schedule: "weekly"
    directory: "/"
    allowed_updates:
      # The default update behavior (for all updates) is weekly.
      - match:
          dependency_type: "all"
          update_type: "all"
```

This looks like it should work, but triggers the validation error:

```console
1. The property '#/update_configs/2' is a duplicate.
   Update configs must have a unique combination
   of 'package_manager', 'directory', and 'target_branch'
```

So I'm assuming what I'd want is not possible without changing the structure of the config.  Oh well, I'll tuck that away as an idea to explore later.

hawkrives:
@rye nice work! darn.

drewvolz:
Thanks for doing this!

PR #4681, opened 2020-05-13
ci: Use GitHub Actions for Data Deployment

rye:
Besides the added last step that uses a `git push` refspec that maps `gh-pages-(SHA)` onto the remote `gh-pages`, the validation and bundling steps are the same, and now the deployment step looks like:

```console
$ git rev-parse --abbrev-ref HEAD
gh-pages-91d6e1dda48e2457f7a4a4d3342cec190ba078cb
$ git commit -m "Automated data deployment at $(date -Is)"
[gh-pages-91d6e1dda48e2457f7a4a4d3342cec190ba078cb (root-commit) f9f051d5] Automated data deployment at 2020-05-13T18:37:45-05:00
 17 files changed, 53 insertions(+)
 create mode 100644 breaks.json
 create mode 100644 building-hours.json
 create mode 100644 bus-times.json
 create mode 100644 chapel.json
 create mode 100644 color-printers.json
 create mode 100644 contact-info.json
 create mode 100644 credits.json
 create mode 100644 dictionary.json
 create mode 100644 faqs.json
 create mode 100644 help.json
 create mode 100644 legal.json
 create mode 100644 news-styles.css
 create mode 100644 news-styles.json
 create mode 100644 pause-menu.json
 create mode 100644 privacy.json
 create mode 100644 transportation.json
 create mode 100644 webcams.json
$ git show --stat HEAD
commit f9f051d518f180bdf41058cefcb9d144a884d4d8 (HEAD -> gh-pages-91d6e1dda48e2457f7a4a4d3342cec190ba078cb)
Author: AAO Data Deployment Bot <>
Date:   Wed May 13 18:37:45 2020 -0500

    Automated data deployment at 2020-05-13T18:37:45-05:00

 breaks.json         |  1 +
 building-hours.json |  1 +
 bus-times.json      |  1 +
 chapel.json         |  1 +
 color-printers.json |  1 +
 contact-info.json   |  1 +
 credits.json        |  1 +
 dictionary.json     |  1 +
 faqs.json           |  1 +
 help.json           |  1 +
 legal.json          |  1 +
 news-styles.css     | 37 +++++++++++++++++++++++++++++++++++++
 news-styles.json    |  1 +
 pause-menu.json     |  1 +
 privacy.json        |  1 +
 transportation.json |  1 +
 webcams.json        |  1 +
 17 files changed, 53 insertions(+)
```

I also temporarily disabled Travis deployments to prevent overwriting of commits made here.

### DIFF
--- a/.github/workflows/deploy-data.yml
+++ b/.github/workflows/deploy-data.yml
@@ -1,0 +1,38 @@
+name: Validate/Deploy Data
+
+on:
+  push: { branches: [master] }
+
+jobs:
+  validate-then-deploy:
+    name: Validate, then deploy data
+    runs-on: ubuntu-latest
+    env: task=GH_PAGES
+    steps:
+      - uses: actions/set-up-node@master
+        with: {node-version: ^12.0}
+      - name: Check out the code
+        uses: actions/checkout@master
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
+      - name: Validate the data
+        run: |
+          npm run validate-bus-data
+          npm run validate-data
+      # Having validated the data, we now prepare a bundle.  This script creates
+      # files in a docs/ directory.
+      - name: Bundle the data
+        run: npm run bundle-data
+      # Notes: actions/checkout@v2 no longer fetches entire history nor enters
+      # detached HEAD state.  We really just need to check out an (orphaned)
+      # branch and then add, commit, and push the appropriate directory.
+      - name: Commit the data
+        run: |
+          git checkout --orphan "gh-pages-$GITHUB_SHA"
+          git --work-tree=docs add .
+          git commit -m "Automated data deployment at $(date -Is)"
+          git show --stat HEAD
+      # If the previous commit successfully happened
+      - name: Deploy the data
+        run:
+          git push --force-with-lease "gh-pages-$GITHUB_SHA:gh-pages"
--- a/.github/workflows/deploy-data.yml
+++ b/.github/workflows/deploy-data.yml
@@ -30,7 +30,7 @@ jobs:
         run: |
           git checkout --orphan "gh-pages-$GITHUB_SHA"
           git --work-tree=docs add .
-          git commit -m "Automated data deployment at $(date -Is)"
+          git commit --author="Data Deployment Bot <>" -m "Automated data deployment at $(date -Is)"
           git show --stat HEAD
       # If the previous commit successfully happened
       - name: Deploy the data
--- a/.github/workflows/deploy-data.yml
+++ b/.github/workflows/deploy-data.yml
@@ -35,4 +35,4 @@ jobs:
       # If the previous commit successfully happened
       - name: Deploy the data
         run:
-          git push --force-with-lease "gh-pages-$GITHUB_SHA:gh-pages"
+          git push --force-with-lease origin "gh-pages-$GITHUB_SHA:gh-pages"
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,11 +34,11 @@ jobs:
         - npm run validate-bus-data
         - npm run validate-data
         - npm run bundle-data
-      deploy:
-        provider: pages
-        skip_cleanup: true
-        github_token: $GITHUB_PAGES_TOKEN
-        local_dir: docs/
+      #deploy:
+      #  provider: pages
+      #  skip_cleanup: true
+      #  github_token: $GITHUB_PAGES_TOKEN
+      #  local_dir: docs/
 
 cache:
   yarn: true


hawkrives (approved):

.github/workflows/deploy-data.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Bundle the data
+        run: npm run bundle-data
+      # Notes: actions/checkout@v2 no longer fetches entire history nor enters
+      # detached HEAD state.  We really just need to check out an (orphaned)
+      # branch and then add, commit, and push the appropriate directory.
+      - name: Commit the data
+        run: |
+          git checkout --orphan "gh-pages-$GITHUB_SHA"
+          git --work-tree=docs add .
+          git commit --author="Data Deployment Bot <>" -m "Automated data deployment at $(date -Is)"
```
drewvolz, line 33: Is the `<>` to denote that there's no email associated with this author?

drewvolz, line 33: Is this coming from `date(1)`? I'm seeing `date: illegal option -- I` when I try to run this.

.github/workflows/deploy-data.yml
```diff
@@ -0,0 +1,38 @@
...
+          git show --stat HEAD
+      # If the previous commit successfully happened
+      - name: Deploy the data
+        run:
+          git push --force-with-lease "gh-pages-$GITHUB_SHA:gh-pages"
```
drewvolz: Is there a reason to have `gh-pages` twice in the branch name here? I notice above that it's only listed once?

```sh
git checkout --orphan "gh-pages-$GITHUB_SHA"
```

.github/workflows/deploy-data.yml
rye: Yes; I have verified that Git seems fine committing this, but it does require the `<>`.  The "deployment bot" is not a real human and cannot be reached at an email, so it makes more sense to just keep the email empty rather than provide another destination for spam emails, I think.

rye: Yes, option is present in GNU coreutils' `date`, which is what we'd have on the `ubuntu-18.04` builder.  Are you using macOS' `date(1)`?  That'd be why this would fail for you.

.github/workflows/deploy-data.yml
rye: Yes, this is one of the advanced features of the `refspec` in `git push <remote> <refspec>`.

The man page for `git-push` says,

```
  <refspec>...
      Specify what destination ref to update with what source object. The
      format of a <refspec> parameter is an optional plus +, followed by the
      source object <src>, followed by a colon :, followed by the destination
      ref <dst>.

      The <src> is often the name of the branch you would want to push, but
      it can be any arbitrary "SHA-1 expression", such as master~4 or HEAD
      (see gitrevisions(7)).

      The <dst> tells which ref on the remote side is updated with this push.
      Arbitrary expressions cannot be used here, an actual ref must be named.
      If git push [<repository>] without any <refspec> argument is set to
      update some ref at the destination with <src> with
      remote.<repository>.push configuration variable, :<dst> part can be
      omitted--such a push will update a ref that <src> normally updates
      without any <refspec> on the command line. Otherwise, missing :<dst>
      means to update the same ref as the <src>.
```

The key here is that typical pushes, `git push origin gh-pages` instruct `git` to update `gh-pages` _on the remote_ with the contents of `gh_pages` _locally_.  The `:` effectively defines a mapping between local ref and remote ref.

Here I'm using `gh-pages-$GITHUB_SHA` as a throwaway branch, but I want to push its contents to the ref `gh-pages` on the remote.  That's why I need `--force-with-lease`, and that's also the reason for the `:`.  Otherwise I'd have to rename or just not use the naming scheme, which I suppose I could do.

(Side note: this is why if you `git push origin :<branch>`, you delete the remote branch; there is no `<src>` so you're overwriting `<branch>` with nothing.)

(Another side note: you can also do cool things like `git push origin :` to push all branches that have a remote-side match, which is helpful for dumping lots of changes at once.)

.github/workflows/deploy-data.yml
drewvolz: Great!

drewvolz: Ah, thanks for clarifying!

.github/workflows/deploy-data.yml
drewvolz: Wow, thanks for explaining that!

drewvolz (approved):
Thank you @rye!

rye:
Alright, I'm going to give this a try.

If data deployment suddenly breaks, I'll probably push a fix as quickly as I can, but this might break downstream temporarily.

PR #4784, opened 2020-07-16
ci: Attribute data deployment commits to github-actions[bot]

rye:
- Remove the per-commit `--author=` setting since Git _requires_ a committer to be specified via at least `user.email`, and will default to the values of those fields for authorship information.

- Use [the API response](https://api.github.com/users/github-actions%5Bbot%5D) to identify the `login` and `id` fields of the `github-actions[bot]` user (which I believe has the permissions allocated to the token included on the runner), and use that user as the author and committer.

I did this mostly to get the avatar to something we control, but also to eliminate the non-existant account we were using previously.  I feel like this is more elegant.

### DIFF
--- a/.github/workflows/deploy-data.yml
+++ b/.github/workflows/deploy-data.yml
@@ -29,8 +29,8 @@ jobs:
       # branch and then add, commit, and push the appropriate directory.
       - name: Commit the data
         run: |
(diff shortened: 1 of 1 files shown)


hawkrives (approved):

drewvolz:
More great improvements, thanks @rye!

PR #4801, opened 2020-07-22
Update android stuff

hawkrives:
Builds on #4800

- update SwipeRefreshLayout, at Android Studio's nudging
- update compileSdkVersion and targetSdkVersion to 30
- update buildToolsVersion to 30.0.0 (pre-installed according to https://github.com/actions/virtual-environments/blob/487c88b726392dda3077c6938c688a9601486763/images/linux/Ubuntu1804-README.md)
- update the gradle plugin from 3.5 to 4.0
- update gradle from v5.5 to v6
- clean up a bit of the github-actions file

### DIFF
--- a/.github/workflows/analyze-and-build.yml
+++ b/.github/workflows/analyze-and-build.yml
@@ -43,8 +43,8 @@ jobs:
           echo 524288 | sudo tee -a /proc/sys/fs/inotify/max_queued_events
           echo 524288 | sudo tee -a /proc/sys/fs/inotify/max_user_instances
           sudo sysctl -p
-      - run: cd android && ./gradlew androidDependencies --console=plain
       - run: echo 'org.gradle.workers.max=2' >> ./android/gradle.properties
+      - run: cd android && ./gradlew androidDependencies --console=plain
       - run: bundle exec fastlane android ci-run
         env:
           FASTLANE_PASSWORD: ${{ secrets.FASTLANE_PASSWORD }}
--- a/.github/workflows/analyze-and-build.yml
+++ b/.github/workflows/analyze-and-build.yml
@@ -35,7 +35,8 @@ jobs:
         with: {ruby-version: ^2.6}
       - uses: actions/checkout@master
       - run: yarn install --frozen-lockfile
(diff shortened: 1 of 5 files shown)


drewvolz (approved):
👍 :shipit:

.github/workflows/analyze-and-build.yml
```diff
@@ -35,11 +35,16 @@ jobs:
...
+      - run: gem install bundler:1.17.2
+      - run: bundle check || bundle install --frozen --path vendor/bundle
       - name: Raise the fs.inotify ulimits to 524288 watches/queued events/user instances
-        run: echo 524288 | sudo tee -a /proc/sys/fs/inotify/max_user_watches && echo 524288 | sudo tee -a /proc/sys/fs/inotify/max_queued_events && echo 524288 | sudo tee -a /proc/sys/fs/inotify/max_user_instances && sudo sysctl -p
-      - run: cd android && ./gradlew androidDependencies --console=plain
+        run: |
+          echo 524288 | sudo tee -a /proc/sys/fs/inotify/max_user_watches
+          echo 524288 | sudo tee -a /proc/sys/fs/inotify/max_queued_events
+          echo 524288 | sudo tee -a /proc/sys/fs/inotify/max_user_instances
+          sudo sysctl -p
```
rye, line 45: This is far more readable! 👍

.github/workflows/analyze-and-build.yml
```diff
@@ -35,11 +35,16 @@ jobs:
...
       - run: echo 'org.gradle.workers.max=2' >> ./android/gradle.properties
+      - run: cd android && ./gradlew androidDependencies --console=plain
```
rye, line 47: Is there any particular reason that `--console=plain` is needed?

We do [have other options](https://docs.gradle.org/current/userguide/command_line_interface.html#sec:command_line_logging), though I don't really have a strong reason to move towards them.

hawkrives: I could do the `-D` version in the config file, but I like their rich terminal output locally; it's just needed here so that Actions can show it nicely, if I recall correctly. This just seemed like the most obvious place to put it so that it would only apply to Actions, I suppose.

PR #4843, opened 2020-08-23
Add a script for parsing the "dining hours" page from reslife

### LINKED ISSUE #4841: Add dorm-specific dining hours info
Dining hours are posted to https://wp.stolaf.edu/reslife/dining-hours/ each day to inform students of when their dorm can get food. It would be nice of AAO could display this information in-app. I'm not sure where this information would be best displayed, but probably under the Stav Hall building hours? Alternatively, each housing location could have its own panel so a student could favorite their own for easy access.

rye:
See #4841. (Closes #4841.)

This PR adds a script that loads and processes the human-entered data on the [dining hours](https://wp.stolaf.edu/reslife/dining-hours/) page. This script _could_ be run automatically via a GitHub Action or something like that, if we wanted to.

(See the updated schedule file on this branch for reference of what this script outputs.)


rye:
I think we should probably rebase or merge once #4845 enters master. I'm going to mark this as a draft until then.

scripts/stav-covid.py
```diff
@@ -0,0 +1,153 @@
...
+parser.add_argument('-o', '--output', type=argparse.FileType('w'), required=True,
+                    help='the file to write the schedule into')
+options = parser.parse_args()
+
+# Fetch the output to a string, `output`.
+url = "https://wp.stolaf.edu/reslife/dining-hours/"
+output = requests.get(url).text
+
+# Now, parse it.
+output = BeautifulSoup(output)
```
drewvolz: One nit, feel free to ignore. 


> GuessedAtParserWarning: No parser was explicitly specified, so I'm using the best available HTML parser for this system ("html.parser"). This usually isn't a problem, but if you run this code on another system, or in a different virtual environment, it may use a different parser and behave differently.
>
> The code that caused this warning is on line 19 of the file s.py. To get rid of this warning, pass the additional argument 'features="html.parser"' to the BeautifulSoup constructor.

```suggestion
output = BeautifulSoup(output, features="html.parser")
```

scripts/stav-covid.py
```diff
@@ -0,0 +1,153 @@
...
+			timespec = cols[idx]
+
+			print("Processing table entry: {} gets {} at {} on {}".format(dorm, meal, timespec, date))
+
+			# Split the timespec on the timespec_re matches
+			times = [time.strip() for time in timespec_re.split(timespec)]
+
+			if len(times) != 2:
+				warnings.warn("timespec {} did not split nicely".format(timespec))
+				continue
```
drewvolz, line 104: So in this case, the script came across one such example (which is why I'm glad this is here!)

> UserWarning: timespec 10:15-10-45 did not split nicely


Is it worth our effort to try and compensate here?

scripts/stav-covid.py
```diff
@@ -0,0 +1,153 @@
...
+				t_close += datetime.timedelta(hours=12)
+
+			# Dinner always occurs in the evening.  If we have something else,
+			# we're probably 12 hours off.
+			if meal == "Dinner" and (t_open.hour < 12 or t_close.hour < 12):
+				t_open += datetime.timedelta(hours=12)
+				t_close += datetime.timedelta(hours=12)
+
+			# The open time should not be the same or even close to the close
+			# time, so we should wrap it around.  (We do this by bumping up the )
```
drewvolz, line 122: By bumping up the... close time? 😄

scripts/stav-covid.py
drewvolz, line 19: A nit –– feel free to ignore.

> GuessedAtParserWarning: No parser was explicitly specified, so I'm using the best available HTML parser for this system ("html.parser"). This usually isn't a problem, but if you run this code on another system, or in a different virtual environment, it may use a different parser and behave differently.
>
> The code that caused this warning is on line 19 of the file s.py. To get rid of this warning, pass the additional argument 'features="html.parser"' to the BeautifulSoup constructor.

```suggestion
output = BeautifulSoup(output, features="html.parser")
```

scripts/stav-covid.py
drewvolz, line 118: Cool, very happy to see this handling here considering there is a case in the data that breaks the regex

> UserWarning: timespec 10:15-10-45 did not split nicely

What are your thoughts on compensating? Does it make our program too brittle if we try to catch this case?

I'm thinking about how `timespec_re` matches for delimiters, and how timeparse splits on colons. If we find a dash where a colon _should_ be (that sounds bad as I type this), we could make a swap?

Eh. My idea sounds super brittle. Perhaps the happy path leaves less room for errors and thorny edge cases of further typos breaking the flow.

scripts/stav-covid.py
drewvolz: Bumping up the ____ (close time)?

```suggestion
			# time, so we should wrap it around.  (We do this by bumping up the 
			# close time)
```

drewvolz (commented):
This is looking very nice, @rye! Thanks a ton.

scripts/stav-covid.py
rye: I was thinking it might be smart to limit the number of splits via `maxsplit` on [`re.split`](https://docs.python.org/3/library/re.html#re.split), (then further splitting any invalid things on word boundaries and rejoining with a `:` before finally giving up) to catch stuff like that.

If you feel like I should, I totally could!

drewvolz: That's a decent algorithm. I'm okay with it as-is, but I wonder what @hawkrives thinks?

scripts/stav-covid.py
rye: Hmm, on my system it actually picks lxml… is there a preference between the two?

hawkrives: I would prefer lxml, but it requires another package. Maybe html.parser?

scripts/stav-covid.py
hawkrives: I'm good with either!

hawkrives (approved):

scripts/stav-covid.py
rye: Yeah, this isn't performance-critical so I'll opt to go with the (batteries-included) Python version.

scripts/stav-covid.py
rye: I think I'll probably just opt to expect the happy path. If anyone complains, we do have the valid excuse that what was entered into the table is _technically_ wrong, and hopefully on the next pass it should get fixed. 😉

scripts/stav-covid.py
drewvolz: - [BeautifulSoup's own list of differences](https://www.crummy.com/software/BeautifulSoup/bs4/doc/#differences-between-parsers)
- [Short breakdown from SO](https://stackoverflow.com/questions/25714417/beautiful-soup-and-table-scraping-lxml-vs-html-parser)

Honestly, feel free to choose lxml.

+- |  html.parser
--|--
Advantages | Batteries included, Decent speed, Lenient (as of Python 2.7.3 and 3.2.)
Disadvantages | Not very lenient (before Python 2.7.3 or 3.2.2)

+- |  lxml
--|--
Advantages | Very fast, Lenient
Disadvantages | External C dependency

+- | html5lib
--|--
Advantages | Extremely lenient, Parses pages the same way a web browser does, Creates valid HTML5
Disadvantages | Very slow, External Python dependency

drewvolz (approved):

scripts/stav-covid.py
```diff
@@ -0,0 +1,167 @@
...
+
+	# We only know the date, so let's back out to that.
+	date = date.date()
+
+	print("Inferred date of {}; processing table now.".format(date))
+
+	body = table.find("tbody")
+
+	# NOTE(rye): Should only have one thead
+	meals = [meal_name_from(th) for th in table.find("thead").find("tr").find_all("th")]
```
hawkrives, line 96: Do we want to `assert len(meals) == 1`, then?

hawkrives (approved):
One thought

scripts/stav-covid.py
rye: If I cared much about error handling, certainly—that'd require doing a separate search for `table.find_all("thead")` instead; I could totally do that, though.

rye:
Merging! Here comes (potential) breakage? But not for long.

PR #4845, opened 2020-08-23
ci: Validate data during JavaScript checks, too

### LINKED ISSUE #4844: GitHub Actions not validating data on PRs
I just realized, on #4843, that I pushed invalid data but the "JavaScript Checks" are green. We don't seem to run `yarn run validate-data`, but I think we probably should, just to make sure valid data gets merged (and eventually deployed).

rye:
What the title says.

Basically, before bundling the data, first make sure that it is valid.

Closes #4844.

### DIFF
--- a/.github/workflows/analyze-and-build.yml
+++ b/.github/workflows/analyze-and-build.yml
@@ -15,6 +15,7 @@ jobs:
       - uses: actions/checkout@master
       - run: yarn install --frozen-lockfile
       - run: yarn run bundle-data
+      - run: yarn run validate-data
       - run: yarn run pretty --no-write --list-different
       - run: yarn run flow check
       - run: yarn run lint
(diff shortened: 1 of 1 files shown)


drewvolz:
Just so I understand, even though we run validate-data in the deploy-data workflow, we should have it run in the analyze-and-build workflow too? https://github.com/StoDevX/AAO-React-Native/blob/ba14670d555e47757b2b2a8d8b9348efaf0c6248/.github/workflows/deploy-data.yml#L20-L22

rye:
That's right… `validate-data` does some checks to make sure the data is formed properly.

Prior to this PR, invalid data could get a green build but would fail to deploy once merged.

drewvolz:
Great, that makes sense. Thanks, Kris!

drewvolz (approved):

rye:
Alright, I'm going to merge since this is a fairly minor (strictening) change and I'd like to get it merged into #4843 as well.

PR #4868, opened 2020-09-10
Flow ➸ TypeScript

rye:
As we look towards TypeScript as a possible replacement (to make React-Native upgrades a bit better), this is one key step.

Here's the general process I followed:

- Rename `.js` files to `.ts`, or if they contain JSX things, rename them to `.tsx`.
- Run `yarn run p`
- If any parsing errors, fix them
- Run `yarn run lint` and fix or disable any lints (didn't really have to disable that many lints)
- Run `yarn run bundle:android` to ensure that we can actually generate a legit bundle (there are some parsing errors that prettier doesn't hit, but the bundler does, for some reason)


rye:
Alright, that's the last of the work (I think?) needed to add typescript, and now eslint is spitting out some valuable errors! (~150 of them)

ESLint is now mostly emitting warnings about the flowtype syntax leftovers (`Parsing error: [...] expected.`) as well as warnings regarding missing return types, poor uses of `any`, etc. — I consider these valid things we should change in our TS code before calling it migrated.

We can, of course, turn off `--max-warnings` so that ESLint exits successfully once all of our _errors_ are resolved, but I haven't made that change yet because I'm not sure if it's the best of all ideas.

drewvolz:
@rye Thanks for kicking this off!

.eslintrc.yaml
```diff
@@ -1,26 +1,24 @@
...
   - react
   - react-native
-  - flowtype
   - react-hooks
 
 extends:
   - eslint:recommended
-  - plugin:flowtype/recommended
   - plugin:react/recommended
+  - 'plugin:@typescript-eslint/recommended'
```
drewvolz, line 17: Chatted with Hawken out of band and he came across [this article](https://github.com/typescript-eslint/typescript-eslint/blob/master/docs/getting-started/linting/TYPED_LINTING.md) on typed linting. 

(It also instructs [how to fix](https://github.com/typescript-eslint/typescript-eslint/blob/master/docs/getting-started/linting/TYPED_LINTING.md#i-get-errors-telling-me-the-file-must-be-included-in-at-least-one-of-the-projects-provided) the two "The file must be included in at least one of the projects provided" errors we currently have for metro.config and babel.config, too.)

Looks like if we enable this plugin inside `.eslintrc.yaml` we gain even more warnings and errors (which is good!)

`1896 problems (1413 errors, 483 warnings)`

```suggestion
  - 'plugin:@typescript-eslint/recommended'
  - 'plugin:@typescript-eslint/recommended-requiring-type-checking'
```

rye: @drewvolz I committed that config change and then uncommitted it; we can always add those lints later, but I think I'd like _valid typescript code_ to be the bar we have to cross in order to merge this PR.

drewvolz: Sounds good to me!

rye:
@drewvolz and @hawkrives, this is now ready for review when you have a chance!

hawkrives (approved):

drewvolz (approved):

PR #4901, opened 2020-10-22
Bump @sentry/react-native to v1.9.0

rye:
In an attempt to get a functional Android nightly out the door soonish, we should try bumping @sentry/react-native to v1.9.0, which we can't do because we're at the max PR limit.

It looks like we actually… already had v1.8.2? But we can do this just to be sure that we have the latest fixes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,7 +99,7 @@
     "@hawkrives/react-native-alternate-icons": "0.6.0",
     "@mapbox/react-native-mapbox-gl": "6.1.3",
     "@react-native-community/netinfo": "5.9.6",
-    "@sentry/react-native": "1.8.2",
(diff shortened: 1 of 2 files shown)


drewvolz:
Fastlane?

> [23:32:38]: Your certificate is not valid, please check end date and renew it if necessary

rye:
I'm pretty sure `MV29PU4H6Z.cer` is expired.

drewvolz (approved):

rye:
@dependabot rebase

rye:
Oh, that's a _me_ PR.

rye (commented):
I think this will be needed for a successful Android nightly, so I'm going to merge this.

✅

PR #4902, opened 2020-10-22
Fastlane: Re-enable Android nightlies

rye:
Per out-of-band discussion, we want to give this a try tonight and see if we can get a nightly out the door on Android in due time.

- If the published build still hasn't gone through review, we shouldn't publish. We should check at 23:00 UTC tomorrow (2020-11-17) to see if the nightly from tonight has been published, and if not, probably publish it.


drewvolz (approved):

rye:
@drewvolz, what do you think, send it?

drewvolz (approved):
Full send.

PR #4919, opened 2020-11-15
Update iOS build environment

rye:
Xcode 11.1 was deprecated on 2020-11-05, which caused builds to start failing.

This PR pins us to the `macos-10.15` build environment, like we pin the Android builders to `ubuntu-18.04`, and changes the version we `xcode-select --install` to `11.7`, the latest `11.x` version available on the macOS build environment.

**This PR may still fail to build due to certificate issues, but it should no longer fail to even start building the iOS code.**

Also in this PR:
- Update code signing identity and provisioning profile
- Update detox simulator to iOS 13.7

### DIFF
--- a/.github/workflows/analyze-and-build.yml
+++ b/.github/workflows/analyze-and-build.yml
@@ -67,7 +67,7 @@ jobs:
     steps:
       - uses: actions/setup-node@master
         with: {node-version: ^12.0}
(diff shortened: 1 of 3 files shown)


rye (commented):
✅

drewvolz (approved):

PR #4935, opened 2020-11-18
Nix wifi reporting

### LINKED ISSUE #4699: Turn off wifi reporting in help tools
An out of band conversation with @hawkrives revealed that the backend for wifi reporting no longer collects the reports and the db entry for aao has been nixed

(a) So we can disable this view from the backend by changing the enabled value here to false. https://github.com/StoDevX/AAO-React-Native/blob/master/data/help/0-wifi.yaml

(b) Additionally, we should remove the location collection logic, and/or this feature + deps + views altogether? 

Thoughts @hawkrives and @rye?

drewvolz:
Closes #4699 

Holding off on updating our privacy policy to reflect the increase until we publish.


data/help/0-wifi.yaml
```diff
@@ -1,18 +0,0 @@
----
-key: wifi
-enabled: true
```
rye, line 3: So that the production version does not break, should we just set `enabled: false` for now until a new version is released?

[The render method](https://github.com/StoDevX/AAO-React-Native/blob/c2904add407f533a4aef46c95917c9d035efb8aa/source/views/help/wifi.js#L77) does not have any error checking in case the value is undefined, as it would be. Not sure if I'm missing some kind of safety mechanism, but if this file just disappears, could that have unintended consequences?

drewvolz: I was debating turning it off vs removing it. You make a good point! I'll make the change you're suggesting about disabling this file instead of removing it.

rye (approved):

PR #5098, opened 2021-04-06
fastlane: Fix `check_build` lane

rye:
We don't typically get fork PRs, so there was a disparity between how we built for normal PRs and fork/restricted PRs.

This PR changes `check_build` for iOS to now use `gym` (like the "normal" `build` lane), which should make maintaining these two sides much simpler. We explicitly _don't_ have certificates available in the restricted case, so we disable signing and archive building and the like to effectively just test the compile.


hawkrives (approved):

drewvolz (approved):
RSLGTM 👍

PR #5123, opened 2021-05-06
Allow for setting of server url in development

drewvolz:
What this does
* Adds a component for getting/setting the server url to localstorage
* Switches the development settings screen to a functional component

General notes
* The demo had a _tiny_ bit of  🎩🪄 magic applied to it. I ended up having to bust `mem`'s cache inside `ccc-server` to get what you see here.
* You may stick urls in there that may or may not be servers. I won't stop you.
* What about android? It works. Don't you worry.

~ | ~ 
--|--
empty | empty
invalid | invalid
valid | valid

Demo

https://user-images.githubusercontent.com/5240843/117249977-bc28fa80-adff-11eb-8d36-8ce2ee8f2c42.mov

### DIFF
--- a/source/lib/constants.ts
+++ b/source/lib/constants.ts
@@ -1,3 +1,3 @@
 export const GH_BASE_URL = 'https://github.com/StoDevX/AAO-React-Native'
 export const GH_NEW_ISSUE_URL = `${GH_BASE_URL}/issues/new`
-export const PRODUCTION_SERVER_URL = 'https://stolaf.api.frogpond.tech/v1'
+export const DEFAULT_URL = 'https://stolaf.api.frogpond.tech/v1'
--- a/source/lib/constants.ts
+++ b/source/lib/constants.ts
@@ -1,2 +1,3 @@
 export const GH_BASE_URL = 'https://github.com/StoDevX/AAO-React-Native'
 export const GH_NEW_ISSUE_URL = `${GH_BASE_URL}/issues/new`
+export const PRODUCTION_SERVER_URL = 'https://stolaf.api.frogpond.tech/v1'
--- a/source/views/settings/screens/overview/developer.tsx
+++ b/source/views/settings/screens/overview/developer.tsx
@@ -1,28 +1,16 @@
 import * as Sentry from '@sentry/react-native'
 import * as React from 'react'
-import {Alert, StyleSheet} from 'react-native'
-import {
-	Section,
-	PushButtonCell,
-	CellTextField,
-	ButtonCell,
-} from '@frogpond/tableview'
+import {Alert} from 'react-native'
+import { Section, PushButtonCell } from '@frogpond/tableview'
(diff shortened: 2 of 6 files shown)


source/lib/constants.ts
```diff
@@ -1,2 +1,3 @@
 export const GH_BASE_URL = 'https://github.com/StoDevX/AAO-React-Native'
 export const GH_NEW_ISSUE_URL = `${GH_BASE_URL}/issues/new`
+export const PRODUCTION_SERVER_URL = 'https://stolaf.api.frogpond.tech/v1'
```
rye: What do you think about the name `DEFAULT_URL` or something like that?

There's a simple "yes or no" relationship between "default" and "(not default)", whereas "production" (if I were reading this for the first time) would lead me to wonder if there's a staging server floating around out there. (which there's not :wink:)

source/views/settings/screens/overview/developer.tsx
```diff
@@ -1,25 +1,40 @@
...
+import {PRODUCTION_SERVER_URL} from '../../../../lib/constants'
 
-type Props = {navigation: NavigationScreenProp<any>}
+type Props = {
+	navigation: NavigationScreenProp<any>
+}
+
+export const DeveloperSection = ({navigation}: Props): React.ReactElement => {
+	const [serverAddress, setServerAddress] = React.useState('')
+	const _serverAddress = React.useRef<CellTextField>(null)
```
hawkrives: Would you mind calling this `serverAddressRef`? I use underscore-prefixed words to mean "variables that must exist but which I don't really care about the contents of".

source/views/settings/screens/overview/developer.tsx
```diff
@@ -32,24 +47,73 @@ export class DeveloperSection extends React.Component<Props> {
...
 				<PushButtonCell
-					onPress={this.sendSentryException}
+					onPress={sendSentryException}
 					title="Send a Sentry Exception"
 				/>
 			</Section>
-		)
-	}
+
+			<Section
```
hawkrives: On the other hand… does it make it significantly cleaner to make a ServerUrlSection component?

hawkrives (changes requested):

source/views/settings/screens/overview/developer.tsx
drewvolz: I'll happily take that naming suggestion!

source/views/settings/screens/overview/developer.tsx
drewvolz: It might make more sense for this to be pulled out. I'll see what I can do!

source/lib/constants.ts
drewvolz: Great suggestion, I'll have that changed to default.

hawkrives (approved):

rye (approved):

PR #5235, opened 2021-07-18
Fix some const-reassignment issues

rye:
This is getting caught by some static analysis tools.  I don't think this is necessarily triggering crashes in the wild, but best to be sure, I suppose.


hawkrives (approved):
Wow, good catches!

PR #5237, opened 2021-07-18
Upgrade @babel/core to v7.14.6, remove old dependency versions

rye:
I... think this should work?  Not sure why we still had `babel-core` and `babel-eslint` here.

If so, closes #5222 by superseding it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -148,7 +148,7 @@
     "wordwrap": "1.0.0"
   },
   "devDependencies": {
-    "@babel/core": "7.12.10",
+    "@babel/core": "7.14.6",
     "@babel/eslint-parser": "7.14.7",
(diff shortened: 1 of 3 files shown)


hawkrives (approved):

hawkrives:
Android:

```
Task :app:bundleReleaseJsAndAssets
error source/lib/login.ts: function (path) {
          return fn.call(state, path, state);
        } could not be cloned. Run CLI with --verbose flag for more details.
Error: function (path) {
          return fn.call(state, path, state);
        } could not be cloned.
    at Object.serialize (v8.js:273:7)
```

hawkrives:
Closing in favor of #5382

PR #5364, opened 2021-09-29
Patch react-native files to fix the build

drewvolz:
Two patches have been added and are called inside of a prepare step as part of the app's build process. Short of a react native upgrade, we can update both `RCTCxxBridge` and `RCTTurboModuleManager` in order to fix the build.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "d": "yarn-deduplicate && yarn",
     "ios": "react-native run-ios",
     "lint": "eslint --report-unused-disable-directives --cache source/ modules/ scripts/ images/",
+    "prepare": "patch -p0 -Nfsi contrib/rn-fixes.patch || true",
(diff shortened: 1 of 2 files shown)


hawkrives (approved):

package.json
```diff
@@ -14,6 +14,7 @@
     "d": "yarn-deduplicate && yarn",
     "ios": "react-native run-ios",
     "lint": "eslint --report-unused-disable-directives --cache source/ modules/ scripts/ images/",
+    "prepare": "patch -p0 -Nfsi contrib/rn-fixes.patch || true",
```
rye, line 17: Minor suggestion in case you want to change it: as prior art, in #3557 we used a wildcard and added a numeric prefix.  If you want to add one, then adding another patch in the future (if necessary) wouldn't require changing this line.

rye (approved):

package.json
drewvolz: Thanks for mentioning this. Funnily enough, I did attempt that approach and had to settle on this implementation. The wildcard and multiple patch approach ended up failing whichever secondary patch was being applied?

rye: Alright! Just thought I'd mention it as a possibility.

PR #5387, opened 2021-10-09
OneSignal Cleanup

rye:
- Remove the unused `ONESIGNAL_KEY` build variable
- Remove the `ic_stat_onesignal_default.png` files.  (They can stay in history in case we ever decide to bring them back.)


drewvolz (approved):
Maybe we'll bring notifications back in a very limited capacity one day, but I don't think we're fans of OneSignal. LGTM 👍

hawkrives (approved):

PR #5413, opened 2021-10-19
strip trailing newline from oles-go

hawkrives:
6A42DD60-0BD0-48F0-B648-FB712E4F4883


drewvolz:
As a follow-up issue, should we add support for showing the end time for a schedule that only has a single entry?

drewvolz (approved):

rye (approved):

PR #5439, opened 2021-10-20
Update 5-3-caas-speaking-space.yaml

brandonjosephcash:
Update CAAS SS

### DIFF
--- a/data/building-hours/5-3-caas-speaking-space.yaml
+++ b/data/building-hours/5-3-caas-speaking-space.yaml
@@ -2,7 +2,7 @@ name: CAAS – Speaking Space
 category: Help and Support
 
 schedule:
-  - title: Hours
+  - title: Appointment Only
     hours: Appointment Only
 
 
--- a/data/building-hours/5-3-caas-speaking-space.yaml
+++ b/data/building-hours/5-3-caas-speaking-space.yaml
@@ -3,7 +3,8 @@ category: Help and Support
 
 schedule:
   - title: Appointment Only
-    hours: Appointment Only
+    hours: []
+    notes: By Appointment Only
 
 
 breakSchedule:
--- a/data/building-hours/5-3-caas-speaking-space.yaml
+++ b/data/building-hours/5-3-caas-speaking-space.yaml
@@ -3,10 +3,8 @@ category: Help and Support
 
 schedule:
   - title: Hours
-    hours:
-      - {days: [Mo, Tu, We, Th], from: '7:00pm', to: '9:00pm'}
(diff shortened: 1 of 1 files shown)


data/building-hours/5-3-caas-speaking-space.yaml
```diff
@@ -3,10 +3,8 @@ category: Help and Support
 
 schedule:
   - title: Hours
-    hours:
-      - {days: [Mo, Tu, We, Th], from: '7:00pm', to: '9:00pm'}
-      - {days: [Sa], from: '2:00pm', to: '5:00pm'}
-      - {days: [Su], from: '6:00pm', to: '9:00pm'}
+    hours: Appointment Only
```
drewvolz: @brandonjosephcash so things that are appointment only: I'm thinking we should list it as open during a certain range of dates/times to prevent showing it as always closed? We technically need a days/times or else we'll list it as closed.

rye: The data validator does not like this as-is — can we use an empty schedule with the "[by] appointment only" moved to a note?

```suggestion
    hours: []
    notes: By Appointment Only
```

rye (dismissed):

data/building-hours/5-3-caas-speaking-space.yaml
drewvolz: @rye I'm thinking we should list it as open during a certain range of dates/times still (if possible) to prevent showing it as always closed. I'm hoping they have a range of hours they accept appointments?

hawkrives: @drewvolz They do, but I think they'd rather it be emphatically by-appointment-only. I'm good with making a schedule named "Appointment Only" with an empty list of hours, and a note, I think.

data/building-hours/5-3-caas-speaking-space.yaml
```diff
@@ -3,10 +3,8 @@ category: Help and Support
 
 schedule:
   - title: Hours
```
hawkrives:
```suggestion
  - title: Appointment Only
```

hawkrives (dismissed):
If we combine this with @rye's suggestion, I think it'll be about as good as we can make it.

data/building-hours/5-3-caas-speaking-space.yaml
drewvolz: Fair enough!

drewvolz (approved):

hawkrives:
Thanks, @drewvolz!

PR #5440, opened 2021-10-20
Update 5-4-caas-writing-desk.yaml

brandonjosephcash:
Update CAAS WD

### DIFF
--- a/data/building-hours/5-4-caas-writing-desk.yaml
+++ b/data/building-hours/5-4-caas-writing-desk.yaml
@@ -2,39 +2,11 @@ name: CAAS – Writing Desk
 category: Help and Support
 
 schedule:
-  - title: Rølvaag 354
+  - title: Hours
     hours:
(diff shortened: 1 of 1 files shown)


data/building-hours/5-4-caas-writing-desk.yaml
```diff
@@ -2,39 +2,11 @@ name: CAAS – Writing Desk
 category: Help and Support
 
 schedule:
-  - title: Rølvaag 354
+  - title: Hours
```
drewvolz, line 5: @brandonjosephcash did the writing desk change to just one location? Is it located elsewhere (not RML 354)?

hawkrives: If it's only one location now, setting it to `Hours` makes sense (plus it won't show up in the list as an extra label)

drewvolz (approved):

PR #5481, opened 2021-10-26
Add a network timeout when installing node_modules on ios ci builds

hawkrives:
yarn fails so often…

### DIFF
--- a/.github/workflows/analyze-and-build.yml
+++ b/.github/workflows/analyze-and-build.yml
@@ -70,7 +70,7 @@ jobs:
       - run: gem install bundler:1.17.2
       - run: bundle check || bundle install --frozen --path vendor/bundle
       # try installing node_modules twice
-      - run: yarn install --frozen-lockfile || yarn install --frozen-lockfile
+      - run: yarn install --frozen-lockfile --network-timeout 60000
       - run: brew tap wix/brew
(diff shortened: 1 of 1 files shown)


.github/workflows/analyze-and-build.yml
```diff
@@ -69,7 +69,8 @@ jobs:
       - run: git fetch --prune --unshallow
       - run: gem install bundler:1.17.2
       - run: bundle check || bundle install --frozen --path vendor/bundle
-      - run: yarn install --frozen-lockfile
+      # try installing node_modules twice
+      - run: yarn install --frozen-lockfile || yarn install --frozen-lockfile
```
drewvolz: Thoughts on trying to use the `--network-timeout` flag instead? Maybe something like `30000` or more if needed?

rye: `--network-timeout` is in milliseconds, so this could be `60000` for example if the timeout should be one minute

hawkrives: Sure, I'm game for that. Who wants to take over the PR to try that?

hawkrives: 😊

drewvolz (approved):

PR #5485, opened 2021-10-27
add type information to credits.tsx

hawkrives:
fix one bug, pass it around… 99 type bugs still up on the wall :-)

### DIFF
--- a/source/views/settings/screens/credits.tsx
+++ b/source/views/settings/screens/credits.tsx
@@ -50,7 +50,7 @@ const Contributors = glamorous(About)({
 	textAlign: 'center',
 })
 
-const formatPeopleList = (arr) =>
+const formatPeopleList = (arr: Array<string>) =>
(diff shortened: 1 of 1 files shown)


source/views/settings/screens/credits.tsx
```diff
@@ -50,7 +50,7 @@ const Contributors = glamorous(About)({
 	textAlign: 'center',
 })
 
-const formatPeopleList = (arr) =>
+const formatPeopleList = (arr: Array<string>) =>
 	arr.map((w) => w.replace(' ', ' ')).join(' • ')
 
 export function CreditsView() {
```
drewvolz: Similarly?

```suggestion
export function CreditsView(): JSX.Element {
```

drewvolz (approved):

source/views/settings/screens/credits.tsx
hawkrives: @rye is Drew's suggestion correct?

PR #5514, opened 2021-10-29
Switch inner-text parsing to textContent from innerText

drewvolz:
* Fixes the Menu descriptions and Streams titles/descriptions
* `innerText` was creating and returning an object whereas `textContent` returns a string

### DIFF
--- a/modules/html-lib/index.ts
+++ b/modules/html-lib/index.ts
@@ -1,7 +1,7 @@
 import {parseDocument} from 'htmlparser2'
 
-import {innerText} from 'domutils'
-export {innerText}
+import {textContent} from 'domutils'
+export {textContent}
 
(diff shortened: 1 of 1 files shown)


hawkrives (approved):

hawkrives:
We should write tests too

drewvolz:
I'm happy to write tests. I'm also starting to think this should be moved to the proxy server instead of it happening client-side.

rye (approved):
I definitely agree with @hawkrives that this seems like a good candidate for testing.

hawkrives:
Let's merge this and file an issue to write tests? I'm also happy to move it to the server, assuming everything we use it on goes through there. (But it still needs tests at that point😁)

PR #5518, opened 2021-10-31
Replace datepicker component

drewvolz:
The building hours editor uses a wrapped datepicker for updating times in the report view. This PR serves to replace that with an updated dependency from the react-native-community org for editing times.

Todo:
- [x] Separate out platform specific components and props into their own files (and satisfy the types)
- [x] Verify why I manually linked for Android but not iOS
- [ ] ~Add ref back for tappable row~ Support for this does not exist upstream
- [ ] Tests

<details>
<summary>📸 WIP Screenshots</summary>

Platform | Keyboard entry | Scroll entry | Waiting for interaction
--|--|--|--
iOS | Not available on iOS | Simulator Screen Shot - iPhone 13 Pro Max - 2021-10-31 at 00 06 56 | Simulator Screen Shot - iPhone 13 Pro Max - 2021-10-31 at 00 06 49
Android | Screenshot_1635663975 | Screenshot_1635663969 | Screenshot_1635663792 

</details>


hawkrives (commented):
Nice work!

hawkrives (approved):

modules/datepicker/basepicker.tsx
```diff
@@ -0,0 +1,83 @@
...
+		if (!selectedDate) {
+			return
+		}
+
+		setDate(moment(selectedDate))
+		props.onDateChange(moment.tz(selectedDate, timezone))
+	}
+
+	const formatDate = (date: Moment): string => {
+		const {mode, format = FORMATS[mode]} = props
```
rye, line 41: This type error is new, I think.

Here I believe `mode` is getting inferred to be `undefined`.

modules/datepicker/basepicker.tsx
```diff
@@ -0,0 +1,83 @@
...
+		// We also need to make the types happy when we negate a negative
+		// MomentZone which is possibly null.
+		let dateInUnixMs = date.valueOf()
+		let momentZone = moment.tz.zone(timezone)
+
+		let offset = momentZone ? -momentZone.utcOffset(dateInUnixMs) : 0
+		return offset
+	}
+
+	let theme: AppTheme = useTheme()
```
rye: `useTheme` can return `AppTheme | undefined`. We should handle that second case, or also update `useTheme` to remove that possible return type. (If the latter, we'd also need to update the rest of the places where we might have handling.)

rye (changes requested):
A couple of new type errors I think we can probably take care of:

```diff
--- tsc-errors.master	2021-11-01 06:37:12.395516067 -0500
+++ tsc-errors.timepicker	2021-11-01 06:38:21.472368642 -0500
@@ -1,34 +1,6 @@
-modules/datepicker/datepicker.android.tsx(73,4): error TS2322: Type '"default" | "calendar" | "spinner"' is not assignable to type '"default" | "clock" | "spinner" | undefined'.
-  Type '"calendar"' is not assignable to type '"default" | "clock" | "spinner" | undefined'.
-modules/datepicker/datepicker.android.tsx(74,11): error TS2345: Argument of type '({ action, hour, minute }: TimePickerResponse) => void' is not assignable to parameter of type '(value: TimePickerAndroidOpenReturn) => void | PromiseLike<void>'.
-  Types of parameters '__0' and 'value' are incompatible.
-    Type 'TimePickerAndroidOpenReturn' is not assignable to type 'TimePickerResponse'.
-      Type 'TimePickerAndroidDismissedAction' is missing the following properties from type 'TimePickerResponse': hour, minute
-modules/datepicker/datepicker.android.tsx(97,13): error TS2345: Argument of type '({ action, year, month, day }: DatePickerResponse) => void' is not assignable to parameter of type '(value: DatePickerAndroidOpenReturn) => void | PromiseLike<void>'.
-  Types of parameters '__0' and 'value' are incompatible.
-    Type 'DatePickerAndroidOpenReturn' is not assignable to type 'DatePickerResponse'.
-      Type 'DatePickerAndroidDismissedAction' is missing the following properties from type 'DatePickerResponse': year, month, day
-modules/datepicker/datepicker.android.tsx(106,13): error TS2345: Argument of type '({ action, hour, minute }: TimePickerResponse) => void' is not assignable to parameter of type '(value: TimePickerAndroidOpenReturn) => void | PromiseLike<void>'.
-  Types of parameters '__0' and 'value' are incompatible.
-    Type 'TimePickerAndroidOpenReturn' is not assignable to type 'TimePickerResponse'.
-      Type 'TimePickerAndroidDismissedAction' is not assignable to type 'TimePickerResponse'.
-modules/datepicker/datepicker.android.tsx(113,13): error TS2345: Argument of type '({ action, year, month, day }: DatePickerResponse) => void' is not assignable to parameter of type '(value: DatePickerAndroidOpenReturn) => void | PromiseLike<void>'.
-  Types of parameters '__0' and 'value' are incompatible.
-    Type 'DatePickerAndroidOpenReturn' is not assignable to type 'DatePickerResponse'.
-      Type 'DatePickerAndroidDismissedAction' is not assignable to type 'DatePickerResponse'.
-modules/datepicker/datepicker.ios.tsx(157,16): error TS2531: Object is possibly 'null'.
-modules/datepicker/datepicker.ios.tsx(164,5): error TS2769: No overload matches this call.
-  Overload 1 of 2, '(props: ModalProps | Readonly<ModalProps>): Modal', gave the following error.
-    Type 'string[]' is not assignable to type '("portrait" | "portrait-upside-down" | "landscape" | "landscape-left" | "landscape-right")[]'.
-      Type 'string' is not assignable to type '"portrait" | "portrait-upside-down" | "landscape" | "landscape-left" | "landscape-right"'.
-  Overload 2 of 2, '(props: ModalProps, context: any): Modal', gave the following error.
-    Type 'string[]' is not assignable to type '("portrait" | "portrait-upside-down" | "landscape" | "landscape-left" | "landscape-right")[]'.
-modules/datepicker/index.tsx(8,46): error TS2307: Cannot find module './datepicker' or its corresponding type declarations.
-modules/datepicker/index.tsx(42,2): error TS2416: Property 'state' in type 'DatePicker' is not assignable to the same property in base type 'Component<Props, State, any>'.
-  Type '{ date: moment.Moment; timezone: string | undefined; }' is not assignable to type 'Readonly<State>'.
-    Types of property 'timezone' are incompatible.
-      Type 'string | undefined' is not assignable to type 'string'.
-        Type 'undefined' is not assignable to type 'string'.
+modules/datepicker/basepicker.tsx(40,33): error TS2538: Type 'undefined' cannot be used as an index type.
+modules/datepicker/basepicker.tsx(57,6): error TS2322: Type 'AppTheme | undefined' is not assignable to type 'AppTheme'.
+  Type 'undefined' is not assignable to type 'AppTheme'.
 modules/event-list/event-detail.android.tsx(85,12): error TS2322: Type 'string' is not assignable to type 'EventType'.
 modules/event-list/event-detail.android.tsx(94,16): error TS2339: Property 'message' does not exist on type 'String'.
 modules/event-list/event-detail.android.tsx(94,25): error TS2339: Property 'disabled' does not exist on type 'String'.
```

modules/datepicker/basepicker.tsx
rye: Inside `export type IOSNativeProps` (inside of the dependency), we have

```typescript
    /**
     * The date picker mode.
     */
    mode?: IOSMode;
```

and the same in `AndroidNativeProps`.

rye: If I try specifying a default,

```diff
diff --git a/modules/datepicker/basepicker.tsx b/modules/datepicker/basepicker.tsx
index b8bfc1eea..f6949bce6 100644
--- a/modules/datepicker/basepicker.tsx
+++ b/modules/datepicker/basepicker.tsx
@@ -37,7 +37,7 @@ export const BaseDateTimePicker = (
        }

        const formatDate = (date: Moment): string => {
-               const {mode, format = FORMATS[mode]} = props
+               const {mode, format = FORMATS[mode || 'date']} = props
                return moment.tz(date, timezone).format(format)
        }
```

then I get a new type error, this one a bit more serious:

```diff
+modules/datepicker/basepicker.tsx(40,25): error TS7053: Element implicitly has an 'any' type because expression of type '"time" | "date" | "datetime" | "countdown"' can't be used to index type '{ date: string; datetime: string; time: string; }'.
+  Property 'countdown' does not exist on type '{ date: string; datetime: string; time: string; }'.
```

So there are probably a couple of routes we could go.

- If we only want to allow `date`, `datetime`, and `time` types, we should encode that into the type definition:
  
  ```diff
  diff --git a/modules/datepicker/types.ts b/modules/datepicker/types.ts
  index 329e90705..796ee2bbf 100644
  --- a/modules/datepicker/types.ts
  +++ b/modules/datepicker/types.ts
  @@ -8,7 +8,8 @@ export interface BaseDatetimePickerProps {
          // dep-specific
          initialDate: Moment
          minuteInterval?: 1 | 2 | 3 | 4 | 5 | 6 | 10 | 12 | 15 | 20 | 30
  -       mode: IOSNativeProps['mode'] | AndroidNativeProps['mode']
  +       mode: (IOSNativeProps['mode'] | AndroidNativeProps['mode']) &
  +               ('date' | 'datetime' | 'time')
          format?: string
          onDateChange: (moment: Moment) => void
  ```
  
  Since doing that also removes the possibility for the value to be `undefined`, you wouldn't need to specify a variant.
  
  We could probably do something with enums or just encode the list of supported format keys as an array somewhere so we don't have to specify this inline.
  
- Otherwise, we should figure out how to make `FORMATS` keys match up one-to-one with the IOSNativeProps and AndroidNativeProps things. This would mean we have support for the other types, but would require a bit more effort, I think.

modules/datepicker/basepicker.tsx
rye: I don't think it's expected for us to ever encounter a case that `useTheme` returns `undefined`, since `setTheme` is called inside the initialization path, but TypeScript is telling us here that our code is written in such a way that if `setTheme` _wasn't_ called, then the default theme we specify in `createTheming` (currently, `undefined`, hence the `| undefined` case) would get returned instead.

- In the case of this branch, if you don't want to touch other code, you could just use an `if` to discard the possibility

  ```typescript
  let theme: AppTheme | undefined = useTheme()

  if (!theme) {
    return <></> // or some other sensible default
  }
  ```
  
- Now that I think about it, couldn't we change out `useTheme` with `getTheme`? I'm under the impression that that call actually would work here — it returns an `AppTheme` which is ultimately set by `setTheme`, so I think that might be right actually.

modules/datepicker/basepicker.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+import {BaseDatetimePickerProps} from './types'
+
+const FORMATS = {
+	date: 'YYYY-MM-DD',
+	datetime: 'YYYY-MM-DD HH:mm',
+	time: 'HH:mm',
+}
+
+export const BaseDateTimePicker = (
```
rye, line 22: More of just a curiosity: any particular reason we aren't doing a `class` or components or something?

modules/datepicker/basepicker.tsx
```diff
@@ -0,0 +1,83 @@
...
+	props: BaseDatetimePickerProps,
+): JSX.Element => {
```
rye, line 24: Is a `JSX.Element` the right thing to return here? Or should we return a `React.ReactNode`? (I'm not super familiar with the standard, so this is more just a question.)

rye (commented):
And a couple more non-blocking thoughts!

modules/datepicker/basepicker.tsx
drewvolz: Good question! This is being turned into a functional component because we can strike with a lighter hammer here. This could remain a class, for sure, but I applied a personal rule when refactoring this: I asked "if this can be done via a functional component, make it one". 

Hooks also offer a lot of niceties if we decide to optimize any bit of this in the future (e.g. useEffect, useCallback, useMemo) and allows us to improve readability by not falling into adding more old-style component lifecycle methods to accomplish the same ideas.

You did inspire another thought! I removed the `ref` which added a tappable callback to the row. It should be added back here via `useRef`.

modules/datepicker/basepicker.tsx
drewvolz: This is a pretty [comprehensive answer](https://stackoverflow.com/a/59840095) which I'll be quoting a section of.

> Why do the render methods of class components return `ReactNode`, and function components return `ReactElement`.
> 
> It's a TS type incompatibility.
> 
> * TS class component returns `ReactNode` with `render()`
> * TS function component returns `JSXElement | null`

modules/datepicker/basepicker.tsx
drewvolz: I’m super thrilled you picked up on this and appreciate the multiple avenues for resolving it.

Also I think at one point I thought about using `Pick` to choose the modes explicitly (since `Exclude` could lead to ones being added that we didn’t approve of before).

modules/datepicker/basepicker.tsx
drewvolz: Woah, I thought I used `getTheme` but you’re right, I did not!

drewvolz: There are a few modules in the app that call `useTheme` that will need what you stated above if they’re kept as-is.

rye (approved):
Thanks for fixing up those new type errors!

drewvolz:
Thanks for the thorough reviews!

PR #5519, opened 2021-11-01
Add 152x152 and 167x167 alt ipad icons

### LINKED ISSUE #5515: Missing iOS icons for iPad Pro
ITMS-90892: Missing recommended icon - The bundle does not contain an alternate app icon for iPad of exactly '167x167' pixels, in .png format for iOS versions supporting iPad Pro. To support older operating systems, the icon may be required in the bundle outside of an asset catalog. Make sure the Info.plist file includes appropriate entries referencing the file. See https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleicons

ITMS-90892: Missing recommended icon - The bundle does not contain an alternate app icon for iPad of exactly '152x152' pixels, in .png format for iOS versions >= 10.0. To support older operating systems, the icon may be required in the bundle outside of an asset catalog. Make sure the Info.plist file includes appropriate entries referencing the file. See https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleicons

drewvolz:
Hopefully closes #5515. Optimized with ImageOptim


hawkrives (commented):
Doing the new files need to be added to the asset catalog?

drewvolz:
I don't think so, I think this only relates to alternative icon sizes. It's possible the old main icon doesn't need updating in this folder and Big Ole is the only one we need to change. We'll see?

hawkrives:
Ah. Sure! Let's see what iTC has to say tonight!

hawkrives (approved):

PR #5529, opened 2021-11-02
enable sentry performance tracing

hawkrives:


hawkrives:
This enables https://docs.sentry.io/platforms/react-native/performance/

source/init/sentry.ts
```diff
@@ -7,7 +7,17 @@ function install() {
...
 
-	Sentry.init({dsn: SENTRY_DSN})
+	Sentry.init({
+		dsn: SENTRY_DSN,
+
+		tracesSampleRate: 0.2,
+
+		integrations: [
+			new Sentry.ReactNativeTracing({
+				tracingOrigins: ['localhost', 'my-site-url.com', /^\//u],
```
drewvolz: Do we need the my-site-url line?

drewvolz (approved):
Nice! One nonblocking comment

rye (approved):

source/init/sentry.ts
hawkrives: Oops, you are correct. I expect that we should change that to frogpond.tech…

PR #5574, opened 2021-11-11
Clear up more types in event-list module

drewvolz:
* Fixes add-to-calendar button (small diff, see commit https://github.com/StoDevX/AAO-React-Native/pull/5574/commits/a4fe8c8f05ff5cda80699f7099a0e224b5fcb2dd)
* Consolidates navigation and prop types
* Adds missing return types to detail views


rye (approved):
```diff
--- tsc-errors.master	2021-11-11 04:59:27.715111723 -0600
+++ tsc-errors.fix-add-to-calendar	2021-11-11 04:59:44.755197966 -0600
@@ -1,13 +1,3 @@
-modules/event-list/event-detail.android.tsx(85,12): error TS2322: Type 'string' is not assignable to type 'EventType'.
-modules/event-list/event-detail.android.tsx(94,16): error TS2339: Property 'message' does not exist on type 'String'.
-modules/event-list/event-detail.android.tsx(94,25): error TS2339: Property 'disabled' does not exist on type 'String'.
-modules/event-list/event-detail.android.tsx(94,35): error TS2339: Property 'onPress' does not exist on type 'String'.
-modules/event-list/event-detail.android.tsx(105,6): error TS2532: Object is possibly 'undefined'.
-modules/event-list/event-detail.android.tsx(106,24): error TS2532: Object is possibly 'undefined'.
-modules/event-list/event-detail.android.tsx(106,47): error TS2532: Object is possibly 'undefined'.
-modules/event-list/event-detail.ios.tsx(56,17): error TS2339: Property 'message' does not exist on type 'String'.
-modules/event-list/event-detail.ios.tsx(56,26): error TS2339: Property 'disabled' does not exist on type 'String'.
-modules/event-list/event-detail.ios.tsx(56,36): error TS2339: Property 'onPress' does not exist on type 'String'.
 modules/event-list/event-list.tsx(14,29): error TS7006: Parameter 'props' implicitly has an 'any' type.
 modules/event-list/event-list.tsx(24,14): error TS2707: Generic type 'NavigationScreenProp<S, P>' requires between 1 and 2 type arguments.
 modules/event-list/event-row.tsx(47,3): error TS2322: Type 'string | null' is not assignable to type 'string'.
```

drewvolz:
@rye I can tackle the remaining ones for this module too. Any preference if that goes in this PR or a followup?

rye:
@drewvolz, if you want to resolve the last ones, that would be fantastic! Otherwise I'm fine with this getting addressed later.

modules/event-list/types.ts
```diff
@@ -1,4 +1,22 @@
...
+export type Props = {
+	navigation: Navigation
+}
+
+export type NavigationHeaderProps = {
+	title: string
+	headerRight: JSX.Element
+}
+
+export type NullableElement = JSX.Element | null
```
hawkrives: My only comment is whether this typealias makes sense. It's short enough that I'm torn about just inlining it everywhere?

drewvolz: I've just added another one below this... If we do this often enough, are we in favor of adding a set of convenience types, or do we prefer them inlined everywhere? I understand the latter approach would involve an import to use it...

drewvolz: Ultimately I'd prefer to have a "named" type for everything to allow for more sharing and readability but I'm open to both approaches.

hawkrives: I think I would rather not have helper functions for `T | null`, at least. 

Once you start getting three options in there, it starts to make more sense, or if they're both custom types, but `| null` I'd like to inline, I think.  

@rye what do you think?

rye: I agree that `T | null` could probably best be handwritten. That makes it clear that it's a one-or-the-other choice rather than a potentially more complex set of options. This way, if you're just reading the code without any extra tools to inline types, you could easily see the two possibilities.

If we wanted to use a type to encode `| null`, I would rather have a generic.

```ts
export type Nullable<T> = T | null
```

drewvolz: Okay! I've walked these back.

hawkrives (approved):

rye (approved):

PR #5586, opened 2021-11-16
ios: Fix broken pod installs

rye:
#5581 broke the iOS build.

This commit updates the `Podfile.lock` file with the correct hash for and version of `react-native-netinfo`.


drewvolz (approved):
So quick! Thanks, @rye!

hawkrives:
How did … how did the last PR merge if the build failed?

PR #5609, opened 2021-11-23
m/fetch/cached: Remove now-unused cached fetch variant

rye:
**I'm PRing this for inclusion in `typescript-fixes`, since it's potentially controversial.** We will need to use administrator privileges to merge this due to the CI being configured by the branch. (So, ignore CI. :grin:)

I don't think we use it, and it was very broken.  I think it broke after we added it.  Some of the types in this file appear to be, like, completely broken?


hawkrives:
Mmm. Sure. I'm sure there's a nicer way of accomplishing that goal anyway.

hawkrives (approved):

drewvolz (approved):

PR #5929, opened 2022-04-18
Re-link the datetime picker for iOS and Android

drewvolz:
Fixes the crash on the platforms when trying to edit datetimes in the building hours editor.

### DIFF
--- a/android/app/src/main/java/com/allaboutolaf/MainApplication.java
+++ b/android/app/src/main/java/com/allaboutolaf/MainApplication.java
@@ -1,8 +1,8 @@
 package com.allaboutolaf;
 
 import android.app.Application;
-import android.net.http.HttpResponseCache;
 import android.content.Context;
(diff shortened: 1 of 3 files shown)


rye (approved):

hawkrives (commented):
Does it really require manual linking on android?

drewvolz:
Android is in a half-linked state prior to this changeset so the button press didn't spawn the datetime picker. We can try removing all of the android linking and see what result we get if you'd like.

hawkrives:
It's supposed to link automatically… so I would like that, if you don't mind. I tried to look at it this afternoon but got stuck on custom tabs again (which I've now forked, so we can fix it up)

drewvolz:
I have now tested removal of the manual linking. I did not see the datetime picker appear in this case.

hawkrives:
Ok. We'll leave it in.

hawkrives (approved):

PR #5947, opened 2022-04-20
Adopt iOS 14 tableview styles

drewvolz:
A sample of what this has accomplished for iOS:

~ | ~ | ~ 
--|--|--
 Simulator Screen Shot - iPhone 13 - 2022-04-19 at 23 35 06 | Simulator Screen Shot - iPhone 13 - 2022-04-19 at 23 25 59 | Simulator Screen Shot - iPhone 13 - 2022-04-19 at 23 32 16
 Simulator Screen Shot - iPhone 13 - 2022-04-19 at 23 34 08 | Simulator Screen Shot - iPhone 13 - 2022-04-19 at 23 34 42

### DIFF
--- a/modules/tableview/index.tsx
+++ b/modules/tableview/index.tsx
@@ -2,17 +2,14 @@ import * as React from 'react'
 import {Platform, StyleSheet} from 'react-native'
 import {sectionBgColor} from '@frogpond/colors'
 
-import {
-	TableView as ActualTableView,
-	Section as ActualSection,
-	Cell,
-} from 'react-native-tableview-simple'
+import * as RNTableView from 'react-native-tableview-simple'
 import type {SectionInterface} from 'react-native-tableview-simple/lib/typescript/components/Section'
 import type {TableViewInterface} from 'react-native-tableview-simple/lib/typescript/components/TableView'
+import {CellInterface} from 'react-native-tableview-simple/lib/typescript/components/Cell'
 export * from './cells'
 
 let Section = (props: SectionInterface): JSX.Element => (
-	<ActualSection
+	<RNTableView.Section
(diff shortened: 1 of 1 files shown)


modules/tableview/index.tsx
```diff
@@ -1,17 +1,34 @@
 import * as React from 'react'
+import {Platform, StyleSheet} from 'react-native'
 import {sectionBgColor} from '@frogpond/colors'
 
 import {
-	TableView,
+	TableView as ActualTableView,
 	Section as ActualSection,
```
rye: Rather than aliasing (which I realize is an existing pattern) — thoughts on renaming our wrapper type to avoid confusion? Or, maybe instead of the `Actual` prefix if we instead did something like `RNTableviewSimpleSection` to make clear what its source was?

hawkrives: I like the idea of `RNTableviewSimpleSection`. Or even 

```ts
import * as RNTableView from 'react-native-tableview-simple'
<RNTableView.Section />
```

rye: Ooo — I forgot about the second `import * as RNTableView` syntax. I definitely prefer that for namespacing. (It mirrors Rust's `<crate>::<item>` disambiguation form and should be easy to remember.)

The one thing I'm not sure about: `import * as RNTableView` or `import * as RNTableview`?

hawkrives: RN does things like `SectionList` and `MaskedViewIOS`, so my vote would be to stick with "*View is a camelcased suffix"?

drewvolz: Renamed!

hawkrives (approved):

PR #5948, opened 2022-04-20
update the appearance and copy of the hours editor

drewvolz:
Before | After
--|--
ios before | ios after
android before |  android after


drewvolz:
I can try to:
- [x] Split up android and iOS styles. The new android view doesn't look improved to me yet, it looks worse than before.
- [x] Improve the padding on the iOS week checkboxes so that they match the padding of the rows below

hawkrives:
I like both of your suggestions.

I also think that the iOS one looks a lot better already!

hawkrives:
@drewvolz This is what Google does in Maps:

hawkrives:
… so we could potentially mimic the iOS buttons (with a different color?) for Android. I think it's OK if the Open/Close remain on different rows.

drewvolz:
Gave it another go and updated the screenshots.

[Updated diff without whitespace](https://github.com/StoDevX/AAO-React-Native/pull/5948/files?diff=split&w=1)

hawkrives (approved):

PR #5966, opened 2022-04-21
Adjust workflows to verify count files

hawkrives:
1. Make the error logs use relative paths, instead of absolute (to reduce diff issues)
2. Add the list of eslint problems to a comparison file


hawkrives:
File paths strike again! 

`/Users/rives/Projects/allaboutolaf` != `/home/runner/AAO-React-Native`

I think that what I should do is to normalize all paths to be relative paths:

1. Find the absolute path to the root of the git repo
2. Run sed on the output file, replacing the git repo root with `./`
3. Profit?

hawkrives:
@drewvolz ready for review!

drewvolz (approved):
They look good to me but I want @rye to have the final say.

rye (approved):

PR #5968, opened 2022-04-22
fix lint 2

hawkrives:
more fixes!

ESLint: 446 → 381 warnings

TS 2698 eliminated!


rye (approved):

modules/food-menu/food-item-detail.tsx
```diff
@@ -53,8 +53,7 @@ export class MenuItemDetailView extends React.Component<Props> {
 					</React.Fragment>
 				) : null}
 
-				{item.nutrition_details &&
-				Object.keys(item.nutrition_details).length > 1 ? (
+				{Object.keys(item.nutrition_details ?? {}).length > 1 ? (
```
drewvolz: Is there a different way we can do this?

drewvolz (approved):
Very nice. One non-blocking comment.

modules/food-menu/food-item-detail.tsx
hawkrives: Object.keys throws on a null input… I could use _.size (which essentially does this) if you'd like. 

That'd probably be nice. I'll do that.

drewvolz (approved):

PR #5975, opened 2022-04-23
Fix-up myriad of items

drewvolz:
* remove duplicate canGoBack prop in home index
* add missing navigation hook dep to contacts list
* export props from report editor
* remove empty navigator file

Recommend [viewing without whitespace](https://github.com/StoDevX/AAO-React-Native/pull/5975/files?w=1)

---

~|~
--|--
ESLint | - 381 problems (0 errors, 381 warnings)<br>+ 380 problems (0 errors, 380 warnings)
TSC | - 339 lines <br> + 304 lines
TSC +- | -3 TS2783<br>-1 TS1208<br>-1 TS2459<br>+2 TS2783


source/views/home/index.tsx
```diff
@@ -91,7 +91,7 @@ export const NavigationKey = 'Home'
 export const NavigationOptions: NativeStackNavigationOptions = {
 	title: 'All About Olaf',
 	headerBackTitle: 'Home',
-	headerRight: (props) => <OpenSettingsButton canGoBack={true} {...props} />,
+	headerRight: (props) => <OpenSettingsButton {...props} />,
```
hawkrives, line 94: Shouldn't we put canGoBack on the right side, to ensure that it's always true?

drewvolz: I think it is already a provided prop on the right side so this would fix a warning about it being added twice. Unless I'm not understanding you correctly and you're saying to replace the value some other way?

hawkrives: Because `props.canGoBack` is a required property, it will always be set, so it will always overwrite our `canGoBack={true}` attribute.

I think I'm proposing that we set it on the right side, after `{…props}` is expanded, so that we ignore the value from `props` and always set it to be `true`.

I don't remember why we put `canGoBack=true` in the first place, but I imagine that somehow we could configure it to be false in `props`… is that really a problem, though?

```tsx
// Current:
	{headerRight: (props) => <OpenSettingsButton canGoBack={true} {...props} />,}
// Proposed:
	{headerRight: (props) => <OpenSettingsButton {...props} />,}
// Alternate Proposal:
	{headerRight: (props) => <OpenSettingsButton {...props} canGoBack={true} />,}
```

hawkrives: Okay. I've convinced myself - let's just do what you have here.

hawkrives:
@drewvolz What are the advantages/disadvantages of `opt to use built-in navigation prop instead of navigation hook`? I think that using the hook (by default) would let us avoid passing `props.navigation` around everywhere, so we may want to default to using the hook.

About useCallback: There's a lint rule somewhere about "avoid creating functions every render", which tells us to use `useCallback` so that we only recreate the function when needed. Thoughts? I think it's probably a good idea…

source/views/home/index.tsx
drewvolz: Alright. If we find a need for it we can come back to this!

drewvolz:
There looks to be at least three ways to access the navigation prop:
- useNavigation hook
- navigation prop
- ref that is passed to RootNavigation

My takeaway is that it looks like we can stick with the hook if we are happy with the hook.

[One doc says](https://reactnavigation.org/docs/use-navigation) to use the hook as you see fit...

> **Access the navigation prop from any component**
> useNavigation is a hook which gives access to navigation object. It's useful when you cannot pass the navigation prop into the component directly, or don't want to pass it in case of a deeply nested child.
> 
> useNavigation() returns the navigation prop of the screen it's inside.

...[the other doc](https://reactnavigation.org/docs/navigating-without-navigation-prop) says to try and use the prop or hook if you are able to and to avoid the ref method when possible.

> **Navigating without the navigation prop**
> Sometimes you need to trigger a navigation action from places where you do not have access to the navigation prop, such as a Redux middleware. For such cases, you can dispatch navigation actions from the navigation container.
> 
> If you're looking for a way to navigate from inside a component without needing to pass the navigation prop down, see useNavigation. **Do not** use this method when you have access to a navigation prop or useNavigation since it will behave differently, and many helper methods specific to screens won't be available.
> 
> You can get access to the root navigation object through a ref and pass it to the RootNavigation

drewvolz:
And you make a good point about keeping useCallback in here. I will add that back.

hawkrives:
Reading those two snippets, it sounds like we should avoid the ref method (okay, we can do that!). I think we should stick with the hook, for consistency, so that we don't have a separate way of doing things for "root components" vs nested components.

hawkrives (approved):

rye (approved):

hawkrives:
I merged in `master` to fix the error count files

PR #5988, opened 2022-04-25
[Viewport] Animated searchbar

drewvolz:
* replace viewport with useWindowDimensions
* convert class to functional component
* add useEffect and useCallback hooks

[View without whitespace](https://github.com/StoDevX/AAO-React-Native/pull/5988/files?w=1)


modules/searchbar/animated.tsx
```diff
@@ -16,30 +16,34 @@ type Props = {
...
-	searchBarTopSpec = {start: 71, end: 10, duration: 200}
-	containerHeightSpec = {start: 125, end: 64, duration: 200}
+	let headerOpacitySpec: AnimatedValueType = React.useMemo(() => {
+		return {start: 1, end: 0, duration: 200}
+	}, [])
 
-	headerOpacity = new Animated.Value(this.headerOpacitySpec.start)
-	searchBarTop = new Animated.Value(this.searchBarTopSpec.start)
-	containerHeight = new Animated.Value(this.containerHeightSpec.start)
+	let searchBarTopSpec: AnimatedValueType = React.useMemo(() => {
```
hawkrives: For these three, can we just pull them out into constants?

modules/searchbar/animated.tsx
```diff
@@ -16,30 +16,34 @@ type Props = {
...
+	let searchBarTop = React.useMemo(() => {
+		return new Animated.Value(searchBarTopSpec.start)
+	}, [searchBarTopSpec.start])
+
+	let containerHeight = React.useMemo(() => {
+		return new Animated.Value(containerHeightSpec.start)
+	}, [containerHeightSpec.start])
 
-	animate = (
+	let animate = (
```
hawkrives: Let's pull this out, too, into a function in the file scope.

modules/searchbar/animated.tsx
```diff
@@ -51,87 +55,91 @@ export class AnimatedSearchBar extends React.Component<Props> {
...
+
+	let searchStyle = [
+		styles.searchBarWrapper,
+		{width: searchBarWidth},
+		{top: showTitle ? searchBarTop : searchBarTopSpec.end},
+	]
+
+	let headerStyle = [
+		styles.header,
+		{
```
hawkrives: Prettier can collapse these if we collapse them now - the line is short enough.

modules/searchbar/animated.tsx
```diff
@@ -51,87 +55,91 @@ export class AnimatedSearchBar extends React.Component<Props> {
...
-		)
-	}
+	let searchBarWidth = viewport.width - 20
+
+	let showTitle = Boolean(props.title)
+
+	let containerStyle = [
+		styles.searchContainer,
+		styles.common,
+		{
```
hawkrives: (Collapse this literal, please)

modules/searchbar/types.ts
```diff
@@ -1,14 +1,22 @@
...
 	backButtonAndroid?: 'search' | boolean
 	backgroundColor?: string
-	onCancel: () => any
-	onChange: (data: string) => any
-	onFocus: () => any
-	onSubmit: () => any
+	onCancel: () => unknown
+	onChange: (data: string) => unknown
+	onFocus: () => unknown
+	onSubmit: () => unknown
```
hawkrives, line 11: Oh by the way. I know I asked for these to be "unknown" the other day, but I've changed my mind (sorry 😅) - I read a post that mentioned that by accepting any return value, you're not using the type checker to its full power, because you may be accidentally returning something. 

So let's switch to "void", and if we need to pass in an async function, we'll cross that bridge when we come to it!

modules/searchbar/animated.tsx
```diff
@@ -1,9 +1,9 @@
 import * as React from 'react'
 
-import {StyleSheet, Animated} from 'react-native'
+import {StyleSheet, Animated, useWindowDimensions} from 'react-native'
 import {white} from '@frogpond/colors'
 import {SearchBar} from './searchbar'
-import {Viewport} from '@frogpond/viewport'
+import {AnimatedValueType} from './types'
 
 type Props = {
```
hawkrives, line 8: Can we replace this with an import of the Props type from types.ts?

hawkrives (dismissed):
Thanks for doing this!

modules/searchbar/animated.tsx
drewvolz: It's not a 1-1 match at the moment, I think there was a prop missing? But I will see what I can do.

modules/searchbar/animated.tsx
```diff
@@ -1,137 +1,104 @@
...
-			if (this.props.active) {
-				this.activateSearch()
-			} else {
-				this.deactivateSearch()
-			}
-		}
-	}
+const headerOpacity = new Animated.Value(headerOpacitySpec.start)
+const searchBarTop = new Animated.Value(searchBarTopSpec.start)
+const containerHeight = new Animated.Value(containerHeightSpec.start)
```
hawkrives, line 18: Does this work? I worry that the instances of Animated.Value need to be inside the render, or they'll be shared between instances of the component?

drewvolz: I'll check if it works. If it does not work, we will need to revert back to the hooks version of these, then.

hawkrives: Let's do that for these three?

drewvolz:
Closing in favor of #6024 

1. Close this `AnimatedSearchbar` refactor PR 
2. Merge the set of viewport changes (noting it will remain for `AnimatedSearchbar`)
3. Open a new PR nixing `AnimatedSearchbar` and adoption of navigation's searchbar

PR #6021, opened 2022-05-01
building hours and student orgs tweaks

hawkrives:
I made a `useMomentTimer` hook! It replaces the common uses of the `<Timer/>` component. Additionally, it can round a moment to the start of the minute, so Hours/Bus can just use that instead of rolling their own code. It won't cause renders more often than you ask for, either.

I switched a few functions in StudentOrgs to memoize… it feels really wasteful to re-compute the `orgToArray` output on every keystroke.

I also fixed `keyExtractor` in StudentOrgs, so that it's not incorporating the index anymore. That should help with perf too.

I'm slowly switching these components to use a `useStudentOrgs`-style data fetching hook, so that they have a single function (returning `{data, isInitial, isPending, error, reload()}`) that we can swap out in the future if we need to switch from e.g. async on-demand network requests to redux-backed with background loading or whatever.

I switched StoPrintErrorView to a function component.

Over in Hours, I switched a `_.flatten` for an Array.prototype.flatMap, and got rid of an `any` in a helper function.

The big change, though, is that I got to merge StatefulHoursList and HoursList together!

```tsx
export function BuildingHoursView(): JSX.Element {

	let {data: buildings, isPending, isInitial, reload} = useBuildingHours()
	let favoriteBuildings = useSelector(
		(state: ReduxState) => state.buildings?.favorites ?? [],
	)

	let grouped = React.useMemo(
		() => groupBuildings(buildings, favoriteBuildings),
		[buildings, favoriteBuildings],
	)

	let {now} = useMomentTimer({intervalMs: 60000, startOf: 'minute'})

	let onPressRow = React.useCallback(
		(building: BuildingType) =>
			navigation.navigate('BuildingHoursDetail', {building}),
		[navigation],
	)

	if (isInitial) {
		return <LoadingView />
	}

	return (
		<SectionList
			ListEmptyComponent={<NoticeView text="No hours." />}
			contentContainerStyle={styles.container}
			keyExtractor={(item) => item.name}
			onRefresh={reload}
			refreshing={isPending && !isInitial}
			renderItem={({item}) => (
				<BuildingRow info={item} now={now} onPress={() => onPressRow(item)} />
			)}
			renderSectionHeader={({section: {title}}) => (
				<ListSectionHeader title={title} />
			)}
			sections={grouped}
		/>
	)
}
```

Look at that simplicity 😇 I got to delete the entire 113-line stateful-list file!


source/views/building-hours/list.tsx
```diff
@@ -20,51 +25,85 @@ const styles = StyleSheet.create({
...
+const emptyList: ReadonlyArray<never> = []
+
+function groupBuildings(
+	buildings: Array<BuildingType> | void,
+	favorites: Array<string>,
+) {
+	if (!buildings) {
+		return emptyList
+	}
+
```
drewvolz, line 43: I think we may need to either change the return type for what group returns or what emptyList is returning.

At line 43 we see the following:

```log
Unexpected any. Specify a different type
```

The return signature for `groupBuildings` is

```js
readonly never[] | {  title: string;
    data: BuildingType[];
}[]
```

source/views/building-hours/list.tsx
```diff
@@ -20,51 +25,85 @@ const styles = StyleSheet.create({
...
+	let favoritesGroup = {
+		title: 'Favorites',
+		data: buildings.filter((b) => favorites.includes(b.name)),
```
drewvolz, line 46: `buildings.filter` is throwing the following error

```console
TypeError: buildings.filter is not a function. (In 'buildings.filter(function (b) {
        return favorites.includes(b.name);
      })', 'buildings.filter' is undefined)
```

source/views/building-hours/lib/get-day-of-week.ts
```diff
@@ -2,4 +2,4 @@ import type {Moment} from 'moment-timezone'
 import type {DayOfWeekEnumType} from '../types'
 
 export const getDayOfWeek = (m: Moment): DayOfWeekEnumType =>
-	m.format('dd') as any as DayOfWeekEnumType
+	m.format('dd') as unknown as DayOfWeekEnumType
```
drewvolz: We may be able to get away with this?

```suggestion
	m.format('dd') as DayOfWeekEnumType
```

source/views/building-hours/lib/get-detailed-status.ts
```diff
@@ -7,10 +7,16 @@ import {isChapelTime, formatChapelTime} from './chapel'
...
+	isActive: boolean
+	label: string | null
+	status: string
+}
+
 export function getDetailedBuildingStatus(
 	info: BuildingType,
 	m: Moment,
-): Array<{isActive: boolean; label: string | null; status: string}> {
+): BuildingStatus[] {
```
drewvolz, line 19: Nice!

source/init/lodash-memoize.ts
```diff
@@ -0,0 +1,3 @@
+import memoize from 'lodash/memoize'
+
+memoize.Cache = WeakMap
```
drewvolz: When filtering the student orgs list via the searchbar we received a red screen with the following message:

```console
TypeError: Attempted to set a non-object key in a WeakMap
```

drewvolz (commented):
This is a huge improvement. I _really_ like where this has taken things! 

#### 🎉  For hours, orgs, and the timing component
- huge improvements in terms of reducing boilerplate:
  - (react-redux) `useSelector`, 
  - (react-async) `useFetch`
- huge improvements in terms of cutting down computations and re-rendering:
  - `useMemo` in the grouping logic

#### 💬  Just a few comments about compiling/running:
- [x] ~left a comment where building hours appears to error out~
- [x] orgs list/detail renders but filtering may have an issue with WeakMap
  - idea: this may be solvable with the course search approach)

source/init/lodash-memoize.ts
hawkrives: gdi I forget that weakmaps can't do primitives for keys. We'll have to revert this change.

source/views/building-hours/lib/get-day-of-week.ts
hawkrives: I don't _think_ so, because you can't go straight from "string" to "DayOfWeekEnum"

modules/timer/hook.ts
```diff
@@ -0,0 +1,44 @@
...
+import {useInterval} from './use-interval'
+
+type BasicProps = {
+	intervalMs: number // ms
+}
+
+type MomentProps = BasicProps & {
+	timezone?: string
+	startOf?: unitOfTime.StartOf
+}
```
rye: Could we use `interface BasicProps` and then `interface MomentProps extends BasicProps`? I think that's equivalent here.

source/views/building-hours/lib/get-day-of-week.ts
drewvolz: I think this is actually okay to do without type erasure. Typescript seems to be happy with it.

source/init/lodash-memoize.ts
drewvolz: What about setting it the way course search sets it?

https://github.com/StoDevX/AAO-React-Native/blob/8b5845dc921b79d802eece2fcb032f85627feee6/source/views/sis/course-search/list.tsx#L57-L60

source/views/building-hours/list.tsx
```diff
@@ -62,7 +62,13 @@ function groupBuildings(
 export function BuildingHoursView(): JSX.Element {
 	let navigation = useNavigation()
 
-	let {data: buildings, isPending, isInitial, reload} = useBuildingHours()
+	let {
+		data: {data: buildings} = [],
```
drewvolz: I've introduced a type error as a result of this destructuring –– tracked in commit https://github.com/StoDevX/AAO-React-Native/pull/6021/commits/5d6d4aed8300dd2022ce79f8202479aea386f22a.

hawkrives: We'll need to edit the return type of useBuildingHours to say `{data: BuildingHour[]}`, I think

drewvolz (approved):
Thanks @hawkrives!

rye (commented):
One comment, otherwise looks good!

source/views/building-hours/lib/get-day-of-week.ts
rye: I think you should be able to tell Typescript that `string` is `DayOfWeekEnumType` — I don't think that's disallowed:

[Playground](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABHADgQwI4gKYAoCUAXIgM5QBOMYA5ogN4BQizi52UI5SAyhVdQDpg5OAFsAwgAs05cXAAmeAEQABJQIjTZC7AEEouAAz5EAakQBGfAG4GAXwYMoATxTZEAUTAhRAFVfuALyISrpKiAA+IQBCSo4ANuyIAEbEXj7+bojBqJg4BIhoJJ7efgG2DBAIJHCJAvFw1LjJNkA).

For strict enum types, I do think it's restricted.

rye (approved):

PR #6025, opened 2022-05-02
Orgs: native searchbar (part 1)

drewvolz:
[View without whitespace](https://github.com/StoDevX/AAO-React-Native/pull/6025/files?w=1)

## Changeset

* Removes dependence upon `@frogpond/searchbar` in the student orgs list.
* Replaces the searchbar with one provided by [react-navigation](https://reactnavigation.org/docs/native-stack-navigator#onchangetext).
* Adds a convenience type for the `onChangeText` type.
* Overrides the searchbar color in the configuration (default is the app's accent color, so darker orange on iOS).
* Adds `contentInsetAdjustmentBehavior="automatic"` to `SectionList` definitions. The docs mention to do so or else the searchbar cuts off the top of the list.

`navigation/types.tsx`

```ts
// convenience type for aliasing the native event
export type OnChangeTextType = {
    nativeEvent: {text: React.SetStateAction<string>}
}
```

`student-orgs/list.tsx`
```ts
React.useLayoutEffect(() => {
    navigation.setOptions({
        headerSearchBarOptions:
            // While we could configure the searchbar options from the top-level routes.tsx... 
            // it appears that every view in that stack would gain a searchbar from setting properties
            // like the background color. Instead we can opt to configure it on each view that needs one.
            barTintColor: white,
           
            // Callback to get the value that's within the search box into the query variable
	    onChangeText: (event: OnChangeTextType) => setQuery(event.nativeEvent.text),
    })
}, [navigation])
```

<details>
<summary>📸 Video</summary>

orgs

</details>

## Quirks

* Both orgs and dictionary will occasionally throw the following message, even though we are using `useLayoutEffect`
```console

```

* Student orgs searchbar briefly flickers on init. The dictionary does not have this issue. I don't know why.

* The searchbar is hidden by default until you scroll, but it's not necessarily a bad thing.


source/navigation/types.tsx
```diff
@@ -57,6 +58,10 @@ export type RootStackParamList = {
 	PrintJobRelease: undefined
 }
 
+export type OnChangeTextType = {
+	nativeEvent: {text: React.SetStateAction<string>}
+}
+
```
rye: Thoughts on making this an `interface` instead of a named `Object` type?

We don't currently have a standard one way or the other, but I thought I'd check.

rye: Actually, looking at how this is used, I wonder if we should split this to to make this more widely-usable?

```typescript
export interface ChangeTextEvent {
    // ...
}
```

```typescript
export interface OnChangeTextHandler {
    onChange: (event: ChangeTextEvent) => void
}
```

drewvolz: Sounds perfect, adding this.

hawkrives (approved):

rye (approved):

PR #6026, opened 2022-05-02
Dictionary: native searchbar (part 2)

drewvolz:
**targets `drew/native-searchbar-orgs` branch**

[View without whitespace](https://github.com/StoDevX/AAO-React-Native/pull/6026/files?w=1)

## Changeset

[new]
* Connects dictionary back to the home view
* Dictionary editor and detail views have been rewritten to be functional components

[repeated from part 1]
* Removes dependence upon `@frogpond/searchbar` in the dictionary list.
* Replaces the searchbar with one provided by [react-navigation](https://reactnavigation.org/docs/native-stack-navigator#onchangetext).
* Overrides the searchbar color in the configuration (default is the app's accent color, so darker orange on iOS).
* Adds `contentInsetAdjustmentBehavior="automatic"` to `SectionList` definitions. The docs mention to do so or else the searchbar cuts off the top of the list.

`dictionary/list.tsx`

```ts
React.useLayoutEffect(() => {
    navigation.setOptions({
        headerSearchBarOptions:
            // While we could configure the searchbar options from the top-level routes.tsx... 
            // it appears that every view in that stack would gain a searchbar from setting properties
            // like the background color. Instead we can opt to configure it on each view that needs one.
            barTintColor: white,
           
            // Callback to get the value that's within the search box into the query variable
	    onChangeText: (event: OnChangeTextType) => setQuery(event.nativeEvent.text),
    })
}, [navigation])
```
<details>
<summary>📸 Video</summary>

dictionary

</details>

## Quirks

[repeated from part 1]

* Both orgs and dictionary will occasionally throw the following message, even though we are using `useLayoutEffect`
```console

```

* Student orgs searchbar briefly flickers on init. The dictionary does not have this issue. I don't know why.

* The searchbar is hidden by default until you scroll, but it's not necessarily a bad thing.


source/views/dictionary/report/editor.tsx
```diff
@@ -94,7 +80,7 @@ const TitleCell = ({text, onChange = () => {}}: TextFieldProps) => (
 	/>
 )
 
-const DefinitionCell = ({text, onChange = () => {}}: TextFieldProps) => (
+const DefinitionCell = ({text, onChange = () => noop}: TextFieldProps) => (
```
hawkrives:
```suggestion
const DefinitionCell = ({text, onChange = noop}: TextFieldProps) => (
```

source/views/dictionary/report/editor.tsx
```diff
@@ -10,80 +10,66 @@ import {
...
+	)
 }
 
-type TextFieldProps = {text: string; onChange: (text: string) => any}
+export {DictionaryEditorView as View}
+
+type TextFieldProps = {text: string; onChange: (text: string) => void}
 
-const TitleCell = ({text, onChange = () => {}}: TextFieldProps) => (
+const TitleCell = ({text, onChange = () => noop}: TextFieldProps) => (
```
hawkrives:
```suggestion
const TitleCell = ({text, onChange = noop}: TextFieldProps) => (
```

source/views/dictionary/report/editor.tsx
```diff
@@ -10,80 +10,66 @@ import {
...
 
-					<Section header="DEFINITION">
-						<DefinitionCell
-							onChange={this.onChangeDefinition}
-							text={definition}
-						/>
-					</Section>
+				<Section header="DEFINITION">
+					<DefinitionCell
+						onChange={(newDefinition) => setDefinition(newDefinition)}
```
hawkrives: How does this read? Too confusing?

```suggestion
						onChange={setDefinition}
```

source/views/dictionary/report/editor.tsx
```diff
@@ -10,80 +10,66 @@ import {
...
+				title="Thanks for spotting a problem!"
+			/>
 
-				<TableView>
-					<Section header="WORD">
-						<TitleCell onChange={this.onChangeTitle} text={term} />
-					</Section>
+			<TableView>
+				<Section header="WORD">
+					<TitleCell onChange={(newTitle) => setTerm(newTitle)} text={term} />
```
hawkrives: How does this read? Too confusing?

```suggestion
					<TitleCell onChange={setTerm} text={term} />
```

source/views/dictionary/report/editor.tsx
```diff
@@ -10,80 +10,66 @@ import {
 import {submitReport} from './submit'
 import type {WordType} from '../types'
 import type {TopLevelViewPropsType} from '../../types'
+import {NativeStackNavigationOptions} from '@react-navigation/native-stack'
+import noop from 'lodash/noop'
 
 type Props = TopLevelViewPropsType & {
```
hawkrives, line 14:
```suggestion
type Props = {
```

source/views/dictionary/list.tsx
```diff
@@ -94,7 +104,7 @@ export function DictionaryView(props: Props) {
 	let renderRow = ({item}: {item: WordType}) => (
```
hawkrives, line 94: Let's inline this, so we can infer the argument type

source/views/dictionary/detail.tsx
```diff
@@ -26,37 +29,46 @@ const styles = StyleSheet.create({
...
+	return (
+		<Container>
+			<Term selectable={true}>{item.word}</Term>
+			<Markdown
+				source={item.definition}
+				styles={{Paragraph: styles.paragraph}}
+			/>
+
+			<Button
+				onPress={() => handleEditButtonPress(item)}
```
hawkrives:
```suggestion
				onPress={handleEditButtonPress}
```

source/views/dictionary/detail.tsx
```diff
@@ -26,37 +29,46 @@ const styles = StyleSheet.create({
...
-				<Button onPress={this.handleEditButtonPress} title="Suggest an Edit" />
-
-				<ListFooter title="Collected by the humans of All About Olaf" />
-			</Container>
-		)
-	}
+	let navigation = useNavigation()
+
+	let handleEditButtonPress = React.useCallback(
+		(item: WordType) =>
```
hawkrives:
```suggestion
		() =>
```

source/views/dictionary/detail.tsx
```diff
@@ -26,37 +29,46 @@ const styles = StyleSheet.create({
...
+			navigation.navigate('DictionaryEditor', {
+				item: item,
+			}),
```
hawkrives:
```suggestion
			navigation.navigate('DictionaryEditor', {item}),
```

source/views/dictionary/detail.tsx
```diff
@@ -26,37 +29,46 @@ const styles = StyleSheet.create({
...
+}
 
-	handleEditButtonPress = () => {
-		let item = this.props.navigation.state.params.item
-		this.props.navigation.navigate('DictionaryEditorView', {
-			word: item,
-		})
-	}
+export let DictionaryDetailView = (props: Props): JSX.Element => {
+	let [item] = React.useState<WordType>(props.route.params.item)
```
hawkrives:
```suggestion
	let item = props.route.params.item
```

source/views/dictionary/list.tsx
```diff
@@ -104,32 +114,32 @@ export function DictionaryView(props: Props) {
...
 	)
 }
-DictionaryView.navigationOptions = {
+
+export {DictionaryView as View}
+
+export const NavigationOptions: NativeStackNavigationOptions = {
 	title: 'Campus Dictionary',
-	headerBackTitle: 'Dictionary',
+	headerBackTitle: 'Back',
```
hawkrives, line 144: Why?

source/views/dictionary/detail.tsx
```diff
@@ -26,37 +29,46 @@ const styles = StyleSheet.create({
...
-		return {
-			title: navigation.state.params.item.word,
-		}
+export const DetailNavigationOptions = (props: {
+	route: RouteProp<RootStackParamList, 'DictionaryDetail'>
+}): NativeStackNavigationOptions => {
+	let {item} = props.route.params
+	return {
+		title: item.word,
+		headerBackTitle: 'Dictionary',
```
hawkrives, line 41: Why is this "Back" sometimes and "Dictionary" sometimes?

hawkrives (dismissed):
Some thoughts- I like it overall!

source/views/dictionary/report/editor.tsx
drewvolz: I did consider that syntax to inline that. What do you think, @rye?

source/views/dictionary/detail.tsx
drewvolz: Are you wanting to avoid using state here?

source/views/dictionary/list.tsx
drewvolz: I can't find the exact documentation I read but the API is (re)written such that the navigation options reflect the current route and the current route only.

If we were to keep "Dictionary" as the back title, we would push to the dictionary list, and the back title would be "Dictionary" which doesn't reflect that back is going to the Home view.

hawkrives: Okay, I was wondering if that was the case!

source/views/dictionary/detail.tsx
hawkrives: Yes, because we never do anything with it.

source/views/dictionary/report/editor.tsx
rye: For singleton `handler={(arg) => someFunction(arg)}` situations like this, I definitely think the `handler={someFunction}` is a good way to go, especially since that's a `useState` delegate, which isn't likely to sprout a new parameter any time soon.

source/views/dictionary/report/editor.tsx
rye: To expand on this: there is no need to extend or try and mix the `TopLevelViewPropsType` into this type. Anything that satisfies the contract of `{route: {params: {item: WordType}}}` would fit here.

There is another pattern we could follow with interfaces, but this is essentially equivalent.

source/views/dictionary/list.tsx
```diff
@@ -53,12 +53,22 @@ const styles = StyleSheet.create({
 	},
 })
 
-type Props = TopLevelViewPropsType
-
-export function DictionaryView(props: Props) {
+function DictionaryView(): JSX.Element {
```
rye, line 56: Thoughts on `JSX.Element` vs `React.ReactNode`? I'd rather go with whatever precedent we have, but I'm not sure what it is :)

source/views/dictionary/report/editor.tsx
drewvolz: Yeah! This is leftover from the non-hook version of navigation and I missed removing it from this file.

source/views/dictionary/list.tsx
drewvolz: This is what I have been relying on as far as a [comprehensive answer](https://stackoverflow.com/a/59840095).

> Why do the render methods of class components return `ReactNode`, and function components return `ReactElement`.
> 
> It's a TS type incompatibility.
> 
> * TS class component returns `ReactNode` with `render()`
> * TS function component returns `JSXElement | null`

hawkrives (approved):

rye (approved):

PR #6037, opened 2022-05-06
add eslint-plugin-jsx-expressions

drewvolz:
There are similar eslint rules out there but this is one of the few rules the perform type checking on the arguments in the expression. There [is an example/explanation](https://github.com/hpersson/eslint-plugin-jsx-expressions/blob/master/docs/rules/strict-logical-expressions.md
) of what would be caught here. I have set this to be a warning for now. Should we consider this an error in the future?

---

> ### Strict JSX expressions for safe conditional renders. (strict-logical-expressions)
>
> Forbids the usage of potentially falsey string or number values in logical && expressions. Oftentimes, these logical expressions are used in jsx to enable conditional rendering behavior.
> 
> However, one can encounter unexpected behavior when a 0 or empty string values is returned as the result of the expression and rendered accidentally.

Examples of incorrect code for this rule:

```ts
// Potentially falsey strings are not allowed
let str
<App>{str && <Foo />}</App>

// Potentially falsey numbers are not allowed
let num
<App>{num && <Foo />}</App>

// Includes types that may be a string or number
let thisOrThat: Record<any, any> | string | number
<App>{thisOrThat && <Foo />}</App>
```

Examples of correct code for this rule:

```ts
// Coalescing to boolean is ok
let str = "Foo"
<App>{!!str && <Foo />}</App>

// Turning into a boolean is ok
let str = "Foo"
<App>{Boolean(str) && <Foo />}</App>

// Constant values are ok
const str = "Foo"
<App>{str && <Foo />}</App>
```

Options

```js
type Options = [
  {
    allowString?: boolean;
    allowNumber?: boolean;
  }
]

defaultOptions: [
  {
    allowString: false,
    allowNumber: false,
  },
]
```

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -8,6 +8,7 @@ parserOptions:
 plugins:
   - '@typescript-eslint'
   - babel
+  - jsx-expressions
   - react
   - react-native
   - react-hooks
@@ -119,3 +120,6 @@ rules:
   # TODO(rye): Follow-up: DENY these
   '@typescript-eslint/explicit-module-boundary-types': warn
   '@typescript-eslint/no-empty-function': warn
+
+  # eslint-plugin-jsx-expressions – https://github.com/hpersson/eslint-plugin-jsx-expressions/
+  jsx-expressions/strict-logical-expressions: warn
--- a/modules/tableview/cells/multi-line-detail.tsx
+++ b/modules/tableview/cells/multi-line-detail.tsx
@@ -17,7 +17,7 @@ export function MultiLineDetailCell(props: Props): JSX.Element {
 				<Text allowFontScaling={true} style={styles.cellTitle}>
 					{title}
 				</Text>
-				{leftDetail && (
+				{Boolean(leftDetail) && (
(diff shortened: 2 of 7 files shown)


hawkrives:
Let's make it an error today!

.eslintrc.yaml
```diff
@@ -119,3 +120,6 @@ rules:
   # TODO(rye): Follow-up: DENY these
   '@typescript-eslint/explicit-module-boundary-types': warn
   '@typescript-eslint/no-empty-function': warn
+
+  # eslint-plugin-jsx-expressions – https://github.com/hpersson/eslint-plugin-jsx-expressions/
+  jsx-expressions/strict-logical-expressions: warn
```
rye, line 125: I agree with @hawkrives — let's make this an error today. Also, missing a newline at the end of this file

rye (approved):

hawkrives (approved):

.eslintrc.yaml
drewvolz: ✅ Error today 
✅ Newline at end of file

Issue #6024, opened 2022-05-02
Use react-navigation searchbars


drewvolz:
- [x] #6025
- [x] #6026
- [x] #6087 
- [x] #3558 
- [x] #6110

drewvolz:
We may have to refactor course search's recent search page (the index/entry into course search) for a few reasons.

First, we do not have a list that the searchbar can attach and insert itself into through the navigation hooks. The current list displays list items without being wrapped by a traditional list.

Second, the searchbar automatically hides itself. There is a prop for android to focus it on load but not on iOS. This could feel strange on a very search oriented view. This leads me to ask if (a) are we OK with that? and (b) would we prefer to remove the recent search view if not? or (c) ...keep the old searchbar and all of its deps?

hawkrives:
I vote we make it work like the bottom of Photos' search page

Bar in the toolbar, then just the list of recent search terms. 

We can do other fancy stuff in the future

drewvolz:
Okay so that means to me that we
- [x] ~Add a SectionList (or two FlatList) to the course-search's recent item list component (it's not currently wrapped)~
  - Update: Issue was with tab navigator. Explanation below in "Challenge 1" as to why we don't need a List to search.
- [x] Add searchbar hook

I can try to see what initialization looks like. I think with no recent searches it may show us the searchbar and with data it may hide the searchbar. I will see what happens.

drewvolz:
Making this work for nested tab navigators was a bit confusing at first but the documentation has been helpful in figuring some of this out.

## Challenge 1
SIS top-level recent searches tab has this setup.

✅ When we are one-level deep in a tab navigator, attaching a search bar to the view will work fine. 
See: Recent course search --> Search results

⚠️  When we are top-level inside a tab navigator, attaching a search bar to the view will not work.
See: Tab.Navigator > Tab.Screen --> Course search

### Solution for challenge 1

That second part was causing the search bar to not appear until I came across react navigation's documentation for `navigation.getParent` as a workaround.

> ### getParent
> This method returns the navigation prop from the parent navigator that the current navigator is nested in. If you have a stack navigator and a tab navigator nested inside the stack, then you can use [`getParent`](https://reactnavigation.org/docs/navigation-prop#getparent) inside a screen of the tab navigator to get the navigation prop passed from the stack navigator.

```tsx
navigation.getParent()?.setOptions({
    headerSearchBarOptions: {
      ...
    }
})
```

[...]

PR #6038, opened 2022-05-07
Fix searchbar with large lists with VirtualizedList optimizations

### LINKED ISSUE #6035: Investigate searchbar jitteriness on PTR
Continuing #6024, we could improve the UX of the new native searchbar.

Ideas:
- Look to see if safearea is causing the flicker like we saw with the tableview
- Is this just PTR?
- Are we refreshing our loading/refreshing variables too quickly?
- Is this caused by the warning that we see on load?

```

```

drewvolz:
## Issue

When using pull-to-refresh in Dictionary or Student Orgs with the new searchbar, the `SectionList` behaves strangely. The navigation header with the embedded searchbar will become jittery and it appears as if performance is taking a hit.

- Fixes #6035 
- Part of #6024 
- Related to #3348

## Investigation

I initially thought that the list rendering issues might be related to calculating the heights of components. While this did not produce a solution, it pointed me further towards what may have been happening.

My test was to take the `contentContainerStyle` of the list and modify the style away from `flexGrow` to see if we were having a height calculation issue.

`student-orgs/list.tsx`

```diff
contentContainer: {
-   flexGrow: 1,
+   flex: 1,
},
```

Testing this out, you'll see that this test trades one problem for another. While this fixes our issue with pull-to-refresh, it makes the height of the list equal to the screen's height which is not a useful change in practice. 

So maybe we're dealing with a list performance issue since we can get both the searchbar and the list to play nice with a smaller sized list.

## Solution

I stumbled upon [a github issue](https://github.com/facebook/react-native/issues/21468) from 2018 in the react-native repo titled 'sticky headers scrolls jumpy when virtualization is enabled'. This tied together several things we were seeing –– a virtualized sectionlist, sticky headers, and jumpy behavior when scrolling a large list.

I thought it seemed familiar. It turns out that we've been here before! #3348 in the all about olaf repo centered around issues with large lists, jumpiness, and section headers in the course search view (to which we [applied an upstream patch](https://github.com/StoDevX/AAO-React-Native/pull/3357) to the list virualizer and called it done).

This solution comes [from this comment](https://github.com/facebook/react-native/issues/21468#issuecomment-533214634):

> Hey, I was running into this too and found a workaround.  I ended up setting a few additional properties, these in particular:
> 
> ```js
>      initialNumToRender={50}
>      maxToRenderPerBatch={50}
>      windowSize={41}
> ```
> 
> I opted out of RN's defaults (because I am dealing with a large list) and this seems to fix the headers from jumping all over when scrolling.  Your mileage may vary.

Looking at this with fresh eyes, we can optimize our larger virtualized lists with props (not inclusive) such as `initialNumToRender`, `maxToRenderPerBatch`, and `windowSize`. See [optimizing flatlist configuration](https://reactnative.dev/docs/optimizing-flatlist-configuration) for the full documentation.

While the suggested increases from the react native defaults are a heuristic, these are not specially selected numbers for each of our virtualized lists... yet. These appear to work for larger lists such as the orgs and dictionary lists, but will it work with course search? Unknown, but this solution seems extensible and inline-able if we find sharing this pattern across other lists needs tweaking.

### [initialNumToRender](https://reactnative.dev/docs/sectionlist#initialnumtorender)

default: 10
override: 50

How many items to render in the initial batch. This should be enough to fill the screen but not much more. Note these items will never be unmounted as part of the windowed rendering in order to improve perceived performance of scroll-to-top actions.

### [maxToRenderPerBatch](https://reactnative.dev/docs/optimizing-flatlist-configuration#maxtorenderperbatch)

default: 10
override: 50

It is a VirtualizedList prop that can be passed through FlatList. This controls the amount of items rendered per batch, which is the next chunk of items rendered on every scroll.

Pros: Setting a bigger number means less visual blank areas when scrolling (increases the fill rate).

Cons: More items per batch means longer periods of JavaScript execution potentially blocking other event processing, like presses, hurting responsiveness.

### [windowSize](https://reactnative.dev/docs/optimizing-flatlist-configuration#windowsize)

default: 21
override: 41

The number passed here is a measurement unit where 1 is equivalent to your viewport height. The default value is 21 (10 viewports above, 10 below, and one in between).

Pros: Bigger windowSize will result in less chance of seeing blank space while scrolling. On the other hand, smaller windowSize will result in fewer items mounted simultaneously, saving memory.

Cons: For a bigger windowSize, you will have more memory consumption. For a lower windowSize, you will have a bigger chance of seeing blank areas.


hawkrives (approved):

rye (approved):
Nice write-up!

PR #6042, opened 2022-05-08
Developer tools

drewvolz:
Because this involves the developer tools, the files look unrelated but are pulled in because of the settings views.

- [x] Bon Appetit Dev Picker
  - [x] Updates the down chevron icons (they switched to arrows)
  - [x] Detail / Nutrition view
- [x] API Tester
  - [x] Text response
  - [x] Parsed response (text view)
  - [ ] Parsed response detail (uses DebugView –– which has navigation issues)
- [ ] DebugView (set to disabled to prevent crash on-tap)
  - [ ] Convert to a functional component and use the navigation hook
  - [ ] Figure out how to get the params with the new navigator for the debug data
- [x] `isDisabled` prop for `PushButtonCell`


rye (approved):

drewvolz:
If it's alright with everyone, let's merge this in. A followup can be done to fix the debug tool.

PR #6043, opened 2022-05-08
More props cleanup

### LINKED ISSUE #6036: Change back button title to default value
We are overriding what back should be as we apply the new navigator APIs. I think leaving it up to the navigator to choose what the value should be might make more sense than us deciding if it should be read as Back, Home, or a custom view title.

drewvolz:
Closes #6036 

**Navigation**
- remove back title and leave it up to the navigator
- pull out title prop explicitly for navigation title setting
- add navigation titles to dev bonapp picker list and detail
- add navigation titles to event detail (radio show, calendar)
- update smoke test to reflect back button being "all about olaf"

**Contacts**
- convert contacts Row and Detail into functional components

**Popover**
- pass the popover ref instead of ref.current to fix warning

**Dictionary and Student Orgs**
- fix initial loading state for loading/empty list

**Lists**
- add emptyList return type

**Share button**
- use share outline instead of filled button


source/views/dictionary/list.tsx
```diff
@@ -70,8 +71,10 @@ function DictionaryView(): JSX.Element {
 		})
 	}, [navigation])
 
-	let {data, error, reload, isPending}: AsyncState<Array<WordType>> =
-		useAsync(fetchDictionaryTerms)
+	let {data, error, reload, isPending}: AsyncState<Array<WordType>> = useAsync(
```
hawkrives, line 74: Isn't there an isInitial field given back to us here?

drewvolz: Yes, but it does not seem to work the way we think it should. 

Using the one provided by useAsync, it looks like it never hits the loading state and initially renders the empty component message. I found this pattern of a separate `isInitialFetch` check already in-place in Student Orgs and it is working over there, too.

I agree with you. I would expect useAsync's to give us the same result, though...

drewvolz: Based upon what we discovered in https://github.com/StoDevX/AAO-React-Native/pull/6021/commits/90baf4e04d0b45862f801c37b8a82a21ae4df9e8, we can use `isLoading` inside of the ListEmptyComponent.

We can refactor the dictionary list to be inline with orgs/hours once #6021 is merged. Does that sound okay?

hawkrives (approved):

rye (approved):

rye (approved):

PR #6059, opened 2022-05-13
Reorganize data fetching for dictionary

drewvolz:
Relevant branches
* This branch targets #6043 
* The data fetching strategy is based upon #6021 

Changeset reorganizes data fetching strategy for the dictionary:
* useFetch to provide data state
* adds loading, empty, and no result states in the list empty component
* move around where grouping/filtering happens


source/views/dictionary/types.ts
```diff
@@ -2,3 +2,8 @@ export type WordType = {
 	word: string
 	definition: string
 }
+
+export interface DiciontaryGroup {
```
hawkrives:
```suggestion
export interface DictionaryGroup {
```

hawkrives (dismissed):

source/views/dictionary/types.ts
drewvolz: Thanks

hawkrives (approved):

PR #6061, opened 2022-05-14
Add new navigation to menus

drewvolz:
[Recommend viewing without whitespace](https://github.com/StoDevX/AAO-React-Native/pull/6061/files?diff=split&w=1)

* useFetch hook for data state in menu-github
* add new tab navigator to menus index


rye (approved):

source/views/menus/carleton-menus.tsx
```diff
@@ -1,96 +1,73 @@
...
-							<Title style={styles.rowText}>{loc.title}</Title>
-						</Row>
-					</ListRow>
-					{i < collection.length - 1 ? (
-						<ListSeparator spacing={{left: 15}} />
-					) : null}
-				</View>
-			))}
+			{carletonCafes.map(
+				(loc: {id: keyof RootStackParamList; title: string}, i, collection) => (
```
hawkrives: Do we need still this type annotation? Or can TS infer it?

drewvolz: Looks like we don't! Good catch

hawkrives (approved):

drewvolz:
There will be a follow-up PR to add useAsync data loading to FancyMenu.

PR #6073, opened 2022-05-20
add navigation back to top-level stoprint (part 1)

drewvolz:
Adds stoprint top-level navigation back

* removes toplevelviewprops
* adds stoprint top-level routes back
* open settings button is broken here but fixed in the follow-up PR


source/views/stoprint/print-jobs.tsx
```diff
@@ -100,13 +94,13 @@ class PrintJobsView extends React.PureComponent<Props, State> {
 
 	keyExtractor = (item: PrintJob) => item.id.toString()
 
-	openSettings = () => this.props.navigation.navigate('SettingsView')
+	openSettings = () => navigation.navigate('SettingsView')
```
hawkrives: Where does `navigation` come from?

drewvolz: I have a bullet point in the PR description but I should have made it a comment in the code.

> open settings button is broken here but fixed in the follow-up PR

I can revert this if you'd like, but I think the plan is to merge part 2 into part 1 before this lands.

hawkrives (approved):

rye (approved):

hawkrives (approved):

PR #6074, opened 2022-05-20
stoprint refactor (part 2)

drewvolz:
[View without whitespace changes](https://github.com/StoDevX/AAO-React-Native/pull/6074/files?diff=split&w=1)

* add new navigation apis
* somewhat implement useRoute to pull out params in a few of the views
* inline view and function calls where it made sense to
* adds plenty of state hooks, effects, and callback optimizations


drewvolz:
Things that were previously broken in this pull request are fixed in the follow-up (#6078)

rye (approved):
Nice!

hawkrives (approved):

PR #6087, opened 2022-05-25
adding new navigation to and refactoring SIS views

drewvolz:
[view without whitespace](https://github.com/StoDevX/AAO-React-Native/pull/6087/files?w=1)

This is the last of the big efforts to bring navigation back to the main views.

Todo: 
- #3986

--- 
While I’ve marked this PR as a draft, there’s a bit more to figure out with types.  I also don’t know how we can even test things out without a login. But if the changes are too much to go through I will go and break this up into smaller pull requests.


hawkrives (commented):
👍

hawkrives:
Error counts need to be updated?

drewvolz:
Merged in the latest. The only change was the order. We will deal with redux typings at some point, either in #5918 or some other changeset?

hawkrives (approved):

source/views/sis/course-search/search.tsx
```diff
@@ -36,87 +33,110 @@ type ReduxDispatchProps = {
...
+		 */
+		navigation.getParent()?.setOptions({
+			headerSearchBarOptions: {
+				barTintColor: c.white,
+				onChangeText: (event: ChangeTextEvent) => {
+					setTypedQuery(event.nativeEvent.text)
+				},
+			},
+		})
+	}, [navigation, typedQuery])
```
drewvolz: After testing this I am seeing UI inconsistencies. The searchbar is added to other views in the tab navigator which is not correct.

This could be happening because of how the tab navigator is nested. I am going through the documentation of how to set this up correctly.

- This in particular: https://reactnavigation.org/docs/screen-options-resolution/
- Hints from someone's wiki: https://github.com/evangreer91/Dev-Notes/wiki/Nesting-navigators
- Helpful reference: https://reactnavigation.org/docs/tab-based-navigation

hawkrives: Or is it time to break SIS up into multiple top-level buttons?

Tab | Button
---|---
Balances | OleCard
TES | TES
Course Search | SIS
Student Work | … Student Work?

drewvolz: Sorry, I meant to say this is **definitely** happening due to `navigation.getParent()?.setOptions()` -- the `.getParent` part being key.

This is our first tab navigation with top-level search in a tab. It works just fine if it's one-level deep, though... meaning Home > SIS tabs > Search > Results, where results has a searchbar.

Breaking these up is still an option, though. We could keep the grouping of things on the home screen if that helps visually separate them from the rest, so like a SectionList of tiled buttons on Home?

drewvolz: I am happy to explore other UI designs, too. How do we feel about a list view?

I think there is a concern about breaking muscle memory on the home screen so if we limited our exploration to the subviews, that could be beneficial?

drewvolz: Moved this out into https://github.com/StoDevX/AAO-React-Native/issues/6109 in the interest of solving this in a follow-up.

PR #6101, opened 2022-05-30
settings (part 1): combine app version info into single view within support

drewvolz:
* combine build/version section with single tableview cell at the bottom of support
* add a function to build the version for consistency between support's footer and app's footer
* remove odds-and-ends section


hawkrives (approved):

source/views/settings/screens/overview/support.tsx
```diff
@@ -24,6 +24,17 @@ const openEmail = () => {
 	})
 }
 
+const getVersion = () => {
+	let version = appVersion()
+	let build = appBuild()
+
+	if (build) {
+		return `${version}-${build}`
```
rye: Any particular reason to make this `-` over `+`?

In a recent execution I think I got `2.8.0-pre` for `appVersion()` and one of our large-number build numbers for `appBuild()`, so I just wonder if this would render as `2.8.0-pre-260000410` for example.

drewvolz: Do you prefer a plus instead of a dash? I was refactoring to keep what we had but we could change it.

drewvolz: A plus makes sense to me 👍

rye (approved):

PR #6102, opened 2022-05-30
settings (part 2): app icon picker as a top-level section

drewvolz:
- inline app icon settings into the root of the settings view
- move icon related logic into an icon screen and out of miscellany section
- remove containing scrollview and tableview from change icon screen
- shorten big ole icon row name to "Big Ole"
- refactor logo component to be functional, use hooks, add types


source/views/settings/screens/overview/app-icon.tsx
```diff
@@ -0,0 +1,18 @@
...
+	React.useEffect(() => {
+		checkIfCustomIconsSupported()
+	}, [])
+
+	let checkIfCustomIconsSupported = async () => {
+		let canChangeIcon = await Icons.isSupported()
+		setCanChangeIcon(canChangeIcon)
+	}
+
+	return canChangeIcon ? <IconSettingsView /> : <></>
```
hawkrives: `: null` and change the return type as well?

drewvolz: Good catch -- I was fighting with the return type.

source/views/settings/screens/overview/app-icon.tsx
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react'
+import * as Icons from '@hawkrives/react-native-alternate-icons'
+import {IconSettingsView} from '../change-icon'
+
+export let AppIconSection = (): JSX.Element => {
```
rye: Is `export let` just for consistency here? I think I generally prefer `export const`, but if there's some reason we go this route I have no complaints.

drewvolz: I know @hawkrives has voiced a thing or two about preferring let over const across the codebase. Do we want to stick to one over the other?

hawkrives: I guess I like `const` at the top level, and `let` within functions?

drewvolz: I am okay with that pattern!

hawkrives (approved):

source/views/settings/screens/overview/app-icon.tsx
rye: That's what I was after — I prefer let by default inside of functions, too, but at the top level it makes sense to prevent accidentally changing the value.

rye (approved):

PR #6111, opened 2022-06-02
move dev bonapp picker to fix navigation transition from modal settings

drewvolz:
Grouping this view with the settings navigation fixes how the navigator pushes it onto the stack, so this fixes the dev bonapp view transition by keeping it on the modal stack.

### DIFF
--- a/source/navigation/routes.tsx
+++ b/source/navigation/routes.tsx
@@ -138,11 +138,6 @@ const HomeStackScreens = () => (
 		</Stack.Group>
 		<Stack.Group>
 			<Stack.Screen
-				component={DevBonAppPickerView}
-				name="BonAppPicker"
-				options={DevBonAppNavigationOptions}
-			/>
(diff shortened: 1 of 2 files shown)


source/navigation/routes.tsx
```diff
@@ -277,6 +272,11 @@ const SettingsStackScreens = () => (
 		<SettingsStack.Screen component={settings.PrivacyView} name="Privacy" />
 		<SettingsStack.Screen component={settings.LegalView} name="Legal" />
 		<SettingsStack.Screen component={settings.APITestView} name="APITest" />
+		<Stack.Screen
```
hawkrives: Stack? Or SettingsStack?

drewvolz: Oh. I missed changing which screen type it is wrapped with. Fun that even appears to work correctly in this setup.

hawkrives (approved):

PR #6152, opened 2022-06-20
🔖 news filtertoolbar

drewvolz:
✨What this is
* A use of filter toolbar on story categories to give a little more depth to the news list.
* Part of the [larger initiative](https://github.com/StoDevX/AAO-React-Native/discussions/6142) to add filtering to more lists.

🗒️ Todo
- [x] make the tests pass
  - Added `@react-native` to `transformIgnorePatterns` via https://github.com/StoDevX/AAO-React-Native/pull/6292/commits/ac12bc7294b4b0b0c3dda303de42edeca3483632 in #6153 
  - Adds `@frogpond` to `transformIgnorePatterns` via 1ff97cf3e106e98656ec88a2a071487f9b877e32
- [x] a cleaner dependency array for buildFilters via e5f56e7e868ee98a7c8c6728f725b205238f2506
- [x] need to fix a few typings (`spec.selected`)
- [x] figure out what filter's `apply: { key: ` should reference

📸 Screenshots
filtering | menu | none | empty
--|--|--|--
filtering | menu | none | empty

🎥 Demo


source/views/news/news-list.tsx
```diff
@@ -50,11 +52,73 @@ export const NewsList = (props: Props): JSX.Element => {
...
+		// to the dependency array. To reliably get this to re-render we need something to retrigger
+		// this callback.
+		//
+		// I recognize that:
+		// 1. disabling the linter for hooks is not great,
+		// 2. excluding two hook dependencies is less than ideal,
+		// 3. including an extraneous dependency is bad practice.
+		//
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, [isPending])
```
drewvolz: This may need some rethinking.

source/views/news/news-list.tsx
```diff
@@ -50,11 +52,73 @@ export const NewsList = (props: Props): JSX.Element => {
...
+
+	React.useEffect(() => buildFilters(), [buildFilters])
+
+	let filterStories = React.useCallback(() => {
+		return entries.filter((story) => {
+			let enabledCategories = filters.flatMap((f: ListType) =>
+				f.spec.selected.flatMap((s) => s.title),
+			)
```
drewvolz: This type needs to be looked at

```
Argument of type '(f: ListType) => (string | number)[]' is not assignable to parameter of type '(this: undefined, value: FilterType, index: number, array: FilterType[]) => string | number | readonly (string | number)[]'.
  Types of parameters 'f' and 'value' are incompatible.
    Type 'FilterType' is not assignable to type 'ListType'.
      Type 'ToggleType' is not assignable to type 'ListType'.
        Types of property 'type' are incompatible.
          Type '"toggle"' is not assignable to type '"list"'.ts(2345)
```

source/views/news/news-list.tsx
```diff
@@ -74,10 +150,17 @@ export const NewsList = (props: Props): JSX.Element => {
 				/>
 			)}
 			ListEmptyComponent={
-				isLoading ? <LoadingView /> : <NoticeView text="No news." />
+				isLoading ? (
+					<LoadingView />
+				) : filters.some((f: ListType) => f.spec.selected.length) ? (
```
drewvolz, line 145: Same with this type

```
Argument of type '(f: ListType) => number' is not assignable to parameter of type '(value: FilterType, index: number, array: FilterType[]) => unknown'.
  Types of parameters 'f' and 'value' are incompatible.
    Type 'FilterType' is not assignable to type 'ListType'.
      Type 'ToggleType' is not assignable to type 'ListType'.ts(2345)
```

source/views/news/news-list.tsx
```diff
@@ -50,11 +52,73 @@ export const NewsList = (props: Props): JSX.Element => {
...
+				key: 'category', // TODO: figure if this is referencing anything or is simply unique
+				enabled: true,
+				spec: {
+					title: 'Categories',
+					options: filterCategories,
+					selected: filterCategories,
+					mode: 'OR',
+					displayTitle: true,
+				},
+				apply: {key: 'category'}, // TODO: figure out what this should reference
```
drewvolz: I am unsure what key this should be referencing

source/views/news/lib/util.ts
```diff
@@ -0,0 +1,19 @@
...
+		.filter((entry) => !entry.content.includes('<form'))
+}
+
+// remove extraneous whitspace
+// decode entities
+// titlecase words
+export const trimStoryCateogry = (label: string): string => {
+	let evenedWhitespace = label.replace(/\s+/gu, ' ')
+	return toLaxTitleCase(decode(evenedWhitespace))
+}
```
drewvolz, line 19: Alternatively we could:
- perform this cleanup on ccc
- provide an array of categories from the server response

source/views/news/news-list.tsx
drewvolz: ~Resolved.~ Not resolved.

source/views/news/lib/util.ts
drewvolz: I am okay with continuing to keep this here, though!

source/views/news/news-list.tsx
drewvolz: Do we see this as able to merge and be iterated upon?

source/views/news/lib/util.ts
hawkrives: Let's keep this here

drewvolz:
Anything more to do here? This is ready for another review.

source/views/news/news-list.tsx
```diff
@@ -50,11 +52,72 @@ export const NewsList = (props: Props): JSX.Element => {
...
+ 		// to the dependency array. To reliably get this to re-render we need something to retrigger
+ 		// this callback.
+ 		//
+ 		// I recognize that:
+ 		// 1. disabling the linter for hooks is not great,
+ 		// 2. excluding two hook dependencies is less than ideal,
+ 		// 3. including an extraneous dependency is bad practice.
+ 		//
+ 		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, [isPending])
```
hawkrives: Just this one. I'll try to take a look at it tonight or tomorrow…?

hawkrives (commented):
… from a month ago

source/views/news/news-list.tsx
drewvolz: This is probably being caused when this `useEffect` both updates and uses a variable in its own dependency array. So by updating `entries` with filtering on category, we re-render infinitely.

1. Remove the `eslint-disable-next-line`
2. Update the dependency array to include `entries` (set it to `[entries, getStoryCategories]`)
3. Observe infinite re-rendering

Note that by commenting out the two calls to `setFilters()` and `setFilters(newsFilters)` on lines 67 and 91, we do not infinitely loop, but we will also not build our filters or be able to filter.

drewvolz: So this problem can be reduced all the way down to problematic code that looks like this:

```ts
React.useEffect(() => {
   setFilters([])
}, [entries])
```

drewvolz: Fixed via e5f56e7e868ee98a7c8c6728f725b205238f2506

source/views/news/news-list.tsx
```diff
@@ -50,11 +52,59 @@ export const NewsList = (props: Props): JSX.Element => {
...
-	// remove all entries with blank excerpts
-	// remove all entries with a <form from the list
-	let entries = data
-		.filter((entry) => entry.excerpt.trim() !== '')
-		.filter((entry) => !entry.content.includes('<form'))
+	let entries = React.useMemo(() => cleanEntries(data), [data])
+
+	let [filters, setFilters] = React.useState<ListType[]>([])
+
+	let getStoryCategories = React.useCallback((story: StoryType) => {
```
hawkrives: Question: why not pull this into a standalone function, outside of the component?

source/views/news/news-list.tsx
```diff
@@ -50,11 +52,59 @@ export const NewsList = (props: Props): JSX.Element => {
...
+					mode: 'OR',
+					displayTitle: true,
+				},
+				apply: {key: 'category'},
+			},
+		]
+		setFilters(newsFilters)
+	}, [entries, getStoryCategories])
+
+	let filterStories = React.useCallback(() => {
```
hawkrives: Similarly here: can we pull this out into a standalone function that takes three parameters?

I'm open to being told "no" :-)

hawkrives: I also feel a little bad about re-filtering the entire list on every render, but I'm sure that we have worse performance pitfalls elsewhere, so I won't complain about it here.

source/views/news/news-list.tsx
drewvolz: Funny enough I did pull them outside when I was making the fixes for this last issue. I left as-is to show the minimal diff and then forgot I had lifted them out. Let's pull it out again, that's a good idea!

source/views/news/news-list.tsx
drewvolz: I won't use the fact that our news lists are small as an excuse since every optimization is worth exploring. Is there a quick way we can make that more performant or do we tackle that later?

hawkrives: use _.memo with a cache size of 1? As a wrapper around our external filter function? Maybe?

drewvolz: Done!

hawkrives:
How sure are you that memoize doesn't use an unbounded cache? It'd be bad if we tried to fix one perf issue only to eat all available ram

drewvolz:
~~I'm not sure about the bounding/unbounding sizes of memoize's cache.~~ 

Memoize's cache would be unbounded, yes.

Although we we have this pattern elsewhere in the codebase for large lists: course catalog and student orgs.

https://github.com/StoDevX/AAO-React-Native/blob/13c66f1b9e0ada193994e1bc1a8365e5badfbf28/source/views/sis/course-search/list.tsx#L63-L85

https://github.com/StoDevX/AAO-React-Native/blob/f422473c201031195e2ac00b76444271edf21e89/source/views/student-orgs/list.tsx#L34-L44

hawkrives:
Ok! :shipit:

hawkrives (approved):

drewvolz:
@rye thoughts? I think we've now got this cleaned up!

rye (approved):

PR #6154, opened 2022-06-21
hours report a problem tweaks

drewvolz:
[`7537c1`](https://github.com/StoDevX/AAO-React-Native/commit/7537c1741048db0d26f3e8f75ced140e07579e99) hours report overview and editor as modal views
* Helps center the context and push less on top of the stack.
* The logic is 2(!) lines thanks to our navigation upgrade.

[`ba5ec71`](https://github.com/StoDevX/AAO-React-Native/commit/ba5ec71f9d6ffaff36697d757576ba94601dc481) add support for overriding the close screen navigation button title
* Sometimes 'Done' sounds like a confirmation when instead we mean to say this is a destructive action, like 'Close'. With modals in the app we may want to override this title for screens like the hours editor where it would be unfortunate if 'Done' was interpreted as an action that saved what you were editing.
* Screens like the hours editor and overview need these changes. It would be unfortunate if 'Done' was interpreted as an action that saved what you were editing when we really meant to say 'Close' which is destructive.

[`f532b9`](https://github.com/StoDevX/AAO-React-Native/commit/f532b9a54a174d435ee28121d9705f6566938a7d) accessory icons to buttoncell's cellAccessoryView
* Particular use-case is hours report row that says "Submit Report"; this adds an icon to the far right side of the cell.
* This enables an icon that the consumer can only control the image of (size, color are all set by ButtonCell).

[`1f0ffd`](https://github.com/StoDevX/AAO-React-Native/commit/1f0ffd80ecd9a70a9426b25e893836dfda8f3c35) make the info header more generic
* More than hours can be corrected (building name, days).
* We have an idea about [adding form support to report new spaces](https://github.com/StoDevX/AAO-React-Native/discussions/5969) so this will read more clearly in the future, too.

[`f3500e`](https://github.com/StoDevX/AAO-React-Native/commit/f3500e885e9423346b5e11e4bb91f379ce3d9598) eslint cleanup 
* 3 issues knocked out:
  * 2 of them were "defined but never used"
  * 1 was "Unexpected empty arrow function"
  
 [`4a2c39`](https://github.com/StoDevX/AAO-React-Native/pull/6154/commits/4a2c39a286f3021c1fb8708bbbe4a8eeed559144) look for unsaved changes
* prompt and prevent dismissing the view with unsaved edits
* prevent dismissal via swiping gesture on hours editor
* change the report back button to "Discard"

---

* (already reviewed as a separate PR) includes #6160
  - classes to functional components
  - useNavigation and useRoute hooks
  - removing toplevelviewprops
  - removing more uses of any

* (already reviewed as a separate PR) includes #6170
  - simplifying the hours rows by condensing the opens and closes times into a single row.
  - providing more context by showing what has changed by showing the previous schedule and changed schedule.

---

Also I won't go for inaccessible white on green next time.    
  
Overview | Editor | Editing
--|--|--
Simulator Screen Shot - iPhone 12 Pro - 2022-06-20 at 22 48 40 | Simulator Screen Shot - iPhone 12 Pro - 2022-06-20 at 22 48 36 |


source/views/building-hours/report/overview.tsx
```diff
@@ -332,4 +336,6 @@ export const NavigationKey = 'BuildingHoursProblemReport'
 
 export const NavigationOptions: NativeStackNavigationOptions = {
 	title: 'Report a Problem',
+	presentation: 'modal',
+	headerRight: () => <CloseScreenButton title="Close" />,
```
drewvolz: I'd like to take a vote on whether the overview (second screenshot) should be a modal. We would need more precautions to prevent accidental swipe-to-dismiss if there's changed data in the form? Thoughts?

hawkrives: I think it should be a modal. Thoughts on labelling it "Discard" instead of "Close"?

drewvolz: Discard seems good to me. Do we want to prevent closing without prompting if changes have been made to the form before discarding?

hawkrives: I think that would be best

drewvolz: My initial attempt at this is to compare two json objects to each other. This works in some cases but not in all.

It fails to work for areas where we've added to the edited data in the following way:

Take either notes or closed-for-chapel. 

If we edit a note, the field and value get added but never removed from the object.

Same with the hours toggle -- we toggle on/off and the field remains false in the data, not removed.

hawkrives: Do we need a "default" object that we can merge the original / edited objects into, so that all properties are guaranteed to exist?

What if we dump both to the yaml strings, like we're sending them to the server? That strips the optional properties, right?

drewvolz: Yeah I was thinking the same thing along the lines of a default object where optional props do not change the shape of the diff.

I can give your idea of dumping the yaml a try.

drewvolz: Nice! That worked great and was added to [`4a2c39`](https://github.com/StoDevX/AAO-React-Native/pull/6154/commits/4a2c39a286f3021c1fb8708bbbe4a8eeed559144).

modules/navigation-buttons/close-screen.tsx
```diff
@@ -21,7 +22,9 @@ export function CloseScreenButton({buttonStyle}: Props): JSX.Element {
 			style={[commonStyles.button, buttonStyle]}
 			testID="button-close-screen"
 		>
-			<Text style={[commonStyles.text, styles.text]}>Done</Text>
+			<Text style={[commonStyles.text, styles.text]}>
+				{title ? title : 'Done'}
```
hawkrives:
```suggestion
				{title ?? 'Done'}
```

source/views/building-hours/report/overview.tsx
```diff
@@ -23,23 +25,62 @@ import type {
...
+	/**
+	 * checking for unsaved edits
+	 *
+	 * noting that we also have `gestureEnabled` set to false in the navigation options
+	 * (ios only) to prevent dismissing the modal without prompting.
+	 * https://reactnavigation.org/docs/preventing-going-back
+	 */
+	React.useEffect(
+		() =>
+			navigation.addListener('beforeRemove', (event) => {
```
hawkrives, line 58: Can we save the return value of `addListener`, and removeListener(?) in the cleanup handler of useEffect?

drewvolz: That is a good question. I am unsure why the RN navigation docs didn't include that initially?

drewvolz:
Anything else we see needing to be done in this PR? This is ready for another review.

hawkrives (approved):

PR #6183, opened 2022-06-28
remove isPending and isInitial checks for list refreshing

drewvolz:
Potentially fixes:
- #6116 

This seemingly fixes the battle that LoadingView and useAsync have going on for showing off their loading indicators.

Do we lose anything by setting the `refreshing` state to false? 

Perhaps LoadingView isn't needed on these views if we have another way to show a loading indicator? I would vote to set refreshing to false (this PR changeset) but I am wary something might be happening under the hood as a side effect. Maybe accessibility is changed by this.

The documentation has this to say about `onRefresh` and `refreshing`. It says "make sure to also set the refreshing prop correctly" and "set this as true while waiting for new data".

> ### [onRefresh](https://reactnative.dev/docs/sectionlist#onrefresh)
> If provided, a standard RefreshControl will be added for "Pull to Refresh" functionality. Make sure to also set the refreshing prop correctly. To offset the RefreshControl from the top (e.g. by 100 pts), use progressViewOffset={100}.

>### [refreshing](https://reactnative.dev/docs/sectionlist#refreshing)
> Set this as true while waiting for new data from a refresh.

### DIFF
--- a/source/views/building-hours/list.tsx
+++ b/source/views/building-hours/list.tsx
@@ -62,8 +62,6 @@ export function BuildingHoursView(): JSX.Element {
 
 	let {
 		data: {data: buildings = []} = {},
-		isPending,
-		isInitial,
 		isLoading,
 		reload,
 	} = useBuildingHours()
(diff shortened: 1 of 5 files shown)


hawkrives:
I think that if we provide onRefresh without updating refreshing, voiceover will get confused…? I'd have to figure out how to trigger a PullToRefresh with VO, though. 

Color me hesitant…

drewvolz:
Fair enough. Let's not approach it this way.

PR #6175, opened 2022-06-27
"more" view to show the A-Z links

drewvolz:
Show the A-Z urls in a grouped section list. 

- Each list row opens a webview.
- The JSON data originates [from here](https://wp.stolaf.edu/wp-json/site-data/sidebar/a-z).

Todo:
- [x] Type fixups
- [x] Camel case renamings
- [x] Search
- Proxy the endpoint on ccc?

---

📸 Screenshots

List | Detail
--|--
Simulator Screen Shot - iPhone 12 Pro - 2022-06-28 at 18 38 35 | detail

PR #6206, opened 2022-07-03
more view (part 2): ccc-ified

drewvolz:
- follow-up to #6175 
- add schema and definition yaml to generate a-to-z json
- utilize the caching server to proxy a new `/a-to-z` endpoint
  - endpoint implementation located at https://github.com/frog-pond/ccc-server/pull/617

### DIFF
--- a/source/init/data.ts
+++ b/source/init/data.ts
@@ -3,6 +3,10 @@ import {insertForUrl} from '@frogpond/fetch'
 
 const dataSets = [
 	{
+		url: '/a-to-z',
+		data: require('../../docs/a-to-z.json'),
+	},
+	{
 		url: API('/spaces/hours'),
 		data: require('../../docs/building-hours.json'),
 	},
--- a/source/init/data.ts
+++ b/source/init/data.ts
@@ -3,7 +3,7 @@ import {insertForUrl} from '@frogpond/fetch'
 
 const dataSets = [
 	{
-		url: '/a-to-z',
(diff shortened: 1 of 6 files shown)


source/init/data.ts
```diff
@@ -2,6 +2,10 @@ import {API} from '@frogpond/api'
 import {insertForUrl} from '@frogpond/fetch'
 
 const dataSets = [
+	{
+		url: '/a-to-z',
```
hawkrives: Doesn't this need an `API()` call?

source/views/more/index.tsx
```diff
@@ -87,26 +87,19 @@ function MoreView(): JSX.Element {
...
 
-	let groupedOriginal = React.useMemo(
-		() =>
-			menuItems.map(({letter, values}) => ({
-				title: letter[0],
-				data: values,
-			})),
-		[menuItems],
-	)
+	let groupedOriginal = React.useMemo(() => data, [data])
```
hawkrives: Can we remove this line entirely?

hawkrives (changes requested):

source/init/data.ts
drewvolz: Did I forget that? I guess I forgot that!

source/views/more/index.tsx
drewvolz: I suppose we could!

hawkrives (approved):

PR #6211, opened 2022-07-03
remove remaining help references

drewvolz:
- follow-up to #6205 where the top-level help view was removed from the list of views
- removes help views, tooling, types, redux methods, and data 

I think we have to leave the schema for backwards compatibility, though.


hawkrives:
What if we commit the current output of the help endpoint? So that the current app doesn't lose a feature?

hawkrives:
oh, I see the 3rd commit now lol

hawkrives (approved):

PR #6235, opened 2022-07-10
github action to update lint and tsc files

drewvolz:
Allow for updating the eslint problems and ts counts/errors files through pull request comments.

---

This may not be working or doing what it says it should do yet.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -119,6 +119,25 @@ jobs:
       - run: npm run validate-data
       - run: npm run bundle-data
 
+  tsc-and-lint-updates:
+    name: Update tsc and eslint files
+    runs-on: ubuntu-20.04
+    if: github.event.issue.pull_request && github.event.comment.body == 'update tsc and lint files'
+    steps:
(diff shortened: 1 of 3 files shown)


drewvolz:
update tsc and lint files

hawkrives:
for what it's worth, my attempt at auto updating doesn't even run anymore lmao - it never triggers for some reason

hawkrives:
but, I also wouldn't be opposed (whichever is easier) to having GitHub automatically push updated eslint/tsc files if they change, even without the magic comment

drewvolz:
I think this may only end up working if this were to be merged into `master`. Based upon @hawkrives comment I don't know if it will work as it should, though.

drewvolz:
I'm willing to back this change out if it doesn't work, but should we see if this works by merging it?

Alternatively, clone the repo and give it a try?

hawkrives (approved):

drewvolz:
@rye any thoughts? I know this is possibly broken at the moment so I can just as easily close this out.

rye (approved):

PR #6286, opened 2022-08-07
Remove browser/tabs dependencies for openUrl

drewvolz:
- Fixes the android build (now the project builds and runs)
- Removes `@frogpond/react-native-custom-tabs`
- Removes `react-native-safari-view` and `@types/react-native-safari-view`
- Does not add `react-native-inappbrowser` (introduced in this PR)


hawkrives (approved):

rye (dismissed):
New `tsc` errors. 

```diff
+modules/open-url/open-url.ts(50,4): error TS2322: Type 'Promise<boolean | BrowserResult>' is not assignable to type 'Promise<boolean>'.
+  Type 'boolean | BrowserResult' is not assignable to type 'boolean'.
+    Type 'BrowserResult' is not assignable to type 'boolean'.
```

Can we just toss the `BrowserResult`? (Always coerce it to `true`?) I'm not sure what the return values (`true` / `false`) are supposed to mean, but if `cancel` | `dismiss` (the values for `BrowserResult` should be treated differently we could do that too.

drewvolz:
The API for `inappbrowser` is a bit rough compared to `custom-tabs`. Do we think we should be following what the dependency's readme suggests as an implementation? 

```ts
try {
    const url = 'https://github.com/proyecto26'
    if (await InAppBrowser.isAvailable()) {
        const result = await InAppBrowser.open(url, {
            // iOS Properties
            ...
            // Android Properties
            ...
        })
        await this.sleep(800);
        Alert.alert(JSON.stringify(result))
    } else {
        Linking.openURL(url)
    }
} catch (error) {
    Alert.alert(error.message)
}
```

hawkrives:
We could wrap that in our openUrl, sure… that is a bit rough, but it seems more accurate too.

drewvolz:
@rye @hawkrives I am noting the scope of this PR has changed to remove _all_ custom in-app browser dependencies. 

Instead we can fallback on our own `openUrl` which opens links natively in their respective platform browsers.

By doing this we are able to remove all dependencies and keep the experience much more consistent.

hawkrives:
I am in favor of this change. 

- It allows us to use the system's default browser
- By doing so, it allows us to re-use any login sessions that the user has
- It allows us to remove a set of dependencies

If, in the future, we want to introduce some sort of symbol to denote links that will open outside of the app, we can do so. 

If we ever want to use AuthenticationSession or something to log in, we'll have to add back a dependency, but I think we're a long way from that.

hawkrives (approved):

hawkrives (approved):

rye (approved):

PR #6333, opened 2022-08-23
ci: Run bundle-data before running tests

rye:
When we split the workflows, we made it so that the files will no longer be present when the tests are run. This led to a test failure when the `.json` files are not generated to be loaded by the tests.

Now, call `npm run bundle-data` to generate the .json files before the tests are run. This leads Jest to pass now.


hawkrives (approved):
😅

drewvolz:
You all are quick! Thanks Kris!

PR #6358, opened 2022-08-30
🔖 streams filterbar

drewvolz:
Replaced with:
- #6412 
- #6413 

---

These changes refactor streams list to accomplish a few ideas:

* refactor data fetching with `useFetch`
* add filter bar to show stream categories
* show athletics streams by default
* Part of #6142 

📸 Screenshots

~ | ~
--|--
Simulator Screen Shot - iPhone 12 Pro - 2022-08-29 at 20 38 38 | Simulator Screen Shot - iPhone 12 Pro - 2022-08-29 at 20 38 52

### DIFF
--- a/source/views/streaming/streams/list.tsx
+++ b/source/views/streaming/streams/list.tsx
@@ -4,6 +4,7 @@ import {timezone} from '@frogpond/constants'
 import * as c from '@frogpond/colors'
 import {ListSeparator, ListSectionHeader} from '@frogpond/lists'
 import {NoticeView, LoadingView} from '@frogpond/notice'
+import {FilterToolbar, ListType} from '@frogpond/filter'
 import {StreamRow} from './row'
 import toPairs from 'lodash/toPairs'
 import groupBy from 'lodash/groupBy'
@@ -12,7 +13,7 @@ import type {Moment} from 'moment-timezone'
 import {toLaxTitleCase as titleCase} from '@frogpond/titlecase'
 import type {StreamType} from './types'
 import {API} from '@frogpond/api'
-import {fetch} from '@frogpond/fetch'
+import {useFetch} from 'react-async'
 
 const styles = StyleSheet.create({
 	listContainer: {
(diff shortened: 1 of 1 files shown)


source/views/streaming/streams/list.tsx
```diff
@@ -23,99 +24,137 @@ const styles = StyleSheet.create({
...
-			renderItem={renderItem}
+			keyExtractor={(item: StreamType) => item.eid}
+			onRefresh={reload}
+			refreshing={isPending && !isInitial}
+			renderItem={({item}: {item: StreamType}) => <StreamRow stream={item} />}
 			renderSectionHeader={({section: {title}}) => (
 				<ListSectionHeader title={title} />
 			)}
-			sections={streams}
+			sections={groupStreams(filterStreams(entries, filters))}
```
drewvolz, line 157: Is there an obvious way we can improve on this?

source/views/streaming/streams/list.tsx
```diff
@@ -23,99 +24,137 @@ const styles = StyleSheet.create({
...
+		API('/streams/upcoming', {
+			sort: 'ascending',
+			dateFrom,
+			dateTo,
+		}),
+		{
+			headers: {accept: 'application/json'},
+		},
+	)
+}
```
rye, line 60: Do these really need to be fat-arrow functions? If so, can they be `const` at least?

source/views/streaming/streams/list.tsx
```diff
@@ -23,99 +24,137 @@ const styles = StyleSheet.create({
...
 
-	let keyExtractor = (item: StreamType) => item.eid
+export const StreamListView = (): JSX.Element => {
+	let {data = [], error, reload, isPending, isInitial, isLoading} = useStreams()
```
rye, line 63: Can we give this an interface? These fields seem magical without a type.

source/views/streaming/streams/list.tsx
```diff
@@ -23,99 +24,137 @@ const styles = StyleSheet.create({
...
-			})
-		} catch (error) {
-			if (error instanceof Error) {
-				setError(error)
-			} else {
-				setError(new Error('unknown error - not an Error'))
-			}
-			return
+let filterStreams = (entries: StreamType[], filters: ListType[]) => {
+	return entries.filter((stream) => {
```
rye, line 33: `entries` could have a more descriptive name like `streamEntries` or `streams`. `streams.filter((stream) => { /* */ })` is a nice pattern imo.

source/views/streaming/streams/list.tsx
```diff
@@ -23,99 +24,137 @@ const styles = StyleSheet.create({
...
+		let enabledCategories = filters.flatMap((f: ListType) =>
+			f.spec.selected.flatMap((s) => s.title),
+		)
+
+		if (enabledCategories.length === 0) {
+			return entries
 		}
```
rye, line 40: Can this get pulled out of the filter function? Seems like you're not doing anything that uses `stream` until the very end.

source/views/streaming/streams/list.tsx
```diff
@@ -23,99 +24,137 @@ const styles = StyleSheet.create({
...
+		if (allCategories.length === 0) {
+			return
+		}
+
+		let categories = [...new Set(allCategories)].sort()
+		let filterCategories = categories.map((c) => {
+			return {title: c}
+		})
+
+		let streamFilters: ListType[] = [
```
rye, line 96: It is possible to tighten `ListType` or rename it to something like `FilterType`? It seems like we're using a type called `List` for a ... list of filters? Shouldn't it just be a list of filters?

drewvolz:
Should I break this PR up into two parts? The first would be the useFetch hook and the second would be the filter bar.

source/views/streaming/streams/list.tsx
drewvolz: Are you asking if `ListType` can be `ListFilterType`? It probably could. This is coming from `modules/filter/types.ts` where we define a whole scope of different `FilterType` definitions.

```ts
export type ToggleType = {
	type: 'toggle'
	key: string
	enabled: boolean
	spec: ToggleSpecType
	apply: ToggleFilterFunctionType
}

export type PickerType = {
	type: 'picker'
	key: string
	enabled: true
	spec: PickerSpecType
	apply: PickerFilterFunctionType
}

export type ListType = {
	type: 'list'
	key: string
	enabled: boolean
	spec: ListSpecType
	apply: ListFilterFunctionType
}

export type FilterType = ToggleType | PickerType | ListType
```

source/views/streaming/streams/list.tsx
drewvolz: Are you asking if lines 34-40 can be extracted into their own function?

source/views/streaming/streams/list.tsx
drewvolz: It does look magical up-front but I guarantee that the types are all pretty clear since this now uses [react-async's `useFetch`](https://docs.react-async.com/api/interfaces?q=usefetc#usefetch-hook).

The type info on line 63 for `useStreams` (or up above on line 46) shows the following:

```ts
let useStreams: (date?: Moment) => AsyncState<StreamType[], FetchRun<StreamType[]>>
```

and hovering over each of these types reflects the [type information that `useFetch`'s state](https://docs.react-async.com/api/state) provides us

```tsx
let data: StreamType[]
let error: Error | undefined
let reload: () => void
let isPending: boolean
let isInitial: false
let isLoading: boolean
```

source/views/streaming/streams/list.tsx
drewvolz: They can definitely be `const`. Is there a reason you had in mind to not make them be bound this way?

hawkrives:
It might be easier to review that way, if you don't mind the work

drewvolz:
Closing this PR in favor of the two-part approach for reviewing these changes.

PR #6388, opened 2022-09-10
Remove react-native-communications

### LINKED ISSUE #6369: Email app is not opening on android
Pixel 6 does not open the email app when tapping:
- Settings > contact us
- Building hours > suggest an edit > report a problem > submit report
- SIS > open jobs > job detail > email
- Probably more I am not thinking of e.g. native directory (not presently merged)

From debugging I have experimented with:
- replacing `react-native-community/email` with `@frogpond/openUrl`.

The email app will open when we pass our `openUrl` wrapper the following url
- `mailto:${to}?cc=${cc}&bcc=${bcc}&subject=${subject}&body=${body}`

If we did choose to go with our wrapper over RNC's, we should add more error checking and data sanitizing. Mainly making sure we call
- `encodeURIComponent().join(',')` on to, cc, and bcc,
- and `encodeURIComponent()` on subject and body.

drewvolz:
This PR accomplishes a few things:
* replaces calls for opening mail and phone apps with `openUrl` – docs show this is [built-in to react native](https://reactnative.dev/docs/linking#built-in-url-schemes)
  * fixes #6369 
  * fixes opening `tel:` uri schemes on android
* tests for email and phone number parsing
* removes `react-native-communications` (not been updated since 2017, causing issues)
* removes `@types/react-native-communications`


source/components/call-phone.ts
```diff
@@ -29,3 +32,15 @@ export function callPhone(phoneNumber: string, opts?: Options): void {
 		)
 	}
 }
+
+const formatNumber = (phoneNumber: string) => {
+	let re = /(\d{3})-?(\d{3})-?(\d{4})/gu
```
hawkrives: Should we handle `+` codes (+1 111 222 3333)?

source/components/send-email.ts
```diff
@@ -13,7 +13,7 @@ type Args = {
 export function sendEmail(args: Args): void {
 	const {to = [], cc = [], bcc = [], subject = '', body = ''} = args
 	try {
-		email(to, cc, bcc, subject, body)
+		openUrl(`mailto:${to}?cc=${cc}&bcc=${bcc}&subject=${subject}&body=${body}`)
```
hawkrives: We need to make sure that we urlencode these segments

source/components/call-phone.ts
hawkrives: If not here, we should file an issue to handle it in the future.

hawkrives (dismissed):

source/components/send-email.ts
drewvolz: Yeah, I agree. I wrote up in the issue that encoding was needed and walked it back incorrectly. I'll try to not muddy the changes too much. Here's at least [how the dependency performed this](https://github.com/anarchicknight/react-native-communications/blob/05bfcdcc123d1e2380156832317a917766699c5c/AKCommunications.js).

source/components/call-phone.ts
drewvolz: We should probably try to handle that in the future. Certain numbers are also showing up as:
- NA
- Without an area code: 7863016

source/components/send-email.ts
hawkrives: I think we can use whatever we use to generate query strings elsewhere, and use it to make this one?

hawkrives: `to=${encodeUriComponent(to)}?{makeMeQueryString({cc, from, subject}, {omitEmpty: true})}` or something like that

source/components/__tests__/email.test.ts
```diff
@@ -0,0 +1,80 @@
...
+			'mailto:test%40domain.com%2Chelp%40domain.com?body=&subject=',
+		)
+	})
+
+	it('should encode the "cc" addressees', () => {
+		expect(formatEmailParts({cc: ['test@domain.com']})).toBe(
+			'mailto:?body=&cc=test%40domain.com&subject=',
+		)
+		expect(formatEmailParts({cc: ['test@domain.com', 'help@domain.com']})).toBe(
+			'mailto:?body=&cc=test%40domain.com&cc=help%40domain.com&subject=',
```
hawkrives: This seems likely problematic. I think we will need to join (but not encode) before we hand it off to openURL?

source/components/__tests__/email.test.ts
```diff
@@ -0,0 +1,80 @@
...
+		)
+	})
+
+	it('should encode the "bcc" addressees', () => {
+		expect(formatEmailParts({bcc: ['test@domain.com']})).toBe(
+			'mailto:?bcc=test%40domain.com&body=&subject=',
+		)
+		expect(
+			formatEmailParts({bcc: ['test@domain.com', 'help@domain.com']}),
```
hawkrives, line 32: Same comment as for CC:

source/components/call-phone.ts
```diff
@@ -29,3 +34,28 @@ export function callPhone(phoneNumber: string, opts?: Options): void {
 		)
 	}
 }
+
+export const formatNumber = (phoneNumber: string): string => {
+	let re = /^(1|)?(\d{3})(\d{3})(\d{4})$/u
```
hawkrives, line 39: International numbers can be anything; isn't Great Britain +83?

drewvolz:
I made a few changes:
- allow for +1 in the phone number, more robust phone number formatting
- uri encode and query stringified all of the email parts
- tests around formatting phone numbers
- tests around encoding and creating email uris

source/components/call-phone.ts
drewvolz: Yes, I agree there are country calling codes other than +1. 

Phone numbers in other countries don't all have exactly ten digits, though, so the formatting here doesn't apply internationally.

drewvolz: So we're not looking to support +83 by replacing` ^(1|)` with `^(\d|)` in the regex and `'+1 '` with `'+${match[1]} '` in the international code due to the ten digits.

source/components/__tests__/email.test.ts
drewvolz: Could you explain a little more about what the issue is?

drewvolz: I think you're questioning what happens when we do the following 

```ts
let encodedTo = encodeURIComponent(['test@domain.com', 'help@domain.com'].join(','))
let query = querystring.stringify({cc, bcc, subject, body})

return `mailto:${encodedTo}?${query}`
```

specifically the encodedTo bit

```ts
 encodeURIComponent(['test@domain.com', 'help@domain.com'].join(','))
 ```
 
 which looks great according to `openUrl`
 
telegram-cloud-photo-size-1-5542185941918132147-y

Comment (hawkrives) [reply source/components/__tests__/email.test.ts]: I think it works well for "To:" - but we aren't doing the .join for "CC:" and "BCC:". We're adding two `&cc=` parameters, instead.

```
mailto:me@me.me?cc=a@b.com&cc=b@c.com
``` 

I just want to know what happens with that. 

I would expect something more like this:

```
mailto:me@me.me?cc=a@b.com,b@c.com
```

source/components/call-phone.ts
hawkrives: Fair!

source/components/__tests__/email.test.ts
drewvolz: Oh that makes sense. I will fix this.

drewvolz: Okay. So oddly enough the implementation as-is does work for your example cc and bcc addresses.

drewvolz:
The latest set of changes provide a sanitizing flow to the query params we generate. [90aa0d2](https://github.com/StoDevX/AAO-React-Native/pull/6388/commits/90aa0d2f5c19abc49db6de96251df03119e7c700)

While the spec allows us to append more cc and bcc values as additional params, it may be simpler overall for us to handle stripping empty arrays and strings from the query altogether. At the very least it makes the tests a little easier to read.

hawkrives (approved):

PR #6389, opened 2022-09-10
Update update-tsc-and-lint-files.sh

drewvolz:
Poking to bring it more in-line with how we do `scripts/push-on-podfile-change.sh`.

If this does not fix it, I'll close this PR and open a new one to revert the scripts entirely via `git revert 52e436 -m 1`

Error that this is trying to fix:

```
if [ -z "${!actor}" ]; then
  echo 'Need to set environment variable "actor"' && exit 1;
fi

scripts/update-tsc-and-lint-files.sh: line 5: !actor: unbound variable
```

### DIFF
--- a/scripts/update-tsc-and-lint-files.sh
+++ b/scripts/update-tsc-and-lint-files.sh
@@ -2,11 +2,11 @@
 
 set -e -v -u -o pipefail
 
-if [ -z "${!actor}" ]; then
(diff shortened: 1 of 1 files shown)


drewvolz:
~~/makefiles~~

Right. This has to be merged before I can test it again.

rye (approved):

hawkrives (approved):

PR #6396, opened 2022-09-12
🧭 Custom tabs refactoring

drewvolz:
- Follow-up to #6286 

## Context

After dogfooding the removal of support for in-app browser dependencies from this app, the experience was not so great. A fair amount of experiences in the app are a bit better if you aren't booted away so suddenly. We've also tried providing alerts that ask for permission to jump apps but that experience is also jarring.

The best of both scenarios does exist, though. This pull request aims to improve the experience by letting the user choose which of the two preferences they like. If the user wants their OS browser to handle all of the linking, great! We'll open with `Linking`. If someone else enjoys how the shipping app today keeps things connected by attaching a browser to the app temporarily, also great! They can continue to do that.

The aim of this UX decision is to keep things in-line with how the app behaves today, to provide smart defaults, and to allow the user control over where their data goes if they so choose. 

## What are these changes?

1. Dependencies! `react-native-inappbrowser-reborn` was added in https://github.com/StoDevX/AAO-React-Native/commit/4cb546a01295d5b43324b6b5f18200a2673b4943

    - unifies the library responsible for creating in-app browsers on iOS and Android
    - very flexible[ in terms of the props](https://github.com/proyecto26/react-native-inappbrowser#usage) we can pass it to customize the experience

2. Refactoring of `openUrl` in https://github.com/StoDevX/AAO-React-Native/commit/08bf992fbbba07b6925bf67ea5d5bc6f82d20181

    - removes platform branching by unifying the call site for opening urls within the app
    - adds additional error checking for opening links from inappbrowser's docs
    - sets custom properties for the in-app browser (enable animation, etc)

3. Allowing the user to save a preference of how links behave in https://github.com/StoDevX/AAO-React-Native/commit/60c3d87d87609ac38e12d5a51dddf7db4fc430f8

    - saves in-app preference to async storage
    - allows for toggling of the preference in app miscellaneous settings
    - uses the saved preference within `@frogpond/open-url` module

Note: Links that open in react native's read-only `TextInput` view that display data detectors for highlighting urls, dates, times (e.g. a calendar detail description field), etc will continue to behave (1) as they do in the shipping app today, and (2) autonomously from this setting. It would take some ugly changes on top of `TextInput` for us to react to the onPress event, find a url within the event props, open that link, and stop the default event from propagating. We can try to tackle that in the future if we decide to go after that behavior.

--- 

🎥  Demos

iOS | Android
--|--
 |


modules/open-url/open-url.ts
```diff
@@ -13,7 +15,29 @@ function genericOpen(url: string): Promise<boolean> {
...
+				animated: true,
+				showTitle: true,
+				enableUrlBarHiding: true,
+				enableDefaultShare: true,
+				modalPresentationStyle: 'currentContext',
+			})
+		} else {
+			// fall back to opening in Chrome / Browser / platform default
+			genericOpen(url)
+		}
```
drewvolz: Does this read cleanly enough? I'm also thinking about this variation for the `try` block:

```ts
try {
    if (await InAppBrowser.isAvailable() === false) {
      // fall back to opening in Chrome / Browser / platform default
      genericOpen(url)
      return true
    }

    await InAppBrowser.open(url, {
      animated: true,
      showTitle: true,
      enableUrlBarHiding: true,
      enableDefaultShare: true,
      modalPresentationStyle: 'currentContext',
    })
} catch (error) {
    ...
}
```

drewvolz: The current implementation may be easier to follow if less returns were added.

modules/open-url/open-url.ts
```diff
@@ -13,7 +15,29 @@ function genericOpen(url: string): Promise<boolean> {
...
+			await InAppBrowser.open(url, {
+				animated: true,
+				showTitle: true,
+				enableUrlBarHiding: true,
+				enableDefaultShare: true,
+				modalPresentationStyle: 'currentContext',
+			})
+		} else {
+			// fall back to opening in Chrome / Browser / platform default
+			genericOpen(url)
```
hawkrives:
```suggestion
			await genericOpen(url)
```
?

drewvolz: Nice catch!

hawkrives (approved):

PR #6407, opened 2022-09-16
Bus Updates

rye:
## Express Bus

- Removed the Express Care Clinic stop, it appears.

The rest of the schedule appears to run on the same times, just with one less stop. I deleted the "column" of times (all ending in `:13`) and that seems to have done the trick.

## Red Line

- Dollar Tree and the YMCA were added.
- Allina Clinic and one of the stops at the Senior Center were removed.

I could use some manual review of this one, as I didn't bother to update the times. I spot-checked, and it _looked_ like all the gaps lined up correctly, but I'm not 100% sure.

## Blue Line

- [Econo Foods is renamed to Family Fare](https://www.southernminn.com/northfield_news/news/new-name-new-look-northfield-econo-planning-remodel/article_f2dd8fc2-bb79-5697-8f7f-4005904a3d48.html). (remodeling)

### DIFF
--- a/data/bus-times/blue-line.yaml
+++ b/data/bus-times/blue-line.yaml
@@ -18,7 +18,7 @@ schedules:
       'Viking Terrace':                  [44.469744, -93.163290]
     stops:
       - City Hall on Washington
-      - Econo Foods
+      - Family Fare
       - Carleton College
       - Northfield Estates
       - Viking Terrace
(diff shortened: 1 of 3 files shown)


hawkrives (approved):

drewvolz (approved):

drewvolz:
I’m trying to remember if we change stop name if the coordinate name has to reflect that, too. There might be a look up that gets the coordinate based on the stop name in source/views/transportation/bus/lib/process-bus-line.ts

PR #6412, opened 2022-09-18
[streams part 1] Refactor streams data fetching with useAsync

drewvolz:
- Part 1 of 2 of modifying the streams refactor data 
- See second half over at #6413 

These changes refactor streams list to modernize the data fetching strategy to utilize a `useFetch` call.


source/views/streaming/streams/list.tsx
```diff
@@ -84,38 +64,33 @@ export const StreamListView = (): JSX.Element => {
...
+			}
 			contentContainerStyle={styles.contentContainer}
-			keyExtractor={keyExtractor}
-			onRefresh={refresh}
-			refreshing={refreshing}
-			renderItem={renderItem}
+			keyExtractor={(item: StreamType) => item.eid}
+			onRefresh={reload}
+			refreshing={isPending && !isInitial}
+			renderItem={({item}: {item: StreamType}) => <StreamRow stream={item} />}
```
hawkrives: Do you need the type annotation here, or is it smart enough to understand the following?

```suggestion
			renderItem={({item}) => <StreamRow stream={item} />}
```

drewvolz: It is smart enough to understand that. Good suggestion.

hawkrives (approved):

source/views/streaming/streams/list.tsx
```diff
@@ -23,52 +23,32 @@ const styles = StyleSheet.create({
...
+	return toPairs(grouped).map(([title, data]) => ({title, data}))
+}
 
-	let getStreams = async (
-		reload?: boolean,
-		date: Moment = moment.tz(timezone()),
-	) => {
-		let dateFrom = date.format('YYYY-MM-DD')
-		let dateTo = date.clone().add(2, 'month').format('YYYY-MM-DD')
+const useStreams = (date: Moment = moment.tz(timezone())) => {
```
rye, line 64: Any particular reason to prefer a `const` `=>` function here?

rye, line 64: What is the return type of this function? I feel like it should be explicit if possible. If you add that, then the call to `useStreams` inside the `StreamListView` definition could get its own type annotation so that the `data`, `error`, `reload`, `isPending`, `isInitial`, and `isLoading` fields aren't just getting pulled out of thin air.

rye (dismissed):

source/views/streaming/streams/list.tsx
hawkrives: As opposed to `function useStreams(date: Moment = moment.tz(timezone())) {`? Not that I can think of.

The main reason to default to arrow functions is that they behave more like you'd expect wrt. `this` (they inherit `this` from the caller's context, instead of the defintion location, IIRC). So there are probably fewer pitfalls if we do everything as `const/let =>`…

hawkrives: It'd just be `typeof useFetch<StreamType[]>`, I think. `useFetch` is doing all of the hard work here, and it's exposing the types to TS' type inference properly IIRC, so those properties aren't quite as invisible as they seem at first.

drewvolz: Right, we're taking advantage of type inference. 

The types can be viewed and become less magical when looking at [react-async's useFetch](https://docs.react-async.com/api/interfaces?q=usefetc#usefetch-hook) documentation. Hovering over each of these types reflects the [type information that is within useFetch's state](https://docs.react-async.com/api/state).

drewvolz:
@rye did your questions get answered in this PR?

PR #6413, opened 2022-09-18
[streams part 2] add category filtering to streams

drewvolz:
I recommend [viewing this diff without whitespace](https://github.com/StoDevX/AAO-React-Native/pull/6413/files?diff=split&w=1)

---

- Part 2 of 2 of modifying the streams refactor data 
- Dependent upon #6412 
- Part of #6142 

These changes refactor streams list to accomplish a few ideas:

* add filter bar to show stream categories
* show athletics streams by default

📸 Screenshots

~ | ~
--|--
Simulator Screen Shot - iPhone 12 Pro - 2022-08-29 at 20 38 38 | Simulator Screen Shot - iPhone 12 Pro - 2022-08-29 at 20 38 52


source/views/streaming/streams/list.tsx
```diff
@@ -28,6 +29,24 @@ const groupStreams = (entries: StreamType[]) => {
...
+	return streams.filter((stream) => {
+		let enabledCategories = getEnabledCategories(filters)
+
+		if (enabledCategories.length === 0) {
+			return stream
+		}
+
+		return enabledCategories.includes(stream.category)
+	})
+}
```
hawkrives:
```suggestion
const filterStreams = (streams: StreamType[], filters: ListType[]) => {
	let enabledCategories = getEnabledCategories(filters)

	if (enabledCategories.length === 0) {
		return streams
	}

	return streams.filter((stream) => enabledCategories.includes(stream.category))
}
```

source/views/streaming/streams/list.tsx
```diff
@@ -47,23 +66,53 @@ const useStreams = (date: Moment = moment.tz(timezone())) => {
...
+				// force title-case on the stream types, to prevent not-actually-duplicate headings
+				category: titleCase(stream.category),
+				date: date,
+				$groupBy: group,
+			}
+		})
+	}, [data])
+
+	React.useEffect(() => {
+		let allCategories = data.flatMap((stream) => titleCase(stream.category))
```
hawkrives:
```suggestion
		let allCategories = data.map((stream) => titleCase(stream.category))
```

hawkrives (changes requested):

source/views/streaming/streams/list.tsx
drewvolz: That does look nicer. Let's go with yours.

hawkrives (approved):

drewvolz:
@rye I'm hoping that this split pull request is helpful for review compared to the original. Let me know if this hasn't addressed your original comments.

source/views/streaming/streams/list.tsx
```diff
@@ -28,6 +29,22 @@ const groupStreams = (entries: StreamType[]) => {
 	return toPairs(grouped).map(([title, data]) => ({title, data}))
 }
 
+const getEnabledCategories = (filters: ListType[]) => {
+	return filters.flatMap((filter: ListType) =>
+		filter.spec.selected.flatMap((spec) => spec.title),
+	)
+}
+
+const filterStreams = (streams: StreamType[], filters: ListType[]) => {
```
rye, line 54: The type of `filters` (`ListType`) doesn't make sense to me when I am reading this. What is that? Are we filtering lists? Can "lists" be passed as filters? I feel like this is an upstream issue with `@frogpond/filter`, so we can ignore that, but that's why I was confused by this code.

(Also the `Type` suffix should get dropped from types as a general rule. Types are types, yes. And a more TypeScript-y way would have us make these full-fledged interfaces, I think, but that's a wide-scale change.)

source/views/streaming/streams/list.tsx
```diff
@@ -28,6 +29,22 @@ const groupStreams = (entries: StreamType[]) => {
 	return toPairs(grouped).map(([title, data]) => ({title, data}))
 }
 
+const getEnabledCategories = (filters: ListType[]) => {
```
rye, line 47: Any particular reason to use a `const` `=>` function here?

source/views/streaming/streams/list.tsx
```diff
@@ -28,6 +29,22 @@ const groupStreams = (entries: StreamType[]) => {
...
+	return filters.flatMap((filter: ListType) =>
+		filter.spec.selected.flatMap((spec) => spec.title),
```
rye: Can we give a `let` binding and a type to some part of this? What type is `filter.spec`? What type is `filter.spec.selected`? No need to be super terse here, imo. If `@frogpond/filter` doesn't expose the necessary types, then we should make it do so, so I'm happy to ignore this if that is the case.

I.e., can we use

```typescript
let filterSelections: /* sometype */ = filter.spec.selected;
filterSelections.flatMap((spec) => spec.title);
```

source/views/streaming/streams/list.tsx
```diff
@@ -47,23 +64,53 @@ const useStreams = (date: Moment = moment.tz(timezone())) => {
...
+			let group = stream.status.toLowerCase() !== 'live' ? dateGroup : 'Live'
+
+			return {
+				...stream,
+				// force title-case on the stream types, to prevent not-actually-duplicate headings
+				category: titleCase(stream.category),
+				date: date,
+				$groupBy: group,
+			}
+		})
```
rye: Could the body of the `map` be turned into a real function? It seems to be doing a fair bit of work.

rye (dismissed):

source/views/streaming/streams/list.tsx
hawkrives: You don't like relying on type inference here? (Just asking for clarity.)

source/views/streaming/streams/list.tsx
hawkrives: I'd be happy with that too.

source/views/streaming/streams/list.tsx
hawkrives: I want to say that back in Flow, there wasn't as much separation between the "type namespace" and "value namespace", so I tended to suffix things with `Type$`. This is my fault for bad API design. We should fix it upstream, yeah.

source/views/streaming/streams/list.tsx
rye: I am fine with type inference, but at more than 2 or 3 levels into a given type I'm left scratching my head trying to figure out which type is being used.

source/views/streaming/streams/list.tsx
drewvolz: Sure, happy to extract this.

source/views/streaming/streams/list.tsx
drewvolz: Borrowing from Hawken's earlier answer in this comment https://github.com/StoDevX/AAO-React-Native/pull/6412#discussion_r979440572. I see it as helpful in avoiding pitfalls when dealing with `this`and caller context.

source/views/streaming/streams/list.tsx
hawkrives: Let's come back around to this once we improve the API of the filters?

source/views/streaming/streams/list.tsx
drewvolz: Tracking this issue at https://github.com/StoDevX/AAO-React-Native/issues/6435

drewvolz:
@rye anything left here?

PR #6553, opened 2022-11-06
🚌 transportation list and data refactor

drewvolz:
Refactoring to update data and provide a sleeker transportation view.

## Changes
Replaces transportation detail screen with `openUrl`
* 963b7b4 allow a flexible amount of detail row lines in other modes list
* 57c3545 replace other-modes detail screen with openUrl

Data updates
* f6912bf add INeedARide alias in place of ride-board
* a8f2bd9 update express bus url
* f299289 link to all transport options

demo | screenshot
--|--
demo | screenshot


rye (commented):
Will this change (and subsequent data deployments) cause crashes on the existing client? We have a large skew at the moment that we should be careful of. Put another way, if our old code is expecting to see the `description` key and we stop publishing data there, how does the current version of the app behave?

drewvolz:
You make a good point that our data isn't versioned.

From the testing I've done, the Markdown component that renders a transportation mode's `description` handles this situation well and the current version of the app behaves fine. The description is encapsulated in the `synopsis` prop and so what little we remove in the shipping app can be found either in that prop or the in the linked website.

So what happens is we're not displaying a description on the detail screen, but that seems fine.

hawkrives (approved):

rye (approved):
Okay, thanks for checking!

drewvolz:
I'll do one last check to make sure we are okay to go without that prop, though. I realize I was verifying this with the latest app and not the shipping app so let me check again.

We can verify this by:
1. Either checking out and building the 2.7.0+46090 build or installing the App Store version 
2. Rewriting the response that comes back from the server to exclude the data.

This should be a more real world approach to see if we are good. I'll follow up.

drewvolz:
Okay we can mitigate the risk for anything that may expect `description` to be provided. This has been updated to play it safe by providing the prop as an empty string. Thanks, @rye!

PR #6571, opened 2022-11-11
Show department results natively in directory view

### LINKED ISSUE #6429: Open department links in directory search
They open a url at the moment. This is a search where we could intercept into a new directory search.

drewvolz:
- Closes https://github.com/StoDevX/AAO-React-Native/discussions/6430
- Closes #6429 

This PR accomplishes a few ideas:
- [x] Tapping on a department fetches department data and renders it natively
- [x] Navigating to a department pushes a new view on to the stack
- [x] Navigating between Query and Department listings works as expected

old | new
--|--
 old| new

### DIFF
--- a/source/views/directory/list.tsx
+++ b/source/views/directory/list.tsx
@@ -16,14 +16,31 @@ import {NoticeView, LoadingView} from '@frogpond/notice'
 import {formatResults} from './helpers'
 import {useFetch} from 'react-async'
 import {List, Avatar} from 'react-native-paper'
-import type {DirectoryItem, SearchResults} from './types'
+import type {DirectorySearchTypeEnum, DirectoryItem, SearchResults} from './types'
 import Icon from 'react-native-vector-icons/Ionicons'
-import {useNavigation} from '@react-navigation/native'
+import {RouteProp, useNavigation, useRoute} from '@react-navigation/native'
 import {NativeStackNavigationOptions} from '@react-navigation/native-stack'
-import {ChangeTextEvent} from '../../navigation/types'
+import {ChangeTextEvent, RootStackParamList} from '../../navigation/types'
 
-const useDirectory = (query: string) => {
-	const url = `https://www.stolaf.edu/directory/search?format=json&query=${query.trim()}`
+const getDirectoryUrl = (query: string, type: DirectorySearchTypeEnum) => {
+	let baseUrl = 'https://www.stolaf.edu/directory/search?format=json'
+
+	switch (type) {
+		case 'Query':
+			return `${baseUrl}&query=${query.trim()}`
+		case 'Department': {
+			let formattedDepartment = query.split(' ').join('+')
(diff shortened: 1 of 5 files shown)


rye (approved):

source/views/directory/list.tsx
```diff
@@ -16,29 +16,69 @@ import {NoticeView, LoadingView} from '@frogpond/notice'
...
+} from '@react-navigation/native-stack'
+import {ChangeTextEvent, RootStackParamList} from '../../navigation/types'
+
+const getDirectoryUrl = (query: string, type: DirectorySearchTypeEnum) => {
+	let baseUrl = 'https://www.stolaf.edu/directory/search?format=json'
+
+	switch (type) {
+		case 'Query':
+			return `${baseUrl}&query=${query.trim()}`
+		case 'Department': {
```
hawkrives, line 36: Add case Major?

source/views/directory/list.tsx
```diff
@@ -47,14 +87,16 @@ export function DirectoryView(): JSX.Element {
...
+	} = useDirectory(searchQuery, searchQueryType)
 
 	React.useLayoutEffect(() => {
 		navigation.setOptions({
 			headerSearchBarOptions: {
 				barTintColor: c.white,
-				onChangeText: (event: ChangeTextEvent) =>
-					setTypedQuery(event.nativeEvent.text),
+				onChangeText: (event: ChangeTextEvent) => {
+					setSearchQueryType('Query')
```
hawkrives, line 107: What happens visually if you're looking at some department results, and start typing?

drewvolz: It behaves the same as searching a Query -- no difference in that UI case.

source/views/directory/list.tsx
drewvolz: I don't really have a way to test that state or that endpoint -- do we want to go down supporting that if we're not using it?

source/views/directory/list.tsx
drewvolz: I have a more in-depth answer:

1. Search > result detail > tap `Department` > `Query` search
    - Initially (debounced by 500ms):
      - Results replace placeholder screen
      - Navigation is unchanged
    - Search (debounced by 500ms):
      - Old results are replaced by new results
      - Navigation is unchanged
    - Tapping cancel:
       - Results are cleared
       - Navigation stack is intact -- are we wanting to cache results to restore them in this case?

source/views/directory/list.tsx
drewvolz: I've now added support for
- department
- first name
- last name
- major
- query
- title
- username

source/views/directory/list.tsx
drewvolz: @hawkrives let me know if there's anything left to change here, otherwise we can tackle additions in a follow-up.

hawkrives (approved):

PR #6573, opened 2022-11-12
Update xcode version to 14.1

drewvolz:
We're now using Xcode 14.1 to build on our machines so we should be able to set CI/CD to match.

- bump xcode to 14.1
- bump macOS to 12
- remove mentions of bitcode

Here are the available versions for future reference:
- https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md#xcode


hawkrives (approved):

drewvolz:
> xcode-select: error: invalid developer directory '/Applications/Xcode_14.1.app'

PR #6593, opened 2022-11-20
Loosen restrictions on TypeScript error codes

### LINKED ISSUE #6592: TypeScript errors are not guaranteed to have 4 digits
The RegEx used to splice out TypeScript error codes in the Makefile as implemented in db87bbb95c669e4c44c25ae73e4db5af8f0273fc is missing a (newly-introduced?) 5-digit error code with the TypeScript v4.9.3 upgrade. We should not depend on the number of digits following the TS prefix.

When adjusting #6586 to work, I noticed the following changes:

```diff
diff --git a/tsc-counts b/tsc-counts
index 74a4b8183..7474bf72f 100644
--- a/tsc-counts
+++ b/tsc-counts
@@ -5,5 +5,5 @@
 2 TS2339
 1 TS7053
 1 TS7016
-1 TS2571
 1 TS2304
+1 TS1804
diff --git a/tsc-errors b/tsc-errors
index 97b418f81..381959006 100644
--- a/tsc-errors
+++ b/tsc-errors
@@ -14,7 +14,7 @@ source/lib/login.ts(26,2): error TS2322: Type 'Promise<false | {} | SharedWebCre
 source/lib/login.ts(29,2): error TS2322: Type 'Promise<void | {}>' is not assignable to type 'Promise<void>'.
   Type 'void | {}' is not assignable to type 'void'.
     Type '{}' is not assignable to type 'void'.
-source/lib/login.ts(77,29): error TS2571: Object is of type 'unknown'.
+source/lib/login.ts(77,29): error TS18046: 'err' is of type 'unknown'.
 source/lib/partition-by-index.ts(5,22): error TS2769: No overload matches this call.
   Overload 1 of 2, '(...items: ConcatArray<never>[]): never[]', gave the following error.
     Argument of type 'T' is not assignable to parameter of type 'ConcatArray<never>'.
```

An error code like `TS18046` is missed by the regex, breaking the `tsc-counts` check and causing these to be misattributed as another error code entirely. While I think these will still be included in the totals, this is confusing.

rye:
With the upgrade to TypeScript v4.9.3, a new 5-digit error is generated (TS18046) which gets snipped by the old `\d{4}` regex component which does not include any word-boundary detection.  This leads to a loss of information when trying to generate the list of TS error codes, and ultimately programmer confusion.

Fixes #6592.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ tsc-errors:
 	-npx tsc | sed "s:$$(git rev-parse --show-toplevel):.:g" | tee ${@}
 
 tsc-counts: tsc-errors
-	-cat ${<} | rg -or '$$1' 'error (TS\d{4})' | sort | uniq -c | sort -nr | sed 's/^ *//' | tee ${@}
+	-cat ${<} | rg -or '$$1' 'error (TS\d+)' | sort | uniq -c | sort -nr | sed 's/^ *//' | tee ${@}
 
 eslint-problems:
 	-npm run lint | sed "s:$$(git rev-parse --show-toplevel):.:g" | tee ${@}
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ tsc-errors:
 	-npx tsc | sed "s:$$(git rev-parse --show-toplevel):.:g" | tee ${@}
 
 tsc-counts: tsc-errors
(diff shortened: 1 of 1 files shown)


hawkrives (approved):

drewvolz:
Great catch with these typescript regex changes, Kris!

PR #6761, opened 2023-01-21
Speed up GitHub Actions

hawkrives:
I've been annoyed at how long our CI pipeline takes to complete. On top of that, if our e2e tests fail, we have to completely recompile the app if we want to try again!

I wanted to fix that. 

> A quick side note: GitHub Actions don't support YAML anchors. I wish they did; it would make sharing cache keys much easier!

## Changes
- Upgrade Linux runners from `ubuntu-20.04` to `ubuntu-22.04`
- Increase granularity of iOS build pipeline by taking advantage of caches 
- Cache the node_modules folder, keyed off of the hash of package-lock.json, completely avoiding `npm ci` on most jobs (e.g., Eslint can just extract one giant tarball; it doesn't need to invoke npm at all.)
- Cache the Cocoapods install folder, keyed off of the hashes of package-lock.json and Podfile.lock
- Cache the compiled iOS app, keyed off of the hashes of package-lock.json, Podfile.lock, and project.pbxproj, to share the app binary between "compile" and "test"
- Cache the .jsbundle to save time
- Inject the .jsbundle into the compiled app, to avoid recompiling when only JS code has changed

## Outcome

> - "Before" times taken from https://github.com/StoDevX/AAO-React-Native/pull/6768
> - "After (no cache)" taken from https://github.com/StoDevX/AAO-React-Native/actions/runs/3987103848
> - "After (cached)" taken from [tbd]

&nbsp; | Before | After
--- | --- | ---
Duration - Wall Clock | 33m 47s | |
Duration - Sum | 51m 40s | |
ESLint | 58s | |
Validate Data | 36s | |
Jest | 1m 25s | |
Prettier | 51s | |
tsc | 46s | |
Android Bundle | 2m 44s | |
Android Build | 7m 43s |
iOS Bundle | 3m 52s | |
iOS Cocoapods Check | 3m 14s | |
iOS Build | 29m 31s | |
iOS Detox | ~(9m 21s) | |


hawkrives:
test comment from local machine

PR #6769, opened 2023-01-23
Support mocking functionality for detox builds

drewvolz:
We are timing out in our Detox tests due to timers going on within Sentry.

When we call `NavigationContainer.onReady` in `App` and instantiate Sentry's instrumentation, we end up in a place where timers are not finishing and we end up timing out our Detox tests.

We can handle this by doing the following:
1. Set up a metro resolver for files ending in `.mock.ts` when `APP_MODE=mocked`
2. Stub out calls and values in the `.mock.ts` file
3. Let metro handle resolving things when it comes across `process.env.APP_MODE='mocked'`

https://github.com/wix/Detox/blob/master/docs/guide/mocking.md

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -510,7 +510,7 @@ jobs:
         if: steps.app-cache.outputs.cache-hit != 'true'
         env:
           SKIP_BUNDLING: 'true'
-        run: APP_MODE=mocked npx detox build e2e --configuration ios.sim.release
+        run: npx detox build e2e --configuration ios.sim.release
 
       - name: Cache the iOS app
         uses: actions/cache/save@58c146cc91c5b9e778e71775dfe9bf1442ad9a12 # v3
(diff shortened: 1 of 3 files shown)


.github/workflows/check.yml
```diff
@@ -510,7 +510,7 @@ jobs:
         if: steps.app-cache.outputs.cache-hit != 'true'
         env:
           SKIP_BUNDLING: 'true'
-        run: npx detox build e2e --configuration ios.sim.release
+        run: APP_MODE=mocked npx detox build e2e --configuration ios.sim.release
```
hawkrives: Don't need this one

.github/workflows/check.yml
```diff
@@ -579,7 +579,7 @@ jobs:
       - run: brew install applesimutils
 
       - name: Run the Detox tests
-        run: npx detox test e2e --configuration ios.sim.release --cleanup --record-logs failing --record-videos failing --record-performance all --capture-view-hierarchy enabled --take-screenshots failing
+        run: APP_MODE=mocked npx detox test e2e --configuration ios.sim.release --cleanup --record-logs failing --record-videos failing --record-performance all --capture-view-hierarchy enabled --take-screenshots failing
```
hawkrives: Don't need this one

hawkrives (changes requested):

hawkrives (approved):

PR #6774, opened 2023-01-23
Remove "glamorous-native"

hawkrives:
I wanted to remove the `glamorous-native` dependency. It's unmaintained, and one of the sources of warnings that we see in the console. 

I'll deal with the react-markdown stuff in a separate PR.


modules/markdown/markdown.tsx
```diff
@@ -22,13 +23,20 @@ import {Link} from './link'
 import {Image} from './image'
 import {List, ListItem} from './list'
 
+const styles = StyleSheet.create({
+	// eslint-disable-next-line react-native/no-color-literals
```
hawkrives: I don't know why line 27 is the right spot for this comment, when it affects line 31 or something, but I've learned not to question ESLint.

modules/markdown/markdown.tsx
```diff
@@ -37,7 +45,7 @@ export type MarkdownProps = {
 		CodeBlock?: StyleProp<TextStyle>
 		Emph?: StyleProp<TextStyle>
 		Heading?: StyleProp<TextStyle>
-		Image?: StyleProp<ViewStyle>
+		Image?: StyleProp<ImageStyle>
```
hawkrives, line 48: While this is "more correct" now… let's not try actually rendering anything, I'm pretty sure that's broken. (The props of img to Image are pretty different)

modules/markdown/markdown.tsx
drewvolz: Do we want to instead import c.black from frogpond/colors?

source/views/settings/screens/credits.tsx
```diff
@@ -27,45 +56,27 @@ const markdownStyles: MarkdownProps['styles'] = {
...
-		android: material.body1Object,
-	}),
-	textAlign: 'center',
-})
+const Contributors = (props: TextProps) => (
+	<Text {...props} style={[styles.contributors, props.style]} />
+)
 
 const formatPeopleList = (arr: Array<string>) =>
 	arr.map((w) => w.replace(' ', ' ')).join(' • ')
```
drewvolz, line 76: What is going on with the replace call?

drewvolz (approved):

source/views/settings/screens/credits.tsx
hawkrives: I believe it replaces spaces with non-breaking spaces, so that people's names don't wrap?

rye (approved):

PR #6789, opened 2023-01-25
Support Dark Mode on iOS and Android

hawkrives:
- WIP rip out our custom theming solution and use PlatformColor and react-navigation's theme
- rework a lot of colors


modules/lists/list-row.tsx
```diff
@@ -3,12 +3,14 @@ import {PropsWithChildren} from 'react'
...
 import {Touchable} from '@frogpond/touchable'
 import {DisclosureArrow} from './disclosure-arrow'
+import * as c from '@frogpond/colors'
 
 const styles = StyleSheet.create({
 	childWrapper: {
 		flex: 1,
 	},
 	container: {
+		backgroundColor: c.secondarySystemGroupedBackground,
```
hawkrives, line 13: Are we able to remove this one now?

source/views/contacts/list.tsx
```diff
@@ -12,7 +12,7 @@ import type {ContactType} from './types'
 
 const styles = StyleSheet.create({
 	listContainer: {
-		backgroundColor: c.white,
+		backgroundColor: c.systemBackground,
```
hawkrives: I'm still not sure when to use these semantic colors - is this systemBackground, or secondarySystemGroupedBackground?

source/views/transportation/other-modes/list.tsx
```diff
@@ -9,7 +9,7 @@ import {LoadingView, NoticeView} from '@frogpond/notice'
 
 const styles = StyleSheet.create({
 	listContainer: {
-		backgroundColor: c.white,
+		backgroundColor: c.systemBackground,
```
hawkrives, line 12: You found the source!

modules/lists/list-row.tsx
drewvolz: Might need more work if we want to remove this. In dark theme, I'm seeing this not work well without this.

source/views/contacts/list.tsx
drewvolz: You'e right, we should be using `secondarySystemGroupedBackground` for these since that matches the background color of the underlying view.

drewvolz (approved):

Issue #6788, opened 2023-01-24
E2E tests should test that every view is visible


hawkrives:
extracted from #4013

---

Courtesy of @drewvolz, I'm adding a to-do list of things I should look at:

### Home
- [x] List of buttons

### Menus
- [x] Stav Hall → #6788
- [x] Cage → #6788
- [x] Pause → #6788
- [x] Burton → #6788
- [x] LDC → #6788
- [x] Weitz → #6788
- [x] Sayles → #6788
- [ ] Menu Item Detail

### SIS
- [x] Balances → #6788
- [x] Open Jobs List → #6788
- [ ] Open Jobs Detail

### Catalog
- [x] Course Search Home → #6788
- [ ] Course Search List
- [ ] Course Search Detail

### Building hours
- [x] Hours List → #6788
- [ ] Hours Detail
- [ ] Hours Problem Report
- [ ] Hours Schedule Editor

### Calendar
- [x] St. Olaf Calendar → #6788
- [x] Oleville Calendar → #6788
- [x] Northfield Calendar → #6788
- [ ] Calendar Event Detail

### Directory
- [x] Directory List → #6788
- [ ] Directory Contact Detail

### Streaming Media
- [x] Events calendar → #6788
- [x] Webcams → #6788
- [x] KSTO Radio → #4013
- [x] KRLX Radio → #4013
- [ ] Radio Schedule

### News
- [x] St. Olaf List → #6788
- [ ] Oleville List
- [ ] Mess List
- [ ] PoliticOle List (if we put it back in)
- [ ] KSTO List

### Map
- [x] Link to Browser → #6788

### Important Contacts
- [x] Contacts List → #6788
- [ ] Contacts Detail

### Transit
- [x] Express Bus
- [x] Red Line
- [x] Blue Line
- [x] Oles Go
- [x] Other Modes
- [ ] Other Modes Detail

### Dictionary
- [x] Dictionary List → #6788
- [ ] Dictionary Detail
- [ ] Dictionary Editor

### Student Orgs
- [x] Orgs List → #6788
- [ ] Orgs Detail

### stoPrint
- [ ] Printer List
- [ ] Print Job Release

### More
- [x] A-Z list → #6788

### Settings
- [x] List of settings
- [ ] Login
- [ ] Contact Us
- [ ] FAQs
- [ ] Reset
- [x] Icon Settings
- [ ] Credits
- [ ] Privacy Policy
- [ ] Legal
[...]

PR #6800, opened 2023-01-27
Add Detox tests for each top-level "module" and tab

hawkrives:
- add detox tests for each module/tab
- test out running multiple simulators on the same macos box (seeing if running 3 simulators is faster than 1, since the macOS boxes are tri-core)

Part of #6788


drewvolz (approved):

hawkrives:
Before I can bring this out of Draft, I need:

- [ ] detox to work on CI
- [ ] to implement shorter timeouts for our tests - no single test takes over 10s, so I propose that instead of the current 120s
- [ ] to figure out if one, two, or three parallel runners work best
- [ ] to figure out if we should shard the runners across machines too

drewvolz (approved):

drewvolz:
As far as I can tell, CI should be passing if we weren't using live data for the menus (Carleton menus show "Closed for spring break")

hawkrives:
Closing in favor of https://github.com/StoDevX/AAO-React-Native/pull/7383

PR #6802, opened 2023-01-27
Revert "Bump @sentry/react-native from 4.13.0 to 4.14.0"

hawkrives:
Reverts StoDevX/AAO-React-Native#6792


hawkrives:
Thanks to @drewvolz, we know that androidx is to blame, not sentry.

PR #6801, opened 2023-01-27
Remove manual linkage of react-native-datepicker

hawkrives:
Part of #6798

As suggested by @drewvolz,

> Our first step could be trying to cherry-pick e7f48ab out of the rn-71 feature branch. It removes the manual linkage of date picker.

I have cherry-picked that commit onto this branch. Let's see what happens!

***

Depends on https://github.com/StoDevX/AAO-React-Native/pull/6802

PR #6806, opened 2023-01-27
drew/gradle lock

### LINKED ISSUE #6798: Fix Android builds
## Device(s) with issues
- Google Pixel 6 (oriole) - Android 12
- Google Pixel 5 (redfin) - Android 11
- Samsung SM-G981U1 (x1q) - Android 10
- Google Pixel 2 (Pixel2Emulator) - Android 12
- Google Redfin 64-bit only (redfin_64) - Android 13
- Google Pixel 2 (Pixel2Emulator) - Android 9
- Google Pixel 2 (Pixel2Emulator) - Android 13

## Exceptions
- java.lang.IllegalStateException: Native module RNDatePickerAndroid tried to override RNDatePickerDialogModule. Check the getPackages() method in MainApplication.java, it might be that module is being created twice. If this was your intention, set canOverrideExistingModule=true. This error may also be present if the package is present only once in getPackages() but is also automatically added later during build time by autolinking. Try removing the existing entry and rebuild.
- java.lang.IllegalStateException: Native module RNDatePickerAndroid tried to override RNDatePickerDialogModule. Check the getPackages() method in MainApplication.java, it might be that module is being created twice. If this was your intention, set canOverrideExistingModule=true. This error may also be present if the package is present only once in getPackages() but is also automatically added later during build time by autolinking. Try removing the existing entry and rebuild.
- java.lang.IllegalStateException: Native module RNDatePickerAndroid tried to override RNDatePickerDialogModule. Check the getPackages() method in MainApplication.java, it might be that module is being created twice. If this was your intention, set canOverrideExistingModule=true. This error may also be present if the package is present only once in getPackages() but is also automatically added later during build time by autolinking. Try removing the existing entry and rebuild.
- java.lang.IllegalStateException: Native module RNDatePickerAndroid tried to override RNDatePickerDialogModule. Check the getPackages() method in MainApplication.java, it might be that module is being created twice. If this was your intention, set canOverrideExistingModule=true. This error may also be present if the package is present only once in getPackages() but is also automatically added later during build time by autolinking. Try removing the existing entry and rebuild.
- java.lang.IllegalStateException: Native module RNDatePickerAndroid tried to override RNDatePickerDialogModule. Check the getPackages() method in MainApplication.java, it might be that module is being created twice. If this was your intention, set canOverrideExistingModule=true. This error may also be present if the package is present only once in getPackages() but is also automatically added later during build time by autolinking. Try removing the existing entry and rebuild.
- java.lang.IllegalStateException: Native module RNDatePickerAndroid tried to override RNDatePickerDialogModule. Check the getPackages() method in MainApplication.java, it might be that module is being created twice. If this was your intention, set canOverrideExistingModule=true. This error may also be present if the package is present only once in getPackages() but is also automatically added later during build time by autolinking. Try removing the existing entry and rebuild.
- java.lang.IllegalStateException: Native module RNDatePickerAndroid tried to override RNDatePickerDialogModule. Check the getPackages() method in MainApplication.java, it might be that module is being created twice. If this was your intention, set canOverrideExistingModule=true. This error may also be present if the package is present only once in getPackages() but is also automatically added later during build time by autolinking. Try removing the existing entry and rebuild.

drewvolz:
We can add gradle dependency locking to the android side to prevent headaches of upstream updates (in particular, androidx) from bumping beta versions and causing build breakages across builds.

This introduces a lockfile that we must keep up to date for our android project's health. It adds a locking configuration  through nebula, [a gradle plugin by Netflix](http://nebula-plugins.github.io). This allows projects using dynamic dependency versions to lock them to specific versions.

- Resolves #6798 
- Supplants #6801
- Related to #6546
- Related to #6547

## General use

This is verbatim recommended guidance [in the wiki](https://github.com/nebula-plugins/gradle-dependency-lock-plugin/wiki/General-Use-Pattern).

### Todo
1. Daily locking of dynamic dependencies
    - [ ] Run a daily job to lock your dependencies. Every day before you show up to work you can have an automated system to pull in the latest dependencies and run your tests. If they pass, your team has a stable dependency graph for the day. If they fail you can have a subset of the team deal with the dependency issues while the rest keep with what they're were doing using yesterday's dependency lock.
    - [ ] Occasionally (and optionally) run `updateLock` to try out a single new dependency.
2. Lock on release
    - [ ] When you're ready to cut that shiny new version of your software create a lock of your dependencies. You'll be able to recreate the exact state of your dependency graph if you ever need to reproduce a build.
3. Deal with IDE issues
    - [ ] Create a global lock with `generateGlobalLock` `saveGlobalLock`.
    - [ ] Watch as you no longer have to deal with one subproject depend on an old version and mess up another's dependency graph
    - [ ] Stop WAR deployments from copying 2 different versions of a dependency into the directory.

## Common Workflows

[See usage here.](https://github.com/nebula-plugins/gradle-dependency-lock-plugin/wiki/Usage)

A `buildscript-gradle.lockfile` is created through the following command:

```
 ./gradlew dependencies --write-locks
 ```
 
How to generate lock (workflow)

```
./gradlew generateLock saveLock
./gradlew test

# if test passes
./gradlew commitLock
```

How to update a lock (the lock must still be saved/committed)

```
./gradlew updateLock -PdependencyLock.updateDependencies=com.example:foo,com.example:bar
```

Reverting to normal gradle behavior even with the plugin applied (command line override)

```
./gradlew -PdependencyLock.ignore=true <tasks>
```

## Running a build with lock state present

The moment a build needs to resolve a configuration that has locking enabled and it finds a matching lock state, it will use it to verify that the given configuration still resolves the same versions.

A successful build indicates that the same dependencies are used as stored in the lock state, regardless if new versions matching the dynamic selector have been produced.

The complete validation is as follows:

- Existing entries in the lock state must be matched in the build
  - A version mismatch or missing resolved module causes a build failure
- Resolution result must not contain extra dependencies compared to the lock stat

## Strict mode

While the default lock mode behaves as described above, two other modes are available:

**Strict mode**: In this mode, in addition to the validations above, dependency locking will fail if a configuration marked as locked does not have lock state associated with it.

**Lenient mode**: In this mode, dependency locking will still pin dynamic versions but otherwise changes to the dependency resolution are no longer errors.

## Links

- [Dependency locking user guide](https://docs.gradle.org/current/userguide/dependency_locking.html)
- [nebula.dependency-lock Plugin](https://plugins.gradle.org/plugin/nebula.dependency-lock)
- [General use pattern wiki](https://github.com/nebula-plugins/gradle-dependency-lock-plugin/wiki/General-Use-Pattern)
- [Usage wiki](https://github.com/nebula-plugins/gradle-dependency-lock-plugin/wiki/Usage)

### CODE REVIEW

File: android/app/build.gradle [hunk 1]
```diff
@@ -348,6 +350,7 @@ if (isNewArchitectureEnabled()) {
 
 dependencyLocking {
     lockAllConfigurations()
+    lockMode = LockMode.STRICT
```
drewvolz: Default is not strict and I'm thinking of removing it. The default behavior is described as follows:

> A successful build indicates that the same dependencies are used as stored in the lock state, regardless if new versions matching the dynamic selector have been produced.
> 
> The complete validation is as:
> - Existing entries in the lock state must be matched in the build
>   - A version mismatch or missing resolved module causes a build failure
> - Resolution result must not contain extra dependencies compared to the lock state

That sounds okay to me?

--- 

**Running in strict mode:**

When running `./gradlew generateLock saveLock` I am getting (correct) warnings about configurations that lack locking specifications. That is because we'd need each configuration to list a lock state.


Here is one for `':app:androidApis`

```
* What went wrong:
Execution failed for task ':app:generateLock'.
> Could not resolve all dependencies for configuration ':app:androidApis'.
   > Locking strict mode: Configuration ':app:androidApis' is locked but does not have lock state.
```
or another failure for `:app:lintChecks`

```
Could not resolve all dependencies for configuration ':app:lintChecks'.
Locking strict mode: Configuration ':app:lintChecks' is locked but does not have lock state.
```

drewvolz:
First, testing the removal of strict lockmode. Attempting to get a build without androidx messing us up.

hawkrives (approved):

PR #6831, opened 2023-01-27
Utilize gradle/gradle-build-action

hawkrives:
switch to the gradle/gradle-build-action action – it offers better Gradle caching stuff.

Also, I'm now bypassing Fastlane for the "check" workflow's Android builds, so I got rid of the "cache bundler for Android" job.


android/gradle.properties
```diff
@@ -22,7 +22,8 @@ org.gradle.parallel=true
 # https://developer.android.com/topic/libraries/support-library/androidx-rn
 android.useAndroidX=true
 # Automatically convert third-party libraries to use AndroidX
-android.enableJetifier=true
+# We do not currently have any libraries that require jetifier!
+android.enableJetifier=false
```
drewvolz: 🎉

drewvolz (approved):

PR #6841, opened 2023-01-28
Publish Gradle build scans automatically (and accept Gradle's TOS in CI)

hawkrives:
I've got the Android Build task set up to auto-publish Gradle scans, but we need to accept the Gradle TOS. I've now got that configured to only happen on CI.

Published Gradle scans expire after 90 days.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -418,7 +418,7 @@ jobs:
   
       - name: Run Gradle
         run: ./gradlew bundleRelease --scan
-        working-directory: androind 
+        working-directory: android
         env:
           SENTRY_ORG: frog-pond-labs
           SENTRY_PROJECT: all-about-olaf
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,7 +1,8 @@
 name: Check
 
 on:
-  push:
+  push: {branches: [master]}
+  pull_request:
 
 concurrency:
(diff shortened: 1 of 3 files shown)


drewvolz:
- A Build Scan that has not been viewed within 3 months of being created will be removed.
- Once a Build Scan has been viewed, it will be available indefinitely.

This sounds good, we can delete the ones we've viewed and let the ones we do not need expire.

drewvolz (approved):

hawkrives:
Oh, that's more nuanced than I realized:

image

hawkrives:
This is supposed to only publish on failure. It is publishing on all builds…

maybe because they include deprecations???

hawkrives:
I wonder if `--scan` is overriding publishOnFailure?

I'd rather publish all of them than none of them, so if I haven't figured this out by ~Monday I'll plan to merge this as-is.

hawkrives:
@drewvolz is pretty sure, after digging, that `--scan` overrides the publishOnFailure command inside the gradle file. Let's see!

hawkrives:
I don't care anymore - let's just always upload it… Gradle's the one paying for the hosting, and they expire after 90 days anyway

hawkrives:
Thinking about this more, I would rather publish all scan logs - this way, if a single job is slow, we can dig in right away and see why.

drewvolz (approved):

PR #6843, opened 2023-01-28
add workflow to validate that a gradle wrapper is official

hawkrives:
I still don't know why Gradle has this wrapper thing in the first place, but they have an official GitHub Action to validate it:

https://github.com/marketplace/actions/gradle-wrapper-validation

Seems like a good idea.


hawkrives:
I'm gonna make this one run on push: and not pr:

drewvolz (approved):

PR #6866, opened 2023-01-30
Codecov in GitHub Actions

hawkrives:
I want Codecov to report itself in the GitHub Checks section of a PR. I also made it not report a comment if the coverage has not changed. 

https://docs.codecov.com/docs/github-checks

I also cleaned up the config a bit. 

Finally, I increased the allowable coverage range to >= 8%, because that’s where we are today. As we increase our coverage, we should increase this number!

Example of what the check looks like:

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -5,7 +5,7 @@ coverage:
 
   status:
     project: false
-    patch: false
+    patch: true
     changes: false
 
 parsers:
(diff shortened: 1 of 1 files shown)


drewvolz (approved):

rye (approved):
All these changes make perfect sense to me! I didn't know we could avoid the report if no change in coverage — that's pretty rad!

PR #6889, opened 2023-01-30
switch from patch to project coverage

hawkrives:
image

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -4,8 +4,8 @@ coverage:
   range: 8...100
 
   status:
-    project: false
-    patch: true
+    project: true
+    patch: false
     changes: false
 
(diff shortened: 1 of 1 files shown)


drewvolz (approved):

rye (approved):
I think we should make it required, then? A nice aspect of the patch requirement is that if you are in an area of less coverage than the project as a whole, you're prompted to make things better. Could we keep both checks, but then just make the project check required? Something like that

hawkrives:
That's a good question. I'll see what I can find?

hawkrives:
It looks like it's a "failing" status that doesn't block merge, because it's not required. I'll turn on the /project status as well, and make /project required.

PR #6916, opened 2023-02-05
More home screen personalization

drewvolz:
Top right home button refactored into a ContextMenu
  - iOS Displays a circle ellipses like the Home app
  - Android cannot open this and uses the old settings button
  
iOS Home screen appearance
  - Large title
  - Transparent header
  - Blur effect
  
iOS Home screen list and button changes
  - Home's scrollview is now wrapped with safeareaview
  - The touchable buttons are no longer wrapped with safeareaview
    - This avoids jitteriness when scrolling and retains the large --> small header transition animation

Photo library image picker
  - `react-native-image-picker` is now an app dependency (cross-platform)
  - iOS: User can select an image from their device to use as the view's background
  - The image uri path is persisted to localstorage, removed when app settings are reset

Considerations
- [x] Photo library picker works on iOS
- [x] Settings works on iOS
- [x] Settings work on Android ~Android probably doesn't display a context menu, so settings are inaccessible~ 

Testing
- #6920
- #6952
- [ ] Jest approach

📸 See it in action

https://user-images.githubusercontent.com/5240843/216795928-051e44ad-8264-4ecf-b78b-b37261731a61.mp4


source/views/home/background.ts
```diff
@@ -0,0 +1,60 @@
...
+export const keys = {
+	background: ['settings', 'background'] as const,
+}
+
+async function chooseBackgroundImage(): Promise<string | undefined> {
+	const {assets, didCancel, errorCode, errorMessage} =
+		await ImagePicker.launchImageLibrary({mediaType: 'photo'})
+
+	if (errorCode || errorMessage) {
+		throw new ImagePickerError()
```
drewvolz, line 22: Bubbling these errors would be good, I don't think we're doing that? As far as I can tell, we're not having to provide permissions for this access, either (tested on iOS).

source/views/home/background.ts
```diff
@@ -0,0 +1,60 @@
...
+	return imagePath
+}
+
+type AppBackgroundMutation = UseMutationResult<
+	string | undefined,
+	unknown,
+	void,
+	unknown
+>
+
```
drewvolz, line 40: Is there a cleaner type available to us?

source/views/home/index.tsx
```diff
@@ -21,54 +28,121 @@ const styles = StyleSheet.create({
...
+							))}
+						</Column>
+					))}
+				</View>
 
-			<UnofficialAppNotice />
-		</ScrollView>
+				<UnofficialAppNotice />
+			</ScrollView>
+			{Boolean(appBackgroundImage) && (
```
drewvolz: I also played with conditionally changing display and opacity but we end up getting a warning if we pass in a blank string as the image uri, so I decided to not allow for rendering without an image.

source/views/home/background.ts
hawkrives: IIRC if you use the system-provided file picker, the user doesn't have to provide access because you only get access to the photo(s) they select.

source/views/home/background.ts
```diff
@@ -0,0 +1,60 @@
...
+export const useUpdateAppBackground = (): AppBackgroundMutation => {
+	return useMutation({
+		mutationFn: chooseBackgroundImage,
+		onSuccess: async (imagePath) => {
+			if (imagePath === undefined) {
+				return
+			}
+
+			await storage.setBackgroundPreference(imagePath)
```
hawkrives, line 49: I feel like this one should move into the mutationFn, but I may have to think about that some more.

source/views/home/index.tsx
```diff
@@ -21,54 +28,121 @@ const styles = StyleSheet.create({
...
+					style={styles.background}
+				/>
+			)}
+		</SafeAreaView>
 	)
 }
 
+const HomeNavbarButtonsEnum = {
+	Settings: 'Settings',
+	ChangeBackground: 'Change background',
```
hawkrives:
```suggestion
	ChangeBackground: 'Change Background',
```

I feel like Title Case is most often used in iOS context menus?

drewvolz:
### Updates on the testing effort

As of iOS 11, apple renders the native image picker view out of the view hierarchy process, which means we don't have access to inspect it with detox (PHPickerViewController). 

I can get it to render the picker while testing but it's just flat out not there as far as accessible elements. 

I am able to mock the callbacks, but this means we can't test picking an image. We test tapping if "Change Background" was tapped successfully and if we fire a callback + get data, I can return the test image data I'm mocking if that'd be useful, but there's not much end-to-end testing we can do here.

~|~
--|--
 | 



- https://medium.com/geekculture/how-to-mock-react-native-image-picker-using-detox-9f7ced41b67f
- https://stackoverflow.com/questions/48208456/e2e-select-an-image-from-a-uiimagepickercontroller-with-wix-detox/55151119

---

### What to do?

So we could do one, either, or both.

A. Take screenshots and check that they're different (to show that we pushed a new VC)
B. Mock out the functionality and check that they were called

source/views/home/button.tsx
```diff
@@ -16,23 +17,32 @@ export function HomeScreenButton({view, onPress}: Props): JSX.Element {
...
+
+	return Platform.select({
+		ios: <Internaltouchable />,
+		android: (
+			<SafeAreaView>
+				<Internaltouchable />
+			</SafeAreaView>
+		),
+		default: <></>,
+	})
```
rye: PascalCase the type name? i.e. `InternalTouchable`
```suggestion
	const InternalTouchable = () => (
		<Touchable
			accessibilityLabel={view.title}
			accessibilityRole="button"
			accessible={true}
			activeOpacity={0.9}
			highlight={false}
			onPress={onPress}
			style={[styles.button, {backgroundColor: view.tint}]}
		>
			<View style={styles.contents}>
				<Icon name={view.icon} size={32} style={[foreground, styles.icon]} />
				<Text style={[foreground, styles.text]}>{view.title}</Text>
			</View>
		</Touchable>
	)

	return Platform.select({
		ios: <Internaltouchable />,
		android: (
			<SafeAreaView>
				<InternalTouchable />
			</SafeAreaView>
		),
		default: <></>,
	})
```

source/views/home/button.tsx
```diff
@@ -16,23 +17,32 @@ export function HomeScreenButton({view, onPress}: Props): JSX.Element {
...
+		>
+			<View style={styles.contents}>
+				<Icon name={view.icon} size={32} style={[foreground, styles.icon]} />
+				<Text style={[foreground, styles.text]}>{view.title}</Text>
+			</View>
+		</Touchable>
 	)
+
+	return Platform.select({
+		ios: <Internaltouchable />,
```
rye: Oops
```suggestion
		ios: <InternalTouchable />,
```

rye (changes requested):

source/views/home/background.ts
drewvolz: @hawkrives any further thoughts?

drewvolz:
Putting this on-hold to focus on v2.8 goals. The bar for merging this is getting the mocked detox test integrated with these changes.

PR #6920, opened 2023-02-06
WIP native visual regression testing

drewvolz:
- Targets #6916 

Playing with visual regression changes from screenshots! 

While we could mock the library actions through jest, we’d be missing an opportunity to provide full testing coverage to this feature. 

Testing with XCUITest gives us access to a full e2e flow with native views like the PHPickerViewController from apple that are not rendered within the view hierarchy and cannot be accessed by detox/jest.

Another bonus: PSPDFKit has some pretty cool posts on how to do pixel regression checking with thresholds on images!

- https://pspdfkit.com/blog/2021/swift-render-tests-in-practice-part-3/
- https://pspdfkit.com/blog/2021/swift-render-tests-in-practice-part-2/
- https://pspdfkit.com/blog/2021/swift-render-tests-in-practice-part-1/

—

Todo
- [ ] Regression checking as-is assumes we want images to be the same, but for changing the background image we want them to be different. So the errors thrown revolve around threshold minimums which fail when an image is different up to a certain amount of pixels. We need to invert the expectation that an image should fail if it is different **for this test case**

📸 Screenshots

~ | ~
--|--
09B34699-EA08-4363-AB5A-803D39D469E7 | 1C1A415F-E82B-4401-9024-AD8C054E4947

### DIFF
--- a/e2e/basic-smoke.spec.ts
+++ b/e2e/basic-smoke.spec.ts
@@ -18,11 +18,13 @@ test('should show the home screen', async () => {
 
 test('should show the settings screen after tap', async () => {
 	await element(by.id('button-open-settings')).tap()
+	await element(by.text('Settings')).tap()
 	await expect(element(by.text('Sign In to St. Olaf'))).toBeVisible()
 })
 
 test('should show home screen after tap to exit settings screen', async () => {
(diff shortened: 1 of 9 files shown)


drewvolz:
Closing in favor of #6952

PR #6938, opened 2023-02-12
Fix credentials loading after logging out

### LINKED ISSUE #6928: Settings attempts to login without credentials
Is this a regression from #6899?

https://user-images.githubusercontent.com/5240843/217707053-d55b8550-4cbf-46f0-ac63-611364640a6d.MOV

drewvolz:
Closes #6928 

## Summary
`useCredentials`  calls `loadCredentials` which throws an error when the credentials are false. This blocks the logout state from finishing. We can fix this by changing two things:

1. Remove the `throw` statement 
2. Convince everything else that consumes `useCredentials` that false is an expected value

Working demo can be seen in screen recording attached in #6939 

## Understanding the problem

If useCredentials --> useQuery --> loadCredentials --> throws

It wouldn't get the new values. Ever. Right?

So things are invalidated, which is why we get the new correct UI after dismissing settings

- We **are** throwing for no credentials in signing in, which is good, and our UI prevents us from hitting that
- We only need to throw for no credentials when signing in (performLogin) but not when requesting the credentials
- loadCredentials should never be allowed to throw
- But we need to still guarantee to the consumers that "false" is an okay value

## Testing the fix

1. Add the following routes to ccc-server:

```js
if (process.env.NODE_ENV !== 'production') {
    api.redirect('/login', 'https://stolaf.api.frogpond.tech')
    api.redirect('/login/error', 'https://stolaf.api.frogpond.tech/?error=login')
}
```

2. Replace the `ky` client in `source/lib/login.ts` with our local client

```diff
+import { client } from '@frogpond/api'

-const loginResponse = await ky(OLECARD_AUTH_URL, {
+const loginResponse = await client.post('login', {
```

3. Start up CCC
```sh
cd ccc-server
npm run stolaf-college
```

4. Configure All About Olaf to use ccc 
> 1. Settings > Server UrL
> 2. Type `http://localhost:3000/v1`
> 3. Enter / Save

5. Switching between responses and observing the outcome
```
ky.post(OLECARD_AUTH_URL)
client.post('login')
client.post('login/error')
```

We see that 
- ky correctly shows us a 403 error with invalid credentials (302 Found > 403 Login Required)
- mocked client lets us authenticate and then logout
- mocked client error prevents login

### DIFF
--- a/source/lib/login.ts
+++ b/source/lib/login.ts
@@ -76,7 +76,8 @@ export async function performLogin(
 	})
 
 	let responseUrl = new URL(loginResponse.url)
-	let responseMessage = responseUrl.searchParams.get('message')
+	// URLSearchParams.get requires a polyfill in react native
+	let responseMessage = responseUrl.href.includes('error=')
 	if (responseMessage) {
 		throw new LoginFailedError(`Login failed: ${responseMessage}`)
 	}
--- a/source/lib/login.ts
+++ b/source/lib/login.ts
@@ -33,9 +33,9 @@ type Credentials = {
 	password: string
 }
 
-async function loadCredentials(): Promise<false | SharedWebCredentials> {
+async function loadCredentials(): Promise<null | SharedWebCredentials> {
 	let credentials = await getInternetCredentials(SIS_LOGIN_KEY)
-	return credentials
+	return credentials ? credentials : null
 }
 
 export async function storeCredentials(
@@ -85,7 +85,7 @@ export async function performLogin(
 	return {username, password}
 }
 
-type QueryFnData = false | SharedWebCredentials
+type QueryFnData = null | SharedWebCredentials
 type DefaultError = null | unknown
 type QueryT<Select> = ReturnType<typeof useCredentials<Select>>
 
--- a/source/lib/login.ts
+++ b/source/lib/login.ts
@@ -33,11 +33,8 @@ type Credentials = {
 	password: string
 }
 
-async function loadCredentials(): Promise<SharedWebCredentials> {
+async function loadCredentials(): Promise<false | SharedWebCredentials> {
 	let credentials = await getInternetCredentials(SIS_LOGIN_KEY)
-	if (credentials === false) {
-		throw new NoCredentialsError()
-	}
 	return credentials
 }
 
@@ -62,10 +59,11 @@ export function resetCredentials(): Promise<void> {
 export async function performLogin(
 	credentials: Credentials | null = null,
 ): Promise<Credentials> {
-	const {username, password} = credentials ?? (await loadCredentials())
-	if (!username || !password) {
+	const saved = credentials ?? (await loadCredentials())
+	if (!saved) {
 		throw new NoCredentialsError()
 	}
+	const {username, password} = saved
 
 	let formData = new FormData()
 	formData.append('username', username)
@@ -86,7 +84,7 @@ export async function performLogin(
 	return {username, password}
 }
 
-type QueryFnData = SharedWebCredentials
+type QueryFnData = false | SharedWebCredentials
 type DefaultError = null | unknown
 type QueryT<Select> = ReturnType<typeof useCredentials<Select>>
 
--- a/source/lib/login.ts
+++ b/source/lib/login.ts
@@ -103,10 +103,10 @@ export function useCredentials<TData = QueryFnData, TError = DefaultError>(
 }
 
 export function useUsername(): QueryT<
-	false | Pick<SharedWebCredentials, 'username'>
+	null | Pick<SharedWebCredentials, 'username'>
 > {
 	return useCredentials({
-		select: (data) => (data ? {username: data.username} : false),
+		select: (data) => (data ? {username: data.username} : null),
 	})
 }
 
--- a/source/lib/login.ts
+++ b/source/lib/login.ts
@@ -102,11 +102,9 @@ export function useCredentials<TData = QueryFnData, TError = DefaultError>(
 	})
 }
 
-export function useUsername(): QueryT<
-	null | Pick<SharedWebCredentials, 'username'>
-> {
+export function useUsername(): QueryT<string | undefined> {
 	return useCredentials({
-		select: (data) => (data ? {username: data.username} : null),
+		select: (data) => data?.username,
 	})
 }
 
--- a/source/views/sis/balances.tsx
+++ b/source/views/sis/balances.tsx
@@ -13,13 +13,14 @@ import {BalancesShapeType, useBalances} from '../../lib/financials'
 import * as c from '@frogpond/colors'
 import {sto} from '../../lib/colors'
 import {useNavigation} from '@react-navigation/native'
-import {NoCredentialsError, useCredentials} from '../../lib/login'
+import {NoCredentialsError, useUsername} from '../../lib/login'
 
 const DISCLAIMER = 'This data may be outdated or otherwise inaccurate.'
 
 export const BalancesView = (): JSX.Element => {
 	let navigation = useNavigation()
-	let credentials = useCredentials()
+	let {data: usernameData} = useUsername()
+	let username = usernameData ? usernameData.username : ''
 
 	let {
 		data = {} as BalancesShapeType,
@@ -28,7 +29,7 @@ export const BalancesView = (): JSX.Element => {
 		isInitialLoading,
 		refetch,
 		isRefetching,
-	} = useBalances(credentials.data?.username)
+	} = useBalances(username)
 
 	let openSettings = () => navigation.navigate('Settings')
 	let refresh = <RefreshControl onRefresh={refetch} refreshing={isRefetching} />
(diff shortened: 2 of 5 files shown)


source/lib/login.ts
```diff
@@ -78,15 +76,16 @@ export async function performLogin(
...
+	let responseMessage = responseUrl.href.includes('error=')
 	if (responseMessage) {
 		throw new LoginFailedError(`Login failed: ${responseMessage}`)
 	}
 
 	return {username, password}
 }
 
-type QueryFnData = SharedWebCredentials
+type QueryFnData = false | SharedWebCredentials
```
drewvolz: After an out of band conversation with Hawken, it looks like the other half of the magic is within the `QueryFnData` type

> If we add " | false" to that line, and put useCredentials back without the throw, what happens?
> 
> - TData and TError are the internal type names used by useQuery
> - TData is the type returned by queryFn, and TError is … the types that "error" can be in the result of useQuery (more or less, "what errors can this throw")
>
> Except that TData is actually the "type returned by select:", which defaults to the identity function. 
> 
> So if we pretend that we have TQueryReturn and TSelectReturn,
> 
> let query = useQuery({
>   queryFn: () => blah, // returns TQueryReturn
>   select: (data: TQueryReturn) // returns TSelectReturn
> 
> query.data: TSelectReturn
> 
> And TData is TSelectReturn
> 
> I hate how much type inference is happening in useCredentials & friends - it makes horrid errors

Nice! This last change makes the contract work.

source/lib/login.ts
```diff
@@ -78,15 +76,16 @@ export async function performLogin(
 	})
 
 	let responseUrl = new URL(loginResponse.url)
-	let responseMessage = responseUrl.searchParams.get('message')
+	// URLSearchParams.get requires a polyfill in react native
+	let responseMessage = responseUrl.href.includes('error=')
```
hawkrives, line 80: I'm OK with this because it's a specific endpoint, and not a general method.

hawkrives (approved):

source/lib/login.ts
hawkrives: (Plus, I suggested that Drew make this change)

drewvolz: Alternative would be dragging in [`react-native-url-polyfill`](https://github.com/charpeni/react-native-url-polyfill/blob/main/js/URLSearchParams.js ) or doing what it does under the hood of importing `URLSearchParams` from whatwg-url-without-unicode.

I'm also okay with this approach for now.

source/lib/login.ts
```diff
@@ -33,11 +33,8 @@ type Credentials = {
...
 }
 
-async function loadCredentials(): Promise<SharedWebCredentials> {
+async function loadCredentials(): Promise<false | SharedWebCredentials> {
 	let credentials = await getInternetCredentials(SIS_LOGIN_KEY)
-	if (credentials === false) {
-		throw new NoCredentialsError()
-	}
 	return credentials
 }
```
rye: I like dropping the throw-point and flattening out the type signature.

`false` still seems like a weird return value for a call like this. Could we do a bit of legwork and remap `false` -> `null`? (Then the return of this is `SharedWebCredentials | null`.)

hawkrives: That makes a lot more sense to me.

drewvolz: Makes sense!

source/lib/login.ts
```diff
@@ -104,10 +103,10 @@ export function useCredentials<TData = QueryFnData, TError = DefaultError>(
...
 
 export function useUsername(): QueryT<
-	false | Pick<SharedWebCredentials, 'username'>
+	null | Pick<SharedWebCredentials, 'username'>
 > {
 	return useCredentials({
-		select: (data) => (data ? {username: data.username} : false),
+		select: (data) => (data ? {username: data.username} : null),
```
hawkrives:
```suggestion
		select: (data) => data?.username,
```

source/views/sis/balances.tsx
```diff
@@ -13,13 +13,14 @@ import {BalancesShapeType, useBalances} from '../../lib/financials'
...
-import {NoCredentialsError, useCredentials} from '../../lib/login'
+import {NoCredentialsError, useUsername} from '../../lib/login'
 
 const DISCLAIMER = 'This data may be outdated or otherwise inaccurate.'
 
 export const BalancesView = (): JSX.Element => {
 	let navigation = useNavigation()
-	let credentials = useCredentials()
+	let {data: usernameData} = useUsername()
+	let username = usernameData ? usernameData.username : ''
```
hawkrives:
```suggestion
	let username = usernameData?.username ?? ''
```

I could go either way on this one.

hawkrives (changes requested):

source/lib/login.ts
drewvolz: The types were unhappy when I tried this earlier

source/views/sis/balances.tsx
drewvolz: I did like this better! Can we centralize this in the username hook, though? I get that changes the expectation to return an empty string, though. I'm trying to figure out what a clean implementation could be for the consumers.

hawkrives: I think I'd rather make the consumers do this… but idk. It's pretty short…

source/lib/login.ts
hawkrives: What was the type error?

drewvolz: I'll bet the explicit return type for `useUsername` has to adjust, too.

```
Type 'UseQueryResult<string | undefined, unknown>' is not assignable to type 'UseQueryResult<Pick<SharedWebCredentials, "username"> | null, unknown>'.
  Type 'QueryObserverRefetchErrorResult<string | undefined, unknown>' is not assignable to type 'UseQueryResult<Pick<SharedWebCredentials, "username"> | null, unknown>'.
    Type 'QueryObserverRefetchErrorResult<string | undefined, unknown>' is not assignable to type 'QueryObserverRefetchErrorResult<Pick<SharedWebCredentials, "username"> | null, unknown>'.
      Type 'string | undefined' is not assignable to type 'Pick<SharedWebCredentials, "username"> | null'.
        Type 'undefined' is not assignable to type 'Pick<SharedWebCredentials, "username"> | null'.ts(2322)
```

I think the difference here becomes

Before:
```
QueryT<null | Pick<SharedWebCredentials, 'username'>> 
```

After:
```
UseQueryResult<string | undefined, unknown>
```

source/views/sis/balances.tsx
```diff
@@ -13,13 +13,14 @@ import {BalancesShapeType, useBalances} from '../../lib/financials'
...
 import {useNavigation} from '@react-navigation/native'
-import {NoCredentialsError, useCredentials} from '../../lib/login'
+import {NoCredentialsError, useUsername} from '../../lib/login'
 
 const DISCLAIMER = 'This data may be outdated or otherwise inaccurate.'
 
 export const BalancesView = (): JSX.Element => {
 	let navigation = useNavigation()
-	let credentials = useCredentials()
+	let {data: usernameData} = useUsername()
```
hawkrives:
```suggestion
	let {data: username = ''} = useUsername()
```

I think we can do this now?

source/views/stoprint/print-release.tsx
```diff
@@ -81,7 +81,7 @@ export const PrintJobReleaseView = (): JSX.Element => {
 	let {job, printer} = route.params
 
 	let {data: usernameData, isLoading: loadingUsername} = useUsername()
-	let username = usernameData ? usernameData.username : ''
+	const username = usernameData ?? ''
```
hawkrives: same here for inlining the default value and renaming from usernameData to username

source/views/stoprint/query.ts
```diff
@@ -31,7 +31,7 @@ export const keys = {
 
 export function usePrintJobs(): UseQueryResult<PrintJobsResponse, unknown> {
 	let {data} = useUsername()
-	let username = data ? data.username : ''
+	const username = data ?? ''
```
hawkrives: Same comment about username

source/views/stoprint/query.ts
```diff
@@ -42,7 +42,7 @@ export function usePrintJobs(): UseQueryResult<PrintJobsResponse, unknown> {
 
 export function useAllPrinters(): UseQueryResult<AllPrintersResponse, unknown> {
 	let {data} = useUsername()
-	let username = data ? data.username : ''
+	const username = data ?? ''
```
hawkrives: Same comment about username

source/views/stoprint/query.ts
```diff
@@ -56,7 +56,7 @@ export function useRecentPrinters(): UseQueryResult<
 	unknown
 > {
 	let {data} = useUsername()
-	let username = data ? data.username : ''
+	const username = data ?? ''
```
hawkrives: Same comment about username

source/views/stoprint/query.ts
```diff
@@ -76,7 +76,7 @@ export function useHeldJobs(
 	printerName: string | undefined,
 ): UseQueryResult<HeldJobsResponse, unknown> {
 	let {data} = useUsername()
-	let username = data ? data.username : ''
+	const username = data ?? ''
```
hawkrives: Same comment about username

hawkrives (changes requested):

hawkrives (approved):

Issue #6850, opened 2023-01-29
Take screenshots of the app with Detox


hawkrives:
Do the magic status bar stuff, make as few changeable things as possible, and ideally generate diffs between screenshots too.

drewvolz:
https://wix.github.io/Detox/docs/guide/taking-screenshots

```js
async function setDemoMode() {
  if (device.getPlatform() === 'ios') {
    execSync(
      'xcrun simctl status_bar "iPhone 11" override --time "12:00" --batteryState charged --batteryLevel 100 --wifiBars 3 --cellularMode active --cellularBars 4'
    );
  } else {
    // enter demo mode
    execSync('adb shell settings put global sysui_demo_allowed 1');
    // display time 12:00
    execSync('adb shell am broadcast -a com.android.systemui.demo -e command clock -e hhmm 1200');
    // Display full mobile data with 4g type and no wifi
    execSync(
      'adb shell am broadcast -a com.android.systemui.demo -e command network -e mobile show -e level 4 -e datatype 4g -e wifi false'
    );
    // Hide notifications
    execSync('adb shell am broadcast -a com.android.systemui.demo -e command notifications -e visible false');
    // Show full battery but not in charging state
    execSync('adb shell am broadcast -a com.android.systemui.demo -e command battery -e plugged false -e level 100');
  }
}
```

PR #6952, opened 2023-02-18
Native visual regression testing (take two)

drewvolz:
- Targets #6916 
- Alternative implementation of #6920 

To see the feature in action, checkout the target PR to view the functionality.

This PR performs an e2e visual regression check by verifying the app is able to successfully set a background image for the app. When the default background image is changed from the initial value to a photo from the simulator's camera roll, `swift-snapshot-testing` quickly performs a pixel check to verify against an image of the desired outcome.

The e2e test introduced in this PR performs an end to end check for:
- [x] Asserting that the context menu works.
- [x] Ensuring the settings screen opens.
- [x] Guaranteeing an image can be chosen.
- [x] Asserting the new background of the app matches the expected outcome pixel for pixel.

The output file we are comparing against lives at the following path (customizable, currently stored with the tests).

```
ios/AllAboutOlafUITests/__Snapshots__/CustomBackgroundImageTests/*.heic
```

### Swift Package Manager additions

Why SwiftPM? Neither of these packages support Cocoapods.

I've made sure to add these two SwiftPM additions to the `AllAboutOlafUITests` target and not `AllAboutOlaf`.

1. Instead of rolling our own regression checking (#6920), we can use [`swift-snapshot-testing`](https://github.com/pointfreeco/swift-snapshot-testing) from pointfree. [It's very customizable](https://github.com/pointfreeco/swift-snapshot-testing/blob/main/Documentation/Available-Snapshot-Strategies.md), well supported, and gives us a lot of control over formats we can diff views in.

2. We are storing the output file as an HEIC instead of a PNG (with [`SnapshotTestingHEIC`](https://github.com/alexey1312/SnapshotTestingHEIC)) because the HEIC storage format reduces file sizes in comparison to PNG. 

Ultimately we will want to store and perform regression checking in a separate repository, but this at least lets us begin performing these types of checks in the codebase today.

### Preventing false positives

To avoid issues of

- homescreen icons changing,
- homescreen footer messages randomizing,
- statusbar datetime text changing (solvable via https://github.com/StoDevX/AAO-React-Native/issues/6850#issuecomment-1407535024),
- and other values that could throw off this test,

this test runs on the component library version of the background image picker. It changes both the library sample and real background of the app, but if something were to change the e2e flow of the real component, perhaps we wouldn't know immediately.

I feel the compromise of knowing that (i) the data has changed, (ii) the hooks are functioning, and (iii) the image is visually checked may be a good set of tradeoffs in lieu of testing the actual homescreen but that is open for discussion!

### Jest vs XCUITest

As described in https://github.com/StoDevX/AAO-React-Native/pull/6916#issuecomment-1417127660, challenges were presented that required us to go native for this test; Apple renders the native image picker view out of the view hierarchy process, which means we do not have access to inspect PHPickerViewController with detox. A native test here-and-there won't hurt, XCUITest doesn't bite!

### Reusability

To get this in state where it could be tested and reusable across the component library and settings, a few changes were made across the app.

- `testID` is now an available prop for `<PushButtonCell />` components
- The navigation bar context menu is separated into its own `<RightHomeContextMenu />` component


drewvolz:
@hawkrives ideas on how to resolve `package.resolved` for Swift Package Manager in our build and caching steps?


```
Resolve Package Graph
xcodebuild: error: Could not resolve package dependencies:
  Package.resolved file is corrupted or malformed; fix or delete the file to continue: unsupported schema version 2
```

hawkrives:
Does

``` 
xcodebuild -workspace ios/AllAboutOlaf.xcworkspace -scheme AllAboutOlaf -configuration Release -destination 'platform=iOS Simulator,name=iPhone 11 Pro,OS=15.0' -derivedDataPath ios/build build
``` 

pass if you run it locally?

Edit: because we may need to update the detox command string.

drewvolz:
I know the pipeline is setup differently but xcodebuild worked fine locally when I built with iOS 16.2 and an iPhone 14 Pro.

That sounds like the problem if v2 package.resolved is invalid; this needs to be created with Xcode 13.

Update: that looked like it did the job to fix the previous error.

drewvolz:
I think you're still correct. We'll need to add a step for `xcodebuild ... test` to see the native e2e test on the pipeline.

drewvolz:
Evaluating this again.

Choosing the detox approach may be better in the end. We can assume the photo picker is out of our control, so if we assume it behaves correctly, we can go the mocked approach of detox and avoid adding native dependencies on SwiftPM and its packages.

Will try again with the approach laid out in https://github.com/StoDevX/AAO-React-Native/pull/6916#issuecomment-1417127660

PR #6983, opened 2023-04-02
Add course search icon to navigation header

### LINKED ISSUE #6980: Move browse all button into nav header for course catalog
It appears to be an important enough action to migrate to the top of the view.

drewvolz:
- Resolves #6980 
- Adds `<SearchButton />` to navigation buttons (icon and text variants)
- Course search is using the `text` variant with the text "Browse" (middle screenshot)
 
Before | [chosen] After (text) | After (icon)
--|--|--
before | after-text | after-icon


hawkrives (approved):
Me like!

PR #7021, opened 2023-08-30
Remove concurrent Renovate PR limit

rye:
We have built up quite a bit of a backlog of PRs. Since we are using a merge queue, why not put it to good use with these dependency updates?

I am not sure if we just never considered this or we turned it off at some point.

For reference, the "rate-limited PR" list is _loooooong_. It extends past the bottom of this screenshot.

image

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -51,4 +51,6 @@
 			rangeStrategy: 'bump',
 		},
 	],
+	// Allow Renovate to open as many PRs as it wants
+	prConcurrentLimit: 0,
 }


rye:
Alternatively, we can click the checkbox and create all the PRs at once 🐳

drewvolz:
When Renovate was first added to the repo, I think we landed on having a concurrent PR limit to reduce the rebase flooding that happens ([quoting @rye](https://github.com/StoDevX/AAO-React-Native/pull/3193#issuecomment-441515078)). If we turn this back on and move away from the schedule, are we signing up for a firehose or since merge queue has been introduced, you're thinking it will potentially be an improved experience?

rye:
If renovate rebases _every single open PR_ every time one gets merged to the primary branch (like dependabot did, IIRC?), that will lead to a polynomial number of pushes and a lot of load on the CI resources.

However, I don't think it does so with a merge queue, at least based on [this documentation](https://docs.renovatebot.com/updating-rebasing/):

>Here is a list of the most common cases where Renovate must update/rebase the branch:
>- When a pull request has conflicts due to changes on the base branch
>- When you have enabled "Require branches to be up to date before merging" on GitHub
>- When you have manually told Renovate to rebase when behind the base branch with `"rebaseWhen": "behind-base-branch"`
>- When a newer version of the dependency is released
>- When you request a manual rebase from the Renovate bot
>- When you use `"automerge": true` and `"rebaseWhen": "auto"` on a branch / pr

Before merging this PR, I'll let Renovate create all the PRs that are stacked up, and then we can check out the impacts on CI.

hawkrives (approved):

PR #7082, opened 2023-10-09
allow section lists in filter popups to expand their content

### LINKED ISSUE #7081: Filters in section-list rows have text and content cutoff, checkbox is unreachable
Issue with filters for menu dietary restriction lists: `flex: 1` applied to [modules/filter/section-list.tsx](https://github.com/StoDevX/AAO-React-Native/blob/master/modules/filter/section-list.tsx) `content` is cut-off and is not allowing us to see the checkmark or the status of the checkmark. 

We need a style change here, flex properties (grow and shrink) should be able to handle this.

Worth looking at while working on the filters for #6765.

drewvolz:
- Closes #7081 

Looks like `flex: 1` is not enough to handle a layout we want to grow in RN.

> `flexShrink` accepts any floating point value >= 0, with 0 being the default value (on the web, the default is 1)

We can update our filter list rows to have a styling of `flex: 1` and `flexShrink: 1` to both (i) layout content and (ii) present overflowing content.

Before | After
--|--
Simulator Screen Shot - iPhone 14 Pro - 2023-10-08 at 18 43 14 | Simulator Screen Shot - iPhone 14 Pro - 2023-10-08 at 18 42 47

### DIFF
--- a/modules/filter/section-list.tsx
+++ b/modules/filter/section-list.tsx
@@ -127,6 +127,7 @@ export function ListSection<T extends object>({
 const styles = StyleSheet.create({
 	content: {
 		flex: 1,
+		flexShrink: 1,
 		paddingVertical: 10,
 	},
 	title: {


hawkrives (approved):
I have no idea why we have to set "shrink" to allow growing 🤷

PR #7083, opened 2023-10-09
ensure items with dietary data are only shown when a diet is enabled

### LINKED ISSUE #6742: Menu filters on cors_icon (an object) allow through items with no cors_icons at all
In other words, meat dishes are showing up for `(V)`

drewvolz:
- Closes #6742 

When filtering by diets in the menu screen, items without dietary data are appearing which could lead to inaccurate results being shown. We don't want to show meat dishes as vegetarian when they aren't explicitly marked as (v) or (vg). 

Following the pattern we have for checking whether specials are enabled, this adds a check to see if dietary filtering is enabled. When grouping items to render, if we are filtering diets, we can exclude items that do not state their dietary data. This should be okay performance-wise.

Bonus performance addition: this also opts FancyMenu's section list into `largeListProps`. Local testing shows this resolves the larger list struggling to keep up while scrolling (the one without specials enabled).

Before | After
--|--
Simulator Screen Shot - iPhone 14 Pro - 2023-10-08 at 20 49 33 | Simulator Screen Shot - iPhone 14 Pro - 2023-10-08 at 20 49 49

### DIFF
--- a/modules/food-menu/fancy-menu.tsx
+++ b/modules/food-menu/fancy-menu.tsx
@@ -10,7 +10,7 @@ import type {
 	StationMenuType,
 } from './types'
 import size from 'lodash/size'
-import {ListSectionHeader, ListSeparator} from '@frogpond/lists'
+import {ListSectionHeader, ListSeparator, largeListProps} from '@frogpond/lists'
 import type {FilterType} from '@frogpond/filter'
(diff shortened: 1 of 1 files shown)


hawkrives (approved):
Thanks!

PR #7087, opened 2023-10-16
🧪  internal developer feature flagging

drewvolz:
Links
- Part of #6106 
- Eventually will impact #6936

[Best viewed without whitespace
](https://github.com/StoDevX/AAO-React-Native/pull/7087/files?diff=split&w=1)

Changes
- Allows for us to define hooks for stating that an internal setting should be toggled in the app.
- Introduces `@frogpond/app-config` module
- Adds a section list of boolean toggles to `Settings > Developer > Feature Flags`
- Output of individual config checks includes a check for `devMode()` enabled
- Adds a `disabled` property to homescreen tiles
- Adds a feature flag for navigating directly to course search without showing the interstitial recents view from the Home Screen
- Adds react-native-testing-library as a dependency

Note

I had this setup with a refactored version of stoPrint but changes were peppered everywhere. This is setup separately from those changes, although the commits are still in the history of this PR. If we accept this method/setup, I'll be sure to squash commit to remove that history.

Intended use 

Change certain sections of the app for more developer exposure and testing to WIP features and layout changes. This is not intended to introduce hidden features or toggles for A/B testing in production.  

Challenges

Hooks are great but they cannot be used everywhere because [hooks have rules](https://legacy.reactjs.org/docs/hooks-rules.html). It will be good to see if this is a useful setup or if it will have to adapt to support areas where hooks cannot be used. One example is the movies tab -- we cannot simply drop this hook into the tab definition to enable/disable it.
 
---

Usage

1. Add types to `modules/app-config/types.ts`

The format is SECTION_KEY where the first underscore will be split at render time for our view to group by, but the entirety of the value (section + key) will be stored.
```tsx
export enum AppConfigType {
	Courses_ShowRecentSearchScreen = 'Courses_ShowRecentSearchScreen',
}
```

2. Add AppConfig definition and feature hook to `modules/app-config/index.tsx`

```tsx
{
	title: 'Show the course search recents screen',
	configKey: AppConfigType.Courses_ShowRecentSearchScreen,
	active: await getFeatureFlag(
		AppConfigType.Courses_ShowRecentSearchScreen,
	),
},

export const useCourseSearchRecentsScreen = (): boolean =>
	useFeature(AppConfigType.Courses_ShowRecentSearchScreen)
```

4.  Use the hook where you'd like
```tsx
import {useCourseSearchRecentsScreen} from '@frog-pond/app-config`

const MyComponent = () => {
  const shouldShowCourseSearchRecents = useCourseSearchRecentsScreen()
  
  if (shouldShowCourseSearchRecents) {
    // show something to developers
  }
}
```


source/lib/storage.ts
```diff
@@ -6,13 +6,29 @@ import {
...
 setStoragePrefix('aao:')
 
+/// MARK: Feature flags
+
+const featureFlagsKey = 'app:feature-flag'
+export function setFeatureFlag(
+	name: AppConfigKey,
+	value: boolean,
+): Promise<void> {
+	let key = `${featureFlagsKey}:${name}`
```
rye: Const?

source/lib/storage.ts
```diff
@@ -6,13 +6,29 @@ import {
...
+	return setItem(key, value)
+}
+export function getFeatureFlag(name: AppConfigKey): Promise<boolean> {
+	let key = `${featureFlagsKey}:${name}`
```
rye: Const?

modules/app-config/types.ts
```diff
@@ -0,0 +1,14 @@
...
+	active: boolean
+}
+
+export enum AppConfigKey {
+	TestDataKey = 'TestData',
+}
+
+export enum AppConfigGroupKey {
+	test = 'test',
+}
```
rye: Couple of things:

- Can we get away with _not_ using the Key suffix in the type name? (e.g. can we use the `keyof` operator in usage sites instead?)
- `group` and `configKey` seem related, but they're in unrelated types and it seems like they would require parallel maintenance. It's not entirely clear to me how I should go about adding a new key based on this interface alone.
  - Can we add a type that contains both the `key` and the `group` and use that as the means to identify the flag? It's not entirely clear to me that that's _better_, but it's an alternative.

modules/app-config/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+			return isDevMode() ? newValue : false
+		},
+	})
+
+	return isDevMode() ? featureValue : false
+}
+
+// exported feature flags
+export const useTestFeature = (): boolean =>
+	useFeature(AppConfigKey.TestDataKey)
```
rye: Is there any way we can add some tests for all this? Seems like some relatively contained logic.

modules/app-config/types.ts
drewvolz: Great points. I thought about this too when writing up the PR readme example and I agree with you that key can go away from the name.

The point of the grouping is to provide a hint wheee the flag is used, and to physically organize/group them in the screen that lists them.

I do not know if there is a way we can achieve this with a singular delimited flag e.g. StoPrint.useTestData would mean we'd be relying on typing the grouping correctly without using an enum.

It is more than I'd like for configuration as you've pointed out. The specialized usage here probably means we can settle for a single key without breaking it up into multiple places.

modules/app-config/index.tsx
drewvolz: I almost didn't open the PR until I wrote some tests knowing I'd see this comment, I'm glad you asked. I definitely will test this!

modules/app-config/types.ts
drewvolz: Let me know if you still see room for improvement with `keyof` here, otherwise I've taken your ideas and implemented the `group` to be included in the `configKey`.

modules/app-config/index.tsx
drewvolz: @hawkrives @rye are we wanting to add `@testing-library/react-hooks` and `react-test-renderer` to the repo? To test the hooks behind `useFeature` [would be to test tanstack query.](https://tanstack.com/query/v4/docs/react/guides/testing)

I can do more looking but it also seems like getting those dependencies installed will involve a bit of dependency management due to `peer react@"^16.9.0 || ^17.0.0" from @testing-library/react-hooks@8.0.1`

```zsh
npm ERR!   1 more (the root project)
npm ERR!   26 more (@frogpond/app-config, @frogpond/badge, ...)
npm ERR!
npm ERR! Could not resolve dependency:
npm ERR! peer react@"^16.9.0 || ^17.0.0" from @testing-library/react-hooks@8.0.1
npm ERR! node_modules/@testing-library/react-hooks
npm ERR!   dev @testing-library/react-hooks@"*" from the root project

npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution
```

rye: I am happy with us adding those, if that's what you're asking.

Are we above those versions for `@testing-library/react-hooks@8.0.1`? or below?

hawkrives (approved):

modules/app-config/index.tsx
drewvolz: I think I figured it out. 

> As part of the changes for React 18, it has been decided that the renderHook API provided by this library will instead be included as official additions to both `react-testing-library` and `react-native-testing-library` with the intention being to provide a more cohesive and consistent implementation for our users.

I had to add `@testing-library/react-native` instead, which has the functionality we need from `@testing-library/react`. And for future note,  `@testing-library/react-hooks` is limited to react 17 but supports 18 now that it is absorbed into its parent package.

drewvolz: Getting closer! Dealing with issues when running the new tests in https://github.com/StoDevX/AAO-React-Native/pull/7087/commits/25867c34682dd3969e834de94bdd0a7473ea4a3b.

Inconsistent results, sometimes passing, errors. I think I have a problem with the mocks. 

```
❯ npm test modules/app-config

ReferenceError: You are trying to `import` a file after the Jest environment has been torn down.

at batchNotifyFn (node_modules/react-native/Libraries/ReactNative/RendererImplementation.js:44:10)
at node_modules/@tanstack/query-core/src/notifyManager.ts:65:9
node_modules/react-native/Libraries/ReactNative/RendererImplementation.js:44

TypeError: require(...).unstable_batchedUpdates is not a function
    at batchNotifyFn (node_modules/react-native/Libraries/ReactNative/RendererImplementation.js:44:51)
    atnode_modules/@tanstack/uery-core/src/notifyManager.ts:65:9
```

sometimes one passes when run individually

```
Test Suites: 1 passed, 1 total
Tests:       1 skipped, 1 passed, 2 total
Snapshots:   0 total
```

rarely both passing
```
❯ npm test modules/app-config

Test Suites: 1 passed, 1 total
Tests:       2 passed, 2 total
Snapshots:   0 total
```

drewvolz: Think I've got these passing now.

modules/app-config/index.tsx
rye, line 1: Does this need to be named `.tsx`? Not seeing any JSX syntax in here.

modules/app-config/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+	return [
+		{
+			title: 'Show the course search recents screen',
+			configKey: AppConfigType.Courses_ShowRecentSearchScreen,
+			active: await getFeatureFlag(
+				AppConfigType.Courses_ShowRecentSearchScreen,
+			),
+		},
+	]
+}
```
rye: Right now you're checking `isDevMode` at the callsite. Can we also stub out `AppConfig` to just return `[]` in the case we're not in dev mode? This should never be hit unless `FeatureFlagsView` runs outside devmode, which IIUC should be controlled at a higher level. This is just an intermediate barrier for the odd caller to `AppConfig` to handle.

modules/app-config/types.ts
```diff
@@ -0,0 +1,19 @@
+export interface FeatureFlagType {
+	configKey: AppConfigType
+	title: string
+	active: boolean
+}
```
rye: We're still using the `Type` suffix here.

Why not just call this a `FeatureFlag`? :)

modules/app-config/types.ts
```diff
@@ -0,0 +1,19 @@
...
+ * __DEV__ app config keys
+ *
+ * Reserved for dev-only flags for internal experimentation, these
+ * will never be A/B flags nor will they return true in production.
+ *
+ * The format is SECTION_KEY where the first underscore will be
+ * split at render time for our view to group by, but the entirety
+ * of the value (section + key) will be stored.
+ */
+export enum AppConfigType {
```
rye: Can we call this... `AppConfigEntry`? I'm fine with calling this a `AppConfigKey` too now.

source/views/home/index.tsx
```diff
@@ -25,6 +25,7 @@ const styles = StyleSheet.create({
 
 function HomePage(): JSX.Element {
 	let navigation = useNavigation()
+	let allViews = AllViews().filter((v) => !v.disabled ?? true)
```
rye: No need to drop the rest of the characters — `view`?

```suggestion
	let allViews = AllViews().filter((view) => !view.disabled ?? true)
```

source/views/settings/screens/feature-flags.tsx
```diff
@@ -0,0 +1,111 @@
...
+import {NativeStackNavigationOptions} from '@react-navigation/native-stack'
+import {CellToggle} from '@frogpond/tableview/cells'
+import {ListEmpty, ListSectionHeader, ListSeparator} from '@frogpond/lists'
+import * as c from '@frogpond/colors'
+import {AppConfig, FeatureFlagType} from '@frogpond/app-config'
+import {groupBy, orderBy} from 'lodash'
+import {LoadingView, NoticeView} from '@frogpond/notice'
+import {toLaxTitleCase} from '@frogpond/titlecase'
+import {Touchable} from '@frogpond/touchable'
+import {commonStyles} from '../../../../modules/navigation-buttons/styles'
```
rye: Can we organize these a little more nicely?

I like:

- Imports from all externals up top
- Imports from our own modules after those
- Imports from everywhere else after that (in this case, `lib/storage`)?

rye (approved):
Approved as-is, with some additional comments.

modules/app-config/index.tsx
drewvolz: Nice idea, handling this here makes sense! Done.

modules/app-config/index.tsx
drewvolz: Updated

modules/app-config/types.ts
drewvolz: I like these name changes! Done.

modules/app-config/types.ts
drewvolz: Changed.

source/views/home/index.tsx
drewvolz: This one got away. Nice find.

source/views/settings/screens/feature-flags.tsx
drewvolz: Organized!

PR #7094, opened 2023-10-21
import rules, groupings, and order enforced by lint

### LINKED ISSUE #7093: Enable eslint-plugin-imports
Via @rye 

> Can we organize these a little more nicely?
> 
> I like:
> 
> - Imports from all externals up top
> - Imports from our own modules after those
> - Imports from everywhere else after that (in this case, lib/storage)?

[`eslint-plugin-import`](https://github.com/import-js/eslint-plugin-import) can do that! We should set it up.

drewvolz:
Closes #7093

- Adds [`eslint-plugin-import`](https://github.com/import-js/eslint-plugin-import/) 
- Adds [`eslint-plugin-simple-import-sort`](https://github.com/lydell/eslint-plugin-simple-import-sort)
  - Example custom grouping [inspired from here](https://github.com/lydell/eslint-plugin-simple-import-sort/blob/main/examples/.eslintrc.js#L68-L98)

Import rules to take into account:
- Imports come first
- No duplicates

Import groupings to ensure:
1. Node built-ins
1. Packages. `react` related packages come first.
1. `react-native-*` community packages
1. React navigation
1. Internal packages and scopes
1. Side effect imports.
1. Parent imports. Put `..` last.
1. Other relative imports. Put same-folder imports and `.` last.
1. Style imports.
1. Image imports

Room for improvement here, plenty of room for feedback.


.eslintrc.yaml
```diff
@@ -125,3 +127,35 @@ rules:
...
+          # Internal packages.
+          ['^(@|@frogpond|@hawkrives|components|modules|util|config)(/.*|$)'],
+          # Side effect imports.
+          ['^\\u0000'],
+          # Parent imports. Put `..` last.
+          ['^\\.\\.(?!/?$)', '^\\.\\./?$'],
+          # Other relative imports. Put same-folder imports and `.` last.
+          ['^\\./(?=.*/)(?!/?$)', '^\\.(?!/?$)', '^\\./?$'],
+          # Style imports.
+          ['^.+\\.s?css$'],
```
hawkrives, line 160: Let's add images here - png, specifically

.eslintrc.yaml
```diff
@@ -125,3 +127,35 @@ rules:
 
   # eslint-plugin-jsx-expressions – https://github.com/hpersson/eslint-plugin-jsx-expressions/
   jsx-expressions/strict-logical-expressions: warn
+
+  # eslint-plugin-import https://github.com/import-js/eslint-plugin-import/
+  import/first: error
+  import/no-duplicates: error
+
+  # eslint-plugin-simple-import-sort https://github.com/lydell/eslint-plugin-simple-import-sort/
+  simple-import-sort/exports: error
```
hawkrives, line 136: Import alone doesn't handle this? I thought it did. I'll have to look at what we use again…

.eslintrc.yaml
```diff
@@ -13,6 +13,8 @@ plugins:
   - 'eslint-plugin-react-native'
   - 'eslint-plugin-react-hooks'
   - '@tanstack/eslint-plugin-query'
+  - 'simple-import-sort'
+  - 'import'
```
hawkrives: I know eslint supports omitting the "eslint-plugin-" prefix, but I prefer to keep it… makes it a lost easier for me to track down which package provides it, especially with scoped packages.

.eslintrc.yaml
drewvolz: It should, however I think the configuration was a lot more involved for module and typescript resolving?

.eslintrc.yaml
drewvolz: Oh hmm I thought I was looking at the docs for that one, I'll take a look again.

hawkrives: I'm in the minority, for sure. 

If you have a plugin called "eslint-plugin-import", you can put either the full name "eslint-plugin-import" or you can omit the "eslint-plugin-" prefix and just put "import". 

For scoped packages… I don't know. "ts-eslint/eslint-plugin" becomes … just "\@ts-eslint"? (See, this is why I use the full form lol)

rye (approved):

hawkrives:
@drewvolz This is what we have for the import plugin:

```js
'import/extensions': ['error', { js: 'never', jsx: 'never', json: 'always', ts: 'never', tsx: 'never' }],
		'import/named': 'error',
		'import/no-cycle': 'off',
		'import/no-namespace': 'warn',
		'import/no-restricted-paths': ['warn', { zones: [{ target: './test', from: './app/actions' }] }],
		'import/no-self-import': 'error',
		'import/order': [
			'warn',
			{
				groups: ['builtin', 'external', 'internal', 'parent', 'sibling', 'index'],
				'newlines-between': 'never',
				alphabetize: { order: 'asc', caseInsensitive: true },
			},
		],
		'import/prefer-default-export': 'off',
``` 

and 

```js
{
			files: ['test/**'],
			env: { node: true },
			extends: ['plugin:@typescript-eslint/recommended-type-checked', 'plugin:@typescript-eslint/stylistic'],
			rules: {
				'import/no-extraneous-dependencies': ['error', { devDependencies: true }],
```

hawkrives (approved):

PR #7108, opened 2024-05-04
Remove period after full month name in menu filter toolbar

drewvolz:
Corrects a display bug in the client that says "May. 4" instead of "May 4"


hawkrives:
Are we OK with having "Dec 12"? I think I prefer that to "May. 5"…

hawkrives (approved):

PR #7110, opened 2024-05-05
small hours updates

drewvolz:
- Add C-Store (cave) hours
- Remove bag lunch line


data/building-hours/1-4-bag-lunch-line.yaml
```diff
@@ -1,9 +1,10 @@
 name: Bag Lunch Line
 category: Food
+links:
+  - {title: 'Bag lunch express order', url: 'https://stolaf.cafebonappetit.com/bag-lunch/'}
 
 schedule:
   - title: Hours
-    notes: Closed until further notice
     hours: []
```
hawkrives: I think they published pick-up hours, right?

hawkrives: "Pick up in Stav between 2-7pm the day before you need your bag lunch."

and

"Bag lunch will take your lunch swipe for the days that you have requested"

hawkrives:
Re-reading the Bag Lunch info on the BonApp website… I propose that we remove Bag Kunch Line from the app. It's no longer a "place with hours"; instead, you have to plan ahead, and there are tradeoffs to using it.

data/building-hours/1-4-bag-lunch-line.yaml
drewvolz: Right, so do we need hours for walk ups? And then hours for 24 hour in advance order pickup?

hawkrives: I vote we delete the bag lunch hours. It feels too specialized now.

hawkrives (approved):

PR #7129, opened 2024-09-02
set a static deployment target in Podfile

hawkrives:
the other one is controlled by RN, and I believe to be the cause of our warnings:

The iOS deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 11.0, but the range of supported deployment target versions is 12.0 to 17.2.99.

Also sets detox to build without codesigning


detox.config.js
```diff
@@ -35,21 +35,32 @@ function findCurrentDeploymentTarget() {
...
 		'-scheme AllAboutOlaf',
 		`-configuration ${configuration}`,
 		`-destination 'platform=iOS Simulator,name=${iPhoneSimulatorDevice},OS=${currentDeploymentTarget}'`,
 		'-derivedDataPath ios/build',
 		'build',
-	].join(' ')
+	]
+
+	if (codeSigningDisabled) {
+		baseCommand.push(
```
hawkrives, line 56: Is it OK for these strings to come at the end of the command? They're not env vars, are they?

drewvolz: Should be fine to have them at the end as they're just individual build settings that control and change the way a target is built. Alternatively, we could try to configure code signing settings through fastlane https://docs.fastlane.tools/actions/update_code_signing_settings/

Separately, noting this
> Note that Detox CLI just executes the build command you specified in your Detox config for a specific configuration. In the other words, it has no extra logic – it just takes the command and runs it, nothing more and nothing less.

Last, I'm wondering how the cached jsbundle is being referenced but we exit out with there being no jsbundle now. Could we see issues installing an archive build without being signed?

I don't see how we'd have impacted the jsbundle creation or referencing through these changes, though.

.github/workflows/check.yml
```diff
@@ -496,6 +496,7 @@ jobs:
         if: ${{ steps.app-cache.outputs.cache-hit != 'true' && !contains(github.event.pull_request.labels.*.name, 'ci/skip-detox') }}
         env:
           SKIP_BUNDLING: 'true'
```
drewvolz, line 499: @hawkrives do we fail to have a bundle because skip_bundling is set to true here?

drewvolz: Never mind I see that's only set in the workflow and used for build.gradle

drewvolz (approved):

PR #7206, opened 2024-09-07
android: Compile with and target Android 14 (API 34)

rye:
In an attempt to resolve the following hard-stop:

image

@hawkrives / @drewvolz I also want to draw your attention to the bottom section there. Apparently we have to ship a new production release? Have we seen this before?


drewvolz:
@rye this sounds familiar to previous "we will stop showing your app if you compile with an older api sdk" notices, including the thread to take our listing off. But it is harsh.

PR #7287, opened 2025-03-04
update boost podfile.lock checksum

drewvolz:
First, cache actions for save/restore required being bumped to 4.2.0 in a footnote of the sunset warning if you're on pinned hashes. It's beyond me why they didn't keep in lockstep with the 4.0 bump for non-pinned. See this link for more info. https://github.com/actions/cache/tree/v4.2.0?tab=readme-ov-file#whats-new

> If you are using pinned SHAs, please use the SHAs of versions v4.2.0 or v3.4.0

Second, our `boost` checksum does not match and this is caused by an upstream issue which react native maintainers said to apply a patch-package to the spec source for react-native's boost. This moved it away from jfrog and to boost.io. Then the checksum should match.  See this link https://github.com/facebook/react-native/issues/42180 for more info.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Extract job definition
         run: yq '.jobs.${{ github.job }}' .github/workflows/check.yml > .github/job.yml
 
-      - uses: actions/cache/restore@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4
+      - uses: actions/cache/restore@1bd1e32a3bdc45362d1e726936510720a7c30a57 # v4.2.0
         id: node-cache
         with:
           path: node_modules/
@@ -43,7 +43,7 @@ jobs:
       - if: steps.node-cache.outputs.cache-hit != 'true'
         run: npm ci
 
-      - uses: actions/cache/save@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4
+      - uses: actions/cache/save@1bd1e32a3bdc45362d1e726936510720a7c30a57 # v4.2.0
         if: steps.node-cache.outputs.cache-hit != 'true'
         with:
           path: node_modules/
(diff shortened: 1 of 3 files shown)


drewvolz:
We shouldn't need the patch package changes as long as our podfile.lock checksum for boost matches the expected value, but I'm seeing the following for [Check / iOS Cocoapods (pull_request)](https://github.com/StoDevX/AAO-React-Native/actions/runs/13647840867/job/38149874919?pr=7287) 

```
 bundle exec pod install --deployment
```
> Verifying no changes
> [!] There were changes to the lockfile in deployment mode:
> SPEC CHECKSUMS:
>   boost:
>     New Lockfile: `57d2868c099736d80fcd648bf211b4431e51a558`
>     Old Lockfile: `7dcd2de282d72e344012f7d6564d024930a6a440`

and I did check the sha256 hashes for the spec source, but that doesn't seem related
```sh
> curl -sL https://archives.boost.io/release/1.76.0/source/boost_1_76_0.tar.bz2 | shasum -a 256
f0397ba6e982c4450f27bf32a2a83292aba035b827a5623a14636ea583318c41  -

> curl -sL https://boostorg.jfrog.io/artifactory/main/release/1.76.0/source/boost_1_76_0.tar.bz2 | shasum -a 256
79e6d3f986444e5a80afbeccdaf2d1c1cf964baa8d766d20859d653a16c39848  -
```

Is the issue related to the checksum mismatch between the Boost tarball sources?

In my commit, I've changed the source URL in the Boost podspec from boostorg.jfrog.io to archives.boost.io while keeping the same SHA256 hash. My SHA256 test confirms that archives.boost.io does provide a file with matching checksum `f0397ba6e982c4450f27bf32a2a83292aba035b827a5623a14636ea583318c41`.

The checksum difference in Podfile.lock (from `7dcd2de2`... to `57d2868c`...) is likely the signature CocoaPods generates for the entire podspec, not just the source file. When the source URL changed, the overall podspec checksum changed, even though I kept the SHA256 hash of the tarball the same.

We're on pods 15.2, and we absolutely do not want to upgrade to 16.2 which fails to update Podfile.lock for regular everyday changes.

???

drewvolz:
Closing in favor of #7289

PR #7297, opened 2025-05-25
Bus schedule detail view and day of week picker

drewvolz:
Implements [#6208](https://github.com/StoDevX/AAO-React-Native/discussions/6208)

- Adds a detail view per row to show all times at a stop
- Adds a context menu to pick a schedule for a different day of the week

List | Detail
--|--
 |


hawkrives:
How hard would it be to add a ✔️ on the currently selected day of the week in the context menu?

hawkrives (approved):

drewvolz:
That might be possible with menuState https://github.com/dominicstop/react-native-ios-context-menu?tab=readme-ov-file#contextmenuview-example-08-1

PR #7299, opened 2025-06-07
add madson construction camera

drewvolz:
https://www.stolaf.edu/multimedia/webcams/?cam=madson#fold

* adds Madson construction camera
* updates m3u8 stream urls


hawkrives:
approved!

PR #7319, opened 2025-10-07
exclude ownership for dependabot notifications

drewvolz:
PRs only targeting dependencies should no longer tag us by dependabot 🤞


hawkrives:
How can we just … turn off dependabot? I've tried before…

hawkrives (approved):

PR #7352, opened 2026-04-06
fix eslint errors

hawkrives:
Fix ESLint errors

### DIFF
--- a/src/app/event/[event].tsx
+++ b/src/app/event/[event].tsx
@@ -1,12 +1,20 @@
 import React from 'react'
 import {Platform} from 'react-native'
 
-import {EventDetail as EventDetailIos} from '@frogpond/event-list/event-detail-ios'
-import {EventDetail as EventDetailAndroid} from '@frogpond/event-list/event-detail-android'
+const EventDetailView = React.lazy(() =>
+	Platform.OS === 'ios'
+		? import('@frogpond/event-list/event-detail-ios').then((m) => ({
+				default: m.EventDetail,
+			}))
+		: import('@frogpond/event-list/event-detail-android').then((m) => ({
+				default: m.EventDetail,
+			})),
+)
 
-const EventDetailView =
-	Platform.OS === 'ios' ? EventDetailIos : EventDetailAndroid
-
-export default function EventDetailScreen(): React.ReactNode {
-	return <EventDetailView />
+export default function EventDetailScreen(): React.JSX.Element {
+	return (
+		<React.Suspense>
+			<EventDetailView />
+		</React.Suspense>
+	)
 }
--- a/src/app/event/[event].tsx
+++ b/src/app/event/[event].tsx
@@ -1,10 +1,11 @@
 import React from 'react'
 import {Platform} from 'react-native'
 
+import {EventDetail as EventDetailIos} from '@frogpond/event-list/event-detail-ios'
+import {EventDetail as EventDetailAndroid} from '@frogpond/event-list/event-detail-android'
+
 const EventDetailView =
-	Platform.OS === 'ios'
-		? require('@frogpond/event-list/event-detail-ios').EventDetail
-		: require('@frogpond/event-list/event-detail-android').EventDetail
+	Platform.OS === 'ios' ? EventDetailIos : EventDetailAndroid
 
 export default function EventDetailScreen(): React.ReactNode {
 	return <EventDetailView />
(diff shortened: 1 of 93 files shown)


src/app/event/[event].tsx
Comment (hawkrives) [inline src/app/event/[event].tsx:1]: I'd rather this change be from require to "await import()" - is that possible?

src/app/sis/job/[job].tsx
```diff
@@ -1,10 +1,11 @@
...
 import {Platform} from 'react-native'
 
+import {JobDetailView as JobDetailViewIos} from '../../../views/sis/student-work/detail-ios'
+import {JobDetailView as JobDetailViewAndroid} from '../../../views/sis/student-work/detail-android'
+
 const JobDetailView =
-	Platform.OS === 'ios'
-		? require('../../../views/sis/student-work/detail-ios').JobDetailView
-		: require('../../../views/sis/student-work/detail-android').JobDetailView
+	Platform.OS === 'ios' ? JobDetailViewIos : JobDetailViewAndroid
```
Comment (hawkrives) [inline src/app/sis/job/[job].tsx]: Is it possible to use "await import()" here?

src/views/stoprint/query.ts
hawkrives, line 16: Give me three alternatives to writing this function.

src/views/home/notice.tsx
```diff
@@ -27,22 +27,22 @@ if (isDevMode()) {
...
 export function UnofficialAppNotice(): React.JSX.Element {
-	const [msg, setMsg] = React.useState(() => sample(messages))
+	const [msg, setMsg] = React.useState(() => sample(messages) ?? messages[0])
 
 	const handlePress = React.useCallback(() => {
 		setMsg((current) => {
 			let next = current
 			while (next === current) {
-				next = sample(messages)!
+				next = sample(messages) ?? messages[0]
```
hawkrives: Revert this and add a targeted disable comment above it

src/views/home/notice.tsx
```diff
@@ -27,22 +27,22 @@ if (isDevMode()) {
 }
 
 export function UnofficialAppNotice(): React.JSX.Element {
-	const [msg, setMsg] = React.useState(() => sample(messages))
+	const [msg, setMsg] = React.useState(() => sample(messages) ?? messages[0])
```
hawkrives: Revert this and use a non-null assertion, and add a targeted disable comment above it

hawkrives (commented):
@claude i left some comments

src/app/event/[event].tsx
Comment (hawkrives) [reply src/app/event/[event].tsx:1]: Done — converted both `[event].tsx` and `[job].tsx` to use `React.lazy()` with dynamic `import()`. Each wraps the result in a `Suspense` boundary.

---

src/app/sis/job/[job].tsx
Comment (hawkrives) [reply src/app/sis/job/[job].tsx]: Done in the same commit — uses `React.lazy()` + dynamic `import()` with `Suspense`.

---

src/views/home/notice.tsx
hawkrives: Reverted — now uses `sample(messages)!` with `// eslint-disable-next-line @typescript-eslint/no-non-null-assertion` above it.

---

src/views/home/notice.tsx
hawkrives: Reverted — now uses `sample(messages)!` with a targeted disable comment above it.

---

src/views/stoprint/query.ts
hawkrives: Three alternatives to the current `QueryOptionsResult` helper type + verbose return annotations:

**1. Drop explicit return types, let TypeScript infer** — this is TanStack Query's intended pattern. Suppress the lint rule per-function:
```ts
// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
export function printJobsQuery(username: string | undefined) {
  return queryOptions({ ... })
}
```

**2. Use `ReturnType` to derive the type** — keeps explicit annotations but avoids hand-writing the complex generic:
```ts
function _printJobsQuery(username: string | undefined) {
  return queryOptions({ ... })
}
export const printJobsQuery: typeof _printJobsQuery = _printJobsQuery
```

**3. Disable `explicit-module-boundary-types` for query files via eslint config** — since these are internal query factories where inference is correct and desirable:
```ts
// in eslint.config.mjs
{ files: ['**/query.ts'], rules: { '@typescript-eslint/explicit-module-boundary-types': 'off' } }
```

I'd recommend option 1 or 3 — the explicit types we added are fragile and will break if query keys change.

---

PR #7422, opened 2026-04-09
more mise less bundler

hawkrives:
Turns out we don't need Bundler to install some of our tools. 

So let's take them out of the Gemfile and put them in Mise!

Now, we do lose the full-tree reproducibility of the Gemfile.lock, but I think that's OK for "tools" (as opposed to libraries).


drewvolz (approved):

hawkrives:
something's going funky with Detox. I want to look at it before merging.

PR #7432, opened 2026-04-10
Clean up patches in contrib/

hawkrives:
- removed three sentry patches that no longer apply
- renamed 0001-rn to a better name which reflects that it applies to callstack's rn theme provider, not rn itself


hawkrives:
Actually, I just realized that 0001 is a node package, and 0003-0005 are Pods. 

I'll think about this some more.

PR #7442, opened 2026-04-10
Upgrade React Query v4 to v5 with queryOptions() pattern

hawkrives:
## Summary

- Upgrade `@tanstack/react-query` and related packages from v4 to v5
- Adopt the `queryOptions()` helper pattern, replacing thin wrapper hooks with exported options objects
- Inline `useQuery(options)` at call sites for explicit data flow
- Audit `isLoading` semantics (v4 → v5 rename), migrate `onSuccess` to `useEffect`

## Behavioral fix

This PR also fixes a pre-existing bug in `source/views/stoprint/query.ts`: `recentPrintersOptions` was using `keys.printers(username)` as its query key, colliding with `allPrintersOptions` and causing the two queries to overwrite each other's cache. It now uses the correct `keys.recentPrinters(username)` key.

https://claude.ai/code/session_01RkJXfT8AJiN19SoK6zMQDJ


source/views/stoprint/print-jobs.tsx
```diff
@@ -21,13 +21,21 @@ import {NativeStackNavigationOptions} from '@react-navigation/native-stack'
...
-		useHasCredentials()
+	let {data: hasCredentials, isLoading: hasCredentialsLoading} = useQuery({
+		...credentialsOptions,
+		select: (data) => Boolean(data),
+	})
+
+	let {data: username = ''} = useQuery({
+		...credentialsOptions,
+		select: (data) => data?.username,
+	})
```
hawkrives: Redo this with one useQuery, that returns the data and isLoading

Then derive hasCredentials from data as a normal variable, and extract username from data if data has username key.

hawkrives: Done in ffcc8a5. Consolidated to a single `useQuery(credentialsOptions)` call and derived `hasCredentials` and `username` as local variables.

---

source/views/stoprint/query.ts
```diff
@@ -29,54 +28,38 @@ export const keys = {
...
-> {
-	let {data: username = ''} = useUsername()
 
-	return useQuery({
+export const recentPrintersOptions = (username: string) =>
+	queryOptions({
 		queryKey: keys.printers(username),
 		enabled: Boolean(username),
 		queryFn: ({signal}) => fetchRecentPrinters(username, {signal}),
 	})
```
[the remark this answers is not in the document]
hawkrives: Good catch — this was a pre-existing bug carried over from the original `useRecentPrinters` hook. Fixed in 83084ed by using `keys.recentPrinters(username)` instead of `keys.printers(username)`.

---

hawkrives:
From @drewvolz 

---

The banner.test.tsx mocking strategy is fragile. It now globally mocks @tanstack/react-query's useQuery:

```ts
 jest.mock('@tanstack/react-query', () => ({
   ...jest.requireActual(...),
   useQuery: jest.fn(),
 }))
``` 

This mocks useQuery for every component in the render tree, not just FaqBanner. Fine for the current isolated render, but it's a footgun if FaqBanner ever gets composed with a component that also calls useQuery in a test. An alternative: mock just the faqsOptions module and let the real useQuery run with a QueryClientProvider.

---

useCourseData loses a tiny bit of deduplication (source/views/sis/course-search/query.ts:67). The original had one createQueryOptions closure; the refactor inlines the ...courseDataOptions(term, levels, gereqs), enabled: terms.length > 0 spread twice (once for filteredTerms, once for terms). Minor, but:

```ts
 let buildQuery = (term: TermType) => ({
   ...courseDataOptions(term, levels, gereqs),
   enabled: terms.length > 0,
 })
 return useQueries({
   queries: (filteredTerms.length ? filteredTerms : terms).map(buildQuery),
 })
``` 

---

Inconsistent options-factory style. Most factories are arrow expressions that return queryOptions(...): export const fooOptions = (x: string) => queryOptions({...})

But heldJobsOptions uses a block body (source/views/stoprint/query.ts:57) and streamsOptions requires call-site invocation streamsOptions() because of the moment default arg (source/views/streaming/streams/list.tsx:74). The latter is a real asymmetry with every other fooOptions export, which are values, not functions. Consider renaming to streamsOptionsFor(date) or similar to signal "this one takes args" — or at minimum, add a comment.

---

serverUrlOptions declared at module scope inside a view file (source/views/settings/screens/overview/server-url.tsx:9). Every other query factory lives in a sibling query.ts. For consistency, move it out.

---


Add a note in the PR body about the recentPrintersOptions queryKey bug fix — it's buried in a commit and reviewers should know it's a behavioral change, not just a rename.

hawkrives:
Thanks for the thorough feedback (and thanks @drewvolz!). Addressed all points in 686a92e:

1. **`banner.test.tsx` mocking strategy** — Replaced the global `useQuery` mock with a real `QueryClientProvider` + `queryClient.setQueryData(['faqs'], ...)`. `../query` is still mocked with a minimal `faqsOptions` to avoid the ESM import chain through `@frogpond/api` → `ky`, but `useQuery` itself now runs for real.

2. **`useCourseData` deduplication** — Extracted a `buildQuery` helper so the `courseDataOptions` spread only appears once:
   ```ts
   let buildQuery = (term: TermType) => ({
     ...courseDataOptions(term, levels, gereqs),
     enabled: terms.length > 0,
   })
   return useQueries({
     queries: (filteredTerms.length ? filteredTerms : terms).map(buildQuery),
   })
   ```

3. **`streamsOptions` asymmetry** — Renamed to `streamsOptionsFor(date)` to signal that it takes args, distinguishing it from the other `*Options` values.

4. **`serverUrlOptions` location** — Moved out of `source/views/settings/screens/overview/server-url.tsx` into a new sibling `source/views/settings/screens/overview/query.ts`, matching the convention used by every other feature.

5. **PR body** — Updated with a "Behavioral fix" section calling out the `recentPrintersOptions` query key fix so reviewers don't miss it.

---

drewvolz (approved):

PR #7443, opened 2026-04-10
Upgrade React Native to 0.75.4

hawkrives:
## Summary

- Upgrade react-native from 0.74.7 to 0.75.4
- Upgrade react from 18.2.0 to 18.3.1
- Upgrade @react-native/babel-preset and @react-native/metro-config to 0.75.4
- Upgrade react-test-renderer to 18.3.1 and @types/react to 18.3.18
- Update react-native peerDependency from `^0.74.0` to `^0.75.0` in all 27 workspace modules

## What's new in RN 0.75

- **Yoga 3.1**: Percentage values for `gap`, `columnGap`, `rowGap`, and border radii
- **Faster autolinking**: ~1.5x faster on iOS
- **Improved Animated performance**

## Validation

- TypeScript: zero errors
- Jest: 200 tests pass (3 skipped)
- ESLint: 0 errors
- Prettier: clean
- Detox E2E: all 3 shards green on macOS runner

## TODO before merging

- [x] Regenerate `ios/Podfile.lock` via `pod install` on macOS
- [x] Verify iOS build succeeds on macOS
- [x] Verify Sentry patches (`contrib/0003-0005`) still apply after pod install

https://claude.ai/code/session_01NjUfL75ZC4HLMHN3vXwrjo


hawkrives:
It's still reusing the cached app… lemme try that again. 

Oh, it's because we haven't pushed the updated Podfile yet. Normally that's part of the upgrade. 

Once https://github.com/StoDevX/AAO-React-Native/issues/7445 is merged, we can have CI update the Podfile for us.

ios/Podfile
```diff
@@ -1,22 +1,8 @@
...
-# force the IPHONEOS_DEPLOYMENT_TARGET to a known value
-platform :ios, deployment_target
-
-# overriding ../node_modules/react-native/scripts/react_native_pods.rb
-# see https://github.com/CocoaPods/CocoaPods/pull/11232
-# prepare_react_native_project!
-install! 'cocoapods',
-	:deterministic_uuids => false,
-	:parallel_pod_downloads => true,
-	:parallel_pod_download_thread_pool_size => 40
```
drewvolz, line 19: In guessing the parallelized pod download was removed for some reason?

hawkrives: I was hoping it was default by now tbh

hawkrives: But I don't know.

.github/workflows/check.yml
```diff
@@ -184,8 +184,18 @@ jobs:
...
         if: steps.app-cache.outputs.cache-hit != 'true'
+        uses: actions/cache@668228422ae6a00e4ad889ee87cd7109ec5666a7 # v5.0.4
+        with:
+          path: ios/Pods
+          key: pods-${{ runner.os }}-xcode@${{ env.xcode_version }}-${{ hashFiles('**/project.pbxproj', '**/Podfile.lock', 'ios/AllAboutOlaf/**', '.github/job.yml') }}
+          restore-keys: |
+            pods-${{ runner.os }}-xcode@${{ env.xcode_version }}-
+
+      - name: Install CocoaPods
+        if: steps.app-cache.outputs.cache-hit != 'true' && steps.pods-cache.outputs.cache-hit != 'true'
```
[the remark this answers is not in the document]
hawkrives: The root concern here (stale Pods tree after a Podfile change that doesn't touch the lockfile) is addressed by adding `ios/Podfile` to the `hashFiles(...)` inputs of the `pods-cache` key in dd885c5. With that in place, any Podfile change invalidates the Pods cache entry, which naturally re-triggers the `mise run pods` step via the existing `steps.pods-cache.outputs.cache-hit != 'true'` condition — so running `pod install` on every cache hit isn't necessary (and would cost ~30-60s per green build for no gain).

---

package.json
```diff
@@ -133,14 +133,14 @@
     "jest": "29.7.0",
     "js-yaml": "4.1.1",
     "junk": "4.0.1",
-    "@react-native/babel-preset": "0.74.89",
+    "@react-native/babel-preset": "0.75.4",
     "minimist": "1.2.8",
```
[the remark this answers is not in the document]
hawkrives: The PR description was wrong, not the babel config. `@babel/plugin-transform-export-namespace-from` is **not** bundled in `@react-native/babel-preset@0.75.4` (I grepped the installed preset and the transform isn't there), and the codebase relies on it — e.g. `export * as studentwork from './student-work'` in source/views/sis/index.tsx:13 and the `export * as EventList` / `export * as EventDetail` entries in modules/event-list/index.ts. Removing the plugin would break the Metro bundle. I've updated the PR description to drop that stale claim; the plugin stays.

---

ios/Podfile
hawkrives: Fwiw `:parallel_pod_downloads` and `:parallel_pod_download_thread_pool_size` aren't standard CocoaPods `install!` options — the stock options in 1.16 are `:deterministic_uuids`, `:integrate_targets`, `:incremental_installation`, etc. ([source](https://github.com/CocoaPods/CocoaPods/blob/1.16.2/lib/cocoapods/installer/installation_options.rb)). CocoaPods would silently ignore the parallel ones, so they were effectively no-ops on the mise-pinned 1.16.2 toolchain and the old block was dead config. I dropped them while re-aligning to the upstream RN 0.75 Podfile template, rather than as a deliberate perf change.

That said, parallel _downloads_ are in fact the default now: `Pod::Downloader.preprocess_options` uses `Parallel.map` with `in_threads` set from the `cocoapods-downloader` gem's thread count, so fetching pod sources already happens concurrently. The part that's still serial in 1.16 is the per-pod _install/prepare_ phase, which `:parallel_pod_downloads` never touched anyway.

---

hawkrives:
From @drewvolz 

---

The banner.test.tsx mocking strategy is fragile. It now globally mocks @tanstack/react-query's useQuery:

```ts
 jest.mock('@tanstack/react-query', () => ({
   ...jest.requireActual(...),
   useQuery: jest.fn(),
 }))
``` 

This mocks useQuery for every component in the render tree, not just FaqBanner. Fine for the current isolated render, but it's a footgun if FaqBanner ever gets composed with a component that also calls useQuery in a test. An alternative: mock just the faqsOptions module and let the real useQuery run with a QueryClientProvider.

---

useCourseData loses a tiny bit of deduplication (source/views/sis/course-search/query.ts:67). The original had one createQueryOptions closure; the refactor inlines the ...courseDataOptions(term, levels, gereqs), enabled: terms.length > 0 spread twice (once for filteredTerms, once for terms). Minor, but:

```ts
 let buildQuery = (term: TermType) => ({
   ...courseDataOptions(term, levels, gereqs),
   enabled: terms.length > 0,
 })
 return useQueries({
   queries: (filteredTerms.length ? filteredTerms : terms).map(buildQuery),
 })
``` 

---

Inconsistent options-factory style. Most factories are arrow expressions that return queryOptions(...): export const fooOptions = (x: string) => queryOptions({...})

But heldJobsOptions uses a block body (source/views/stoprint/query.ts:57) and streamsOptions requires call-site invocation streamsOptions() because of the moment default arg (source/views/streaming/streams/list.tsx:74). The latter is a real asymmetry with every other fooOptions export, which are values, not functions. Consider renaming to streamsOptionsFor(date) or similar to signal "this one takes args" — or at minimum, add a comment.

---

serverUrlOptions declared at module scope inside a view file (source/views/settings/screens/overview/server-url.tsx:9). Every other query factory lives in a sibling query.ts. For consistency, move it out.

---


Add a note in the PR body about the recentPrintersOptions queryKey bug fix — it's buried in a commit and reviewers should know it's a behavioral change, not just a rename.

hawkrives:
@hawkrives @drewvolz — I think this feedback may have landed on the wrong PR. None of the files drewvolz mentions are in #7443's diff:

- `banner.test.tsx` — not touched
- `source/views/sis/course-search/query.ts` — not touched
- `source/views/stoprint/query.ts` (`heldJobsOptions`) — not touched
- `source/views/streaming/streams/list.tsx` (`streamsOptions`) — not touched
- `source/views/settings/screens/overview/server-url.tsx` (`serverUrlOptions`) — not touched
- `recentPrintersOptions` queryKey fix — no such change here

This PR's file list is limited to the RN 0.75 upgrade: `package.json` / `package-lock.json`, the 27 `modules/*/package.json` peerDep bumps, `ios/AllAboutOlaf/AppDelegate.mm`, `ios/Podfile`, `ios/Podfile.lock`, `ios/AllAboutOlaf.xcodeproj/project.pbxproj`, `mise.lock`, and `.github/workflows/check.yml`. No `source/views/**` or test changes.

Sounds like this review was meant for a different PR — possibly a React Query `queryOptions` / factory refactor. Happy to address the concerns on that PR once I know which one you mean.

---

drewvolz (approved):

PR #7560, opened 2026-04-19
upgrade CI to Xcode 26.4 / macos-26 / iPhone 17e

hawkrives:
## Summary

- Bump GitHub Actions runners from `macos-15` to `macos-26` in `build-and-deploy.yml`, `check.yml`, and `cocoapods.yml`
- Switch `xcode_version` env / explicit `xcode-select` calls from `Xcode_16.4` to `Xcode_26.4`
- Swap the UITest target simulator device from `iPhone 16 Pro` to `iPhone 17e`
- Rebase iOS app + ccache cache keys onto `host=macOS@26` so we don't read stale `macOS@15` artifacts

## Notes

- The `macos-26` image ships Xcode 26.4 (17E192), iOS 26.4 simulator runtime, and the iPhone 17e device — verified against [`actions/runner-images` macos-26-arm64 readme](https://github.com/actions/runner-images/blob/main/images/macos/macos-26-arm64-Readme.md).
- `cache_epoch` is left at `1` because the host/xcode fragments in the cache key already vary on runner image + Xcode bump, so reads naturally miss and the first run on macos-26 will repopulate.

## Test plan

- [ ] `Check` workflow passes — especially `ios-build` (xcodebuild build-for-testing with Xcode 26.4) and both `ios-uitest` shards (booting iPhone 17e / iOS 26.4)
- [ ] Scheduled `Build and Deploy` run or a manual `workflow_dispatch` builds on macos-26 without Fastlane / match regressions
- [ ] `Check & Update Cocoapods` still resolves `pod install` under Xcode 26.4 when triggered


contrib/0003-fmt-disable-consteval.patch
```diff
@@ -0,0 +1,14 @@
+--- ios/Pods/fmt/include/fmt/base.h
++++ ios/Pods/fmt/include/fmt/base.h
+@@ -124,9 +124,9 @@
+ #elif FMT_MSC_VERSION && FMT_MSC_VERSION < 1929
+ #  define FMT_USE_CONSTEVAL 0  // consteval is broken in MSVC VS2019 < 16.10.
+ #elif defined(__cpp_consteval)
+-#  define FMT_USE_CONSTEVAL 1
++#  define FMT_USE_CONSTEVAL 0  // Xcode 26 Clang rejects FMT_STRING in consteval (StoDevX patch)
```
drewvolz: Is this patching another patch, based upon the comment? Oh or we are saying it's our patch and not upstream

hawkrives: Your second reading was the intent — "this comment is from our patch, not upstream fmt." Agreed it reads ambiguously though, so I dropped the `(StoDevX patch)` suffix in `2ab9a66`. The sentinel substring matched by `apply-patches.sh` is preserved. The patch itself applies directly to upstream fmt 11.0.2's `include/fmt/base.h`.

---

drewvolz (approved):

ios/AllAboutOlafUITests/ModuleCampusMapTests.swift
```diff
@@ -16,8 +16,14 @@ class ModuleCampusMapTests: XCTestCase {
...
 		XCTAssertTrue(homescreen.waitForNonExistence(timeout: 30))
 
-		// dismiss SFSafariViewController
-		app.buttons["Done"].tap()
+		// dismiss SFSafariViewController; wait for the Done button since
+		// Safari view load timing varies (iOS 26 sim renders it slower than
+		// iOS 18 did, so tapping immediately would miss it)
+		let doneButton = app.buttons["Done"]
+		XCTAssertTrue(
+			doneButton.waitForExistence(timeout: 30),
```
drewvolz, line 24: This looks high

hawkrives: Agreed — 30s was the app-launch budget, but Safari's Done button is modal chrome that appears immediately on presentation, so 10s matches the in-view pattern used elsewhere in the suite. Lowered in `83c50e0`.

---

drewvolz (approved):

PR #7573, opened 2026-04-25
Add athletics modules

drewvolz:
This pull request introduces an Athletics section, providing users with an interface for viewing and filtering St. Olaf athletics scores. The implementation includes navigation integration, data fetching and grouping, UI components for listing scores, filtering by sport, and persistent filter preferences. Additionally, the app is configured to allow insecure HTTP loads from the athletics API domain for development purposes.

The most important changes are:

**Navigation Integration**
* Added the `Athletics` screen to the navigation stack and defined its navigation key and options (`routes.tsx`, `types.tsx`, `index.ts`). 

**Data Fetching and Grouping**
* Implemented data fetching from the athletics API and grouped scores by date for display, using a new query and utility functions (`query.ts`, `utils.ts`, `types.ts`). 

**UI Components**
* Created UI components for the Athletics section, including the main list view, tab bar for switching between date sections, row display for individual games, and a filter panel for selecting sports (`list.tsx`, `tabbar.tsx`, `row.tsx`, `filters.tsx`). 

**User Preferences and State Management**
* Added a persistent filter store using Zustand and AsyncStorage to save users' selected sports locally (`store.ts`).


source/views/athletics/list.tsx
```diff
@@ -0,0 +1,216 @@
...
+		const menSports = uniqueSports.filter((s) => s.includes("Men's"))
+		return [
+			{title: "Women's Sports", data: womenSports},
+			{title: "Men's Sports", data: menSports},
+		]
+	}, [data])
+
+	// Seed the filter store with all sports on first load
+	React.useEffect(() => {
+		if (selectedSports.length === 0 && sports.some((s) => s.data.length > 0)) {
```
hawkrives: I thought we saved the user's filter? Won't this bypass their saved filter on load and reset to no-filter?

source/views/athletics/utils.ts
```diff
@@ -0,0 +1,104 @@
...
+	'July',
+	'August',
+	'September',
+	'October',
+	'November',
+	'December',
+]
+
+/**
+ * Parse the API's "M/D/YYYY h:mm:ss AM/PM" date format into a JS Date.
```
hawkrives, line 29: Doesn't the API return a date_utc field? Is that one ISO?

source/views/athletics/utils.ts
```diff
@@ -0,0 +1,104 @@
...
+	const todayStart = startOfDay(now)
+	const yesterdayStart = new Date(todayStart)
+	yesterdayStart.setDate(yesterdayStart.getDate() - 1)
+
+	const yesterday: Score[] = []
+	const today: Score[] = []
+	const upcoming: Record<string, Score[]> = {}
+
+	for (const score of scores) {
+		const date = parseGameDate(score.date_utc)
```
hawkrives: Oh. I guess not. 

Should we parse this once, at query time? And/or transform it server-side?

ios/AllAboutOlaf/Info.plist
```diff
@@ -77,6 +77,11 @@
 		<true/>
 		<key>NSExceptionDomains</key>
 		<dict>
+			<key>athletics.stolaf.edu</key>
+			<dict>
+				<key>NSTemporaryExceptionAllowsInsecureHTTPLoads</key>
+				<true/>
+			</dict>
```
hawkrives: Given that the app is configured to load through CCC now, should we go ahead and take out this exception?

drewvolz: Good point!

drewvolz: @copilot please handle this

source/views/athletics/query.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const keys = {
+	all: ['athletics', 'scores'] as const,
+}
+
+export const athleticsOptions = queryOptions({
+	queryKey: keys.all,
+	queryFn: ({signal}): Promise<Score[]> =>
+		client.get('athletics/scores', {signal}).json<Score[]>(),
+	select: (scores): ProcessedScore[] =>
```
hawkrives, line 14: 🔴 **Critical: production date parsing fails on-device.**

This `select` builds `parsedDate` with `new Date(score.date_utc)`, but `utils.ts` ships a hand-written `parseGameDate` whose own doc-comment says Hermes (the RN engine) **cannot** parse the API's `"M/D/YYYY h:mm:ss AM/PM"` format via `Date.parse`.

So on a real device every `parsedDate` becomes `Invalid Date`: `isSameDay` and `date > todayStart` are all false in `groupScoresByDate`, and Yesterday/Today/Upcoming all render empty (headers read `undefined, undefined NaN`). The unit tests stay green only because they call `parseGameDate` directly under Node/V8 — the shipped path is never exercised.

**Fix:** `select: (scores) => scores.map((s) => ({...s, parsedDate: parseGameDate(s.date_utc)}))`.

— Claude (automated review)

source/views/athletics/row.tsx
```diff
@@ -0,0 +1,152 @@
...
+										accessibilityIgnoresInvertColors={true}
+										source={{uri: item.hometeam_logo}}
+										style={styles.teamLogo}
+									/>
+								) : null}
+								<Text style={styles.teamName}>{item.hometeam.trim()}</Text>
+							</View>
+
+							<View style={styles.gameInfo}>
+								{item.status.indicator === 'A' ? (
```
hawkrives: 🟠 **Finalized games never display their score.**

The score panel below is gated on `item.status.indicator === 'O'` (ongoing) only. The Today FINALIZED bucket in `list.tsx` selects `indicator !== 'O' && result !== ''`, so finalized games take the `indicator === 'A'` branch here and render `item.time` (or nothing) — the final `team_score`/`opponent_score` and the W/L `result` are never shown for a completed game.

Consider rendering the score panel whenever a score exists (e.g. `result !== ''` or both scores present), not just for live games.

— Claude (automated review)

source/views/athletics/list.tsx
```diff
@@ -0,0 +1,219 @@
...
+		const menSports = uniqueSports.filter((s) => s.includes("Men's"))
+		return [
+			{title: "Women's Sports", data: womenSports},
+			{title: "Men's Sports", data: menSports},
+		]
+	}, [data])
+
+	// Seed the filter store with all sports on first load — only after
+	// hydration is complete so we don't overwrite a previously saved filter.
+	React.useEffect(() => {
```
hawkrives, line 56: 🟡 **Deselecting all sports silently reverts to "show all".**

This seeding effect re-seeds `selectedSports` with every sport whenever `selectedSports.length === 0`. So when a user clears all filters (via the section "All" toggle or by toggling each off), `setSelectedSports([])` immediately triggers a re-seed on the next render and the UI snaps back to all-selected. There's no way to persist an empty selection, and the filter predicate already treats `length === 0` as "show all" — so the clear-all affordance is effectively dead.

— Claude (automated review)

source/views/athletics/utils.ts
```diff
@@ -0,0 +1,104 @@
...
+		return new Date(NaN)
+	}
+	const [datePart, timePart, ampm] = parts
+	const [month, day, year] = datePart.split('/').map(Number)
+	const [hours, minutes, seconds] = timePart.split(':').map(Number)
+	let hour24 = hours % 12
+	if (ampm === 'PM') {
+		hour24 += 12
+	}
+	return new Date(year, month - 1, day, hour24, minutes, seconds)
```
hawkrives, line 51: 🟡 **`date_utc` is interpreted as local time.**

`new Date(year, month-1, day, hour24, ...)` constructs a Date in the **device's local zone**, but the source field is named `date_utc`. For a UTC instant like `1/27/2025 2:00:00 AM`, a user at UTC-6 is really at 8 PM on Jan 26 — but this buckets and labels the game as Jan 27. Games near midnight UTC land in the wrong Today/Yesterday/Upcoming section with the wrong day-name header for any non-UTC user.

If the intent is to show the school's local wall-clock time, that's defensible — but then the field/handling should be documented as such rather than named `_utc`.

— Claude (automated review)

source/views/athletics/row.tsx
```diff
@@ -0,0 +1,152 @@
+import * as React from 'react'
+import {Image, PixelRatio, StyleSheet, Text, View} from 'react-native'
+import * as c from '@frogpond/colors'
+import {Score} from './types'
+
+type Props = {
+	data: Score[]
+	/** When true, show a date prefix above each game (used in Upcoming view). */
+	includePrefix?: boolean
```
hawkrives: 🟡 **`includePrefix` is declared and passed but never used.**

This prop is documented ("show a date prefix above each game") and `list.tsx` passes `includePrefix={selectedSection === Constants.UPCOMING}`, but `AthleticsRow` destructures only `{data}` and never reads it. The documented Upcoming date-prefix behavior doesn't exist. Either wire it up or drop the prop (and the per-row comparison in `list.tsx`).

— Claude (automated review)

source/views/athletics/list.tsx
```diff
@@ -0,0 +1,219 @@
...
+				.flatMap((s) => s.data)
+			const byDate: Record<string, ProcessedScore[]> = {}
+			for (const score of upcomingScores) {
+				const key = formatDateString(score.parsedDate)
+				if (!byDate[key]) {
+					byDate[key] = []
+				}
+				byDate[key].push(score)
+			}
+			return Object.keys(byDate)
```
hawkrives: 🟡 **Upcoming sections render in insertion order, not chronological order.**

`byDate` is keyed by formatted strings like `"Sunday, January 26"`. `Object.keys()` iterates non-integer string keys in **insertion order**, which here equals the order scores arrive from the API. If `athletics/scores` returns games grouped by sport (or otherwise unsorted), the Upcoming tab shows dates out of order (e.g. Feb 1 above Jan 28). Sort the keys (or the upcoming scores) by `parsedDate` before building sections.

— Claude (automated review)

hawkrives (commented):
Automated code review of the Athletics module — high-effort pass focused on correctness, with some cleanup notes. Two of these break core functionality on a real device; the rest are medium UX/correctness issues and minor cleanups. Details are left as inline comments.

**🔴 Critical**
- `query.ts` parses dates with `new Date(date_utc)` instead of the Hermes-safe `parseGameDate`, so all date grouping silently fails on-device (tests pass only under Node).
- `row.tsx` never displays a final score for finalized games (score panel is gated on `indicator === 'O'`).

**🟡 Medium:** can't persist an empty filter; `date_utc` treated as local time; unused `includePrefix` prop; non-chronological Upcoming ordering; brittle length-only filter-hint check.

**⚪ Cleanup:** duplicated Upcoming grouping; dead types; unrelated `.gitignore` entry.

— Posted by Claude (automated review), on behalf of @hawkrives

source/views/athletics/store.ts
```diff
@@ -0,0 +1,49 @@
...
+			version: 1,
+			onRehydrateStorage: () => (state) => {
+				state?.setHasHydrated(true)
+			},
+		},
+	),
+)
+
+export function selectShowChangeFiltersMessage(state: FilterState): boolean {
+	const {selectedSports, totalSports} = state
```
hawkrives: 🟡 **Filter-hint check compares only lengths, not sets.**

`selectedSports.length === totalSports` treats "all selected" as a count match. But `selectedSports` is persisted across sessions and can hold sports no longer in the current data, while `totalSports` counts only current sports. If a season drops one sport and adds another, the lengths can coincide while the sets differ — suppressing the hint even though the user is filtering out a current sport (and vice-versa). Comparing membership (e.g. every current sport is in `selectedSports`) would be robust.

— Claude (automated review)

source/views/athletics/list.tsx
```diff
@@ -0,0 +1,219 @@
...
+		}))
+	}, [baseData, selectedSports])
+
+	// Build sections to render depending on the selected tab
+	const sections = React.useMemo(() => {
+		if (selectedSection === Constants.FILTER) {
+			return []
+		}
+
+		if (selectedSection === Constants.UPCOMING) {
```
hawkrives, line 83: ⚪ **Duplicated date grouping.**

`groupScoresByDate` (utils.ts) already emits one section per upcoming day keyed by `formatDateString`. This branch flattens those sections back into a list and rebuilds the same `Record<string, ProcessedScore[]>` by `formatDateString` — re-implementing the shared util's grouping. The two copies can drift (key format, ordering). Consider reusing the upcoming sections `groupScoresByDate` already produced.

— Claude (automated review)

source/views/athletics/types.ts
```diff
@@ -0,0 +1,82 @@
...
+	opponent_score: string
+	result: GameResult
+	ip_time: string
+	prescore_info: string
+	postscore_info: string
+	links: Links
+	coverage: Coverage
+}
+
+export interface AthleticsResponse {
```
hawkrives: ⚪ **Dead types.**

`AthleticsResponse`, `AthleticsData`, and `GroupedScores` are exported but never imported anywhere in the module. `AthleticsResponse` also describes a `{scores: Score[]}` envelope that `query.ts` never uses (`queryFn` does `.json<Score[]>()`), and `GroupedScores` (`Score[]`) parallels `DateGroupedScores` (`ProcessedScore[]`). Removing them avoids a future reader having to reconcile two "grouped scores" shapes.

— Claude (automated review)

.gitignore
```diff
@@ -76,4 +76,5 @@ fastlane/play-private-key.json
 
 # generated logs files in CI
 /logs
+.worktrees/
```
hawkrives: ⚪ **Unrelated to this feature.**

`.worktrees/` looks like a personal git-worktree / dev-tool artifact with no connection to the Athletics module. Bundling it into a feature PR muddies `.gitignore` blame. Worth splitting out (and possibly into a personal global gitignore instead).

— Claude (automated review)

drewvolz:

hawkrives (approved):

drewvolz (commented):
✅

PR #7585, opened 2026-05-09
discover local ccc-server via mDNS on the Server URL screen

drewvolz:
## Summary

In dev mode, the Settings → Server URL screen now scans for a `_ccc-server._tcp` mDNS service on the local network and presents any discovered servers as tappable cells. Tapping fills the URL field; the user still presses Save to apply and restart.

Companion PR: https://github.com/frog-pond/ccc-server/pull/1066

## How it works

A new `useServerDiscovery` hook wraps `react-native-zeroconf` and:
- Returns empty immediately if not in dev mode — zero cost for production users
- Guards against the native module being null (before `pod install`) so the screen doesn't crash
- Scans for `_ccc-server._tcp` services on mount, stops on unmount

The `ServerUrlSection` renders a **LOCAL SERVERS** section (dev mode only, hidden until at least one server is found) with `PushButtonCell` rows.

## Setup

```sh
# package install, podfile install, pods linkage
...

# Start ccc-server with mDNS advertisement
mise run stolaf-college:mdns  # in the ccc-server repo

# Run a debug build on a device on the same network
# Navigate to Settings → Server URL
```

Screenshot 2026-05-09 at 1 04 21 PM

## Changes
- `source/views/settings/screens/overview/use-server-discovery.ts` — new hook
- `source/views/settings/screens/overview/server-url.tsx` — LOCAL SERVERS section
- `types/react-native-zeroconf.d.ts` — type declaration for untyped package
- `ios/AllAboutOlaf/Info.plist` — `NSBonjourServices` + `NSLocalNetworkUsageDescription` (required by iOS 14+)
- `CLAUDE.md` — documents the local discovery workflow

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -64,6 +64,17 @@ mise run tsc    # Type check
 mise run pods   # Install cocoapods, even on Linux
 ```
 
+### Local Server Discovery
+
+In dev mode (debug builds, or with the dev-mode override enabled in Settings), the Settings → Server URL screen will automatically discover a `ccc-server` instance running on the same network via mDNS. Discovered servers appear as tappable cells; tapping one fills the URL field.
+
+To use this:
+1. Start `ccc-server` with mDNS advertisement enabled: `mise run stolaf-college:mdns` (in the `ccc-server` repo)
+2. Run a debug build of the app on a device on the same network
+3. Navigate to Settings → Server URL — the server will appear automatically
+
+The feature uses `react-native-zeroconf` (native pod). If the pod hasn't been linked yet (`mise run pods`), discovery is silently skipped — the screen won't crash.
+
 ## Agent Workflow
 
 **Session startup:** Always run `mise run agent:setup` at the start of every session. This installs dependencies and bundles data files.
--- a/ios/AllAboutOlaf.xcodeproj/project.pbxproj
+++ b/ios/AllAboutOlaf.xcodeproj/project.pbxproj
@@ -1158,6 +1158,7 @@
 				"${PODS_CONFIGURATION_BUILD_DIR}/React-cxxreact/React-cxxreact_privacy.bundle",
 				"${PODS_CONFIGURATION_BUILD_DIR}/Sentry/Sentry.bundle",
 				"${PODS_CONFIGURATION_BUILD_DIR}/boost/boost_privacy.bundle",
+				"${PODS_CONFIGURATION_BUILD_DIR}/glog/glog_privacy.bundle",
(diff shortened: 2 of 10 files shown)


hawkrives:
## Code review

I pulled the `react-native-zeroconf@0.14.0` source/README to verify the new hook against the actual API. Most of the diff is lockfiles and the autogenerated `project.pbxproj`/`Podfile.lock`, which I treated as generated and out of scope. Four findings, most-severe first:

### 1. `onRemove` wipes the entire server list on any single removal — `use-server-discovery.ts:43`
The `remove` event fires per-service with the removed service's name, but the handler ignores `_name` and calls `setServers([])`. With two or more ccc-servers on the network (or normal mDNS churn where one entry is re-announced), losing one server clears *all* discovered servers, including ones still present.

```ts
const onRemove = (name: string) => {
	setServers((prev) => prev.filter((s) => s.name !== name))
}
```

### 2. URL built from `service.host` (an mDNS `.local.` hostname) instead of `addresses[0]` — `use-server-discovery.ts:35`
The package README documents `host` as e.g. `'XeroxPrinter.local.'` — a hostname *with a trailing dot* — while `addresses` holds the resolved IPs (`'192.168.1.23'`). The cell produces `http://hostname.local.:port/v1/`; RN's fetch is not guaranteed to resolve a trailing-dot `.local.` name, so tapping a discovered server can yield a URL that won't load even though discovery "worked." Prefer `service.addresses[0]` (bracket it if IPv6).

### 3. Empty-string `txt.path` defeats the `/v1/` default — `use-server-discovery.ts:34`
`service.txt?.path ?? '/v1/'` only substitutes the default for `null`/`undefined`. If a server advertises `path=''`, the empty string passes through and the URL loses its path entirely (`http://host:port`). Use `service.txt?.path || '/v1/'` if an empty path should fall back.

### 4. Redundant dev-mode gate and duplicate `useIsDevMode()` call — `server-url.tsx:23,68`
`useServerDiscovery()` already returns `[]` whenever `!isDevMode`, so `discoveredServers.length > 0` implies dev mode. The extra `isDevMode &&` in the render guard plus the second `useIsDevMode()` call in the screen are redundant — the hook is the single source of truth. (Cleanup, not a crash.)

For what it's worth, the `OTHER_LDFLAGS` rewrite to `("$(inherited)", " ")` and the added `glog_privacy.bundle` are CocoaPods-generated artifacts, not hand-authored, so I didn't flag them.

---
🤖 Reviewed by Claude (Claude Code), at Wren's request.

drewvolz:

hawkrives (approved):

PR #7598, opened 2026-05-27
Reddit post detail: context menu, link detection, and swipe-to-dismiss

drewvolz:
## What's changed

**Context menu on post detail header**
Replaces the single "open in browser" button with an ellipsis context menu that offers two actions on tap:
- **Open in Browser** (safari SF Symbol)
- **Share** (square.and.arrow.up SF Symbol)

This avoids assuming the user wants to open the URL immediately — they may just want to share it.

**Tappable links in post body**
Switches the post body from `htmlToFormattedText` (plain text, no links) to `htmlToSegments`, the same approach already used in comment rows. Links in post content are now highlighted and tappable, matching comment behavior.

**Swipe-to-dismiss browser**
Changes `expo-web-browser` `presentationStyle` from `CURRENT_CONTEXT` → `PAGE_SHEET` so the in-app browser can be dismissed with a swipe down.

**Extended `@frogpond/context-menu`**
`actions` now accepts `ContextMenuAction` objects `{ key, title?, icon? }` in addition to plain strings, making SF Symbol icons possible per item. Fully backward-compatible — all existing callers are unaffected.

---

https://github.com/user-attachments/assets/42150555-8116-4606-833b-7111da90e679


source/views/reddit/useRedditLinkHandler.ts
```diff
@@ -0,0 +1,75 @@
...
+	return React.useCallback(
+		async (url: string) => {
+			const redditMatch = parseRedditPostUrl(url)
+			if (!redditMatch) {
+				openUrl(url)
+				return
+			}
+
+			try {
+				const post = await fetchRedditPost(redditMatch.postUrl)
```
hawkrives: **Fire-and-forget fetch with no `AbortSignal` — double-tap and mid-navigation race.**

Each link tap runs this async handler to completion: `await fetchRedditPost(...)` then `navigation.push(...)`. A double-tap fires two handlers and pushes two identical `RedditPostDetail` screens. Navigating away while the fetch is in flight still triggers `navigation.push` when it resolves.

`fetchRedditPost` already accepts a `signal?: AbortSignal` that's never passed here — wiring it to an `AbortController` (cancelled on unmount/next tap) plus an in-flight guard would close both gaps.

— Claude Code

source/views/reddit/post-detail.tsx
```diff
@@ -127,8 +155,8 @@ export function PostDetailView(): React.ReactNode {
 	]
 		.filter((part): part is string => Boolean(part))
 		.join(' · ')
-	const bodyText = sanitizeBodyText(
-		contentHtml ? htmlToFormattedText(contentHtml) : '',
+	const bodySegments = sanitizeBodySegments(
```
hawkrives: **Body parse runs unmemoized on every render.**

`htmlToSegments(contentHtml)` + `sanitizeBodySegments` re-parse the entire post HTML on every render — and this view re-renders on every collapse toggle, fullscreen-index change, and header menu state update. `menuItems` is memoized but the body isn't. Wrapping in `useMemo(..., [contentHtml])` avoids the repeated DOM parse.

— Claude Code

source/views/reddit/post-detail.tsx
```diff
@@ -206,9 +233,21 @@ export function PostDetailView(): React.ReactNode {
 				) : null}
 
 				{/* Body text */}
-				{bodyText ? (
+				{bodySegments.length > 0 ? (
 					<Text selectable={true} style={styles.body}>
-						{bodyText}
+						{bodySegments.map((seg, i) =>
```
hawkrives: **Segment-rendering JSX duplicated with `comment-row.tsx`.**

This `segments.map(seg => link/text <Text>)` block is essentially copy-pasted from `comment-row.tsx`, and the two copies already diverge — `comment-row` uses `(onLinkPress ?? openUrl)(seg.url)` while this uses `handleLinkPress(seg.url)`. Extracting a shared `<SegmentedText segments onLinkPress>` component would keep link rendering in one place.

— Claude Code

source/views/reddit/post-detail.tsx
```diff
@@ -385,6 +427,10 @@ const styles = StyleSheet.create({
 		lineHeight: 22,
 		marginTop: 4,
 	},
+	link: {
```
hawkrives: **`styles.link` duplicated verbatim.**

This is identical to `styles.link` in `comment-row.tsx` (`{color: c.link, textDecorationLine: 'underline'}`). Co-locating it with the shared segment component (see other comment) means a theme/decoration change happens once instead of in two StyleSheets.

— Claude Code

source/views/reddit/post-detail.tsx
```diff
@@ -139,8 +167,7 @@ export function PostDetailView(): React.ReactNode {
 		return []
 	})()
 
-	const isCrosspost =
-		postType === 'crosspost' || (!bodyText && !thumbnail && !linkUrl)
+	const isCrosspost = postType === 'crosspost' || crosspostParent != null
```
hawkrives: **Redundant clause + a dropped fallback worth confirming.**

`parsePost` only ever sets `crosspostParent` when `postType === 'crosspost'`, so `crosspostParent != null` can never be the deciding operand — this reduces to `isCrosspost = postType === 'crosspost'`.

Separately, this replaces the old `(!bodyText && !thumbnail && !linkUrl)` heuristic, which rendered the crosspost card for content-less posts. Those posts will now render blank instead. If that's intentional, the second operand can just be dropped; if not, the empty-post case needs another home.

— Claude Code

source/views/reddit/post-detail.tsx
```diff
@@ -56,10 +59,10 @@ function flattenComments(
 	})
 }
 
-function sanitizeBodyText(raw: string): string {
-	const trimmed = raw.trimStart()
-	if (trimmed.toLowerCase().startsWith('submitted by')) return ''
-	return raw
+function sanitizeBodySegments(segments: Segment[]): Segment[] {
```
hawkrives, line 63: **`submitted by` strip only inspects the first *text* segment.**

The old `sanitizeBodyText` checked the start of the fully-flattened text. This finds the first segment whose `type === 'text'` and checks its `trimStart()`. If the post HTML leads with an anchor (link segment) before the `submitted by` text, `firstText` won't start with `submitted by` and the boilerplate line is no longer removed. Rare, but a behavior change from the string version.

— Claude Code

source/views/reddit/reddit-api.ts
```diff
@@ -240,6 +240,23 @@ export async function fetchRedditPosts(
 	return parseRedditPostsJson(json)
 }
 
+export async function fetchRedditPost(
+	postUrl: string,
+	signal?: AbortSignal,
+): Promise<RedditPostType | null> {
+	const parsed = new URL(postUrl)
+	const jsonPath = parsed.pathname.replace(/\/$/u, '') + '.json'
```
hawkrives: **`.json` URL construction duplicated with `fetchRedditComments`.**

This block (parse URL → strip trailing slash → append `.json` → add `?raw_json=1`, plus the `User-Agent` header on the fetch) is repeated almost verbatim a few lines down in `fetchRedditComments`. A small `redditJsonUrl(postUrl)` helper would keep the endpoint scheme in one place.

— Claude Code

hawkrives (commented):
Automated code review by **Claude Code** (Opus 4.8), run on request. High-effort, recall-biased pass over the diff. No blocking bugs — the items below are one likely UX bug plus correctness edge cases and cleanup. Inline comments follow.

drewvolz:

hawkrives (approved):

hawkrives (approved):

PR #7610, opened 2026-08-05
Expo 57, RN 0.86, @expo/ui, expo-prebuild, and CNG, oh my

hawkrives:
- [x] we've gone from React Native 0.81.5 to 0.86.2
- [x] we've gone from Expo 54 to 57
- [x] the home screen is now SwiftUI (vie `@expo/ui`)
- [x] we've dropped the `react-native-context-menu` dependency for `@expo/ui`
- [x] we've dropped the `@react-native-community/datepicker` dependency for `@expo/ui`
- [x] we've dropped the `react-native-clipboard` dependency for `@expo/ui`
- [x] we're now using `expo prebuild` (think `xcodegen`) to automatically generate the Xcode project for us, including the Podfile and Podfile.lock
- [x] app variants (dev builds installed alongside the prod build!)

followups:
- [ ] migrate Settings to SwiftUI (FormGroup and friends)
- [ ] npm to pnpm


drewvolz (approved):

CLAUDE.md
```diff
@@ -4,13 +4,21 @@
...
 - **Fastlane** for CI/CD
 - Monorepo with internal packages in `modules/`
 
+## Commit Messages
+
+No conventional-commit prefixes. Write `Drop the migration planning docs`, not
+`chore: drop the migration planning docs`.
+
+Use the imperative mood, capitalised, with no trailing full stop. Explain in the
+body why the change was needed, not what the diff already shows.
```
drewvolz: Eh, likely okay? but the robot thought to say something

>CLAUDE.md:19  makes a commit body sound mandatory: “Explain in the body why…”. Most recent repository commits are subject-only, and the section never explicitly changes that practice. This leaves contributors and agents uncertain whether every commit now requires a body. Change it to: “When including a body, explain why the change was needed rather than restating the diff.”
> 
> The policy otherwise has no conflict with commit tooling: the beta trigger ( [ci run beta] ) is matched anywhere in the commit message and remains usable.

.claude/skills/build-to-device/SKILL.md
drewvolz, line 1: I like it!

images/icons/app-icon-development.png
drewvolz, line 1: 📸

plugins/__tests__/fixtures/Podfile
drewvolz, line 1: Had to look this one up

> The fixture tests config plugins that modify the generated Podfile.
> 
> Even though Expo generates the initial Podfile, this project’s plugins still patch it—for example, to preserve custom native settings that Expo does not express directly.  plugins/__tests__/fixtures/Podfile  is a stable input representing Expo’s generated output, letting tests verify those patches remain valid and idempotent without running  expo prebuild  during every Jest run.
> 
> It should be updated whenever the relevant Expo-generated Podfile structure changes; otherwise, it can mask a broken plugin anchor.

source/views/home/button.tsx
```diff
@@ -1,70 +1,125 @@
...
 }
 
+export const CELL_MARGIN = 10
+const CELL_RADIUS = 17
+const cellVerticalPadding = 8
+const cellHorizontalPadding = 4
+
+/// SwiftUI has no "fill the available width" constant reachable from JS, so we
+/// cap the frame at a width no phone reaches and let the stack divide the space.
+export const FILL_WIDTH = 10_000
```
drewvolz, line 30: 🤔

source/views/reddit/index.tsx
```diff
@@ -1,13 +1,33 @@
...
+	font,
+	foregroundColor,
+	padding,
+	shapes,
+} from '@expo/ui/swift-ui/modifiers'
 import * as c from '@frogpond/colors'
 
+/// The Ionicons glyphs come from the app-wide font registered through
+/// `UIAppFonts` in Info.plist; SwiftUI addresses it by its family name.
+const ICON_FONT_FAMILY = 'Ionicons'
```
drewvolz, line 26: Is there a way to reuse this constant that's defined 3 times?

source/views/transportation/bus/components/day-picker.tsx
```diff
@@ -183,25 +218,65 @@ export const DayPickerHeader = ({
...
+						modifiers={[
+							padding({horizontal: 8, vertical: 6}),
+							background(c.systemBackground, headerButtonShape),
+							contentShape(headerButtonShape),
+							strokeBorder({
+								color: accentColor,
+								cornerRadius: HEADER_BUTTON_RADIUS,
+								shape: 'roundedRectangle',
+								style: {lineWidth: 1},
+							}),
```
drewvolz, line 245: SwiftUI (in swift itself) lets you create reusable modifiers, I'm assuming we could do the same with these if we have common props we're applying?

uitests/UITestCase.swift
```diff
@@ -11,6 +11,11 @@ class UITestCase: XCTestCase {
 
 		app = XCUIApplication()
 		app.launchArguments.append(TestIdentifiers.LaunchArguments.uiTesting)
+		// Reset persisted state for every test. Without this, UserDefaults and
+		// AsyncStorage carry over between tests in a run, so a test's result can
+		// depend on what ran before it -- testLongPressNoticeTogglesDevMode
+		// inverts if dev mode is already on, and failed only in long runs.
+		app.launchArguments.append(TestIdentifiers.LaunchArguments.resetState)
```
drewvolz, line 18: 🥳

app.config.ts
```diff
@@ -0,0 +1,193 @@
...
+		scheme: 'AllAboutOlaf',
+		icon: './images/icons/app-icon.png',
+	},
+	development: {
+		displayName: 'AAO Dev',
+		bundleIdentifier: `${BUNDLE_ID}.dev`,
+		scheme: 'AllAboutOlafDev',
+		// The shipping icon with a diagonal DEV ribbon across the top-right
+		// corner, which crosses sky rather than the building.
+		icon: './images/icons/app-icon-development.png',
```
drewvolz, line 40: I want to swap these icons and make the windmill the default again 😄

package-lock.json
```diff
@@ -9456,134 +7931,131 @@
...
+      "peerDependenciesMeta": {
+        "react-native-worklets": {
+          "optional": true
+        }
+      }
+    },
+    "node_modules/expo-modules-jsi": {
+      "version": "57.0.4",
+      "license": "MIT",
+      "peerDependencies": {
```
drewvolz, line 7975: 🤖 
> package-lock.json:7972  locks  expo-modules-jsi@57.0.4 , which fails under the project’s Xcode 26 toolchain: its Swift sources declare  weak let  properties. A clean CNG prebuild,  pod install , and  xcodebuild build-for-testing  exits 65, blocking simulator and release builds. Upgrade to a fixed Expo modules release or patch  weak let  to  weak var  through the existing patch framework.

.github/workflows/check.yml
```diff
@@ -152,15 +156,20 @@ jobs:
...
+      # without touching any tracked iOS file. Omitting them meant a patch-only
+      # change reused a cached app built without that patch.
       - name: Check for cached iOS app
         id: app-cache
         uses: actions/cache/restore@668228422ae6a00e4ad889ee87cd7109ec5666a7 # v5.0.4
         with:
           lookup-only: true
           path: ios/build/Build/Products/
-          key: ios-app/e=${{ env.cache_epoch }}${{ env.cache_bust }}/host=macOS@26/target=ios/xcode=${{ env.xcode_version }}/ref=${{ github.head_ref || github.ref_name }}/hash=${{ hashFiles('**/project.pbxproj', 'ios/Podfile', '**/Podfile.lock', 'ios/AllAboutOlaf/**', 'ios/AllAboutOlafUITests/**') }}
+          key: ios-app/e=${{ env.cache_epoch }}${{ env.cache_bust }}/host=macOS@26/target=ios/xcode=${{ env.xcode_version }}/ref=${{ github.head_ref || github.ref_name }}/hash=${{ hashFiles('app.config.ts', 'plugins/**', 'uitests/**', 'package-lock.json', 'contrib/**', 'scripts/apply-patches.sh', '.mise.toml') }}
```
drewvolz: 🤖 
>  omit  images/**  / asset inputs from the CNG app-cache key. An icon or other prebuild asset change can reuse a stale  .app , skip iOS compilation, and let UI tests pass against old native assets. Include all prebuild-consumed assets or use an Expo fingerprint.

.github/workflows/check.yml
```diff
@@ -152,15 +156,20 @@ jobs:
...
           restore-keys: |
-            ios-app/e=${{ env.cache_epoch }}${{ env.cache_bust }}/host=macOS@26/target=ios/xcode=${{ env.xcode_version }}/ref=master/hash=${{ hashFiles('**/project.pbxproj', 'ios/Podfile', '**/Podfile.lock', 'ios/AllAboutOlaf/**', 'ios/AllAboutOlafUITests/**') }}
+            ios-app/e=${{ env.cache_epoch }}${{ env.cache_bust }}/host=macOS@26/target=ios/xcode=${{ env.xcode_version }}/ref=master/hash=${{ hashFiles('app.config.ts', 'plugins/**', 'uitests/**', 'package-lock.json', 'contrib/**', 'scripts/apply-patches.sh', '.mise.toml') }}
```
drewvolz: 🤖 
>  omit  images/**  / asset inputs from the CNG app-cache key. An icon or other prebuild asset change can reuse a stale  .app , skip iOS compilation, and let UI tests pass against old native assets. Include all prebuild-consumed assets or use an Expo fingerprint.

plugins/with-binary-stripping.ts
```diff
@@ -0,0 +1,85 @@
...
+		Object.assign(settings, STRIPPING_SETTINGS)
+	}
+	return project
+}
+
+const withBinaryStripping: ConfigPlugin = (config) =>
+	withXcodeProject(config, (mod) => {
+		mod.modResults = applyStripping(mod.modResults, APP_TARGET)
+		return mod
+	})
```
drewvolz, line 83: 🤖 
> no longer applies  STRIP_INSTALLED_PRODUCT=YES  to Pod targets, so release frameworks retain symbols and increase the shipped app size. Restore the Pod-target setting in an idempotent Podfile hook.

plugins/with-xcuitest-target.ts
```diff
@@ -0,0 +1,297 @@
...
+	if (!appTargetKey) {
+		throw new Error(
+			`with-xcuitest-target: no \`${APP_TARGET}\` target to test. The app target is renamed or missing.`,
+		)
+	}
+
+	// `unit_test_bundle` is the closest type the xcode package knows; the product
+	// type is corrected below. Both are .xctest bundles, so the product file it
+	// creates is already right.
+	let target = project.addTarget(name, 'unit_test_bundle', name, BUNDLE_ID)
```
drewvolz, line 176: 🤖 
> leaves the generated UI-test target’s product reference as  AllAboutOlafUITests.mdimporter  instead of  .xctest . Rewrite the generated  PBXFileReference  and add a regression assertion.

source/views/home/index.tsx
```diff
@@ -39,40 +38,59 @@ function HomePage(): React.ReactNode {
...
+		<Host
+			matchContents={false}
+			modifiers={[accessibilityIdentifier('screen-homescreen')]}
+			style={styles.host}
 		>
-			<FaqBannerGroup style={styles.banner} target={FAQ_TARGETS.HOME} />
+			<ScrollView>
+				<VStack
+					modifiers={[
+						padding({all: CELL_MARGIN}),
```
drewvolz, line 53: 🤖 
> adds SwiftUI padding around a hosted FAQ banner that already has margins, producing unintended double (20pt) insets. Remove one spacing layer.

drewvolz (approved):

plugins/with-binary-stripping.ts
hawkrives: apparently most of our dependencies ship static frameworks now, which are stripped at app link time and as such don't respect the old setting, but are in fact stripped by the app-target stripping

source/views/home/index.tsx
hawkrives: i didn't see this one during manual tests… verifying…

hawkrives: Besides the invisible text, looks OK to me?

app.config.ts
hawkrives: lol ok!

source/views/reddit/index.tsx
hawkrives: Yeah, but I'd rather move these towards being SF Symbols and away from ionicons personally

source/views/transportation/bus/components/day-picker.tsx
hawkrives: Yep, you can apparently just import and export them as usual.

source/views/home/button.tsx
hawkrives: For now, I'd like to leave it. I'd love to look into using a SwiftUI grid or something after this lands.

package-lock.json
hawkrives: apparently this would be a problem if we were compiling said framework ourselves, but because we're consuming the binary framework instead, it's not an issue.

source/views/home/index.tsx
drewvolz: The default background/colors I chose need tweaking. Thanks for checking

plugins/with-xcuitest-target.ts
hawkrives: fixed

.github/workflows/check.yml
hawkrives: thanks, fixed!

.github/workflows/check.yml
hawkrives: thanks, fixed!

CLAUDE.md
hawkrives: updated!

PR #7619, opened 2026-08-06
ci: bundle data before committing in the cocoapods workflow

hawkrives:


drewvolz (approved):

hawkrives:
Superseded by #7620.

Rather than adding `mise run bundle-data` to cocoapods.yml, the `tsc` mise task now declares `depends = ["prepare", "bundle-data"]`. `source/views/faqs/local-faqs.ts` imports `docs/faqs.json`, so that dependency is real and now encoded rather than documented.

That fixes the failure wherever `tsc` runs — including the hk pre-commit hook this workflow triggers via push-on-podfile-change — instead of only in the one workflow that happened to be missing the step. Which is how the bug arrived: check.yml had `bundle-data` before its checks and cocoapods.yml did not.

bundle-data takes ~0.1s and is idempotent.

PR #7630, opened 2026-08-07
Replace the manual ios-cache-check hash with expo-fingerprint

hawkrives:
## Summary

Started as a diagnostic spike comparing [`expo-fingerprint`](https://docs.expo.dev/versions/latest/sdk/fingerprint/) against the hand-maintained `hashFiles()` glob list gating `ios-cache-check` in `.github/workflows/check.yml`. The spike data supports the swap, so this now wires it in for real.

### What the spike found

- `@expo/fingerprint`'s default iOS output recursively hashes autolinked native-module directories under `node_modules` (not just `package.json`/lockfile versions), so it already caught the exact bug class the manual hash needed a special fix for: applying `contrib/0004-change-icon-legacy-module.patch` (which edits `node_modules/react-native-change-icon/ios/ChangeIcon.h`) changed the fingerprint hash, confirmed both locally and in CI (`31d0dcd...` both times, reproducible across Node 22 locally and Node 24 in CI).
- A `uitests/**`-only change left the fingerprint hash unchanged while the manual `hashFiles()` key changed — confirmed in CI ([run comparison](https://github.com/StoDevX/AAO-React-Native/actions/runs/31141626312)): `f8f8df73...` → `bda7adfc...` (manual) vs `31d0dcdf...` → `31d0dcdf...` (fingerprint, unchanged). Expected: expo-fingerprint has no notion of XCUITest targets.

### What changed

- `ios-cache-check`'s cache key now hashes `steps.fingerprint.outputs.hash` (from `@expo/fingerprint fingerprint:generate --platform ios`, run after `npm ci` + `mise run prepare` so patches are applied first) combined with `hashFiles('uitests/**', '.mise.toml', 'contrib/**', 'scripts/apply-patches.sh')` for what the fingerprint can't see:
  - `uitests/**` — XCUITest sources, not part of native project generation
  - `.mise.toml` — pins the Ruby/fastlane toolchain, independent of what `prebuild` generates
  - `contrib/**`, `scripts/apply-patches.sh` — `0003-fmt-disable-consteval.patch` targets `ios/Pods/fmt/include/fmt/base.h`, a CocoaPods-generated file that doesn't exist in `node_modules` for the fingerprint step to see, so a change to that patch would otherwise be invisible
- Dropped `app.config.ts`, `plugins/**`, and `package-lock.json` from the manual hash — all covered by the fingerprint now (config plugins and `app.config.ts` show up as direct fingerprint sources; native dependency changes get picked up via the post-`npm ci` `node_modules` state the fingerprint hashes).
- Pinned `@expo/fingerprint` as a direct `devDependency` (`0.20.6`) rather than relying on it riding along transitively through `expo`, so an unrelated Expo version bump can't silently change or drop it.

### Trade-off

`ios-cache-check` goes from a near-instant `hashFiles()` lookup to needing `npm ci` + `mise run prepare` before it can compute the fingerprint — roughly a minute added to every `check` run, cache hit or not.

## Test plan

- [x] Confirmed locally and in CI that the fingerprint hash changes on the `react-native-change-icon` patch and is stable across Node versions
- [x] Confirmed in CI that a `uitests/**`-only change leaves the fingerprint hash unchanged while the manual key changes
- [ ] Let this run across a few more real PRs/pushes to confirm no unexpected cache misses/hits, including a native-config-only change (`app.config.ts`/`plugins/**`) and a `.mise.toml`-only change
- [ ] Mark ready for review once confidence is high


.github/workflows/check.yml
```diff
@@ -155,21 +155,46 @@ jobs:
...
+      # node_modules/react-native-change-icon/ios/ChangeIcon.h without
+      # touching any tracked iOS file.
+      - name: Compute native-project fingerprint
+        id: fingerprint
+        run: |
+          npm ci
+          mise run prepare
+          npx --no-install fingerprint fingerprint:generate --platform ios > /tmp/expo-fingerprint.json
+          hash=$(node -pe "JSON.parse(require('fs').readFileSync('/tmp/expo-fingerprint.json')).hash")
+          echo "hash=$hash" >> "$GITHUB_OUTPUT"
```
hawkrives, line 171:
```suggestion
      - uses: jdx/mise-action@1648a7812b9aeae629881980618f079932869151 # v4.0.1
      - run: npm ci
      - run: mise run prepare

      # @expo/fingerprint hashes what prebuild would actually generate:
      # app.config.ts, plugins/**, and (after `prepare` applies contrib/**'s
      # patches, above) the patched contents of any autolinked native module
      # under node_modules, including package-lock.json-driven version bumps.
      - name: Compute native-project fingerprint
        id: fingerprint
        run: |
          npx --no-install fingerprint fingerprint:generate --platform ios > /tmp/expo-fingerprint.json
          hash=$(jq .hash /tmp/expo-fingerprint.json)
          echo "hash=$hash" >> "$GITHUB_OUTPUT"
```

hawkrives: Agreed on splitting the steps and switching to `jq` — took that as-is in `b27e0ab`.

One fix: `jq .hash` without `-r` returns the JSON-encoded string, quote characters included (`"31d0dcdfc6e8a8e78b130d7b7287a5ae6e823232"` rather than `31d0dcdfc6e8a8e78b130d7b7287a5ae6e823232`), which would land in the cache key as literal `"` characters. Every other `jq`-into-`$GITHUB_OUTPUT` call already in this file (the simulator-selection steps further down) uses `-r` for exactly that reason, so I used `jq -r .hash` to match.

---

drewvolz (approved):

hawkrives:
@claude check for merge conflicts please

hawkrives:
Yes — real conflicts, and bigger than a simple rebase. #7610 (this PR's old base) merged, GitHub retargeted this PR to `master`, and `master` now also has #7631 (npm → pnpm migration), which touches the exact same territory this PR does:

- `npm ci` → `pnpm install --frozen-lockfile` everywhere in `check.yml`
- `package-lock.json` → `pnpm-lock.yaml`
- `.mise.toml` → `mise.toml`
- `contrib/**` + `scripts/apply-patches.sh` are gone entirely — patches moved to pnpm's native `patchedDependencies` (`patches/*.patch` + `pnpm-workspace.yaml`), applied automatically during install, no separate step
- `contrib/0003-fmt-disable-consteval.patch` was dropped as stale (confirmed: gone from the tree, no replacement)
- `contrib/0004-change-icon-legacy-module.patch` survived as `patches/react-native-change-icon@5.0.0.patch`, same fix, applied by pnpm now

Master's own `ios-cache-check` hash already adapted to this independently (now hashes `pnpm-lock.yaml` instead of the old patch-file glob). Given that our whole rationale for keeping `contrib/**`/`scripts/apply-patches.sh` in the supplementary hash was specifically the now-gone fmt patch, resolving this isn't a mechanical merge — it changes what this PR should even look like. Checking with the thread on how you'd like to proceed before I touch it.

---

hawkrives:
Closing this spike. While porting it to pnpm (following the #7631 migration), I found a hard blocker: `@expo/fingerprint@0.20.6` silently returns `hash: null` for 32 of 44 sources under pnpm's `node_modules` layout — including `react-native` itself and every autolinked native module (`react-native-change-icon`, `expo-modules-core`, etc.), since they live under the `.pnpm` virtual-store/symlink paths the tool apparently can't hash.

Confirmed concretely: I hand-reverted the `react-native-change-icon` patch (the exact regression this whole spike was built around catching) back to its unpatched state and regenerated the fingerprint — the hash didn't change at all. Under pnpm, this cache key would essentially never invalidate on native changes, which is the stale-cache failure class this repo has already hit twice.

Master's own `ios-cache-check` key (now hashing `pnpm-lock.yaml` directly) doesn't have this problem and is the right approach for now. No newer stable `@expo/fingerprint` release exists to retest against; there's a related-but-different open upstream issue (expo/expo#44808, spurious hash *changes* under pnpm) but not this null-hash/false-negative bug specifically.

---

PR #7667, opened 2026-08-08
Strip HTML tags with the parser instead of a regex

hawkrives:
Resolves the CodeQL `js/polynomial-redos` alert on `modules/html-lib/index.ts:21`.

`@frogpond/html-lib` already parses HTML with `htmlparser2`, so this now uses the same parser rather than adding a dependency. 

Resolves https://github.com/StoDevX/AAO-React-Native/security/code-scanning/1

---
_Generated by [Claude Code](https://claude.ai/code/session_01D1tMgYUBZ4mQygyAi2BZN5)_

### DIFF
--- a/modules/html-lib/index.ts
+++ b/modules/html-lib/index.ts
@@ -17,12 +17,39 @@ export function innerTextWithSpaces(elem: AnyNode): string {
 	return textContent(elem).split(/\s+/u).join(' ').trim()
 }
 
-export function removeHtmlWithRegex(str: string): string {
-	return str.replace(/<[^>]*>/gu, ' ')
+function collectText(nodes: ChildNode[], out: string[]): void {
+	for (const node of nodes) {
+		if (isText(node)) {
+			out.push(node.data)
+			continue
+		}
+
+		if (!isTag(node)) continue
+
+		// `<script>` and `<style>` bodies are code, not prose; dropping them
+		// keeps stylesheets and JS out of what we render as text.
+		const tag = node.name.toLowerCase()
+		if (tag === 'script' || tag === 'style') continue
+
+		collectText(node.children, out)
+	}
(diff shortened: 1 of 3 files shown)


modules/html-lib/index.ts
```diff
@@ -17,12 +17,39 @@ export function innerTextWithSpaces(elem: AnyNode): string {
...
+		if (!isTag(node)) continue
+
+		// `<script>` and `<style>` bodies are code, not prose; dropping them
+		// keeps stylesheets and JS out of what we render as text.
+		const tag = node.name.toLowerCase()
+		if (tag === 'script' || tag === 'style') continue
+
+		collectText(node.children, out)
+	}
+}
```
hawkrives, line 44: can we make this a loop instead of a recursive function?

hawkrives: Done in 5bb2f4a — it's an explicit stack now, children pushed in reverse so they pop back off in document order.

Worth doing on its own merits, too: nesting depth is uncontrolled input just like the tag soup, and the recursive version overflowed the stack between 5k and 20k levels. htmlparser2 builds the tree iteratively and parses that input without complaint, so the walk was the only thing that couldn't take it:

```
depth  5000 recursive: ok -> "deep"
depth  5000 iterative: ok -> "deep"
depth 20000 recursive: threw RangeError: Maximum call stack size exceeded
depth 20000 iterative: ok -> "deep"
```

Added a test at 20k nested `<div>`s to hold that. I left `walkNodes` and `walkSegments` recursive — same exposure, but they're untouched by this PR and converting them is a bigger change than a ReDoS fix should carry. Happy to do it here or in a follow-up, whichever you prefer.

---

drewvolz (approved):

PR #7674, opened 2026-08-09
expo router home stoprint

hawkrives:


hawkrives:
### Home screen

Home screen with seven tiles

### stoPrint

stoPrint not-logged-in empty state

Not reachable without live St. Olaf credentials in this environment: the jobs list, printer list, and release-with-printer happy-path screens. The not-found branches on the two nested detail routes were reachable via deep link and confirmed working:

Printer selection not-found state

Release job not-found state

hawkrives:
Follow-up fix: `support.tsx` (used by stoPrint's error view for the Report button) still imported `useNavigation`/`NavigationProp` as a runtime value from `@react-navigation/native`, tripping the same Metro SDK56 check the earlier `modules/navigation-buttons` fix addressed. Same fix applied — `useNavigation` from `expo-router`, `NavigationProp` kept type-only. Confirmed via Metro's own import-stack trace and confirmed this is the only consumer of `support.tsx`, so no other open PR in this stack is affected.

drewvolz (approved):

PR #7679, opened 2026-08-09
expo router home streaming

hawkrives:


hawkrives:
### Home screen

Home screen with nine tiles

### Streaming Media

Streaming Media tab bar with header and back button

Webcams tab with live thumbnails

KSTO tab

The KSTO/KRLX radio-schedule screens are deliberately not wired in this PR — their shared calendar dependency has its own unrelated react-navigation import that needs to be handled as part of Calendar's own group PR later in this stack. Tapping "Schedule" shows expo-router's built-in Unmatched Route screen until then, not a crash:

Unmatched Route screen for the deferred schedule route

drewvolz (approved):

PR #7692, opened 2026-08-09
expo router home faq

hawkrives:

### DIFF
--- a/app/(home)/Faq.tsx
+++ b/app/(home)/Faq.tsx
@@ -1,0 +1,19 @@
+import * as React from 'react'
+import {Stack} from 'expo-router'
+
+import {View as FaqView, NavigationOptions} from '../../source/views/faqs'
+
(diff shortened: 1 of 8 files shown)


hawkrives:
SIS Balances tab — the FAQ banner is now tappable (previously a no-op fallback since the SIS group PR, since this route didn't exist yet):



Faq screen, reached by tapping the banner (real tap, not a deep link):



This is the 15th and final group PR in checkpoint 2's stack — every home-grid-tile group has now been migrated from React Navigation to expo-router. One pre-existing, out-of-scope bug found during verification: the tapped FAQ card's highlight border (`c.tintColor`, a PlatformColor) doesn't render visibly — confirmed unrelated to this PR's changes (swapping the color to a literal renders fine, isolating it to that PlatformColor's resolution, likely a missing AccentColor asset). Flagging for visibility, not fixed here.

drewvolz (approved):

PR #7697, opened 2026-08-09
expo router settings static

hawkrives:

### DIFF
--- a/app/(component-library)/_layout.tsx
+++ b/app/(component-library)/_layout.tsx
@@ -1,0 +1,6 @@
+import * as React from 'react'
+import {Stack} from 'expo-router'
+
+export default function ComponentLibraryLayout(): React.ReactNode {
+	return <Stack />
+}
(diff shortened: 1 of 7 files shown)


hawkrives:
Home screen — unaffected by this PR:



Credits — deep-linked, modal sheet presentation:



Privacy — pushed within the same sheet (back button reads "Credits"), confirming the nested-group architecture works as designed:



Legal — pushed within the same sheet again (back button reads "Privacy"):



First PR of checkpoint 4's 8-PR stack (Settings + Component Library migration). Both new route groups are intentionally unreachable until PR 8 (`SettingsRoot`) adds a real entry point — verified via direct deep link, not a tap-through, which is expected for this PR.

One thing worth flagging: final review caught a real bug (cold-start deep link into a modal group mounted nothing beneath the sheet) — fixed and verified live. A second, narrower issue surfaced during that same verification (dismissing a cold-start deep-linked sheet lands on the Menus screen rather than the real Home tile grid) — a fix attempt was tried, verified live, found not to work, and reverted rather than left in as an unverified claim. That issue is real but doesn't block this PR (every screen here is deep-link-only for now), and is documented as an open follow-up in the fix commit's message.

drewvolz (approved):

PR #7851, opened 2026-09-05
Quiet bundle-data, and document the simulator workflows

hawkrives:
Working on reducing the context usage for robots

1. Add a skill for running ui tests
2. Add CLAUDE.md in uitests/
3. Make bundle-data quiet by default

### DIFF
--- a/.claude/skills/run-on-simulator/SKILL.md
+++ b/.claude/skills/run-on-simulator/SKILL.md
@@ -1,0 +1,87 @@
+---
+name: run-on-simulator
+description: Use when you need to look at the app itself on a simulator - checking a change by eye, screenshotting a screen, or reproducing something by hand, rather than running the XCUITests or building to a phone
+---
+
+# Run the App on a Simulator
+
+For seeing a change with your own eyes. To *assert* something about a screen,
+write an XCUITest instead — see `run-uitests`. For a physical iPhone, see
+`build-to-device`.
+
+## The whole thing, headless
+
+```bash
+# 1. Build and install. Slow the first time (30+ min), incremental after.
+mise run ios          # expect it to fail at the last step; see below
+
+# 2. Boot a simulator.
+xcrun simctl list devices available | grep iPhone
+xcrun simctl boot <UDID>; xcrun simctl bootstatus <UDID> -b
+
+# 3. Serve the JavaScript, and wait until it answers.
+npx expo start --port 8081 > /tmp/metro.log 2>&1 &
+until curl -sf http://localhost:8081/status | grep -q running; do sleep 1; done
+
+# 4. Launch it pointed at Metro, and look.
+xcrun simctl openurl <UDID> \
+  "NFMTHAZVS9.com.drewvolz.stolaf://expo-development-client/?url=http%3A%2F%2Flocalhost%3A8081"
+xcrun simctl io <UDID> screenshot /tmp/shot.png    # then Read the png
+```
+
+No GUI is needed at any point. `simctl` drives a booted simulator whether or
+not anything is on screen, and `io screenshot` reads the framebuffer directly.
+
+## The two steps that mislead
+
+**`mise run ios` fails at the end, after doing the useful work.** It builds,
+signs and installs, then tries to raise the Simulator GUI and reports:
+
+```
+CommandError: Simulator app did not open fast enough.
+Try opening Simulator first, then running your app.
+```
+
+Take its advice only once you have checked the GUI exists. Some Xcode installs
+have no `$(xcode-select -p)/Applications/` directory at all, so
+`open -a Simulator` answers `Unable to find application named 'Simulator'` and
+no amount of retrying will help. **The app is installed either way** — the
+failure is in raising a window, not in the build. Ignore it and go to step 2.
+
+**`simctl launch` gets you a red screen.** Launching by bundle id alone:
+
+```bash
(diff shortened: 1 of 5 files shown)


drewvolz:
## Skill Evaluation Report

Analyzed both skills against the [skill-creator](https://github.com/anthropics/claude-plugins-official/tree/main/skill-creator) guidelines.

---

### run-on-simulator (87 lines) — **Grade: B+**

| Criterion | Score | Notes |
|-----------|-------|-------|
| Description quality | 7/10 | Good scope ("simulator, not tests, not phone") but could be pushier for triggering |
| Progressive disclosure | 9/10 | Compact, under 500 lines |
| Why explanations | 9/10 | Excellent on the misleading steps (Simulator error, red screen) |
| Structure | 8/10 | Clear sections, good cross-refs to `run-uitests` and `build-to-device` |
| Actionability | 7/10 | Copy-paste commands, but `<UDID>` placeholder could use capture guidance |
| Bundled resources | 5/10 | No scripts despite repetitive 4-step sequence |

**Strengths:** The "Two steps that mislead" section is excellent — it explains *why* the error messages are deceptive rather than just saying "ignore them." The bundle identifier table is practical.

**Weaknesses:** The description says what it's for but doesn't push hard on triggering. The 4-step workflow could be a bundled `scripts/run-simulator.sh` with UDID as an argument.

---

### run-uitests (160 lines) — **Grade: A-**

| Criterion | Score | Notes |
|-----------|-------|-------|
| Description quality | 8/10 | Strong scope definition, mentions red-green proving |
| Progressive disclosure | 9/10 | Good length, clear hierarchy |
| Why explanations | 10/10 | Outstanding — "Why step 3 is what makes this affordable" is the gold standard |
| Structure | 9/10 | Four commands section, troubleshooting, writing new tests |
| Actionability | 8/10 | Practical workflow including "watch the test fail first" |
| Bundled resources | 5/10 | Complex commands still inline |

**Strengths:** 
- The "Why step 3 is what makes this affordable" section is exemplary skill writing — it explains the reasoning so the reader can adapt, not just follow robotically
- The red-green workflow (revert fix, watch fail, restore, watch pass) teaches discipline
- The `capture()` advice ("put it *before* the assertion") prevents a common mistake with clear reasoning

**Weaknesses:** Same description pushiness issue. The 4-step command sequence is long and error-prone to copy; a helper script would reduce friction.

---

### uitests/CLAUDE.md (77 lines) — **Grade: A**

Not a skill, but well-crafted directory context:
- Layout table gives instant orientation
- Code example shows ideal style (method chaining, no raw queries)
- Conventions explain *why* (drift, element-type variance, tap-drop)
- Two gotchas are genuinely the two things that will bite

---

## Recommendations

1. **Add triggering push to descriptions** — Make them slightly more aggressive about when to activate

2. **Bundle the command sequences** — A `scripts/build-uitest.sh` and `scripts/run-uitest.sh` would reduce copy-paste errors and make the skills more actionable

3. **Add UDID capture helper** — Both skills use `<UDID>` placeholders; a one-liner like:
   ```bash
   UDID=$(xcrun simctl list devices available | grep 'iPhone 16' | head -1 | grep -oE '[A-F0-9-]{36}')
   ```
   would eliminate manual lookup

4. **The skills complement each other well** — The cross-references are appropriate and the scope boundaries are clear

---

**Overall:** Solid, well-reasoned skills. The run-uitests skill in particular is unusually good at explaining *why* things work the way they do. Main gap is bundled scripts — these workflows are complex enough that helper scripts would significantly improve usability.

---
https://claude.ai/code/session_01BrD1biXGFFFGdR1xURU5km

drewvolz (approved):

hawkrives:
Claude says:

---

Thanks for the read. Taking these one at a time, because I want to push back on two of the three.

**Bundle the command sequences into scripts** — I'd rather not. The arguments vary every run (the `-only-testing` target, the UDID, the result bundle path), so a script would end up passing most of them straight through, and what's left is the part the skill exists to convey: *why* each step is there. Step 3 is the clearest case. `SKIP_BUNDLING=true` is the reason a JavaScript edit needs no rebuild, and that fact is what makes the red-green loop cheap enough to bother with. Behind a script it becomes a step you run without knowing that. If the copy-paste friction is the real problem, the repo's answer is a mise task — `mise run uitest -- <filter>` — not a `scripts/` directory inside `.claude/skills/`.

**Add a UDID capture helper** — agreed in principle, but the proposed one-liner doesn't work:

```bash
UDID=$(xcrun simctl list devices available | grep 'iPhone 16' | head -1 | grep -oE '[A-F0-9-]{36}')
```

`grep 'iPhone 16'` also matches "iPhone 16 Pro", "iPhone 16 Pro Max" and "iPhone 16 Plus", so `head -1` picks whichever Xcode happened to list first — and the hardcoded model name rots at every Xcode release. Showing `simctl list devices available` and letting the reader pick was deliberate: it makes you choose a device that actually exists on your machine, which is also the difference between a local run and CI's pinned `iPhone 17e`.

**Push harder on triggering in the descriptions** — mildly fair, and the easiest to act on, but both already follow the `Use when …` convention with explicit scope boundaries ("simulator, not tests, not phone"). I'd rather leave them than make them shouty to satisfy a rubric.

One thing the review didn't cover: it graded the three markdown files and skipped `scripts/bundle-data.mjs` and `bundle-data-dir.mjs`, which are the only executable change here and the only part that can break `agent:setup` or the pre-commit hook. Worth a second pair of eyes on the `--verbose` split and the `-h` / `--help` branch in particular.

PR #7854, opened 2026-09-05
Fix ongoing events and filter out past events

drewvolz:
Fixes calendar issues preventing rendering due to missing categories. Other fixes are detailed below.

## Summary
- `isOngoing` now requires both `startTime < today` AND `endTime > now` (events that ended are no longer marked ongoing)
- Filter out past events in query select - only events with `endTime > now` are shown
- Export `now()` from timer module that respects frozen date in UI testing mode
- Jest runs with `isUITesting: true` so tests use the frozen date consistently
- Guard against undefined categories in Calendar.tsx

## Test plan
- [x] Jest tests pass (1118 tests)
- [x] Verify ongoing section only shows events that haven't ended
- [x] Verify past events (before today) are not shown in the calendar

### DIFF
--- a/app/(home)/Calendar.tsx
+++ b/app/(home)/Calendar.tsx
@@ -25,14 +25,14 @@ export default function CalendarPage(): React.ReactNode {
 	let {selectedCategory, selectCategory} = useCalendarFilterStore()
 
 	let availableCategories = useMemo(() => {
-		let cats = new Set(events.flatMap((e) => e.event.categories))
+		let cats = new Set(events.flatMap((e) => e.event.categories ?? []))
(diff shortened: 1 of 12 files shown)


hawkrives (approved):

drewvolz:
@hawkrives hmmm this seems unrelated 
> Test Suite 'ModuleDirectoryTests' started at 2026-09-05 23:23:01.326.
Error:     testCancelledSwipeBackKeepsTheQuery, Failed to get matching snapshot: No matches found for first query match sequence: `Descendants matching type SearchField`, given input App element pid: 23701

drewvolz:
☝️ That's now being fixed in https://github.com/StoDevX/AAO-React-Native/pull/7859

PR #7855, opened 2026-09-05
Replace News toolbar with unified source+category picker

drewvolz:
## Summary

- Port the Calendar picker pattern to News: a single `@expo/ui Menu` with sections per source and category toggles
- Selecting a category implicitly selects its source; deselecting returns to all stories from that source
- Move news filter state from Redux to zustand (matching Calendar)
- Fetch both sources to populate the picker with all categories

## Bug fixes

- **Loading state**: now renders centered (`LoadingView` outside `List`) instead of a thin bar at top
- **Filter button lag**: fixed by using stable `@expo/ui Menu` button style instead of `Stack.Toolbar.Menu` with variant toggling
- **Missing categories**: added `?? []` guard for stories without categories (matches #7854)

## Test plan

- [x] Open News, verify loading spinner is centered
- [x] Open picker, verify both source sections appear with categories
- [x] Tap a category from St. Olaf News → stories filter to that category
- [x] Tap same category again → filter clears, shows all St. Olaf stories
- [x] Tap a category from The Olaf Messenger → switches source and filters
- [x] Verify picker button turns blue when filter is active

https://github.com/user-attachments/assets/42cac856-9e57-48de-bf26-f5052d5335b1

### DIFF
--- a/app/(home)/News.tsx
+++ b/app/(home)/News.tsx
@@ -2,52 +2,56 @@ import * as React from 'react'
 import {Stack} from 'expo-router'
 import {useQueries} from '@tanstack/react-query'
 
-import {extractCategories, NewsList} from '../../source/features/news/news-list'
+import {NewsList} from '../../source/features/news/news-list'
 import {NewsPicker} from '../../source/features/news/news-picker'
+import {
+	combineNewsResults,
+	type NewsFeedQuery,
+	type NewsFeeds,
+} from '../../source/features/news/lib/combine'
+import {resolveCategory} from '../../source/features/news/lib/util'
 import {namedNewsOptions} from '../../source/features/news/query'
 import {NEWS_SOURCES} from '../../source/features/news/sources'
 import {useNewsFilterStore} from '../../source/features/news/store'
+
+const NEWS_SOURCE_IDS = NEWS_SOURCES.map((s) => s.id)
+
+// Defined out here so react-query can memoize the fold; a combine rebuilt on
+// every render is re-run on every render.
+const combineNewsFeeds = (results: NewsFeedQuery[]): NewsFeeds =>
+	combineNewsResults(NEWS_SOURCE_IDS, results)
 
 export default function NewsPage(): React.ReactNode {
 	let {selectedSource, selectedCategory, select} = useNewsFilterStore()
 	let source = NEWS_SOURCES.find((s) => s.id === selectedSource) ?? NEWS_SOURCES[0]
 
-	// Fetch all sources to populate the picker with categories
-	let queries = useQueries({
+	// Every source is fetched, not only the one on screen: the picker lists the
+	// categories each source offers, and the Menu has no "opened" callback to
+	// defer the others to. The cost is one extra feed per cold open.
+	let {entriesBySource, categoriesBySource, queryBySource, unavailableSources} = useQueries({
 		queries: NEWS_SOURCES.map((s) => namedNewsOptions(s.id)),
+		combine: combineNewsFeeds,
 	})
 
-	// The query for the currently selected source
-	let currentQueryIndex = NEWS_SOURCES.findIndex((s) => s.id === selectedSource)
-	let currentQuery = queries[currentQueryIndex >= 0 ? currentQueryIndex : 0]
-
-	// Build categories by source from fetched data
-	// Destructure data arrays to satisfy @tanstack/query(no-unstable-deps)
-	let dataArrays = queries.map((q) => q.data)
-	let categoriesBySource = React.useMemo(() => {
-		let result: Record<string, string[]> = {}
-		NEWS_SOURCES.forEach((s, i) => {
-			let data = dataArrays[i] ?? []
-			result[s.id] = extractCategories(data)
-		})
-		return result
-		// eslint-disable-next-line react-hooks/exhaustive-deps
-	}, dataArrays)
+	let categories = categoriesBySource[source.id] ?? []
+	let category = resolveCategory(selectedCategory, categories)
 
 	return (
 		<>
 			<Stack.Screen options={{title: source.title}} />
 			<NewsList
-				query={currentQuery}
-				selectedCategory={selectedCategory}
+				entries={entriesBySource[source.id] ?? []}
+				query={queryBySource[source.id]}
+				selectedCategory={category}
 				thumbnail={source.thumbnail}
 			/>
 			<NewsPicker
 				categoriesBySource={categoriesBySource}
 				onSelect={select}
-				selectedCategory={selectedCategory}
-				selectedSource={selectedSource}
+				selectedCategory={category}
+				selectedSource={source.id}
 				sources={NEWS_SOURCES}
+				unavailableSources={unavailableSources}
 			/>
 		</>
 	)
--- a/app/(home)/News.tsx
+++ b/app/(home)/News.tsx
@@ -1,27 +1,53 @@
 import * as React from 'react'
 import {Stack} from 'expo-router'
-import {useQuery} from '@tanstack/react-query'
-import {useDispatch, useSelector} from 'react-redux'
+import {useQueries} from '@tanstack/react-query'
 
-import {NewsList} from '../../source/features/news/news-list'
+import {extractCategories, NewsList} from '../../source/features/news/news-list'
+import {NewsPicker} from '../../source/features/news/news-picker'
 import {namedNewsOptions} from '../../source/features/news/query'
 import {NEWS_SOURCES} from '../../source/features/news/sources'
-import {selectNewsSource, setNewsSource} from '../../source/redux/parts/settings'
+import {useNewsFilterStore} from '../../source/features/news/store'
 
 export default function NewsPage(): React.ReactNode {
-	let dispatch = useDispatch()
-	let selectedId = useSelector(selectNewsSource)
-	let source = NEWS_SOURCES.find((candidate) => candidate.id === selectedId) ?? NEWS_SOURCES[0]
+	let {selectedSource, selectedCategory, select} = useNewsFilterStore()
+	let source = NEWS_SOURCES.find((s) => s.id === selectedSource) ?? NEWS_SOURCES[0]
+
+	// Fetch all sources to populate the picker with categories
+	let queries = useQueries({
+		queries: NEWS_SOURCES.map((s) => namedNewsOptions(s.id)),
+	})
+
+	// The query for the currently selected source
+	let currentQueryIndex = NEWS_SOURCES.findIndex((s) => s.id === selectedSource)
+	let currentQuery = queries[currentQueryIndex >= 0 ? currentQueryIndex : 0]
+
+	// Build categories by source from fetched data
+	// Destructure data arrays to satisfy @tanstack/query(no-unstable-deps)
+	let dataArrays = queries.map((q) => q.data)
+	let categoriesBySource = React.useMemo(() => {
+		let result: Record<string, string[]> = {}
+		NEWS_SOURCES.forEach((s, i) => {
+			let data = dataArrays[i] ?? []
+			result[s.id] = extractCategories(data)
+		})
+		return result
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, dataArrays)
 
 	return (
 		<>
 			<Stack.Screen options={{title: source.title}} />
 			<NewsList
-				onSelectSource={(id) => dispatch(setNewsSource(id))}
-				query={useQuery(namedNewsOptions(source.id))}
-				selectedSourceId={source.id}
-				sources={NEWS_SOURCES}
+				query={currentQuery}
+				selectedCategory={selectedCategory}
 				thumbnail={source.thumbnail}
+			/>
+			<NewsPicker
+				categoriesBySource={categoriesBySource}
+				onSelect={select}
+				selectedCategory={selectedCategory}
+				selectedSource={selectedSource}
+				sources={NEWS_SOURCES}
 			/>
 		</>
 	)
--- a/source/features/news/news-list.tsx
+++ b/source/features/news/news-list.tsx
@@ -7,33 +7,28 @@ import {LoadingView, NoticeView} from '@frogpond/notice'
 import {openUrl} from '@frogpond/open-url'
 import type {StoryType} from './types'
 import {NewsRow} from './news-row'
-import {cleanEntries, trimStoryCateogry} from './lib/util'
+import {filterByCategory} from './lib/util'
 import {emptyStateProps} from './lib/empty-state'
-import {UseQueryResult} from '@tanstack/react-query'
+import type {NewsFeedQuery} from './lib/combine'
 
 type Props = {
-	query: UseQueryResult<StoryType[]>
+	query: NewsFeedQuery
+	/** The selected source's stories, cleaned by the same pass that built the picker's categories */
+	entries: StoryType[]
 	thumbnail: false | ImageResolvedAssetSource
 	selectedCategory: string | null
 }
 
-let getStoryCategories = (story: StoryType) => {
-	return (story.categories ?? []).map((category) => trimStoryCateogry(category))
-}
-
 export const NewsList = (props: Props): React.ReactNode => {
-	let {data = [], error, refetch, isError, isLoading} = props.query
+	let {error, refetch, isError, isLoading} = props.query
+	let {entries, selectedCategory} = props
 
-	let entries = React.useMemo(() => cleanEntries(data), [data])
+	let filteredEntries = React.useMemo(
+		() => filterByCategory(entries, selectedCategory),
+		[entries, selectedCategory],
+	)
 
-	let filteredEntries = React.useMemo(() => {
-		if (props.selectedCategory === null) return entries
-		return entries.filter((story) =>
-			getStoryCategories(story).includes(props.selectedCategory as string),
-		)
-	}, [entries, props.selectedCategory])
-
-	let hasActiveFilter = props.selectedCategory !== null
+	let hasActiveFilter = selectedCategory !== null
 
 	if (isLoading) {
 		return <LoadingView />
@@ -77,13 +72,6 @@ export const NewsList = (props: Props): React.ReactNode => {
 			</VStack>
 		</Host>
 	)
-}
-
-/** Extract unique categories from stories, sorted A-Z */
-export function extractCategories(stories: StoryType[]): string[] {
-	let cleaned = cleanEntries(stories)
-	let cats = new Set(cleaned.flatMap((story) => getStoryCategories(story)))
-	return [...cats].sort()
 }
 
 const styles = StyleSheet.create({
--- a/source/features/news/news-list.tsx
+++ b/source/features/news/news-list.tsx
@@ -1,43 +1,24 @@
 import * as React from 'react'
 import {StyleSheet, type ImageResolvedAssetSource} from 'react-native'
-import {Stack} from 'expo-router'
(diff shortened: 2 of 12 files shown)


hawkrives:
Request: add All Stories category to each source

source/features/news/news-picker.tsx
```diff
@@ -0,0 +1,88 @@
...
+	categoriesBySource,
+	selectedSource,
+	selectedCategory,
+	onSelect,
+}: Props): React.ReactNode {
+	let isActive = selectedCategory !== null
+	// Keep the modifier list structurally identical every render — only the
+	// colour value changes. Swapping modifier types/count rebuilds the native
+	// Menu and closes it mid-interaction.
+	let menuModifiers = [LABEL, foregroundStyle(isActive ? c.systemBlue : c.label)]
```
hawkrives, line 44: **The same Menu-collapse bug is still live in `CalendarPicker`.**

Commit `50b05e17d` fixed it here by keeping the modifier list structurally identical. But `NewsPicker` was forked from `modules/ccc-calendar/calendar-picker.tsx`, and that file still has the exact construct the commit message calls out:

```ts
const ACTIVE_STYLE = [LABEL, buttonStyle('borderedProminent'), tint(c.systemBlue), foregroundStyle(c.white)]
const INACTIVE_STYLE = [LABEL, foregroundStyle(c.label)]
let menuModifiers = isActive ? ACTIVE_STYLE : INACTIVE_STYLE
```

Four modifiers vs two — types and count both change. Repro on Calendar: open the picker, tap the selected category to clear it → `isActive` flips to false → the native Menu rebuilds and closes mid-interaction, the same defect this PR fixes for News.

Fixing the shared component instead of the copy would also close the visual gap this leaves: Calendar's active picker is a filled blue pill, News' is now just a blue glyph.

source/features/news/store.ts
```diff
@@ -0,0 +1,24 @@
...
+type NewsFilterStore = {
+	selectedSource: string
+	selectedCategory: string | null
+	select: (source: string, category: string | null) => void
+}
+
+export const useNewsFilterStore = create<NewsFilterStore>()(
+	persist(
+		(set) => ({
+			selectedSource: 'stolaf',
```
hawkrives, line 14: **The category filter is now persisted forever, and nothing ever validates it against the feed.**

The deleted `news-list.tsx` code kept the chosen categories in `React.useState` on purpose — the removed comment said the categories on offer come from the feed, so the reader's narrowing shouldn't be tied to the story list. It reset every launch. This store persists it to AsyncStorage instead.

Scenario: reader filters to "Athletics"; two weeks later the feed no longer carries that category. On launch, `filteredEntries` is empty → `ContentUnavailableView` with "No stories to show. Try changing the filters." In the picker, no `Toggle` matches (`selectedCategory === cat` is false for every category) **and** "All Stories" is off (`selectedCategory !== null`). Every switch reads off while the list is filtered to nothing, with only the blue glyph as a tell — permanently, across launches, until the reader guesses to tap "All Stories".

Dropping `selectedCategory` when it isn't in `categoriesBySource[selectedSource]` would fix it.

app/(home)/News.tsx
```diff
@@ -1,28 +1,54 @@
...
-				sources={NEWS_SOURCES}
+				query={currentQuery}
+				selectedCategory={selectedCategory}
 				thumbnail={source.thumbnail}
 			/>
+			<NewsPicker
+				categoriesBySource={categoriesBySource}
+				onSelect={select}
+				selectedCategory={selectedCategory}
+				selectedSource={selectedSource}
```
hawkrives: **The picker gets the raw persisted id while the list gets the fallback, so the two desync.**

Line 13 and line 22 both fall back to `NEWS_SOURCES[0]` when the persisted `selectedSource` isn't a known id; the picker gets `selectedSource` unresolved.

Scenario: a persisted `selectedSource` of `'oleville'` (a source id that gets renamed or removed later — `sources.ts` already notes Oleville exists in `data/sources.yaml`). The title bar and story list show St. Olaf News, but inside the picker `isSourceSelected` is false for **both** sections, so no category is on and both "All Stories" toggles read off. The reader sees St. Olaf stories with a picker claiming nothing is selected.

```suggestion
				selectedSource={source.id}
```

source/features/news/store.ts
```diff
@@ -0,0 +1,24 @@
...
+import {create} from 'zustand'
+import {persist, createJSONStorage} from 'zustand/middleware'
+
+type NewsFilterStore = {
+	selectedSource: string
+	selectedCategory: string | null
+	select: (source: string, category: string | null) => void
+}
+
+export const useNewsFilterStore = create<NewsFilterStore>()(
```
hawkrives, line 11: **No tests for any of the new logic, and five existing ones were deleted.**

This PR removes `describe('news source selection')` (5 tests) from `source/redux/parts/__tests__/settings.test.ts` and adds nothing. Untested and newly added: this store, `extractCategories`, and the `selectedCategory` filter predicate — all pure JS.

That is exactly what the repo CLAUDE.md asks Jest to cover: *"assert what is genuinely decided in JavaScript: pure functions, reducers, parsers, data shaping, hooks, and which branch a component chose."* The global CLAUDE.md adds *"Reducing test coverage is worse than failing tests"* and requires red/green TDD for every feature.

The pattern already exists in the repo — `source/features/calendar/store.ts`, which this store mirrors, has `source/features/calendar/__tests__/store.test.ts` covering default, set, change, and clear. `source/features/news/lib/__tests__/` covers `cleanEntries`, `trimStoryCateogry`, and `emptyStateProps`. Nothing here is hard to test; it just wasn't.

app/(home)/News.tsx
```diff
@@ -1,28 +1,54 @@
...
+	})
+
+	// The query for the currently selected source
+	let currentQueryIndex = NEWS_SOURCES.findIndex((s) => s.id === selectedSource)
+	let currentQuery = queries[currentQueryIndex >= 0 ? currentQueryIndex : 0]
+
+	// Build categories by source from fetched data
+	// Destructure data arrays to satisfy @tanstack/query(no-unstable-deps)
+	let dataArrays = queries.map((q) => q.data)
+	let categoriesBySource = React.useMemo(() => {
```
hawkrives: **`useQueries` has a `combine` option built for exactly this; using it removes the deps hack.**

`dataArrays` + a non-literal `useMemo` dep list + a suppression comment is a workaround for a problem the library already solves. `combine` receives every result and memoizes the derived value:

```ts
let {currentQuery, categoriesBySource} = useQueries({
	queries: NEWS_SOURCES.map((s) => namedNewsOptions(s.id)),
	combine: (results) => ({
		currentQuery: results[Math.max(0, NEWS_SOURCES.findIndex((s) => s.id === selectedSource))],
		categoriesBySource: Object.fromEntries(
			NEWS_SOURCES.map((s, i) => [s.id, extractCategories(results[i]?.data ?? [])]),
		),
	}),
})
```

That drops lines 20–35 entirely, along with the disable directive. Worth noting the directive is broader than it looks: it silences `react(use-memo)` (error severity) as well as `react-hooks(exhaustive-deps)` (warn) — I verified both fire on this line without it. It also uses `eslint-disable-next-line`, where every other suppression in this repo is `oxlint-disable-next-line react/exhaustive-deps`, and the comment above it names `@tanstack/query(no-unstable-deps)`, which is not either of the rules being suppressed.

source/features/news/news-picker.tsx
```diff
@@ -0,0 +1,88 @@
...
+	categoriesBySource: Record<string, string[]>
+	selectedSource: string
+	selectedCategory: string | null
+	onSelect: (source: string, category: string | null) => void
+}
+
+const STAYS_OPEN = [menuActionDismissBehavior('disabled')]
+const LABEL = accessibilityLabel('News filter')
+
+export function NewsPicker({
```
hawkrives, line 32: **This is `CalendarPicker` with sections, copied rather than shared.**

Against `modules/ccc-calendar/calendar-picker.tsx`, these are byte-identical: the `STAYS_OPEN`/`LABEL` constants, the JSDoc shape, the `Stack.Toolbar placement="bottom"` → `Spacer` → `Toolbar.View` → `Host matchContents` → `Menu label={<Image systemName=…/>}` → `Section modifiers={STAYS_OPEN}` → `Toggle` tree, and (until commit `50b05e17d`) the `ACTIVE_STYLE`/`INACTIVE_STYLE` pair. The only real differences are the glyph, the label text, and one vs many sections.

The cost is already visible in this PR: the collapse fix landed on the copy and not on the original, so Calendar still ships the bug. A single component in `modules/` taking `sections: {title, items}[]`, `selected`, and `onSelect` would carry one fix to both.

source/features/news/news-list.tsx
```diff
@@ -1,216 +1,91 @@
...
+						))
+					)}
+				</List>
+			</VStack>
+		</Host>
 	)
 }
 
+/** Extract unique categories from stories, sorted A-Z */
+export function extractCategories(stories: StoryType[]): string[] {
```
hawkrives: **`extractCategories` is a pure function living in a `.tsx` component module that imports `@expo/ui/swift-ui`.**

The convention next door is the opposite: `lib/empty-state.ts` was pulled out of this very file with a comment saying it was *"extracted from the list so the choice is testable on its own,"* and `cleanEntries` / `trimStoryCateogry` — which this function calls — live in `lib/util.ts` with tests in `lib/__tests__/`.

As written, a Jest test for `extractCategories` has to import `news-list.tsx` and load `@expo/ui/swift-ui`, `@frogpond/colors`, and `NewsRow` to reach it. Moving it to `lib/util.ts` beside `cleanEntries` makes it testable the same way its two callees already are, and stops the route file importing a component module for a data helper.

source/features/news/news-picker.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+	return (
+		<Stack.Toolbar placement="bottom">
+			<Stack.Toolbar.Spacer />
+			<Stack.Toolbar.View>
+				<Host matchContents={true}>
+					<Menu label={<Image systemName="newspaper" />} modifiers={menuModifiers}>
+						{sortedSources.map((source) => {
+							let categories = categoriesBySource[source.id] ?? []
+							let sortedCategories = [...categories].sort((a, b) => b.localeCompare(a))
```
hawkrives: **Categories are sorted twice per render, in opposite directions.**

`extractCategories` (news-list.tsx:86) already returns `[...cats].sort()` — A-Z — and this line immediately re-sorts the same array Z-A. The first sort is pure waste, and the two disagree on comparator: `Array.sort()` orders by UTF-16 code unit, `localeCompare` by locale, so the A-Z sort isn't even the inverse of the Z-A one for accented or cased category names.

Sort once. Either drop the `.sort()` in `extractCategories` and keep this one, or have `extractCategories` return the reversed order the Menu wants (what `Calendar.tsx:30` does) and drop this one.

app/(home)/News.tsx
```diff
@@ -1,28 +1,54 @@
...
+		let result: Record<string, string[]> = {}
+		NEWS_SOURCES.forEach((s, i) => {
+			let data = dataArrays[i] ?? []
+			result[s.id] = extractCategories(data)
```
hawkrives: **`cleanEntries` runs twice over the selected source's stories on every data change.**

`extractCategories` calls `cleanEntries(stories)` internally (news-list.tsx:84), and `NewsList` calls `cleanEntries(data)` again on the same array (news-list.tsx:27). For the selected source both run over the identical `StoryType[]`, each doing a `.trim()` per entry plus an `includes('<form')` scan of every entry's full HTML `content` — the expensive half of the pass, over full post bodies.

Hoisting the cleaned list to this component and passing it down (or having `extractCategories` take already-cleaned entries) removes the second pass.

app/(home)/News.tsx
```diff
@@ -1,28 +1,54 @@
...
 
 export default function NewsPage(): React.ReactNode {
-	let dispatch = useDispatch()
-	let selectedId = useSelector(selectNewsSource)
-	let source = NEWS_SOURCES.find((candidate) => candidate.id === selectedId) ?? NEWS_SOURCES[0]
+	let {selectedSource, selectedCategory, select} = useNewsFilterStore()
+	let source = NEWS_SOURCES.find((s) => s.id === selectedSource) ?? NEWS_SOURCES[0]
+
+	// Fetch all sources to populate the picker with categories
+	let queries = useQueries({
```
hawkrives: **Both feeds are fetched on every News open, when only one is rendered.**

The second feed exists solely to label rows in a menu the reader may never open. Each is a manifest resolution plus a full feed download, and `source/init/tanstack-query.ts` persists the query cache to AsyncStorage with a 24h `gcTime`, so both feeds' full post bodies now get written to disk instead of one.

If the picker's categories are worth that, fine — but it is a real doubling of cold-open network and storage on cellular, and it is not mentioned as a tradeoff anywhere. Fetching the non-selected source only once the Menu is opened would get the same picker for half the traffic.

source/features/news/news-picker.tsx
```diff
@@ -0,0 +1,88 @@
...
+	sources: NewsSource[]
+	/** Categories keyed by source id */
+	categoriesBySource: Record<string, string[]>
+	selectedSource: string
+	selectedCategory: string | null
+	onSelect: (source: string, category: string | null) => void
+}
+
+const STAYS_OPEN = [menuActionDismissBehavior('disabled')]
+const LABEL = accessibilityLabel('News filter')
```
hawkrives, line 29: **`uitests/TestIdentifiers.swift` still declares the two accessibility labels this PR deleted.**

```swift
/// The bottom-right toolbar menu's accessibilityLabel, in news-list.tsx.
static let picker = "News Sources"

/// The bottom-left toolbar menu's accessibilityLabel, in news-list.tsx.
static let categoryFilter = "Categories"
```

Both labels are gone (replaced by the single `'News filter'` on this line), and `news-list.tsx` no longer renders a toolbar at all. Nothing references them today, so nothing fails — but the next person writing a News XCUITest reaches for `TestIdentifiers.News.picker` and gets a query that can never match. Update them to `"News filter"` pointing at `news-picker.tsx`, or delete them.

source/features/news/store.ts
```diff
@@ -0,0 +1,24 @@
...
+	persist(
+		(set) => ({
+			selectedSource: 'stolaf',
+			selectedCategory: null,
+			select: (source, category) => set({selectedSource: source, selectedCategory: category}),
+		}),
+		{
+			name: 'news-filter-preferences',
```
hawkrives, line 19: **Readers who picked The Olaf Messenger get silently reset to St. Olaf on upgrade.**

The old choice lives in redux-persist under the `root` key as `settings.selectedNewsSource`; the new one is a separate zustand store under `news-filter-preferences`, defaulting to `'stolaf'`. Nothing reads the old value, so on first launch after the update every Messenger reader lands back on St. Olaf News with no explanation.

Small blast radius, but the deleted test suite existed specifically to pin this field's rehydration behaviour, so the loss looks unintended rather than accepted. A one-time read of the persisted redux value in the store's initializer (or an explicit note that the reset is fine) would settle it.

app/(home)/News.tsx
```diff
@@ -1,28 +1,54 @@
...
+		queries: NEWS_SOURCES.map((s) => namedNewsOptions(s.id)),
```
hawkrives, line 32: **A failure in the non-selected source's fetch is swallowed.**

Only `currentQuery`'s `isError` reaches `NewsList`. If the other source's query rejects, `dataArrays[i]` is `undefined`, `extractCategories([])` returns `[]`, and its Section renders with just "All Stories" — indistinguishable from a source that genuinely has no categories.

Scenario: the Messenger feed is down while reading St. Olaf. The reader opens the picker, sees the Messenger section with no categories, taps "All Stories" under it to switch, and only then gets "A problem occured while loading". The picker had the information a moment earlier and showed nothing.

source/features/news/news-picker.tsx
hawkrives: I think this is resolved in another PR

source/features/news/store.ts
hawkrives: I don't understand the issue here

source/features/news/store.ts
hawkrives: Fixed. `resolveCategory` (source/features/news/lib/util.ts) drops a category the current source no longer offers, and `News.tsx` resolves it before handing it to either the list or the picker — so a category that disappears from the feed stops filtering instead of persisting into an empty list. It comes back on its own if the feed carries that category again. Covered in `lib/__tests__/categories.test.ts`.

app/(home)/News.tsx
hawkrives: Fixed — the picker gets `source.id`, the same resolved value the title bar and the list use.

source/features/news/store.ts
hawkrives: Added, all pure JS: `source/features/news/lib/__tests__/categories.test.ts` (extractCategories, filterByCategory, resolveCategory), `lib/__tests__/combine.test.ts` (the fold: cleaning, per-source keying, failed sources), and `source/features/news/__tests__/store.test.ts` (default, select, clear, switch) mirroring the calendar store suite.

app/(home)/News.tsx
hawkrives: Done. The `useMemo`, the `dataArrays` hack and the suppression are all gone; `useQueries` takes a `combine`. The fold itself is `combineNewsResults` in `lib/combine.ts` so it can be tested without a component, and `News.tsx` passes a module-level wrapper so react-query can memoize it — a combine rebuilt each render is re-run each render.

source/features/news/news-list.tsx
hawkrives: Moved. `extractCategories` now sits in `lib/util.ts` beside `cleanEntries`, along with `filterByCategory` (the category predicate the list used to hold inline). Both are tested in `lib/__tests__/categories.test.ts`; `news-list.tsx` exports only the component again.

source/features/news/news-picker.tsx
hawkrives: Sorted once. `extractCategories` sorts A-Z with `localeCompare`; the picker just reverses that array for the Menu, so the two orders cannot disagree.

app/(home)/News.tsx
hawkrives: Fixed. `combineNewsResults` cleans each source once and both the entries and the categories come out of that pass; `NewsList` takes the cleaned `entries` instead of cleaning again.

app/(home)/News.tsx
hawkrives: Kept, but written down. There is no signal to defer on: `MenuProps` in @expo/ui 57.0.14 is `label`, `systemImage`, `onPrimaryAction`, `children` — no opened callback, and `onPrimaryAction` takes tap-to-open away rather than reporting it. A comment in `News.tsx` now says the second feed is the picker every source list, and what it costs.

source/features/news/news-picker.tsx
hawkrives: Updated — the two are now one `News.picker = "News filter"`, pointing at `news-picker.tsx`.

app/(home)/News.tsx
hawkrives: Fixed. The fold names the sources whose fetch failed, and the picker renders a disabled "Couldn’t load stories" row in that section, so a source that is down no longer reads as a source with no categories. Flagging one limit honestly: I could not exercise the failed-feed path on the simulator — nothing here injects a fetch failure — so the row is verified by the fold test, not by eye.

hawkrives (approved):

PR #7859, opened 2026-09-06
Fix day picker strip and tests for date/times in the calendar

drewvolz:
JS and Native

Both #7858 and #7855 will need these test fixes (and anything new)

commit | about
--|--
ce414428f | deriveDays → whole weeks through the last event's Saturday; component gets a trailing scroll inset + snap-set filter so every week's Sunday reaches the leading edge without empty cells. deriveDays tests rewritten to the new contract.   
2d497ae02 | verifySundayLeadsTheStrip measures the expected week from UITEST_FROZEN_DATE (mirrored as frozenNow), not live Date()

Test fixes
- TestIdentifiers.swift — added frozenNow, mirroring UITEST_FROZEN_DATE.
- CalendarScreen.swift — verifySundayLeadsTheStrip derives the week from frozenNow, not Date(); removed the temp debug string. campusTimeZone stays (the calendar and the dayCell formatter must agree on zone or the ISO string is off by a day).

Swipe/snap-to fixes
- day-picker-strip.tsx — deriveDays now ends on the Saturday of the last event's week (whole weeks); component adds trailingInset so the last Sunday pins to the leading edge without empty day cells, plus a weekStarts reachability guard.
- day-picker-strip.test.tsx — deriveDays tests rewritten to the whole-weeks contract. 103 event-list tests pass, tsc clean.

### DIFF
--- a/modules/event-list/__tests__/day-picker-strip.test.tsx
+++ b/modules/event-list/__tests__/day-picker-strip.test.tsx
@@ -6,124 +6,96 @@ import {describe, expect, jest, test} from '@jest/globals'
 import {DayPickerStrip, deriveDays} from '../day-picker-strip'
 import type {SourcedEvent} from '../types'
 
+// A Sunday, so "this week" runs 2026-08-23 (Sun) through 2026-08-29 (Sat).
 const NOW = moment('2026-08-23T12:00:00Z')
+
+function event(iso: string, isOngoing = false): SourcedEvent {
+	return {
+		sourceId: 'a',
+		key: iso,
+		event: {startTime: moment(iso), isOngoing},
+	} as unknown as SourcedEvent
+}
+
+function isoDays(days: ReturnType<typeof deriveDays>): string[] {
+	return days.map((d) => d.format('YYYY-MM-DD'))
+}
 
 describe('deriveDays', () => {
 	test('returns empty array when events is empty', () => {
-		let result = deriveDays([], NOW)
-		expect(result).toEqual([])
+		expect(deriveDays([], NOW)).toEqual([])
 	})
 
-	test('returns today when only today has events', () => {
-		let events = [
-			{
-				sourceId: 'a',
-				key: 'k',
-				event: {startTime: moment('2026-08-23T14:00:00Z'), isOngoing: false},
-			},
-		]
-		let result = deriveDays(events as unknown as SourcedEvent[], NOW)
-		expect(result).toHaveLength(1)
-		expect(result[0].isSame(NOW, 'day')).toBe(true)
+	test('returns empty array when every event is in the past or ongoing', () => {
+		let events = [event('2026-08-22T10:00:00Z'), event('2026-09-15T10:00:00Z', true)]
+		expect(deriveDays(events, NOW)).toEqual([])
 	})
 
-	test('returns days in order starting from today', () => {
-		let events = [
-			{
-				sourceId: 'a',
-				key: '1',
-				event: {startTime: moment('2026-08-25T10:00:00Z'), isOngoing: false},
-			},
-			{
-				sourceId: 'a',
-				key: '2',
-				event: {startTime: moment('2026-08-23T10:00:00Z'), isOngoing: false},
-			},
-			{
-				sourceId: 'a',
-				key: '3',
-				event: {startTime: moment('2026-08-24T10:00:00Z'), isOngoing: false},
-			},
-		]
-		let result = deriveDays(events as unknown as SourcedEvent[], NOW)
-		expect(result).toHaveLength(3)
-		expect(result[0].format('YYYY-MM-DD')).toBe('2026-08-23')
-		expect(result[1].format('YYYY-MM-DD')).toBe('2026-08-24')
-		expect(result[2].format('YYYY-MM-DD')).toBe('2026-08-25')
+	test('spans whole weeks, Sunday through Saturday', () => {
+		let result = deriveDays([event('2026-08-25T10:00:00Z')], NOW)
+		expect(result).toHaveLength(7)
+		expect(result[0].day()).toBe(0)
+		expect(result[6].day()).toBe(6)
+		expect(isoDays(result)[0]).toBe('2026-08-23')
+		expect(isoDays(result).at(-1)).toBe('2026-08-29')
 	})
 
-	test('excludes days before today', () => {
-		let events = [
-			{
-				sourceId: 'a',
-				key: '1',
-				event: {startTime: moment('2026-08-22T10:00:00Z'), isOngoing: false},
-			},
-			{
-				sourceId: 'a',
-				key: '2',
-				event: {startTime: moment('2026-08-23T10:00:00Z'), isOngoing: false},
-			},
-		]
-		let result = deriveDays(events as unknown as SourcedEvent[], NOW)
-		expect(result).toHaveLength(1)
-		expect(result[0].format('YYYY-MM-DD')).toBe('2026-08-23')
+	test('leads with Sunday of the current week regardless of the first event', () => {
+		let result = deriveDays([event('2026-08-27T10:00:00Z')], NOW)
+		expect(isoDays(result)[0]).toBe('2026-08-23')
 	})
 
-	test('excludes ongoing events from day derivation', () => {
-		let events = [
-			{
-				sourceId: 'a',
-				key: '1',
-				event: {startTime: moment('2026-08-20T10:00:00Z'), isOngoing: true},
-			},
-			{
-				sourceId: 'a',
-				key: '2',
-				event: {startTime: moment('2026-08-23T10:00:00Z'), isOngoing: false},
-			},
-		]
-		let result = deriveDays(events as unknown as SourcedEvent[], NOW)
-		expect(result).toHaveLength(1)
-		expect(result[0].format('YYYY-MM-DD')).toBe('2026-08-23')
+	test("extends through the Saturday of the last event's week", () => {
+		// Last event is Wed 2026-09-02, so the strip runs to Sat 2026-09-05.
+		let result = deriveDays([event('2026-08-23T10:00:00Z'), event('2026-09-02T10:00:00Z')], NOW)
+		expect(isoDays(result).at(-1)).toBe('2026-09-05')
+		expect(result).toHaveLength(14)
 	})
 
-	test('deduplicates days with multiple events', () => {
-		let events = [
-			{
-				sourceId: 'a',
-				key: '1',
-				event: {startTime: moment('2026-08-23T09:00:00Z'), isOngoing: false},
-			},
-			{
-				sourceId: 'a',
-				key: '2',
-				event: {startTime: moment('2026-08-23T14:00:00Z'), isOngoing: false},
-			},
-		]
-		let result = deriveDays(events as unknown as SourcedEvent[], NOW)
-		expect(result).toHaveLength(1)
+	test('returns a continuous run of unique ascending days', () => {
+		let result = deriveDays(
+			[event('2026-09-02T10:00:00Z'), event('2026-08-23T10:00:00Z'), event('2026-08-23T18:00:00Z')],
+			NOW,
+		)
+		let iso = isoDays(result)
+		expect(new Set(iso).size).toBe(iso.length)
+		for (let i = 1; i < result.length; i++) {
+			expect(result[i].diff(result[i - 1], 'days')).toBe(1)
+		}
 	})
 
-	test('fills in gap days between events', () => {
+	test('excludes past and ongoing events from the range end', () => {
 		let events = [
-			{
-				sourceId: 'a',
-				key: '1',
-				event: {startTime: moment('2026-08-23T10:00:00Z'), isOngoing: false},
-			},
-			{
-				sourceId: 'a',
-				key: '2',
-				event: {startTime: moment('2026-08-26T10:00:00Z'), isOngoing: false},
-			},
+			event('2026-08-22T10:00:00Z'),
+			event('2026-09-20T10:00:00Z', true),
+			event('2026-08-24T10:00:00Z'),
 		]
-		let result = deriveDays(events as unknown as SourcedEvent[], NOW)
-		expect(result).toHaveLength(4)
-		expect(result[0].format('YYYY-MM-DD')).toBe('2026-08-23')
-		expect(result[1].format('YYYY-MM-DD')).toBe('2026-08-24')
-		expect(result[2].format('YYYY-MM-DD')).toBe('2026-08-25')
-		expect(result[3].format('YYYY-MM-DD')).toBe('2026-08-26')
+		let result = deriveDays(events, NOW)
+		// Only the 2026-08-24 event counts, so the strip is just this week.
+		expect(isoDays(result)).toEqual([
+			'2026-08-23',
+			'2026-08-24',
+			'2026-08-25',
+			'2026-08-26',
+			'2026-08-27',
+			'2026-08-28',
+			'2026-08-29',
+		])
+	})
+
+	test('fills gap days between events', () => {
+		let result = deriveDays([event('2026-08-24T10:00:00Z'), event('2026-08-27T10:00:00Z')], NOW)
+		expect(isoDays(result)).toContain('2026-08-25')
+		expect(isoDays(result)).toContain('2026-08-26')
+	})
+
+	test('measures the week in the zone of the moment it is given', () => {
+		// 02:00 UTC Sunday is still 21:00 Saturday in this zone, so the week --
+		// and its leading Sunday -- is the earlier one. deriveDays must not
+		// silently reinterpret `now` in UTC.
+		let now = moment.tz('2026-09-06T02:00:00Z', 'America/Chicago')
+		let result = deriveDays([event('2026-09-10T10:00:00Z')], now)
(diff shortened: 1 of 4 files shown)


modules/event-list/day-picker-strip.tsx
```diff
@@ -59,14 +61,13 @@ export function deriveDays(events: readonly SourcedEvent[], now: Moment): Moment
...
-	// Start from Sunday of the current week so the strip always opens on a
-	// week boundary.
 	let sunday = today.clone().startOf('week')
+	let rangeEnd = lastDay.clone().endOf('week')
 
 	let days: Moment[] = []
 	let current = sunday.clone()
 
-	while (current.isSameOrBefore(lastDay, 'day')) {
+	while (current.isSameOrBefore(rangeEnd, 'day')) {
```
hawkrives, line 70: **Trailing empty days break the invariant `dayWithEventsFrom` depends on.**

`event-list.tsx` documents it explicitly (line 127):

> The range ends on a day that has events, so looking forward always lands somewhere.

That is no longer true. The range now ends on the Saturday of the last event's week, so unless the last event happens to fall on a Saturday there are 1-6 trailing days with no section behind them, and `days.find(...) ?? day` falls through to the `?? day` branch:

```ts
let dayWithEventsFrom = React.useCallback(
	(day: Moment) =>
		days.find((d) => !d.isBefore(day, 'day') && sectionKeys.has(sectionKeyFor(d))) ?? day,
```

Concretely: last event Wed 2026-09-16, user taps Fri 2026-09-18 in the strip.

1. `target` = Fri 2026-09-18 (no section).
2. `scrollTarget.value = '2026-09-18'` — SwiftUI `scrollPosition` gets an id no view carries, so the list does not move.
3. `setChosenDay(target)` makes `chosenDayValid` false, so `selectedDay` reverts to `defaultDay` — the tapped cell never shows as selected.
4. If a valid day had been chosen before, `prevChosenDayValid.current` was true, so the effect at `event-list.tsx:109` fires `stripRef.current?.scrollToDay(defaultDay)` and yanks the strip back to the top week, fighting the `scrollToDay(target)` that `handleSelectDay` just issued.

Also note this is arguably the wrong layer to fix the snapping at: the day *range* (a data concern) is being reshaped to satisfy a scroll-snap constraint (a view concern), which is exactly why the data invariant broke. Either make `dayWithEventsFrom` fall back to the last day that does have events, or keep `deriveDays` ending on the last event day and let the strip treat a partial final week as its own snap stop.

modules/event-list/day-picker-strip.tsx
```diff
@@ -59,14 +61,13 @@ export function deriveDays(events: readonly SourcedEvent[], now: Moment): Moment
 		return []
 	}
 
-	// Start from Sunday of the current week so the strip always opens on a
-	// week boundary.
 	let sunday = today.clone().startOf('week')
+	let rangeEnd = lastDay.clone().endOf('week')
```
hawkrives, line 65: **`rangeEnd` is measured in the event's zone while the loop walks in `now`'s zone, so `days.length` is not always a multiple of 7.**

The two moments genuinely live in different zones today:

- `now` comes from `useMomentTimer({..., timezone: timezone()})` → `America/Chicago`.
- `entry.event.startTime` comes from `moment(event.startTime)` in `modules/ccc-calendar/query.ts:26` — plain `moment`, so **device-local**.

`lastDay.clone().endOf('week')` is therefore Saturday 23:59:59.999 *in the device's zone*, while `current.isSameOrBefore(rangeEnd, 'day')` evaluates `startOf`/`endOf` in `current`'s zone (`now`'s). When the device is west of campus the loop runs one iteration too far.

Concrete: device in `America/Los_Angeles`, last event `2026-09-19T15:00:00-05:00`.
- `lastDay` = Sat 2026-09-19 PDT → `rangeEnd` = 2026-09-19T23:59:59.999-07:00 = 2026-09-20T01:59:59 CDT.
- `current` = Sun 2026-09-20 00:00 CDT: `startOf('day') <= rangeEnd <= endOf('day')` → `isSame` is true → pushed.
- Result: 22 days, not 21.

That breaks the whole-weeks contract the rest of this PR is built on — `trailingInset`'s `7 * CELL_TOTAL_WIDTH` assumption and the `weekStarts` guard both take it as given. Symmetrically, a device east of campus can push `lastDay` onto the next day and add a whole extra empty week.

Deriving the end from `sunday` instead of from `lastDay` keeps it exact regardless of zone:

```ts
let sunday = today.clone().startOf('week')
let weeks = Math.ceil((lastDay.diff(sunday, 'days') + 1) / 7)
let rangeEnd = sunday.clone().add(weeks * 7 - 1, 'days')
```

modules/event-list/__tests__/day-picker-strip.test.tsx
```diff
@@ -6,124 +6,96 @@ import {describe, expect, jest, test} from '@jest/globals'
...
+	})
+
+	test('measures the week in the zone of the moment it is given', () => {
+		// 02:00 UTC Sunday is still 21:00 Saturday in this zone, so the week --
+		// and its leading Sunday -- is the earlier one. deriveDays must not
+		// silently reinterpret `now` in UTC.
+		let now = moment.tz('2026-09-06T02:00:00Z', 'America/Chicago')
+		let result = deriveDays([event('2026-09-10T10:00:00Z')], now)
+		expect(result[0].format('YYYY-MM-DD')).toBe('2026-08-30')
 	})
```
hawkrives, line 99: **Nothing verifies the behaviour this PR exists to fix.**

The stated goal is "every week's Sunday reaches the leading edge without empty cells" — a scroll-offset/snap outcome. Per `CLAUDE.md`:

> Appearance and interaction — tint, spacing, alignment, truncation, hit targets, menu and sheet presentation, safe-area behaviour — belong in an XCUITest under `uitests/`

But the Jest suite here only covers `deriveDays` (pure data), and `uitests/ModuleCalendarTests.swift` has no test that swipes the strip: `testDayPickerStripLeadsWithSunday` only reads the leftmost cell at launch, `testDayCellsMeetTheMinimumTapTarget` reads frames, and `testScrollingTheListMovesTheStripSelection` drives the *list*, not the strip. `trailingInset`, the `weekStarts` reachability guard, and the snap offsets ship with no assertion behind them.

Worth a `swipeStripLeft()` on `CalendarScreen` plus an assertion that `dayCells().first` is again a Sunday (`2026-09-06` / `2026-09-13` against the frozen clock) after the strip settles. The fixture's last event is Sat 2026-09-19, which also means the trailing-empty-day case flagged above can never appear in a UI-test run — worth moving one fixture event so it can.

modules/event-list/day-picker-strip.tsx
```diff
@@ -150,9 +151,14 @@ export let DayPickerStrip = React.forwardRef<DayPickerStripHandle, Props>(functi
 		setContainerWidth(event.nativeEvent.layout.width)
 	}, [])
 
+	// Trailing room so the last week's Sunday can still pull to the leading
+	// edge -- scroll inset, not day cells, so there is no empty week to swipe
+	// into. A full week already fills a phone; wider screens need the rest.
+	let trailingInset = Math.max(0, containerWidth - 7 * CELL_TOTAL_WIDTH)
```
hawkrives, line 157: **`trailingInset` overshoots by 12pt because it ignores the content padding.**

Solve for the inset that makes `maxScroll` land exactly on the last Sunday's offset. With `L = days.length - 7`:

```
offsetForIndex(L) = PADDING_HORIZONTAL + L*52 - CELL_MARGIN = L*52 + 4
maxScroll         = PADDING_HORIZONTAL*2 + T + N*52 - cw = N*52 + 16 + T - cw
```

Setting them equal gives `T = cw - 376`, i.e. `cw - (7*CELL_TOTAL_WIDTH + PADDING_HORIZONTAL*2 - CELL_MARGIN)`. The code uses `cw - 364`, so `maxScroll` sits 12pt beyond the final snap offset on every screen width.

Two visible effects:

- At the end of the strip you can always drag 12pt into blank space, and it rubber-bands back.
- A one-week strip that previously could not scroll at all now can. On a 393pt screen with 7 days: `T = 29`, content `= 8 + 364 + 8 + 29 = 409 > 393`, so `maxScroll = 16` while the only snap offset is `4`. The strip renders at offset 0 and jumps 4pt on the first touch.

```suggestion
	let trailingInset = Math.max(
		0,
		containerWidth - (7 * CELL_TOTAL_WIDTH + PADDING_HORIZONTAL * 2 - CELL_MARGIN),
	)
```

modules/event-list/day-picker-strip.tsx
```diff
@@ -165,15 +171,21 @@ export let DayPickerStrip = React.forwardRef<DayPickerStripHandle, Props>(functi
...
+	 * cell always stays, even on a strip too short to scroll.
 	 */
 	let weekStarts = React.useMemo(() => {
 		return days
 			.map((day, index) => ({day, index}))
-			.filter(({day, index}) => index === 0 || day.day() === 0)
+			.filter(({day, index}) => {
+				if (index === 0) return true
+				if (day.day() !== 0) return false
+				return PADDING_HORIZONTAL + index * CELL_TOTAL_WIDTH - CELL_MARGIN <= maxScroll
```
hawkrives, line 185: **Duplicated offset arithmetic.**

`PADDING_HORIZONTAL + index * CELL_TOTAL_WIDTH - CELL_MARGIN` is copied verbatim out of `offsetForIndex` (line 166). The two are load-bearing on each other — the filter's whole job is to decide whether `offsetForIndex` would clamp — so if either changes the guard stops meaning what it says, silently.

Extract the raw form once and have both use it:

```ts
let rawOffsetForIndex = React.useCallback(
	(index: number) => PADDING_HORIZONTAL + index * CELL_TOTAL_WIDTH - CELL_MARGIN,
	[],
)

let offsetForIndex = React.useCallback(
	(index: number) => Math.max(0, Math.min(rawOffsetForIndex(index), maxScroll)),
	[rawOffsetForIndex, maxScroll],
)
```

then the filter reads `return rawOffsetForIndex(index) <= maxScroll`.

uitests/Screens/CalendarScreen.swift
```diff
@@ -240,13 +240,15 @@ struct CalendarScreen: Screen {
 
 		var calendar = Calendar(identifier: .gregorian)
 		calendar.locale = Locale(identifier: "en_US_POSIX")
+		calendar.timeZone = TestIdentifiers.Calendar.campusTimeZone
```
hawkrives, line 243: **The expected Sunday rides on an implicit `firstWeekday`.**

Round-tripping `[.yearForWeekOfYear, .weekOfYear]` only yields a Sunday if `calendar.firstWeekday == 1` and `minimumDaysInFirstWeek == 1`. Neither is set here; both are inherited from whatever `Locale(identifier: "en_US_POSIX")` resolves to on the simulator, and `Calendar(identifier:)` on iOS 16+ also picks up the device's *First Day of Week* setting through `Locale.current` before that assignment. If `firstWeekday` ever came out as 2, week 36 of 2026 would round-trip to Sun 2026-09-06 rather than Sun 2026-08-30, and the test would fail against a perfectly correct strip.

The app side is not locale-derived at all — `deriveDays` runs under moment's default `en` locale and `weekStarts` hardcodes `day.day() !== 0` — so the test should hardcode to match rather than infer:

```suggestion
		calendar.timeZone = TestIdentifiers.Calendar.campusTimeZone
		// The app's week starts on Sunday unconditionally (moment's default `en`
		// locale, and `day() === 0` in the strip). Pin it here rather than
		// inheriting it from the simulator's region settings.
		calendar.firstWeekday = 1
		calendar.minimumDaysInFirstWeek = 1
```

modules/event-list/day-picker-strip.tsx
```diff
@@ -165,15 +171,21 @@ export let DayPickerStrip = React.forwardRef<DayPickerStripHandle, Props>(functi
...
+	 * dropped -- its days still render, they are just not a snap stop. The first
+	 * cell always stays, even on a strip too short to scroll.
 	 */
 	let weekStarts = React.useMemo(() => {
 		return days
 			.map((day, index) => ({day, index}))
-			.filter(({day, index}) => index === 0 || day.day() === 0)
+			.filter(({day, index}) => {
+				if (index === 0) return true
+				if (day.day() !== 0) return false
```
hawkrives, line 184: **Two different definitions of "week start" in one component.**

`deriveDays` uses `startOf('week')` / `endOf('week')`, which are locale-driven (`moment.localeData().firstDayOfWeek()`), and `scrollToDay` does the same on line 196. This filter instead hardcodes `day.day() !== 0`. They agree only because nothing in the app ever calls `moment.locale(...)`, so the default `en` locale (dow 0) is in force.

The moment anything sets a locale whose week starts on Monday — a future i18n pass, or a transitive dependency importing `moment/locale/*` — `deriveDays` emits Mon→Sun weeks while this filter keeps every *Sunday*, which now sits at index 6, 13, 20. Every snap offset lands six cells into a week, and `scrollToDay`'s `days.findIndex` for the Sunday of a week returns a day that is no longer the strip's leading edge. It would look like a swipe bug, not a locale bug.

Worth making the week start explicit in one place — a `WEEK_START_DAY` constant used by both, or `startOf('isoWeek')`-style handling — so the two cannot drift.

modules/event-list/__tests__/day-picker-strip.test.tsx
```diff
@@ -6,124 +6,96 @@ import {describe, expect, jest, test} from '@jest/globals'
...
-				key: '2',
-				event: {startTime: moment('2026-08-23T10:00:00Z'), isOngoing: false},
-			},
-		]
-		let result = deriveDays(events as unknown as SourcedEvent[], NOW)
-		expect(result).toHaveLength(1)
-		expect(result[0].format('YYYY-MM-DD')).toBe('2026-08-23')
+	test('leads with Sunday of the current week regardless of the first event', () => {
+		let result = deriveDays([event('2026-08-27T10:00:00Z')], NOW)
+		expect(isoDays(result)[0]).toBe('2026-08-23')
```
hawkrives, line 45: **This test cannot fail for the reason its name gives.**

`NOW` is itself Sunday 2026-08-23, so `'2026-08-23'` is simultaneously "Sunday of the current week" and "today". An implementation that started the range at `today` — the exact behaviour the rename to whole weeks is guarding against — passes this assertion unchanged.

The only place the leading-Sunday contract is genuinely exercised is the timezone test at the bottom, incidentally. Give this one a mid-week `now`:

```ts
test('leads with Sunday of the current week regardless of the first event', () => {
	let wednesday = moment('2026-08-26T12:00:00Z')
	let result = deriveDays([event('2026-08-27T10:00:00Z')], wednesday)
	expect(isoDays(result)[0]).toBe('2026-08-23')
})
```

That also covers the days-before-today case the old `excludes days before today` test claimed to cover (and, being anchored to a Sunday `NOW`, never did either).

uitests/TestIdentifiers.swift
```diff
@@ -165,11 +165,26 @@ struct TestIdentifiers {
...
+		/// campus dates; a test reasoning about "today" or a week boundary works
+		/// in the same zone.
+		static let campusTimeZone = TimeZone(identifier: "America/Chicago")!
+
+		/// The instant the app freezes its clock to under `isUITesting`, so a
+		/// test reasons about "today" the way the app does rather than off the
+		/// live wall clock. Mirrors `UITEST_FROZEN_DATE` in
+		/// `modules/timer/index.ts` (`2026-09-05T12:00:00-05:00`, a Saturday).
+		static let frozenNow = Foundation.Calendar(identifier: .gregorian).date(
+			from: DateComponents(timeZone: campusTimeZone, year: 2026, month: 9, day: 5, hour: 12)
```
hawkrives, line 179: **The mirror is reconstructed rather than parsed, so it can drift by an hour without anyone noticing.**

`UITEST_FROZEN_DATE` is `2026-09-05T12:00:00-05:00` — a fixed offset. This rebuilds it as "12:00 in `America/Chicago`", which only agrees while the date sits in CDT. Move the frozen date into November and the JS constant is `-05:00` (as written) while Swift resolves `America/Chicago` to `-06:00`; the two instants differ by an hour, and on a date near midnight that is a different day and a silently wrong `dayCell` expectation.

Parsing the same string the app parses removes the second source of truth:

```suggestion
		static let frozenNow = ISO8601DateFormatter().date(from: "2026-09-05T12:00:00-05:00")!
```

modules/event-list/day-picker-strip.tsx
```diff
@@ -247,7 +259,10 @@ export let DayPickerStrip = React.forwardRef<DayPickerStripHandle, Props>(functi
 	return (
 		<View style={styles.container} onLayout={handleLayout}>
 			<ScrollView
-				contentContainerStyle={styles.scrollContent}
+				contentContainerStyle={[
+					styles.scrollContent,
+					{paddingRight: PADDING_HORIZONTAL + trailingInset},
```
hawkrives, line 264: The comment above calls this the trailing inset, but `paddingRight` is a physical edge. Under an RTL layout direction the strip's trailing edge is the left one, so the inset would land on the side the strip starts from and the last week's Sunday still could not reach the leading edge — the exact bug this change is fixing, mirrored.

`paddingEnd` is direction-aware and reads the same as the comment. Note it also needs to beat `styles.scrollContent`'s `paddingHorizontal`, which it does by being later in the array, same as now.

```suggestion
					{paddingEnd: PADDING_HORIZONTAL + trailingInset},
```

PR #7861, opened 2026-09-06
Keep the event sheet's toolbar items on every presentation

### LINKED ISSUE #7827: Add to calendar CTA text can show up as blank
I tried hardcoding most things (tint, text, etc) and I might point at the modal presentation? I do not know.

hawkrives:
Fixes #7827.

## What was wrong

Reopening an event sheet left the **Add to Calendar** button drawn but with no
text. First open looked fine; every open after it was blank.

The cause is in expo-router 57.0.17 (unchanged in 57.0.19), not in our code —
which is why hardcoding the tint and the text in `EventDetail` changed nothing.

react-native-screens reuses **one navigation controller for every presentation
of a modal**. `RouterToolbarHostView` cleared its bar items on the way out but
left the toolbar *showing*, so the next screen presented into that controller
found it already unhidden. Its own `setToolbarHidden(false, animated:)` became a
no-op, and its bar items never reached the bar.

The logs show every presentation handing UIKit an identical, correct item — the
only thing that differs is `isToolbarHidden` on entry:

| | round 1 (renders) | round 2 (blank) |
| --- | --- | --- |
| screen VC | `0x148d11400` | `0x148d13200` (new) |
| nav controller | `0x10a9de400` | `0x10a9de400` (**same**) |
| `UIToolbar` | `0x14a1a6580` | `0x14a1a6580` (**same**) |
| `isToolbarHidden` on entry | `yes` | **`no`** |
| items handed over | `[nil, "Add to Calendar", nil]` | `[nil, "Add to Calendar", nil]` |

Hiding the toolbar on teardown restores the state the next presentation expects.
The `viewControllers.count <= 1` guard limits that to a modal's own controller —
a pushed screen shares its stack with the screen underneath, whose toolbar has
to stay up.

## Also here

`AddToCalendar.addEvent` measured its half-second floor from before the message
was set to right after, which is nothing, so it always waited the full 500ms and
*then* started saving. Every save cost half a second it did not need. Measuring
around the save keeps what the floor was for while letting a slow save satisfy
it.

That component could not be loaded under Jest at all — `delay` and its
transitive deps ship ESM-only and were not in the transform allowlist, which is
why `lib.ts` had tests and the component had none. `jest.config.js` unblocks it.

## Verifying

`testAddToCalendarSurvivesReopeningTheSheet` opens the sheet, dismisses it, and
opens it again. Confirmed red without the patch and green with it, on builds
proven fresh by a runtime `NSLog` marker; all 11 `ModuleCalendarTests` pass.
Jest: 121 suites, 1123 tests.

Jest cannot see this one — it has no layout pass and no compositor, so the blank
button is only visible to an XCUITest against the simulator.

| before | after |
| --- | --- |
| button drawn, no title | "Add to Calendar" |

## Worth knowing before merging

- **The patch needs `mise run prebuild`.** `pnpm patch` gives the package a new
  `.pnpm/…_patch_hash=…` directory, and `ios/Pods` keeps pointing at the
  pre-patch path until prebuild reruns. Until then Xcode silently compiles the
  unpatched copy. This cost me three misleading test runs, so there is a
  `CLAUDE.md` note about it.
- **Regenerating the patch reintroduces a bad hunk.** `pnpm patch-commit`
  emitted a spurious deletion of `ExpoRouterModule.kt`; I stripped it by hand.
- **The UIKit mechanism is inferred, not observed.**
  `navigationController.toolbar.items` is empty even when the CTA renders
  correctly, so on iOS 26 the bar actually drawn is private and I could not
  inspect it. What is proven is the causal link: leaving the toolbar unhidden
  breaks the next presentation, re-hiding it fixes it, deterministically.
- Worth reporting upstream so the patch can eventually be dropped.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -161,6 +161,17 @@ The feature uses `react-native-zeroconf` (native pod). If the pod hasn't been li
 
 **Before committing:** Always run `mise run agent:pre-commit` before committing any changes. This formats code with oxfmt, runs oxlint, checks TypeScript types, and runs Jest tests. Do not commit if any step fails.
 
+**Patching a native dependency:** `pnpm patch` gives the patched package a new
+`node_modules/.pnpm/<name>@<version>_patch_hash=…` directory, but `ios/Pods`
+still points at the pre-patch path until you run `mise run prebuild`. Until then
+Xcode compiles the unpatched copy and edits under `node_modules/<name>` are
+silently ignored — builds succeed, tests pass, and nothing says why. After
+adding or changing a patch that touches native code, run `mise run prebuild` and
+confirm `ios/Pods/Pods.xcodeproj/project.pbxproj` names the `patch_hash` path.
+To prove a build really picked a change up, put an `NSLog` in the patched source
+and watch for it with `xcrun simctl spawn <UDID> log stream`; grepping the
+source file only tells you what is on disk, not what got compiled.
+
 **Dependency upgrades:** Whenever you upgrade a dependency whose version is mentioned in this file (e.g., React Native, React Navigation, React Query, Redux Toolkit, TypeScript, Jest, Fastlane), update the version reference in CLAUDE.md as part of the same change. Stale version references in this file mislead future sessions about the project's current state.
 
 ## Superpowers Skills Framework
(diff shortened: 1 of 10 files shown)


hawkrives:
Replaced by https://github.com/StoDevX/AAO-React-Native/issues/7863

PR #7872, opened 2026-09-07
Rebuild PR #7866 from a cold ccache

hawkrives:
Test branch for #7866. Same tree as the PR head (a4b9e07) with `cache_epoch` bumped, so every cache key misses and the iOS app is compiled from scratch.

If the calendar and Carleton map UITests pass here, the failures on #7866 came from ccache serving objects compiled against the older expo headers, not from the dependency bump itself.

Not for merge.


hawkrives:
Cold cache made no difference — the same four calendar tests fail identically, so ccache was not the cause. Bisecting the dependencies instead in #7873, #7874 and #7875.

Issue #7921, opened 2026-09-10
Move the remaining screens off react-native-tableview-simple, FlatList, and SectionList


hawkrives:
`@expo/ui/swift-ui` is already the house style — ~40 files use it, and the `Host` + `List` + `Section` pattern is settled (see `source/features/building-hours/list/building-list.tsx`). 29 files still use `react-native-tableview-simple`, `SectionList`, or `FlatList`.

**Constraint:** `@expo/ui`'s `List` is `List { Children() }` ([ListView.swift](https://github.com/expo/expo/blob/main/packages/expo-ui/ios/ListView.swift)) — every child mounts as a native SwiftUI view up front, with no windowing. Measured on device, 1,000 rows mount without noticeable delay.

**`app/(home)/CourseSearchResults.tsx` stays on `SectionList`.** Its result set is the whole course catalogue, uncapped, and it is the only screen that can exceed that ceiling. It keeps `@frogpond/lists` and `largeListProps` alive for exactly one caller. Everything else migrates.

Logic worth testing moves into a `lib/` file as it goes — `@frogpond/lists` cannot be imported under Jest. Appearance and hit targets belong in `uitests/`.

### Stage 1 — small static lists
Sets the pattern; do these first. **Done** — see the commits on `expo-ui-stage-1-static-lists`.

The five disclosure-style screens share one `DisclosureRow` in `source/components/rows.tsx`; `@expo/ui` has no `NavigationLink`, so the chevron and its `contentShape` are hand-drawn and worth writing once. The athletics filter tab was not a disclosure list at all — it is now a `List(selection:)` under an active edit mode, matching `modules/filter/filter-sheet.tsx`.
- [x] `app/(home)/Transportation/other-modes.tsx`
- [x] `app/(home)/More/index.tsx`
- [x] `app/(home)/SIS/student-work.tsx`
- [x] `source/features/athletics/filters.tsx`
- [x] `app/(settings)/APITest.tsx`
- [x] `app/(home)/PrintJobs/index.tsx`

### Stage 2 — medium section lists
Grouping and filtering logic moves to `lib/`.
- [x] `app/(home)/Athletics.tsx`
- [x] `app/(home)/Streaming Media/index.tsx`
- [x] `app/(home)/PrintJobs/[jobId]/printers.tsx`
- [x] `app/(home)/StudentOrgs/index.tsx` (~250 rows, inside the ceiling)

### Stage 3 — retire `@frogpond/tableview`
[...]

PR #7925, opened 2026-09-10
Move Student Orgs and the printer list to @expo/ui

hawkrives:
Stage 2 (part one) of #7921. **Stacked on #7924** — review that first; this branch contains its commits.

### Commits

- `84de8fbb7` Let a disclosure row carry several detail lines and a leading image
- `1f4eea6c2` Move Student Orgs and the printer list to @expo/ui
- `67065c7e5` Finish mocking stoPrint, and reach the printer list with it

### Screens on the simulator

| Student Orgs | Select Printer |
| --- | --- |
| Student Orgs | Printers |

The printer rows with no second line are printers whose `location` is blank in the fixtures — the row drops an empty detail rather than drawing a gap.

### Three stoPrint bugs, found by trying to photograph that second screen

- **`fetchColorPrinters` and `cancelPrintJobForUser` had no mock guard**, so a mocked run still called the live server.
- **The credentials gate hung** the printer screen on its spinner. A mocked run prints as the stand-in account whatever the keychain holds, so it no longer waits on that read. The real path still does.
- **The stand-in account was applied inside a query's `select`**, which React Query skips when a query has no data — exactly the empty-keychain case it existed for. This one is on #7924 and is user-facing: it showed "Could not find this print job."

A failing UI test now attaches a screenshot of wherever it stopped (`UITestCase.tearDownWithError`). All three were invisible in the logs.

### Not fixed here

`GET https://stolaf.api.frogpond.tech/v1/color-printers` answers **404**. That is a server response rather than a network failure, so a colour print job on the real API likely hits the same "Connection Issue" screen. Unrelated to this migration; worth an issue against `ccc-server`.

The new detail-array and thumbnail props have no consumer yet — Streaming Media and Athletics are the next PR. The thumbnail path is Jest-tested but has not been seen on a device.

### DIFF
--- a/app/(home)/PrintJobs/[jobId]/printers.tsx
+++ b/app/(home)/PrintJobs/[jobId]/printers.tsx
@@ -34,10 +34,8 @@ type PrinterListViewProps = {
 function PrinterListView({job}: PrinterListViewProps): React.ReactNode {
 	let router = useRouter()
 
-	let {data: username = ''} = useQuery({
-		...credentialsOptions,
-		select: (data) => stoprintUsername(data, isStoprintMocked),
-	})
+	let {data: credentials} = useQuery(credentialsOptions)
+	let username = stoprintUsername(credentials, isStoprintMocked)
 
 	let {
 		data: allPrinters = [],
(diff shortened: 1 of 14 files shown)


drewvolz:
> https://stolaf.api.frogpond.tech/v1/color-printers answers 404

What happened?

hawkrives:
> > https://stolaf.api.frogpond.tech/v1/color-printers answers 404
> 
> What happened?

The data moved to a /printers namespace at some point, inside ccc. That endpoint has likely been broken for years… 

There's another PR up to use the correct endpoint.

drewvolz (approved):

PR #7929, opened 2026-09-10
Move Streaming Media and the athletics scores to @expo/ui

hawkrives:
Stage 2 (part two) of #7921, and the last of the stage-2 screens. **Stacked on #7925**, which is stacked on #7924.

### Commits

- `add6d5617` Move Streaming Media to @expo/ui
- `f67cb8d30` Move the athletics scores to @expo/ui
- `b4e132fd7` Give the athletics tabs a week of fixtures to show
- `338144afc` Read the athletics clock the way the fixtures are written

### Screens on the simulator

| Streaming Media | Athletics — Today |
| --- | --- |
| Streaming Media | Athletics |

Streaming Media is the first screen to draw a **remote** thumbnail through the shared row — those posters are real network images hosted inside a SwiftUI list, which is what the `RNHostView` path in #7925 was added for and could not be proven in Jest.

The athletics crests are a grey square on purpose: the fixtures inline a 1×1 PNG, and something visible is what distinguishes a hosted image from one that failed to load.

### Two bugs found by looking

- **Athletics measured "today" from `new Date()`**, so under UI testing it ignored the frozen clock and its Today tab followed the wall clock. It reads `now()` now, as the calendar does.
- The screen previously drew a **card inside a card** — its own rounded, shadowed row inside what is now an `insetGrouped` list.

### Athletics fixtures

Live scores are whatever St. Olaf played that day, so a UI test could open Today and find nothing — an empty tab proves nothing about the screen. Under `isUITesting` the scores now come from a fixed week filling every bucket the tabs show, the way the dictionary and building hours already work. A test asserts that pairing, since fixtures drifting out of their buckets would quietly empty the tabs again.

### Known, and not introduced here

- A row with no thumbnail does not align with rows that have one — visible on the two Streaming Media entries without posters. The old `ListRow` behaved the same way; it just reads more obviously in a grouped list.
- The `@expo/ui` `List`-under-`NativeTabs` bottom inset, as described on #7924.

Streaming Media still uses Moment: its `StreamType.date` is built as one by the query, so moving to `date-fns` means changing that parsing and belongs in its own change.

### DIFF
--- a/app/(home)/Athletics.tsx
+++ b/app/(home)/Athletics.tsx
@@ -1,9 +1,10 @@
 import * as React from 'react'
-import {SectionList, StyleSheet, View} from 'react-native'
+import {StyleSheet, View} from 'react-native'
+import {Host, List, Section} from '@expo/ui/swift-ui'
+import {listStyle, refreshable} from '@expo/ui/swift-ui/modifiers'
 import {Stack} from 'expo-router'
 import {useQuery} from '@tanstack/react-query'
 
-import {ListSectionHeader} from '@frogpond/lists'
 import {LoadingView, NoticeView} from '@frogpond/notice'
 import * as c from '@frogpond/colors'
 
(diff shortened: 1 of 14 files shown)


hawkrives:
@drewvolz note that the grey logo boxes are only for UITest mode; I have a screenshot somewhere showing that the logos still load and work, too.

drewvolz (approved):

PR #7970, opened 2026-09-11
Move the building-hours decision logic out of the views and into pure functions

hawkrives:
The robots had been introducing logic decisions inside of the ui views, which made them hard to test (especially under SwiftUI).

This is a start at pulling out what we can into pure functions.


hawkrives:
Screenshots from the new XCUITest, for the description.

Swiping a row reveals the favourite action, and tapping it grows a Favorites section at the top of the list.

| Swiped | Favourited |
| --- | --- |
| swipe reveals the favorite action | Favorites section at the top of the list |

Both come from `testSwipingARowFavoritesTheBuilding`, watched red before green: with the `SwipeActions.Actions` block removed it fails on "Swiping a row should reveal its Add to Favorites action".

drewvolz (approved):

PR #7985, opened 2026-09-12
Give every building status a glyph, and a status to services that are not doors

hawkrives:
Two problems, one cause.

**The dot could not say what it needed to.** The row had five statuses and three colours: `color-helpers.ts` mapped Open to green and Closed to red, and everything else fell through to `systemYellow`. Yellow therefore meant Almost Open, Almost Closed and Chapel at once, and the whole signal was lost to a colourblind reader.

**A set that is not a door had no way to say it was open.** SARN's advocate line runs 8pm to 8am, and `contextualStatus` skipped those sets entirely:

| When | Row said | Dot | Actually |
| --- | --- | --- | --- |
| Tue 20:30 | Closed today | red | Line staffed |
| Sat 22:00 | Closed today | red | Line staffed |
| Tue 07:00 | Opens at 11:10 AM | red | Line staffed until 8am |

## What changed

Every status now has a glyph as well as a colour. All of them share the circle silhouette, so the column stays aligned and shape carries what colour could not:

| Status | Glyph | Colour |
| --- | --- | --- |
| Open | `circle.fill` | green |
| Almost Open | `circle.lefthalf.filled` | yellow |
| Almost Closed | `circle.righthalf.filled` | yellow |
| Chapel | `bell.circle` | yellow |
| Service | the set's own symbol | green |
| Closed | `circle` | red |

A schedule set can describe the service behind it, via the `status` block that #7986 adds to the schema and the two data files. This branch is the code that reads it: `name` is what the status line says, and `symbol` is the glyph, falling back to a filled circle when absent.

**Availability first, kind second.** The service glyph appears only when no door is open, so a building open at the door while staffing a phone still reads Open. But a service open *now* outranks a door that opens *later*: at 3am SARN's line is what you can actually reach.

**`BuildingStatusType` is real now.** It was declared and never used, and never matched what the code produced — `getScheduleStatusAtMoment` returned sentences like `"Opens in 17 minutes"` that nothing rendered, formatted for every row on every minute tick and thrown away. It now returns a status, the colour map is keyed by that type, and a missing case is a type error rather than a silent fall-through to yellow. `getAccentTextColor`, `FG_COLORS` and two barrel aliases went with it, all of which had zero consumers.

## Depends on the data landing first

The data half is split out into #7986 so ccc-server can pick the fields up separately. `fetchBuildings` uses this repository's bundled hours only when `isUITesting`; every other run fetches `spaces/hours` over the wire, so a field reaches a device only after the server has it.

Until then this branch is inert for the Service status specifically: a dev build shows SARN as a plain red circle, which is correct given the data it was handed. Everything else here — the glyphs, the typed status, the Almost states — works immediately.

## Test plan

- [x] `npx jest` — 199 suites, 1797 passing
- [x] `npx tsc --noEmit` — clean
- [x] `npx oxlint` — clean
- [x] `npx oxfmt --check .` — clean
- [x] UITest on a simulator: Open and Closed glyphs render and are distinguishable
- [x] Dev build on a physical iPhone: installs, launches, survives
- [ ] The half-filled, bell and phone glyphs seen on a device

The last box needs either the clock moved or the server updated, and `simctl` has no time command. That gap is what the dev-overrides spec is for.

https://claude.ai/code/session_01MAehhCQ3YMSnedwQHnrB1i

### DIFF
--- a/source/features/building-hours/lib/__tests__/contextual-status.test.ts
+++ b/source/features/building-hours/lib/__tests__/contextual-status.test.ts
@@ -243,4 +243,20 @@ describe('contextualStatus', () => {
 
 		expect(contextualStatus(building, now).short).toBe('Opens at 1 PM')
 	})
+
+	it('names the service when only the service is open', () => {
+		// data/building-hours/7-3-sarn.yaml; the advocate line runs 8pm to 8am.
+		let building = makeBuilding([
+			{title: 'Office', hours: [{days: ['Tu'], from: '7:00pm', to: '8:00pm'}]},
+			{
+				title: 'Phone',
+				isPhysicallyOpen: false,
+				status: {symbol: 'phone.circle', name: 'Phone'},
+				hours: [{days: ['Mo', 'Tu', 'We', 'Th', 'Fr', 'Sa', 'Su'], from: '8:00pm', to: '8:00am'}],
+			},
+		])
+		let now = moment.tz('2026-09-08 22:00', timezone) // Tuesday 10pm
+
+		expect(contextualStatus(building, now).short).toBe('Phone until 8 AM')
+	})
 })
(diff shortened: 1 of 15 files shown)


drewvolz (approved):

hawkrives:
TODO

- flip the open-soon and close-soon icons; they read backwards
- try SARN without a circle

drewvolz (approved):

PR #7995, opened 2026-09-12
Store and query calendar events in on-device SQLite

hawkrives:
## Summary

_Left for @hawkrives to write — per CLAUDE.md, the description is yours._

## Outstanding before merge

Acceptance was deliberately deferred so review could start in parallel. Jest cannot answer any of these — the list view is an `@expo/ui` SwiftUI `ScrollView`/`LazyVStack`, so where it opens and how it scrolls is only observable on a simulator.

- [x] `ModuleCalendarTests` (~36% of the UITest suite by time)
  - Green in CI on 27126b2e0, all three shards. `testAttributionOnlyOnTheDetailScreen` was failing because the read hooks passed the device's clock to `hydrate` while the screens pick their day from `now()`, which UI testing freezes: once the real date passed the event's end it stopped counting as ongoing, fell off the frozen day, and the test opened a different event whose caption sits below the sheet's resting detent. Fixed in c6ec35848.
- [ ] Does the opening scroll show a frame at the top of last month before it moves, and does it animate?
  - The list now opens on today (`testTheUpcomingListOpensOnToday`, seen red first), but only settled screenshots were checked; the first frames were not recorded.
- [ ] Does SwiftUI resolve a `scrollPosition` id belonging to a lazily-built `LazyVStack` section far down the stack?
- [x] Does the first Today press do anything when `scrollTarget` already holds today's key?
  - Scrolling writes the leading section back into `scrollTarget`, so after a scroll the first press returns to today (`testTodayReturnsTheUpcomingListToTheTop`). Without a scroll the press does nothing, and the list is already on today.
- [ ] Where does the reader land when today has no events and the nearest section is in the past?
- [ ] The failed-read notice and its Try Again, as drawn on `EventList`, `DayView`, and `EventDetail`
- [ ] Cold launch offline shows the previous window rather than an empty list
- [ ] Selecting a category narrows the list, and the menu's tallies match what the list shows
  - Narrowing passes (`testResetFiltersClearsTheFilter`), and reset leaves the reader where they were. The tallies were not compared against the list's rows. The fixture's Academic Year tally is 7: the window keeps Welcome Convocation, which finished that morning.

## Device calendars are gone

The list reads its events from SQLite, which device events never entered — so they were still being fetched and then thrown away. The feature was dev-mode only and behind an EventKit grant, so rather than teach the database about EventKit ids, the source is deleted: `device-calendar.ts`, the four EventKit queries, the access check, the picker's "Add Device Calendars…" row, the detail screen's device branch, and `CalendarSource.kind`, which nothing read once `'device'` was no longer one of its values.

`@frogpond/add-to-device-calendar` — putting an event *onto* the phone's calendar — is a separate module and is untouched, so `expo-calendar` and `NSCalendarsUsageDescription` both stay.

## Known, accepted, documented

- Retention runs one day wider than the read window. `end_date` is exclusive, so an occurrence ending exactly at the cutoff genuinely finished before the window opens; the window is right to exclude it and retention is simply more generous. Safe direction, recorded in the design note.
- An event cancelled upstream *after it has already finished* stays in the retained past until it ages out. The feed stops mentioning it, so there is no signal to act on without an ingest change. One cancelled while still upcoming or running does disappear.
- A tag carried only by an out-of-window copy of a cross-source duplicate is printed and filterable but not tallied. Omits a menu value rather than offering a dead one.

## Deliberately not in scope

FTS5 search; venue/building queries; switching St. Olaf ingest to `/tec/v1`; the `per_page` truncation (the app fetches page one of 13, so it shows 50 of 642 St. Olaf events — deferred because fixing it would invalidate the equivalence test's baseline); Presence's unused `tags` array; making `eventKey` zone-stable.

A follow-up task is queued for a pre-existing dedupe/window interaction that can hide a cross-source duplicate whose higher-ranked copy falls outside the read window.


drewvolz:
## Code Review Findings (xhigh)

### Correctness Issues

**1. Empty string organization tags** (`source/database/calendar/rows.ts:87`)
If upstream sends `organization: ['']`, it passes through all filters and displays `['']` instead of no organization. The `sponsorMap` empty-string check has the same root cause.

**2. Device events never displayed** (`modules/ccc-calendar/use-merged-events.ts:23`)
Dev-mode only: device calendar events are fetched but `useMergedEvents` discards the data (only uses metadata). Events never reach the display layer.

**3. convertEvents called per-row** (`source/database/calendar/hydrate.ts:44`)
Called once per row instead of batched — for 100 events that's 100 function calls instead of 1.

### Cleanup Items

**4. Duplicated localDate function** (`source/database/calendar/read.ts:48`)
Same function in both `read.ts` and `write.ts` with a comment acknowledging it.

**5. NO_WINDOW sentinel workaround** (`app/(home)/EventDetail.tsx:68`)
`useNeighbours` lacks an `enabled` parameter, so a 1970-01-01 window stands in for "skip this query".

**6. Triple-slash comments** (`modules/event-list/keys.ts:3`)
Uses `///` instead of JSDoc `/** */` per CLAUDE.md convention.

---
https://claude.ai/code/session_01RdzQbBJGHqmadwLXE2499Z

hawkrives:
All six addressed, plus the shard-3 failure. CI is green on 27126b2e0.

**1. Empty string organization tags** — `tagRows` now drops any value that is blank after trimming, so a feed sending `organization: ['']` writes no tag row at all. Fixed at the write rather than at each reader: the menu, the filter and the detail caption all read the same rows.

`sponsorMap` needs no matching guard. `organizationsQuery` groups over `event_tag`, so a dedupe key with no organization rows produces no row at all — with blanks gone at the write, `orgs` can never be the empty string that `split` would turn into `['']`. (94f887feb)

**2. Device events never displayed** — real regression, and the fix was to delete the feature rather than restore the merge. Device events were still being fetched and discarded; putting them back meant either a second in-memory merge path beside the SQL one, or teaching the database about EventKit ids. Dev-mode only, behind a grant, so: `device-calendar.ts`, the four EventKit queries, the access check, the picker's "Add Device Calendars…" row, the detail screen's device branch, and `CalendarSource.kind` — which nothing read once `'device'` stopped being one of its values. `@frogpond/add-to-device-calendar` is a different module and is untouched. −760 lines. (27126b2e0)

**3. convertEvents called per-row** — `hydrate` parses every row's wire JSON first and converts the batch in one call. (3a0740234)

**4. Duplicated localDate** — moved to `source/database/calendar/dates.ts`; read and write both import it. They have to agree — the retention cutoff and the read window's back edge name the same day — so two copies could have drifted without a test noticing. (792884374)

**5. NO_WINDOW sentinel** — `useOccurrences` now takes `Window | null` and uses React Query's `skipToken`. The ternary that picks it is also what narrows `window` to a non-null `Window` for the read, so no assertion is needed. Covered by a test that fails if the query runs — asserting the SQL runner went uncalled alone would have passed either way, since `occurrencesQuery` throws on a null window before it reaches the runner. (ab726c391)

**6. Triple-slash comments** — JSDoc in `keys.ts` and `rows.ts`. `plugins/with-app-delegate-customizations.ts` has a `///` block too, but it predates this branch, so I left it. (116bb5a9c)

---

**The failing shard** — `testAttributionOnlyOnTheDetailScreen`. The read hooks passed the device's clock to `hydrate` while the calendar screens choose their day from `now()`, which UI testing freezes. Once the real date passed the event's end it stopped counting as ongoing and fell off the frozen day, so the test opened a different event whose caption sits below the sheet's resting detent. Both hooks now hydrate against `now()`. (c6ec35848)

No test for item 3: asserting `convertEvents` was called once is asserting a mock, not behaviour. The existing `hydrate` tests cover what it produces and stayed green.

drewvolz (approved):

PR #8021, opened 2026-09-16
Format times and dates in the device locale

### LINKED ISSUE #8001: Campus times should reflect device time formatting preferences
Eg 24hr mode

hawkrives:
Closes https://github.com/StoDevX/AAO-React-Native/issues/8001


drewvolz:
## Code Review

Reviewed the `@frogpond/time-format` module and its call sites at `xhigh` effort. 11 findings, most-severe first.

### Correctness — crashes

1. **`modules/time-format/index.ts:19`** — `localeWithHourCycle()` appends its own `-u-hc-` Unicode extension without checking whether the tag already carries a `-u-` extension, producing an invalid doubled-singleton BCP47 tag.
   If `expo-localization` ever reports a languageTag that already has a Unicode extension (e.g. an iOS user set Settings → General → Language & Region → Calendar to a non-Gregorian calendar, yielding `ja-JP-u-ca-japanese`), `deviceLocale()` builds `ja-JP-u-ca-japanese-u-hc-h23`. Verified: `new Intl.DateTimeFormat('ja-JP-u-ca-japanese-u-hc-h23', {hour:'numeric'})` throws `RangeError: Invalid language tag`. Since `deviceLocale()` is the default locale for nearly every formatter in the app, this crashes date/time rendering app-wide (building hours, calendar, course schedule, bus times, streaming, food menu) the moment such a user opens any screen.

2. **`app/(home)/CourseDetail.tsx:72`** — `formatWeekday(moment(day, 'dd'), 'long')` has no fallback for an unrecognized `day` value, unlike the deleted `format-day.ts`'s `default: return day`.
   `ScheduleDay.day` is a plain `string` taken straight from the SIS feed's `offering.day` with no enum validation. Any value moment's `'dd'` parser doesn't recognize makes `moment(day, 'dd')` an Invalid Moment; `.toDate()` is Invalid Date, and `Intl.DateTimeFormat.format()` throws `RangeError: Invalid time value` (verified), crashing the course detail screen where the old code just echoed the raw string back.

3. **`source/features/transportation/bus/components/times.tsx:17`** — `formatTime(time)` throws for an Invalid Moment produced by `parse-time.ts`'s strict-mode parse, where the old `.format()` call degraded gracefully.
   `parse-time.ts` calls `moment.tz(time, 'h:mma', true, timezone())` in strict mode; a bus-feed time string that doesn't match `'h:mma'` exactly returns an Invalid Moment (not `null`), which flows unfiltered through `processBusLine`/`processBusSchedule` into `DepartureTimeList`. The deleted code, `time?.format(TIME_FORMAT) ?? 'None'`, rendered `'Invalid date'` harmlessly (moment's `.format()` never throws). The new `time ? formatTime(time) : 'None'` treats the truthy Invalid Moment as valid and throws, crashing that stop's row.

### Correctness — locale leaks the review's own tests lock in as "passing"

4. **`source/features/building-hours/lib/summarize-days.ts:65`** — `formatFullDay()` hardcodes the English word `'and'` before the last day regardless of `locale`, even though every day name it joins goes through the locale-aware `formatWeekday()` — and the PR's own test comment claims this branch "follows the locale too."
   The new `describe('follows the locale')` block asserts `summarizeDays(['Mo', 'We', 'Sa'], true, 'ja-JP')` equals `'月曜日, 水曜日, and 土曜日'` — an English word spliced into a Japanese string — locking in the defect as expected, passing behavior.

5. **`source/features/building-hours/lib/summarize-days.ts:43`** — The `'Weekdays'` / `'Weekends'` / `'Every day'` shorthand strings stay hardcoded English no matter what `locale` is passed, unlike every other branch of `summarizeDays`, and none of the new "follows the locale" tests exercise this shorthand path.
   A Japanese-locale device viewing a building open Monday–Friday sees the English word `'Weekdays'` embedded in an otherwise-Japanese building-hours line, and nothing in the new locale-coverage suite would catch a regression or confirm the gap exists.

6. **`modules/event-list/times.ts:23`** — `detailTimes()`'s same-day branch overwrites a localized end time with the hardcoded English literal `'Midnight'`, the same defect as `format-times.ts`'s `formatSingleTime()`.
   `start`/`end` are built via `formatDateTime`/`formatTime` using the caller's locale, but when `event.endTime` is exactly hour 0 the code unconditionally sets `end = 'Midnight'` with no locale awareness — a Japanese-locale device with an event ending at midnight gets a mixed-language summary.

### Correctness — stale cache

7. **`modules/time-format/index.ts:28`** — `deviceLocale()` memoizes the device language + 24-hour-clock preference once into a module-level `cachedDeviceLocale` and never invalidates it, but the module's own comment on `localeWithHourCycle()` explains the 24-Hour Time switch is a distinct OS setting `Intl` can't see on its own — and that switch doesn't force an app relaunch the way a language change does.
   A user backgrounds the app, flips Settings → General → Language & Region → 24-Hour Time, and returns without the JS context reloading. Every `formatTime`/`formatDate`/`formatDateTime` call across building-hours, bus times, course schedule, and event details keeps rendering with the stale hour-cycle for the rest of the session.

### Test coverage

8. **`modules/event-list/__tests__/times.test.ts:107`** — `detailTimes()`'s same-day/multi-day/zero-length tests assert the exact comma-joined string `'Aug 20, 2026, 5:30 PM'`, without the caveat `source/features/streaming/streams/__tests__/row-detail.test.ts` already documents for this identical format shape: Apple's on-device ICU renders `'at'` instead of a comma, so this Jest assertion verifies an output real iPhones don't produce.
   Anyone trusting these green tests as proof of the event share-sheet's actual on-device text will find it reads `'Aug 20, 2026 at 5:30 PM'` on a real iPhone — the same class of gap the PR's own `row-detail.test.ts` shows the authors already knew to flag elsewhere, just missed here.

### Maintainability

9. **`scripts/jest-setup.js:17`** — The global `jest.mock('expo-localization', ...)` default (`languageTag: 'en-US'`, `uses24hourClock: false`) is now silently load-bearing for pre-existing tests that never declare a locale dependency, because `formatTime`/`formatDate` now route through `deviceLocale()` by default wherever no explicit locale is passed.
   A later, unrelated change to this shared mock (or a merge conflict resolving it differently) would flip tests across building-hours, bus, course-schedule, and streaming — e.g. `contextual-status.test.ts`'s `'Open until 8 PM'` — to 24-hour output with no visible connection back to `jest-setup.js`.

### Conventions (CLAUDE.md: "No Moment.js — use date-fns or Day.js")

10. **`modules/time-format/index.ts:1`** — The entire new `@frogpond/time-format` module is built around moment-timezone's `Moment` type and value-imports `moment-timezone` for its `.tz()` patch. This is brand-new code, not pre-existing usage the diff merely touches, and every exported function takes a `Moment` as its primary parameter — entrenching Moment.js in new code the project's own conventions explicitly forbid.

11. **`app/(home)/CourseDetail.tsx:9`** — The diff adds a fresh `import moment from 'moment-timezone'` to a file that previously had zero date-library dependency, solely to build a throwaway Moment for `formatWeekday(moment(day, 'dd'), 'long')`, where the deleted `format-day.ts` needed no date library at all. The same throwaway-Moment pattern is duplicated in `summarize-days.ts` instead of being centralized as a single locale-aware helper.

---

hawkrives:
Went through all 11 findings. Fixes landed on the stacked branch (#8023), not committed directly here, since that's where this module's follow-on work has been happening.

**Fixed — verified crashes:**
1. `localeWithHourCycle()` doubling a `-u-` extension → `Intl` throws `Invalid language tag`. Fixed: c36bd8e13
2. `CourseDetail.tsx`'s unvalidated `day` → `Intl` throws `Invalid time value` on an unrecognized value. Fixed: 42dfdeb7d
3. Bus departure row crashing on an Invalid Moment (still truthy, so `time ? ... : 'None'` didn't catch it). Fixed: 9b139f693, plus the actual root cause — `parseTime` now returns `null` for unparseable data instead of an Invalid Moment, matching its own `null | Moment` signature: a267c2f00

**Fixed — documentation:**
8. Added the same Node-ICU-vs-Apple-ICU comment `row-detail.test.ts` already had to the equivalent `times.test.ts` assertions, and moved it to cover all three sibling tests pinning the same shape: 6821a4af9, 3c5ba2998
9. Documented that `jest-setup.js`'s `expo-localization` mock values are silently load-bearing for tests that never intended to depend on locale mocking: 6821a4af9

**Not fixing, by design:**
4–6. The hardcoded English words (`formatFullDay`'s `"and"`, `Weekdays`/`Weekends`/`Every day`, `event-list/times.ts`'s `'Midnight'`) are the same category as `Midnight`/`Noon`/status prose already called out as intentionally untranslated copy — accepted as out of scope for this pass.
10–11. Moment.js in new code: the whole surrounding ecosystem (building-hours, bus, course-search) already used `moment-timezone` before any of this work started — migrating to `date-fns` would be an unrelated, larger framework change, not a fix for this PR's scope.

**Already addressed, separately:**
7. The stale-cache finding is fixed on the stacked branch too, tied to the device's actual zone/locale rather than cleared unconditionally on every resume.

drewvolz (approved):

PR #8027, opened 2026-09-17
Size the day-picker strip from the viewport, and take @expo/ui 57.0.18

hawkrives:


drewvolz:
Built this one locally. Interestingly enough we lost our colors

> What broke: expo/expo#49838 (merged into @expo/ui 57.0.17) widened tint, border, strokeBorder, and containerBackground to accept any ShapeStyle instead of just Color, by erasing every value to AnyShapeStyle. That erasure always binds to SwiftUI's tint<S: ShapeStyle>(_:) overload (iOS 16+) instead of the plain tint(_ tint: Color?) overload (iOS 15+) — and the generic overload silently fails to render in UIKit-bridged contexts (list/form row action buttons, swipe actions, widgets), even for a bare color.
> 
> A same-day fix, expo/expo#50187 ("Keep the Color overload for the tint modifier"), was opened by an Expo maintainer on 2026-09-15 and closed without merging — so it never shipped, and 57.0.18 still has the bug.
> 
> There's no JS-side workaround: resolveShapeStyle('red') already sends {type: 'color', color: 'red'} correctly — the erasure happens natively in ViewModifierRegistry.swift, regardless of what JS sends.
> 
> Options to actually fix this:
> 1. Cherry-pick #50187's diff into patches/@expo__ui@57.0.18.patch (small, 3-file native patch)
> 2. Pin @expo/ui back to 57.0.16 (last version before #49838 landed)

PR #8031, opened 2026-09-17
Raise the minimum iOS to 27

hawkrives:
If we're going to raise it, we might as well raise it to latest.

Also, iOS 18 to 26 only dropped the XS/XR, and iOS 27 drops *nothing* compared with iOS 26.


app.config.ts
```diff
@@ -211,9 +211,11 @@ const config: ExpoConfig = {
 			'expo-build-properties',
 			{
 				ios: {
-					// Expo SDK 57 needs at least 16.4; the project has run ahead of
-					// that since SDK 56, and the Podfile must not drift from it.
-					deploymentTarget: '18.6',
+					// The oldest iOS the app installs on. Expo SDK 57's own floor is
+					// 16.4; the project has run ahead of that since SDK 56, and the
+					// Podfile must not drift from it. A build needs an SDK at least this
+					// new, so CI runs the Xcode 27 image -- see ios.yml.
```
hawkrives:
```suggestion
```

drewvolz (approved):

PR #8059, opened 2026-09-19
Mount the Upcoming list a step at a time, from today

hawkrives:
Switching the calendar between Day and Upcoming froze the screen: taps and swipes did nothing until the switch finished.

## Why

Upcoming rendered every event in the read window at once: 295 rows on the live calendars, about a month of them already over. `LazyVStack` only puts off SwiftUI's drawing. React still creates every row's `@expo/ui` views, and each view's props run a SwiftUI layout pass on the main thread. Going back to Day then tore all of those views down.

Instruments' hang detection showed this (iOS 27 simulator, profiler attached):

| | Before | After |
|---|---|---|
| Day → Upcoming, JS render + commit | ~300 ms | ~33 ms |
| Day → Upcoming, main thread frozen | 2.31 s | no hang over 250 ms |
| Upcoming → Day, main thread frozen | 1.45 s | no hang over 250 ms |
| Scrolling while the list grows | — | no hang over 250 ms |

## What changed

- **Upcoming leaves out events that are over.** `Ongoing` stays, and so does anything on an earlier day that hasn't ended yet, such as a KSTO show that started before midnight, whose `isOngoing` flag was set when the schedule was fetched. The read window keeps its month of past events for Day's strip and the event detail's timeline.
- **A window whose events are all over shows the empty notice** (with Try Again when a calendar failed), instead of a blank list.
- **Upcoming mounts 15 rows, then 15 more whenever a scroll comes to rest within a screen of the end.** It mounts whole days only, and always through the section the list opens on, so `Ongoing` can't push today out of reach.
- `ScheduleView` (KSTO/KRLX) uses the same list, so it now mounts in steps too. Its data already has no finished events.

## Tests

- Jest: `upcomingSections`, `sectionsToMount` and `isNearEnd` in `sections.test.ts`, plus checks that the list leaves out finished days and shows the notice when nothing is left. Each fails when the code it covers is broken.
- The failure-note placement test used past days to sit above today. It now uses an `Ongoing` run, the one thing that can still sit there.
- UITest `testScrollingUpcomingToTheEndReachesTheLastEvent`: fails with growth switched off (the list stops at Sep 14) and passes with it on. All 21 `ModuleCalendarTests` pass locally.

## Not in this PR

The first ⋯ menu open after launch hangs for about 260 ms in a Debug build. It's Swift runtime and UIKit menu setup, with no app code involved, and it needs checking on a Release build before anything is done about it.


drewvolz (commented):
## Code Review Findings

Two issues identified:

### 1. Empty-check misses filtered sections (modules/event-list/event-list.tsx:153)

The early return checks `props.events.length === 0` but doesn't account for `upcomingSections` filtering all past events out. If a user has only past events in the read window (no ongoing, no today, no future), `props.events.length > 0` passes the guard, but `upcomingSections` returns `[]`, causing the list to render an empty scrollable area instead of the "No upcoming events" NoticeView.

**Suggested fix:** Add a check for `sections.length === 0` after the existing guards.

### 2. Stale closure in growNearEnd (modules/event-list/event-list.tsx:100)

`growNearEnd` captures `hasMore` at render time. If a scroll phase change fires after render but before `hasMore` updates from a concurrent refresh, the closure sees a stale value. This can cause a one-cycle delay in list growth (stale `false` prevents budget growth) or a no-op state update (stale `true`). Self-corrects on next render, so practical impact is minimal.

**Note:** This is low-severity since the functional updater ensures state updates are correct and the issue self-corrects within one render cycle.

hawkrives:
Re: [review](https://github.com/StoDevX/AAO-React-Native/pull/8059#pullrequestreview-5254684637)

1. Fixed in e46f4b748 — the empty check now counts the events left after filtering, and passes those to `emptyNotice`, so a failed calendar still gets its "Could not load" notice when everything else is over.
2. Not changing this one: `ScrollView` rebuilds its `onGlobalEvent` handler from the modifiers on every render (`createViewModifierEventListener`), so a phase event always reaches the latest listener, and its `hasMore` matches what is mounted.

drewvolz (approved):

PR #8060, opened 2026-09-19
Condense duplicate day labels in building hours

drewvolz:
## Summary

- Groups schedule entries by their day label so "Weekdays" or "Weekends" appears once per section with multiple time slots stacked beneath
- The accent bar shows when any entry in the group is active
- Active times are bolded
- Normalizes the list row title weight to regular

## Test plan

- [x] Open a building detail sheet and verify duplicate day labels are condensed
- [x] Verify the accent bar appears next to the day label when open
- [x] Verify active time slots are bolded

Before | After
--|--
 | Simulator Screenshot - iPhone 17 Pro - 2026-09-18 at 22 05 50
Simulator Screenshot - iPhone 17 Pro - 2026-09-18 at 21 37 23 | Simulator Screenshot - iPhone 17 Pro - 2026-09-18 at 21 37 11


source/features/building-hours/detail/schedule-row-swiftui.tsx
```diff
@@ -17,63 +10,59 @@ import {formatBuildingTimes, summarizeDays} from '../lib'
...
+			<Text
+				modifiers={[
+					font({textStyle: 'body', weight: hasActiveEntry ? 'semibold' : 'regular'}),
+					foregroundStyle(c.label),
+				]}
+			>
+				{days}
+			</Text>
+			<Spacer />
+			<VStack alignment="leading" spacing={2}>
```
hawkrives: **Times column is leading-aligned, so its right edge is ragged.**

The VStack sits after the `Spacer`, so its widest line meets the trailing edge and every shorter line stops short of it. Every single-entry row is flush right, so a condensed group no longer lines up with the rows around it. The PR's own after-screenshot shows it: "9 AM — Noon" ends left of "3 PM — 10 PM" and of "5 PM — 10 PM" in the section above.

```suggestion
			<VStack alignment="trailing" spacing={2}>
```

source/features/building-hours/detail/schedule-row-swiftui.tsx
```diff
@@ -17,63 +10,59 @@ import {formatBuildingTimes, summarizeDays} from '../lib'
...
-					modifiers={[
-						font({textStyle: 'body', weight: isActive ? 'semibold' : 'regular'}),
-						foregroundStyle(c.secondaryLabel),
-					]}
-				>
-					{times}
-				</Text>
-			</HStack>
+	return (
+		<HStack alignment="top" spacing={BAR_GAP}>
```
hawkrives, line 27: **Dropping the inner `padding({vertical: BAR_OVERSHOOT})` removes the bar's overshoot.**

That padding made the bar clear the text by 3pt at each end, the Calendar.app idiom `event-list-row.tsx:174` and `event-detail-header.tsx:79` still use. Now the bar tracks the content height exactly, so it no longer matches the other accent bars in the app.

Also, with `maxHeight: Infinity` the bar now spans every time in the group, not only the active one — in the screenshot it runs beside "9 AM — Noon" while "3 PM — 10 PM" is the open window. The JSDoc says the bar sits "to the left of the day label", which is not what it draws.

source/features/building-hours/detail/building-detail.tsx
```diff
@@ -119,23 +120,38 @@ export function BuildingDetailSwiftUI({building, now, campus}: Props): React.Rea
...
-							/>
-						))}
-					</Section>
-				))}
+				{schedules.map((schedule) => {
+					let groups = new Map<
+						string,
+						Array<{schedule: (typeof schedule.hours)[0]; isActive: boolean}>
+					>()
+					for (let set of schedule.hours) {
```
hawkrives: **The `Map` merges rows that are not next to each other, which reorders the section.**

Hours of `[Weekdays 7–10am, Saturday 10–2, Weekdays 5–9pm]` render as Weekdays (7–10am, 5–9pm) then Saturday: the second Weekdays row jumps above Saturday. Times inside a group also keep data order rather than clock order.

I checked the bundled St. Olaf YAML and `carls-app/carls`, and no venue does this today, so this is latent. But the screen renders server data, so either group only runs of adjacent rows or sort each group's times on purpose, and say which in a comment.

source/features/building-hours/detail/schedule-row-swiftui.tsx
```diff
@@ -17,63 +10,59 @@ import {formatBuildingTimes, summarizeDays} from '../lib'
...
+					foregroundStyle(c.label),
+				]}
+			>
+				{days}
+			</Text>
+			<Spacer />
+			<VStack alignment="leading" spacing={2}>
+				{entries.map(({schedule, isActive}) => (
+					<Text
+						key={`${schedule.from}-${schedule.to}`}
```
hawkrives: **`${from}-${to}` is not unique inside a group.**

Two rows with the same days and times — `['Mo','Tu']` and `['Tu','Mo']` both 9–5, or a row duplicated in the data — land in one group with one key. React warns about the duplicate key and can drop or merge a row when the list updates. The old key had the same weakness, but it is now the only thing telling entries apart inside a row.

source/features/building-hours/detail/building-detail.tsx
```diff
@@ -119,23 +120,38 @@ export function BuildingDetailSwiftUI({building, now, campus}: Props): React.Rea
...
-								accentColor={accentColor}
-								isActive={isScheduleRowActive(schedule, set, now)}
-								now={now}
-								schedule={set}
-							/>
-						))}
-					</Section>
-				))}
+				{schedules.map((schedule) => {
+					let groups = new Map<
```
hawkrives: **The grouping has no test.**

Condensing rows is data shaping decided in JavaScript, which the repo's CLAUDE.md says belongs in Jest: *"In Jest, assert what is genuinely decided in JavaScript: pure functions, reducers, parsers, data shaping, hooks, and which branch a component chose."* The only fixture in `building-detail.test.tsx` has one hours row, so nothing checks that two Weekdays rows become one, that the bar shows when *any* entry is active, or that only the active time is bold. The PR's test plan is manual only.

hawkrives: **The grouping belongs in `lib/`, not inline in the JSX map.**

A pure `groupHoursByDays(set, now): Array<{label, entries}>` beside `isScheduleRowActive` keeps the screen declarative and can be tested without the `@expo/ui` stand-in. It also removes the inline type and the second `summarizeDays` call noted below.

source/features/building-hours/detail/building-detail.tsx
```diff
@@ -119,23 +120,38 @@ export function BuildingDetailSwiftUI({building, now, campus}: Props): React.Rea
...
+						string,
+						Array<{schedule: (typeof schedule.hours)[0]; isActive: boolean}>
```
hawkrives: **This inline type repeats `ScheduleEntry` from `schedule-row-swiftui.tsx`.**

`(typeof schedule.hours)[0]` is `SingleBuildingScheduleType`. Export `ScheduleEntry` and use `new Map<string, ScheduleEntry[]>()` so the two cannot drift.

source/features/building-hours/detail/schedule-row-swiftui.tsx
```diff
@@ -17,63 +10,59 @@ import {formatBuildingTimes, summarizeDays} from '../lib'
...
-				<VStack
-					modifiers={[
-						frame({minWidth: 4, maxWidth: 4, maxHeight: Infinity}),
-						...(isActive ? [background(accentColor), clipShape('capsule')] : []),
-					]}
-				>
-					{null}
-				</VStack>
-			) : null}
+	let days = summarizeDays(entries[0].schedule.days)
```
hawkrives: **The day label is computed twice.**

The parent already has it as the `Map` key (`building-detail.tsx:129`); this recomputes it from `entries[0]` and relies, without saying so, on every entry sharing the same days. Pass the label in as a prop and drop this call.

source/features/building-hours/detail/schedule-row-swiftui.tsx
```diff
@@ -17,63 +10,59 @@ import {formatBuildingTimes, summarizeDays} from '../lib'
...
-	schedule,
-	now,
-	isActive,
-	accentColor,
-	showAccentBar = true,
-}: Props): React.ReactNode {
-	let days = summarizeDays(schedule.days)
-	let times = formatBuildingTimes(schedule, now)
+export function ScheduleRowSwiftUI({entries, now, accentColor}: Props): React.ReactNode {
+	if (entries.length === 0) return null
```
hawkrives: **This guard can never fire.**

The only caller builds each group with at least one entry. The dead branch suggests an empty group is a real case when it is not.

source/features/building-hours/detail/schedule-row-swiftui.tsx
drewvolz: Dismissing, this is intentional

source/features/building-hours/lib/__tests__/group-hours-by-days.test.ts
```diff
@@ -0,0 +1,79 @@
...
+	test('keeps non-adjacent rows with the same label as separate groups', () => {
+		let schedule = makeSchedule([
+			{days: ['Mo', 'Tu', 'We', 'Th', 'Fr'], from: '7:00am', to: '10:00am'},
+			{days: ['Sa'], from: '10:00am', to: '2:00pm'},
+			{days: ['Mo', 'Tu', 'We', 'Th', 'Fr'], from: '5:00pm', to: '9:00pm'},
+		])
+
+		let groups = groupHoursByDays(schedule, now)
+
+		expect(groups).toHaveLength(3)
```
hawkrives: **These tests check counts and labels, never which row landed where.**

Nothing here reads `sourceIndex`, `startIndex`, or `entries[n].schedule`. All five tests would pass if the grouping filed the 7am row under the third group, or if `groupStartIndex` never updated and every group reported `0`. Both indexes become React keys (`building-detail.tsx`, `schedule-row-swiftui.tsx`), so a stuck `startIndex` gives duplicate keys within a section.

Asserting the whole shape covers it in one line per test:

```ts
expect(groups.map((g) => [g.label, g.startIndex, g.entries.map((e) => e.sourceIndex)])).toEqual([
	['Weekdays', 0, [0]],
	['Saturday', 1, [1]],
	['Weekdays', 2, [2]],
])
```

The active-entry test has the same gap for the flush after the loop: it never puts the active row in a group that follows a label change.

source/features/building-hours/lib/group-hours-by-days.ts
```diff
@@ -0,0 +1,56 @@
...
+ * Groups schedule hours into runs of adjacent rows with the same day label.
+ * Non-adjacent rows with the same label remain separate groups.
+ */
+export function groupHoursByDays(schedule: NamedBuildingScheduleType, now: Moment): GroupedHours[] {
+	let groups: GroupedHours[] = []
+	let currentLabel: string | null = null
+	let currentGroup: ScheduleEntry[] = []
+	let groupStartIndex = 0
+
+	for (let i = 0; i < schedule.hours.length; i++) {
```
hawkrives: **The loop keeps three variables that mirror the last group, and flushes twice.**

`currentLabel`, `currentGroup`, and `groupStartIndex` shadow what `groups.at(-1)` already holds. The flush appears inside the loop and again after it, and the `currentLabel !== null` check exists only to narrow the type. Appending to the last group does the same job:

```ts
let groups: GroupedHours[] = []
schedule.hours.forEach((set, i) => {
	let label = summarizeDays(set.days)
	let entry = {schedule: set, isActive: isScheduleRowActive(schedule, set, now), sourceIndex: i}
	let last = groups.at(-1)
	if (last?.label === label) {
		last.entries.push(entry)
	} else {
		groups.push({label, entries: [entry], startIndex: i})
	}
})
return groups
```

source/features/building-hours/lib/group-hours-by-days.ts
```diff
@@ -0,0 +1,56 @@
...
+	isActive: boolean
+	/** Original index in schedule.hours, for stable React keys. */
+	sourceIndex: number
+}
+
+export type GroupedHours = {
+	label: string
+	entries: ScheduleEntry[]
+	/** Index of the first entry in the original hours array. */
+	startIndex: number
```
hawkrives: **`startIndex` always equals `entries[0].sourceIndex`.**

Every group has at least one entry, so this field copies data the group already carries, and `groupStartIndex` exists only to keep the copy in step. Keying the row on `group.entries[0].sourceIndex` in `building-detail.tsx` would let both go.

source/features/building-hours/lib/__tests__/group-hours-by-days.test.ts
```diff
@@ -0,0 +1,79 @@
+import {groupHoursByDays} from '../group-hours-by-days'
+import moment from 'moment-timezone'
+import type {NamedBuildingScheduleType, DayOfWeekEnumType} from '../../types'
+
+function makeSchedule(
+	hours: Array<{days: DayOfWeekEnumType[]; from: string; to: string}>,
+): NamedBuildingScheduleType {
+	return {
+		title: 'Test',
+		hours: hours.map((h) => ({...h, timezone: 'America/Chicago'})),
```
hawkrives: **`timezone` is not a field of `SingleBuildingScheduleType`.**

A row carries `days`, `from`, and `to`; `findOpenWindow` takes its zone from `timezone()` in `@frogpond/constants`. TypeScript lets the extra key through only because it comes out of a `map` callback, not a literal. It does nothing except suggest each row picks its own zone. `makeSchedule`'s parameter already fits the hours type.

```suggestion
		hours,
```

source/features/building-hours/lib/group-hours-by-days.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+/**
+ * Groups schedule hours into runs of adjacent rows with the same day label.
+ * Non-adjacent rows with the same label remain separate groups.
+ */
```
hawkrives, line 22: **The JSDoc covers the grouping but not the other half: marking which rows are open at `now`.**

That is the only reason the function takes `now`, and a caller reading `groupHoursByDays(schedule, now)` has to open this file to learn it.

```suggestion
/**
 * Groups schedule hours into runs of adjacent rows with the same day label,
 * and marks each row that is running at `now`.
 * Non-adjacent rows with the same label remain separate groups.
 */
```

source/features/building-hours/lib/__tests__/group-hours-by-days.test.ts
```diff
@@ -0,0 +1,79 @@
+import {groupHoursByDays} from '../group-hours-by-days'
+import moment from 'moment-timezone'
```
hawkrives: **The local import comes before the third-party one.**

The repo's CLAUDE.md: "**Imports:** React → React Native → third-party → local."

```suggestion
import moment from 'moment-timezone'
import {groupHoursByDays} from '../group-hours-by-days'
```

source/features/building-hours/lib/__tests__/group-hours-by-days.test.ts
```diff
@@ -0,0 +1,79 @@
...
+	hours: Array<{days: DayOfWeekEnumType[]; from: string; to: string}>,
+): NamedBuildingScheduleType {
+	return {
+		title: 'Test',
+		hours: hours.map((h) => ({...h, timezone: 'America/Chicago'})),
+	}
+}
+
+describe('groupHoursByDays', () => {
+	let now = moment.tz('2026-09-18T15:00:00', 'America/Chicago')
```
hawkrives, line 15: **`./moment.helper` already builds Chicago moments for these tests.**

`is-schedule-row-active.test.ts` and the other tests in this folder use `plainMoment` rather than calling `moment.tz(…, 'America/Chicago')` themselves:

```ts
let now = plainMoment('2026-09-18T15:00:00', 'YYYY-MM-DD[T]HH:mm:ss')
```

That keeps the campus zone in one place and drops the direct `moment-timezone` import.

source/features/building-hours/lib/__tests__/group-hours-by-days.test.ts
```diff
@@ -0,0 +1,79 @@
...
+		expect(groups[2].entries).toHaveLength(1)
+	})
+
+	test('marks the active entry when now falls within its hours', () => {
+		let schedule = makeSchedule([
+			{days: ['Mo', 'Tu', 'We', 'Th', 'Fr'], from: '9:00am', to: '12:00pm'},
+			{days: ['Mo', 'Tu', 'We', 'Th', 'Fr'], from: '3:00pm', to: '10:00pm'},
+		])
+
+		// Friday at 3pm is within the second slot
```
hawkrives: **The comment says 3pm; the moment is 3:30pm.**

3pm is the second slot's opening boundary, a different case from the one this test checks.

```suggestion
		// Friday at 3:30pm is within the second slot
```

source/features/building-hours/detail/schedule-row-swiftui.tsx
```diff
@@ -1,79 +1,61 @@
...
-	frame,
-	padding,
-} from '@expo/ui/swift-ui/modifiers'
+import {background, clipShape, font, foregroundStyle, frame} from '@expo/ui/swift-ui/modifiers'
 import * as c from '@frogpond/colors'
 import type {Moment} from 'moment-timezone'
-import type {SingleBuildingScheduleType} from '../types'
-import {formatBuildingTimes, summarizeDays} from '../lib'
+import type {ScheduleEntry} from '../lib'
+import {formatBuildingTimes} from '../lib'
```
hawkrives: **Two imports from `../lib`.**

```suggestion
import {formatBuildingTimes, type ScheduleEntry} from '../lib'
```

source/features/building-hours/list/building-list-row.tsx
```diff
@@ -93,7 +93,7 @@ export const BuildingListRow = React.memo(function BuildingListRow({
 				>
 					<Text
 						modifiers={[
-							font({textStyle: 'body', weight: 'medium'}),
+							font({textStyle: 'body', weight: 'regular'}),
```
hawkrives: **`weight: 'regular'` is the default.**

The status text beside it is plain `font({textStyle: 'body'})`. Spelling out the default reads as an override, and makes two identical styles look different.

```suggestion
							font({textStyle: 'body'}),
```

source/features/building-hours/lib/group-hours-by-days.ts
hawkrives: this feels complicated. me no like.

drewvolz: We're thinking this is too fancy of a suggestion.

drewvolz: We probably like this better now compared to the current state.

hawkrives (approved):

Issue #8067, opened 2026-09-19
More list warns about duplicate keys: three A-Z entries share one URL


hawkrives:
The More list keys its rows by `link.url` (`app/(home)/More/index.tsx:68`), and three St. Olaf A–Z entries — "Academic Advising", "Academic Success Center" and "Center for Advising and Academic Support" — all point at `https://wp.stolaf.edu/academic-success/`. React logs a duplicate-key warning and LogBox raises a toast over the list on a debug build.

The data is upstream and arguably correct: three names for one destination. The client should not assume a URL is unique. `key={${link.url}-${link.label}}` is enough.

Pre-existing; surfaced while screenshotting that screen for #8061.

Issue #8066, opened 2026-09-19
Balances' "Log in with St. Olaf" row does nothing


hawkrives:
`source/features/sis/balances.tsx:32` defines `openSettings` as `() => {}`, and line 89 puts it behind a live `DisclosureRow` titled "Log in with St. Olaf". Tapping it does nothing at all — the row looks tappable, responds to touch, and goes nowhere.

The stub is documented as waiting on Settings moving to expo-router.

Found while auditing row accessories for #8061. Left alone there: a dead row is a real bug but has nothing to do with which glyph it draws.

drewvolz:
I remembered finding this but forgot to log it. Thanks! Separately, is that login button designed to not appear on first load of SIS (not the agreement screen, just regular load)? I did PTR and then saw it, which is how it became forgotten

Issue #8065, opened 2026-09-19
Reddit link card can't know its destination before the tap


hawkrives:
The link card at `app/(home)/RedditPostDetail.tsx:198` draws `chevron.right`, but its destination is only decided when you press it: `parseRedditPostUrl` pushes a `RedditPostDetail` for a Reddit post URL and opens the browser for anything else. Even a matching URL falls back to the browser when the fetch fails or returns nothing.

So no single accessory is honest for every press. Under the rule in #8061 a push takes `chevron.right` and a web page takes `arrow.up.right`, and this row is sometimes each.

Two ways out:

- Classify at render. `parseRedditPostUrl` is already a pure function in `useRedditLinkHandler.ts`; exporting it lets the card draw the right glyph for the URL it holds. The chevron would still lie on a network failure, which is an already-degraded case.
- Always `arrow.up.right`. The card shows a bare domain and a `globe` glyph at `:195` and already declares `accessibilityRole="link"` at `:190`, all of which argue the web is its honest default.

Left out of #8061 deliberately, as a product call rather than a mechanical fix.

drewvolz:
Defer this one. We're going to remove the community views.

PR #8068, opened 2026-09-19
Name each row's destination instead of always drawing a chevron

### LINKED ISSUE #8061: adjust chevron on lists
- chevrons for stack push
- no chevron for sheet or modal open
- no chevron for SFSafariWebViewController

hawkrives:
Closes #8061.

Rows drew `chevron.right` whenever they were tappable — whether the tap pushed a screen, opened a mail composer, placed a call, or opened a web page. The chevron means one of those, so it told you nothing.

## The rule

| Destination | Means | Accessory |
| --- | --- | --- |
| `push` | another screen in this navigation stack | `chevron.right` |
| `action` | the row is the thing, and tapping does it | none; the label is tinted |
| `external` | the row points at a document elsewhere | `arrow.up.right` |

The line between the last two is task against document, not which process draws the result. Calling a number or writing to an address completes what the row names and returns you where you were, so those are actions even though `tel:` and `mailto:` hand off to Phone and Mail. A web page is external even though an in-app browser renders it.

The tint stands in for a missing accessory rather than doubling one. `DetailRow` is the exception — its value is `secondaryLabel`-coloured text that reads as static, so it tints whenever it is tappable and not a push.

component-library-rows

## Why Calendar and Maps look like counter-examples

Both tap through with no accessory at all. Neither pushes: Calendar's event opens a sheet dismissed with an ✕, and a Maps search result replaces the sheet's own contents. Mail and Notes are the control — both push, both carry a chevron on every row, and Notes draws a count beside it, which is the shape `DetailRow` uses. Contacts is the one push list without a chevron, and its cause is documented: the A–Z index owns the trailing edge.

Marking the in-app browser at all is a deliberate step past Apple's practice, which leaves such rows unmarked in Settings, the App Store and Notes. A web page is a different context whatever frame it arrives in, and the arrow says so before the tap rather than after.

## What changed

`DisclosureRow` and `DetailRow` take `destination?: 'push' | 'action' | 'external'`, defaulting to `'push'`. A single `RowAccessory` owns the mapping. Twelve rows then say what they do:

- **external** — Streaming, Transportation, More, a student org's website, a person's office hours and profile
- **action** — org contacts and advisors, a person's email and phone

Two rows don't use the shared components. The news row builds its own because the accessory sits inside the title row so the excerpt below runs full width; it swaps glyph and now matches `RowAccessory`'s colour and size. Dictionary's sub-sense rows were an `@expo/ui` `Button` with `systemImage`, which draws on the **leading** edge — they're `DisclosureRow`s now.

## Screens

| More — every row external | Directory — action beside external |
| --- | --- |
| more | directory-person |

| Streaming | News | Dictionary — chevron now trailing |
| --- | --- | --- |
| streaming | news | dictionary-sense |

In Directory, a long value makes `LabeledContent` stack and the accessory trails the text rather than sitting at the trailing edge. Left alone: that is SwiftUI's own layout, and `LabeledContent` was chosen to follow the platform rather than fight it.

On News the accessory is a trailing column of the row rather than a sibling of the headline, so it reads as the row's accessory and the excerpt stops short of it. It stays pinned to the row's top edge rather than centred.

## Testing

Seven Jest tests assert which accessory each destination chooses, on both row types, plus that an untappable `DetailRow` draws nothing whatever it says. Each was mutation-checked — breaking `RowAccessory` three different ways failed exactly the expected tests.

No test asserts the tint, the glyph's size, or which edge it sits on. Those aren't decided in JavaScript, and a Jest assertion on them would pass while the screen was broken. They're verified in the screenshots above, which is also why this adds a **Rows page to the component library**: `DisclosureRow`'s action tint had no call site reachable with live data, since no org in the API has contacts.

## Left for their own issues

- #8065 — the Reddit link card can't know its destination before the tap
- #8066 — Balances' "Log in with St. Olaf" row is a no-op stub
- #8067 — three A–Z entries share a URL, so More warns about duplicate keys


source/features/news/news-row.tsx
```diff
@@ -141,17 +147,23 @@ export const NewsRow = (props: Props): React.ReactNode => {
...
-							color={c.secondaryLabel}
-							modifiers={CHEVRON_MODIFIERS}
-							size={20}
-							systemName="chevron.right"
-						/>
-					</HStack>
+					<Text modifiers={TITLE_MODIFIERS}>{story.title}</Text>
 					<Text modifiers={EXCERPT_MODIFIERS}>{story.excerpt}</Text>
 				</VStack>
+				{/* A `Spacer` beneath the glyph, inside its own `VStack`, is what
```
drewvolz, line 151: > The trailing accessory moved from an inline, title-anchored position into a new sibling VStack+Spacer that relies on SwiftUI cross-axis sizing to pin it to the row's top edge, and this purely visual layout change ships with no on-device verification.
> 
> For a story with no thumbnail (`thumb === null`), the only other sibling in the outer HStack is the text VStack (title+excerpt). If that column's resolved height isn't propagated to the new accessory VStack the way the comment assumes, the `arrow.up.right` glyph could render vertically centered or clipped instead of pinned to the top. Jest can't see this (correctly, per this repo's testing philosophy), and `uitests/ModuleNewsTests.swift` only calls `verifyNewsRowsAppear()`, which checks reachability, not glyph position — so a misplaced/overlapping icon would ship undetected.
> 
> ---
> _Generated by [Claude Code](https://claude.ai/code)_

source/features/news/news-row.tsx
```diff
@@ -141,17 +147,23 @@ export const NewsRow = (props: Props): React.ReactNode => {
...
+				    pins it to the row's top edge: the outer `HStack` is
+				    `alignment="center"`, and a `VStack` stretched to the row's full
+				    height settles non-expanding content -- the image -- at its own
+				    top before the `Spacer` claims the rest. */}
+				<VStack>
+					<SwiftUIImage
```
drewvolz, line 157: > This hand-rolls the same trailing-accessory glyph styling (`color: c.tertiaryLabel`, `size: 14`) that `source/components/rows.tsx`'s new `RowAccessory` component now centralizes, linked only by a comment rather than a shared constant or the component itself.
> 
> If `RowAccessory`'s color or size changes later (e.g. a future design tweak in `rows.tsx`), there's no compiler, lint, or test signal that this independent copy needs the same change, so the two accessories could silently diverge in appearance.
> 
> ---
> _Generated by [Claude Code](https://claude.ai/code)_

drewvolz (commented):
Reviewed at `xhigh` effort: diffed the full PR against `origin/master`, read every changed file plus enclosing functions/callers, cross-checked all `DisclosureRow`/`DetailRow` call sites for `destination` semantics, and ran the full Jest suite (1936 passed, only 5 pre-existing unrelated failures from missing bundled fixtures), `tsc --noEmit`, `oxlint`, and `oxfmt --check` — all clean on touched files.

The diff is well-disciplined overall: the `RowDestination` union (`push`/`action`/`external`) is threaded consistently through Directory, StudentOrgs, More, Transportation, and streaming call sites, tinting logic matches its documented intent, defaults preserve prior behavior for untouched callers, and the new Jest tests correctly assert branch selection rather than layout. No crash-level or logic bugs found — two lower-severity findings left as inline comments on `news-row.tsx`, both about unverified visual/layout changes and un-shared styling.

---

drewvolz (approved):

PR #8072, opened 2026-09-19
Turn Transportation into one screen of bus line widgets

### LINKED ISSUE #7991: Proposal: Transporation becomes a widget screen
Transporation becomes a widget screen, and you tap on a widget to open a formSheet with the full schedule.

but you can see the next few stops and the running/not status of each bus line at a glance, and scroll sideways through the current timetable index, or scroll downwards for Other Modes.

hawkrives:
Closes #7991.

Transportation was five native tabs, one per bus line plus Other Modes, each
drawing a full timetable. Answering "is the Express running, and when does it
next reach my stop" meant picking a tab and reading a column; comparing two
lines meant switching tabs and holding the first one in your head.

Now it is one screen. Each line is a widget: its name, what it is doing right
now, and a sideways-scrolling strip of its route. Tapping a widget anywhere
opens the full timetable in a sheet.

Transportation widgets

## The strip

Each stop shows its departure for the round the bus is on, and the strip says
where the bus is without anyone reading a time: the stops behind it are filled
discs on a solid rail, the ones ahead are rings on a faint one, and the bus is
a bead between them, drawn at its true fraction of the leg and pulsing so the
position reads as live rather than printed.

It opens on the stop the bus has just left, so the leg it is on has both ends
in view, and scrolls either way from there. It comes to rest on a stop rather
than mid-gap: every cell is one width, so `viewAligned` snapping means whatever
the strip shows is a run of whole stops rather than a column of times with a
name sliced in half at each edge.

The rail begins at the first stop and ends at the last — a route has two ends
and the rail should say so. Past the end sits one more slot: the next round's
first departure when another round follows, or, on the day's last round, an
empty ring standing off the end of the rail. Scrolling to the end of the route
answers a question instead of stopping dead.

The strip scrolled to the end of the route

A line that is not running collapses to its header. It has nothing to show, and
a placeholder card for each idle line pushed the one running line off screen.

## The sheet

It opens at half height and pulls to full. The timetable is SwiftUI throughout
— rows, rail, dots — rather than React Native views hosted inside the list,
which bled grey through at the medium detent and zoomed as the sheet resized.

While a bus is out, the stops it has already left collapse into a single row
naming how many, so the sheet opens on where the bus is; tapping that row shows
them. The day picker lives in its own toolbar, and tapping a stop opens that
stop's departures, pushed inside the sheet so there is a way back. A line with
nothing to show today gets a `ContentUnavailableView` rather than an empty
list.

| Timetable | One stop | Another day |
| --- | --- | --- |
| Line timetable | Stop schedule | Sunday schedule |

## Other Modes

Now below the widgets rather than behind a tab, and regrouped. "Bus" and "Bike"
named the vehicle, which is not what anyone chooses between — the question is
whether you are getting around Northfield or getting to an airport. The St.
Olaf transportation page carries no category: it points at everything rather
than being one of the options, so its section draws no heading.

Other Modes

## The line's colour

Each header leads with a `bus.fill` glyph in its line's own colour, bringing
back the cue the native tab bar used to carry.

It is tinted with `colors.bar`, not `colors.dot`. Despite the names, `bar` is
the line's colour — the rail *and* every stop's dot — while `dot` is only the
highlight for the stop a bus is currently sitting at, darkened so it reads
against the rail beneath it. Tinting with `dot` drew the glyph darker than the
stops beside it, and rendered the Blue Line's in a near-black that read as no
colour at all. `BusLineColors` now says so, so the next reader does not have to
work it out from the call site.

## Hiding a line without deleting it

Oles Go — now Campus-Go — is on-demand only, so it is an information entry
rather than a line with a timetable, and it moves to Other Modes.

Its feed document cannot just be deleted: released app versions still fetch it
and would break. So `data/_schemas/bus-times.yaml` grows an optional
`hidden: true`, which newer versions honour and older ones have never heard of.
`2-oles-go.yaml` keeps its `line`, `notice` and `schedules` untouched for
exactly that reason.

The filter is in the screen rather than the query: `busLineOptions` shares a
cache key and looks a single line up by name for the sheet, so filtering at
fetch would break that lookup.

## A patch to `@expo/ui`

The strip fought the reader's finger: a drag moved it, then jerked it
backwards. `@expo/ui` binds `.scrollPosition(id:)` straight to a shared
`ObservableState`, a scroll view reports its leading item as the reader drags,
and a held id is re-applied on every body pass -- which re-aligns the content
mid-gesture.

Measured from a screen recording, frame by frame, on a 60-step drag:

| | during the drag | at release |
| --- | --- | --- |
| before | 52 frames forward **and a 120px jump backwards** | no deceleration |
| after | 52 frames forward, no reversal | fling decelerates |

`patches/@expo__ui@57.0.14.patch` now holds the id only while a scroll written
from JavaScript is on its way, and lets go once it lands. A report from the
scroll view still reaches JavaScript; it is just never held as a target. The
calendar's event list uses the same modifier and gets the same fix.

## Notes for review

- The widget is drawn entirely in SwiftUI via `@expo/ui` — no `RNHostView`, no
  `matchContents`, so none of the unbounded-measure trouble the timetable had.
- Every decision either surface makes is a pure function with tests:
  `buildStopStrip`, `deriveLineState`, `findBusTarget`, `collapseEarlierStops`
  and `stripAnchorIndex`. The SwiftUI cells take props.
- Bus lines and transit modes are ordered by a numeric filename prefix in
  `data/`, the convention `data/help/` and `data/webcams/` already use. Both
  feed a UI-test fixture, so the ordering and the category names are this
  checkout's rather than whatever the server last deployed.
- The data changes land in production only once ccc-server picks them up; the
  app changes are inert until then, so the two can ship in either order.
- `ModuleTransportationTests` covers what Jest cannot see here: the sideways
  scroll, the sheet presenting, the cells' hit targets, and reaching the
  unlabelled Other Modes section. Six tests, all passing; every screenshot
  above is from that run.

Follow-up filed as #8073 — Other Modes and Important Contacts are converging on
the same shape and should share an `InfoDocument` schema and an `InfoSheet`,
which is also what Campus-Go needs before it can carry more than one line.


drewvolz:
Automated review at extra-high recall, against `8776960` (10 finder angles, deduped, swept). Errs toward surfacing — a few of these are judgement calls rather than certainties, and the last four are cleanup rather than bugs.

## Correctness

### 1. The strip draws the bus twice, on top of itself — `source/features/transportation/bus/widget.tsx:476`

`busPropsForRow` returns a fraction for **both** stops either side of the leg. That works in the timetable because a `List` row clips each half; the widget's `LazyHStack` inside a `ScrollView` clips nothing.

With the bus at progress 0.5, cell N gets `busFraction: -0.5` and cell N-1 gets `+0.5`. Both resolve to the same absolute x, so two `BusGlyph`s render at one point. Each has its own `pinging` state driven by its own `onAppear`, and `LazyHStack` instantiates the two cells at different moments — so the two expanding rings run out of phase, and where they do coincide the 0.45-opacity rings stack to roughly 0.70. `widget.test.tsx` asserts `toHaveLength(2)` and locks the duplicate in.

### 2. The anchor latch shares state with the scroll view's own reports — `widget.tsx:390`

`openOnTheBus` guards on `scrollTarget.get() === null`, but the patched `report(_:)` in `patches/@expo__ui@57.0.14.patch` does `state.value = id` for **every** id the scroll view reports. Two consequences:

- Any report that lands before `useScrollGeometryChange` fires with `contentWidth > 0` leaves `scrollTarget.get()` non-null, so the anchor is never set and the strip opens at stop 0 — the exact thing the feature exists to prevent.
- `RouteEndCell` (line 479) is the only child of the `scrollTargetLayout()` stack with no `id()` modifier. Whenever SwiftUI anchors on it the binding goes `nil`, `report(nil)` writes `state.value = nil`, and the next geometry tick sees `null` again and calls `scrollTarget.set(anchorId)` — the strip yanks itself back to the bus mid-scroll.

### 3. A skipped stop breaks the solid rail behind the bus — `widget.tsx:173`

`trailingActive = isPassed` and `previousPassed` both test only for `'after'`, so both rail halves either side of a `'skip'` cell draw faint even when the bus is well past them.

Stops A, B, C with `['2:00pm', false, '2:10pm']` (the fixture in `build-stop-strip.test.ts`) at 2:05pm: cells are `after / skip / before`. A's trailing half is solid and B's leading half is solid, but B's trailing and C's leading are both faint. `findBusTarget` puts the bus at progress 0.5 between B and C — so the bead rides a faint rail, contradicting the comment at line 167 that the leg the bus is on is solid all the way across.

### 4. Other Modes' error and loading states were dropped in the merge — `app/(home)/Transportation/index.tsx:43`

Only `data` and `refetch` are destructured. Offline, or `transit/modes` 500s while `transit/bus` is cached: `otherModes` falls back to `[]`, the sections silently vanish, and the screen shows bus widgets followed by an empty footer section. The deleted `other-modes.tsx` handled all three cases — `isError` → `NoticeView`, `isLoading` → `LoadingView`, `data.length === 0` → a "No Other Modes" `ContentUnavailableView`. None is re-established anywhere.

### 5. A bus-feed error now hides Other Modes entirely — `index.tsx:62`

`if (busesErrored) return <NoticeView .../>` short-circuits before the `List` renders. When `transit/bus` fails but `transit/modes` is fine, Hiawathaland Transit, Ole Bikes, Campus-Go, the airport shuttles and the St. Olaf page are all unreachable. Behind their own tab they used to survive a bus outage. Same for `busesLoading` at line 58.

### 6. `stop.tsx` re-derives the schedule from a different clock than the stop came from — `app/(home)/Transportation/line/stop.tsx:52`

`BusStopPage` resolves `stop` from `deriveLineState({line, now: momentForDay})`, but `BusStopDetailInternal` calls `processBusLine` / `getScheduleForNow` / `getCurrentBusIteration` with its own real-time `now` from the timer at line 139. Pick Sunday on a Wednesday and the `stopStatus` in the no-departures branch describes Wednesday's bus against Sunday's stop.

It is also duplicated work — the parent's `deriveLineState` already returns `status`, `currentBusIteration` and `schedule` — and the screen ends up with two independent 60 s timers, which is the opposite of what `index.tsx`'s own header comment says: *"One timer drives every widget: a clock per line would have them ticking over at slightly different moments."*

## Efficiency

### 7. Every widget re-parses every schedule, every minute, unmemoized — `widget.tsx:359`

`deriveLineState` → `processBusLine` maps `parseTime` over *all* of a line's schedules (not just today's), and `parseTime` does a strict `moment.tz(...)` parse plus three mutations per entry. Against the bundled feed that is **502 time cells across the four lines**, re-parsed on the JS thread every 60 s from `useMomentTimer`, plus again on every other parent render. On master this ran for one line at a time, because each line was its own tab. `BusLineWidget` has no `React.memo` and no `useMemo`; hoisting the parse into the query's `select`, or memoising on `[line, now.format('YYYY-MM-DD HH:mm')]`, removes essentially all of it.

## Simplification

### 8. The collapse fires at a hidden count of one — `source/features/transportation/bus/lib/collapse-earlier-stops.ts:25`

`targetIndex: 2` gives `firstVisibleIndex: 1, hiddenCount: 1`, so a "1 previous stop" row replaces exactly one `TimetableRow`. Same list length, one extra tap. Returning `{firstVisibleIndex: 0, hiddenCount: 0}` when `targetIndex - 1 < 2` is the one-line fix.

## Comments and dead code

### 9. Orphaned JSDoc — `source/features/transportation/bus/components/timetable-row.tsx:337`

Two `/** … */` blocks sit back to back at lines 336–346. The first ("One row of a bus timetable… Pressable when given `onPress`") describes `TimetableRow`, but now annotates `CollapsedStopsRow`; `TimetableRow` at line 392 has none. `CLAUDE.md`: *"JSDoc (`/** … */`) to annotate a declaration."*

### 10. Stale comment contradicting its own code — `timetable-row.tsx:243`

It says `busFraction` *"never exceeds half a row -- so the glyph stays inside the row drawing it, which a `List` row would otherwise clip."* `busPropsForRow` returns `progress - 1` (−1…0) and `progress` (0…1), and its own docstring says the opposite — that both rows draw the bus *because* the row clips. At progress 0.2 the target row's copy sits 0.8 rows above its dot and is clipped away entirely, which is the intended mechanism.

### 11. Dead `default:` branch carried into a new file — `source/features/transportation/bus/lib/derive-line-state.ts:73`

`BusStateEnum` has five members and the switch handles all five, so `default` can never run and the `subtitle = 'Error'` seed can never be read. Its body is `// TODO(rye): Find a replacement for this.` and `// ;(status: empty)` — commented-out pre-TypeScript code. Extracting this function was the moment to drop it.

### 12. Two unused mocks — `source/testing/expo-ui-mock.tsx:161` and `:644`

`symbolEffect` and `Capsule` are referenced nowhere outside their own definitions (and two prose comments explaining why a `Rectangle` was used *instead* of a capsule). `BusGlyph` abandoned SF Symbol effects for a scaling `Circle`, which is what made `symbolEffect` unnecessary.

## Test coverage

### 13. The strip-swipe test can pass without the strip moving — `uitests/ModuleTransportationTests.swift:16`

It swipes twice then waits for "El Tequila", with no assertion that the cell was absent beforehand. The strip opens at `stripAnchorIndex`, partway along the route, so the target may already be realised — in which case the test passes even if the swipe was swallowed by the enclosing list, which is precisely what its docstring says it exists to catch. The same PR applies the right rigour two tests down (`verifyStopListsDepartures` is documented as the precondition for `verifyStopSkipped`).

### 14. The `hidden` filter has no test at all — `index.tsx:48`

`visibleBusLines = busLines.filter((line) => !line.hidden)` lives in a route file with no Jest test, and `testEveryLineHasAWidget` only asserts Express Bus and Red Line *are* present. Nothing asserts Oles Go is absent, even though `2-oles-go.yaml` sets `hidden: true` and is in the fixture the UI tests read. Deleting the filter — or moving it into `busRoutesOptions`, which the comment warns against — is a silent regression.

### 15. Import order in the new `stop.tsx` — `stop.tsx:30`

`@frogpond/notice` and `@frogpond/colors` are imported at lines 30–31, below the block of local `source/features/transportation/bus/...` imports at 12–28. `CLAUDE.md`: *"**Imports:** React → React Native → third-party → local."* New file, so no surrounding-style exemption.

---

Also worth noting: the description says *"The calendar's event list uses the same modifier and gets the same fix."* `scrollPosition` is used nowhere else in the tree — `widget.tsx` is its only consumer, so the patch currently has one caller, not two.

drewvolz (approved):

PR #8077, opened 2026-09-19
Prevent double-tap navigation by switching to router.navigate

drewvolz:
## Summary

- Replaces `router.push` with `router.navigate` throughout the app
- `navigate()` won't push duplicate screens onto the stack if the same route is already at the top, preventing multi-screen pushes from rapid tapping during navigation transitions
- Updates test mocks to provide `navigate` instead of `push`

## Test plan

- [ ] Rapidly tap a home screen tile — only one screen should push
- [ ] Verify navigation still works normally throughout the app
- [ ] Run Jest tests (done, passing)


hawkrives (dismissed):
I'd like us to ban router.push in Oxlint, like we do for other items here: https://github.com/StoDevX/AAO-React-Native/blob/master/.oxlintrc.json#L80

drewvolz:
Would a library wrapper exposing what we want it to use instead of linter rules around what not to use be a different way to encourage the positive pattern?

hawkrives:
> Would a library wrapper exposing what we want it to use instead of linter rules around what not to use be a different way to encourage the positive pattern?

I guess? But then we have to maintain custom wrappers and types, and oxlint gets caught by the pre-commit hooks too, right?

drewvolz:
Rules updated!

hawkrives (approved):

